PR #49, opened 2014-12-22
add on-click; improve ascii_bar; ditch regex

AdamBSteele:
Sorry for making this two pull requests.

In this change, I've:
 - Added an on-click function which uses notify-msg to create a notification to display the time remaining until full charge/discharge.
 - Implemented simple global variables for customization of ascii_bar mode
 - Swapped out regex operations for string & list operations, which slightly reduces overhead and improves legibility.

Thanks for your time,
Adam

### DIFF
--- a/modules/battery-level.py
+++ b/modules/battery-level.py
@@ -4,7 +4,6 @@ from __future__ import division  # python2 compatibility
 from time import time
 
 import math
-import re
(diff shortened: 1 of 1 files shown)


ultrabug:
This has been merged into the battery_level module. Thanks again !

PR #54, opened 2015-01-13
Bar / ASCII bar also available for text mode format in battery_level.py

ZeiP:
I wanted to show the bar icon in addition to the percentage in battery_level text mode. This adds the possibility of using both bar and ascii_bar contents in the text mode format option also.


ultrabug:
Obsolete, fixed on #54

PR #55, opened 2015-01-13
Support time/tztime definitions with extra parts in the section name

ZeiP:
The time/tztime handler currently only supports definitions with section names that match the exact string. I have two time definitions, ”tztime local” and ”tztime utc”; this fixes the section name detection to only use the first space-separated part to detect if the section defines a time segment.


ultrabug:
Good catch and point once again @ZeiP, thanks a lot.

Please update your issue #54 as you've also included this patch there.

PR #56, opened 2015-01-18
Customizable state characters, mpc-like templating, no regex hack

Zopieux:
The PR implements the following features:

- Hard coded `STATE_CHARACTERS` is replaced with three new customizable parameters.
- Added a "fallback" mechanism, so if the required metadata are not available, we fallback to another  formatter. By default we want title and artist to be available, if not we fallback to file. To me this is the sanest option.
- Because of the implementation of the previous feature, the use of regex is not required anymore. Instead we use a custom subclass of built-in `Formatter`. A blessing!


ultrabug:
This is an elegant enhancement for the module, thanks @Zopieux !

I'll merge it soon and it will land into py3status v2.2.

py3status/modules/mpd_status.py
```diff
@@ -9,41 +9,67 @@
...
 
     # available configuration parameters
     cache_timeout = 2
     color = None
     format = '{state} №{pos}. {artist} - {title} [{time}] | {next_title}'
+    format_fallback = '{state} №{pos}. {file} [{time}] | {next_title}'
+    fallback_if_empty = 'artist,title'
+    state_play = "[play]"
+    state_pause = "[pause]"
+    state_stop = "[stop]"
```
ultrabug: Should those `state_xxx` really be configurable ? Maybe they should be placed in the __init__(self) below ?

shrimpza: Just to comment here, customisable state strings would be very useful, since you can put play (">"), pause, stop icons in their place, rather than the textual representations of the states.

ultrabug: Oh right, ok thanks a lot @shrimpza :) I'm not a mpd user so I couldn't test it myself.

Maybe you did test the proposed module update ? If so, your opinion would be great !

Zopieux:
Wait a sec! I just noticed mpc actually has a very powerful syntax to do more or less what I am already doing with the fallback thing:

    [[%artist% - ]%title%]|[%file%]

this will try to display `%artist - %title%`, if no artist it will display `%title%` only, if neither it will display `%file%` as a final fallback. See `mpc(1)`. It's even more powerful than my patch and is already builtin. I'll update this PR to remove the fallback stuff but keep the small refactorizations.

ultrabug:
Ok @Zopieux I'll wait then :) maybe set a default format using this mpc format ?

As I said, I'm not a mpd user so it's hard for me to be helpful.. But please let me know if you need more time, then I'll merge this after py3status v2.2 which I intend to release tomorrow.

Zopieux:
I guess you can release without this PR, it will take some time and I'm not available in the coming week!

ultrabug:
Hi @Zopieux 

Any news about this mate ? Next version is coming soon so if you want to see this issue come true, update me ;)

Zopieux:
Hello, I'll try to do something in the coming days. No big deal if it can't make the next release though!

ultrabug:
Hi @Zopieux, please tell me what you wanna do with this PR, cheers !

Zopieux:
Wow, this is getting super old, sorry about that. Lemme resolve conflicts first and I'll have a look.

Zopieux:
Okay, this is a bit longer but much more powerful! We can now use mpc-like templating in `format`. I also cleaned up the code a bit. I tested it thoroughly, it should be ready for the next release.

Zopieux:
Legacy (backward compatible) support for `{}` now added.

PR #60, opened 2015-01-26
Bitcoin price checker

dosera:
Added a module to display bitcoin prices in US $ in the py3status bar.
Currently 4 sites, btc-e.com bitstamp.com bitfinex.com and bitpay.com, are supported.


ultrabug:
Thanks a lot @dosera :)

PR #65, opened 2015-02-21
New bluetooth module

jmdana:
I wrote a new module which shows the status of the bluetooth connection ("BT: OFF" vs. "BT: \<DEVICE_NAME\>").

### DIFF
--- a/py3status/modules/bluetooth.py
+++ b/py3status/modules/bluetooth.py
@@ -43,12 +43,17 @@ class Py3status:
         p1 = subprocess.Popen(cmd1, stdout=subprocess.PIPE)
         p2 = subprocess.Popen(cmd2, stdin=p1.stdout, stdout=subprocess.PIPE)
 
-        mac = p2.communicate()[0].strip().decode("utf-8")
+        macs = p2.communicate()[0].strip().decode("utf-8").split()
 
-        if mac != "":
-            cmd3 = shlex.split("hcitool name %s" % mac)
-            p3 = subprocess.Popen(cmd3, stdout=subprocess.PIPE)
-            output += p3.communicate()[0].strip().decode("utf-8")
+        if macs != []:
+            names = []
+            for mac in macs:
+                cmd3 = shlex.split("hcitool name %s" % mac)
+                p3 = subprocess.Popen(cmd3, stdout=subprocess.PIPE)
+                names.append(p3.communicate()[0].strip().decode("utf-8"))
+
+            output += "|".join(names)
+
             color = self.color_good or i3s_config['color_good']
         else:
             output += "OFF"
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks for this nice contribution @jmdana !

I'll permit myself some QA annotations on the code you're submitting. I try to keep the modules as much alike as they can be. If you don't have time or cba doing it, I'll take the liberty to do it after the merge so there's nothing mandatory here, just say so ;)

Cheers mate

jmdana:
Sure. Please make the annotations in the code and I will have a look at them.

py3status/modules/bluetooth.py
```diff
@@ -0,0 +1,82 @@
...
+    - hcitool
+    - sed
+
+@author jmdana <https://github.com/jmdana>
+@license GPLv3 <http://www.gnu.org/licenses/gpl-3.0.txt>
+"""
+
+from time import time
+import subprocess
+import shlex
```
ultrabug: please, order your imports alphabetically

py3status/modules/bluetooth.py
```diff
@@ -0,0 +1,82 @@
...
+    color_good = None
+    color_bad = None
+
+    def __init__(self):
+        pass
+
+    def kill(self, i3s_output_list, i3s_config):
+        pass
+
+    def on_click(self, i3s_output_list, i3s_config, event):
```
ultrabug: __init__, kill and on_click should be removed if they don't actually do anything please

py3status/modules/bluetooth.py
```diff
@@ -0,0 +1,82 @@
...
+        pass
+
+    def get_status(self, i3s_output_list, i3s_config):
```
ultrabug: convention is to name the main module's method as the module name (simply bluetooth in your case)

py3status/modules/bluetooth.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    from time import sleep
+    x = Py3status()
+    config = {
+        "color_good": "#00FF00",
+        "color_bad": "#FF0000",
+        }
+
+    while True:
+        print(x.get_status([], config))
```
ultrabug: obvious renaming as per above comment, cheers

jmdana:
Done.

I've also removed the dependency on "sed" so it uses Python regular expressions, just like I did for keyboard_layout.

Please keep the good work, this is a great project!

ultrabug:
Thanks a lot @jmdana, I'm always glad to get some nice contributions !

On a future enhancement of this module, maybe a format configuration parameter would be nice :)

jmdana:
Just point me to a module that I can use as a template (so we make sure that I get the right one) and I'll have a look at it.

ultrabug:
The net_rate module is one of the most configurable one. I find the placeholders and format very elegant, you surely will too :)

https://github.com/ultrabug/py3status/blob/master/py3status/modules/net_rate.py

PR #66, opened 2015-02-21
[DPMS] Activate/deactivate the screensaver as well.

jmdana:
The DPMS module allows activation/deactivation of DPMS and does it well.
However, the purpose of deactivating DPMS is, in most cases, to avoid
the screen going blank. If the screensaver is not deactivated then the
screen will go blank anyway. This commits allow simultaneous
activation/deactivation of both.

### DIFF
--- a/py3status/modules/dpms.py
+++ b/py3status/modules/dpms.py
@@ -38,8 +38,8 @@ class Py3status:
         if event['button'] == 1:
             if self.run:
                 self.run = False
-                system("xset -dpms")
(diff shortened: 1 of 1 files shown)


ultrabug:
I totally agree with you @jmdana, thanks a lot for your contribution :)

@dosera : is that okay with you too mate ?

dosera:
Sure, go for it @ultrabug.

ultrabug:
Thanks a lot mates ! :)

PR #70, opened 2015-02-24
[bugfix] incorrect parsing of "setxkbmap -query"

jmdana:
The setxkbmap method assumes that the layout appears at the end of the "setxkbmap -query" command, which is not always true (e.g.  options defined with "setxkbmap -option" may appear after the layout).

The module has been modified so it uses "sed" for the parsing.


ultrabug:
Hi @jmdana, thanks for this proposal.

Wouldn't it then be more efficient to use a regular expression from python instead of hacking through the shell ?

jmdana:
A new pull request was made. This was automatically closed when I rebased the code.

PR #71, opened 2015-02-24
[bugfix] incorrect parsing of "setxkbmap -query"

jmdana:
The setxkbmap method assumes that the layout appears at the end of the "setxkbmap -query" command, which is not always true (e.g. options defined with "setxkbmap -option" may appear after the layout).

The module has been modified so it uses a regular expression for parsing.

### DIFF
--- a/py3status/modules/keyboard_layout.py
+++ b/py3status/modules/keyboard_layout.py
@@ -13,6 +13,8 @@ Requires:
 
 from subprocess import check_output
 from time import time
+import shlex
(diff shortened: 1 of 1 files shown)


jmdana:
The previous pull request dissapeared when I did the rebase. Sorry about that.
I have modified the code so it uses a Python regular expression.

ultrabug:
This is a nice and elegant PR, thanks a lot @jmdana :)

PR #72, opened 2015-02-24
[Bluetooth] configurable formatting

jmdana:
* `separator`: the string separator that will be used between devices
* `format_prefix`: the prefix string that will be shown when there are active connections.
* `format`: the string that will be shown when there are active connections. Accepts the placeholders `{name}` and `{mac}`.
* `format_no_conn_prefix`: the prefix string that will be shown when there are no connections.
* `format_no_conn`: the string that will be shown if there are no connections.


ultrabug:
Awesome ! Thanks mate :)

PR #74, opened 2015-03-10
added check for data.state.lastchange

0xphk:
module fails if lastchange is not existent in spaceapi .json data files and displays no status

PR #76, opened 2015-03-14
fix for Space APIs without 'lastchange' field

rixx:
Space APIs are not required to provide a 'lastchange' field. Spaces
without 'lastchange' timestamp will now always only show 'open' or
'closed' without a date.

Instead of PR #74

### DIFF
--- a/py3status/modules/spaceapi.py
+++ b/py3status/modules/spaceapi.py
@@ -28,9 +28,10 @@ class Py3status:
     # available configuration parameters
     cache_timeout = 60
     closed_color = None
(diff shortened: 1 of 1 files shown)


ultrabug:
Yeah seems like quite a few of you guys are frustated by this problem. @gidsi again, friendly ping. Is this PR good enough to solve the same problem reported by @0xphk ?

gidsi:
i told @rixx about the problem and asked if she could fix it, couldn't do it better :)

ultrabug:
Oh well, thank you very much mates :)

PR #77, opened 2015-03-15
IMAP module may now check several IMAP folders

rixx:
Returns sum of unseen mails in given folders. Former behaviour with only
one given directory is unchanged.

(I wanted the mail number to reflect a mailing list directory of mine, hence the change)

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -1,7 +1,8 @@
 # -*- coding: utf8 -*-
 """
 Module displaying the number of unread messages
-on an IMAP inbox (configurable).
+on an IMAP inbox (configurable, may also be a 
+comma-separated list of IMAP folders).
 
 @author obb
 """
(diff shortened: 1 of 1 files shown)


py3status/modules/imap.py
```diff
@@ -42,13 +43,19 @@ def check_mail(self, i3s_output_list, i3s_config):
...
-            connection.select(self.mailbox)
-            unseen_response = connection.search(None, self.criterion)
-            mails = unseen_response[1][0].split()
-            mail_count = len(mails)
+            mail_count = 0
+            directories = self.mailbox.split(',')
+            
+            for directory in directories:
+                connection = imaplib.IMAP4_SSL(self.imap_server, self.port)
+                connection.login(self.user, self.password)
```
ultrabug: Maybe we could only connect once and login once before the loop (and close after) and instead only select the given directory in the loop ?

ultrabug:
Thanks @rixx, this is indeed a good thing to propose. I permited myself with a comment to save resources on your code, please review ?

rixx:
Yeah, sorry, should have thought of that right away. Worked it in, here you go.

py3status/modules/imap.py
```diff
@@ -42,13 +43,19 @@ def check_mail(self, i3s_output_list, i3s_config):
...
             connection.login(self.user, self.password)
-            connection.select(self.mailbox)
-            unseen_response = connection.search(None, self.criterion)
-            mails = unseen_response[1][0].split()
-            mail_count = len(mails)
+            
+            for directory in directories:
+                connection.select(directory)
+                unseen_response = connection.search(None, self.criterion)
+                connection.close()
```
ultrabug: little glitch here and we're done, should be closed outside the `for`, just before the return I guess plz

ultrabug:
Thx @rixx, only one forgotten line as per comment and we're done :)

rixx:
d'oh, sorry, fixed.

ultrabug:
Thanks a lot (again) mate :)

PR #78, opened 2015-03-16
Free output format for unread mails

rixx:
Gives the user the choice of formatting the mail notification (including where to put the number of unread mails). Additionally provides the option to just not show the mail notification if there are no unread mails.

Default behaviour doesn't change.

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -1,7 +1,8 @@
 # -*- coding: utf8 -*-
 """
 Module displaying the number of unread messages
-on an IMAP inbox (configurable).
+on an IMAP inbox (configurable, may also be a 
+comma-separated list of IMAP folders).
 
 @author obb
 """
@@ -42,13 +43,19 @@ class Py3status:
 
     def _get_mail_count(self):
         try:
-            connection = imaplib.IMAP4_SSL(self.imap_server, self.port)
-            connection.login(self.user, self.password)
-            connection.select(self.mailbox)
-            unseen_response = connection.search(None, self.criterion)
-            mails = unseen_response[1][0].split()
-            mail_count = len(mails)
+            mail_count = 0
+            directories = self.mailbox.split(',')
+            
+            for directory in directories:
+                connection = imaplib.IMAP4_SSL(self.imap_server, self.port)
+                connection.login(self.user, self.password)
+                connection.select(directory)
+                unseen_response = connection.search(None, self.criterion)
+                connection.close()
+                mails = unseen_response[1][0].split()
+                mail_count += len(mails)
             return mail_count
+
         except:
             return 'N/A'
 
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -46,7 +46,7 @@ class Py3status:
             if self.hide_if_zero:
                 response['full_text'] = ''
             else:
-                response['full_text'] = self.name.replace('%unseen', '0')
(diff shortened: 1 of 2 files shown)


py3status/modules/imap.py
```diff
@@ -27,17 +29,24 @@ def check_mail(self, i3s_output_list, i3s_config):
...
+            self.new_mail_color = i3s_config['color_bad']
 
         if mail_count == 'N/A':
-            response['color'] = check_failed_color
+            response['color'] = ''
+            response['full_text'] = mail_count
         elif mail_count != 0:
-            response['color'] = new_mail_color
+            response['color'] = self.new_mail_color
+            response['full_text'] = self.name.replace('%unseen', str(mail_count))
```
ultrabug: I think that using format() placeholders would be more elegant, plz check https://github.com/ultrabug/py3status/blob/master/py3status/modules/net_rate.py for usage.

ultrabug:
What an improvement spree you're into @rixx thx a lot :) I permitted myself again to point you to using format() placeholders instead of a *dirty* replace ?

py3status/modules/imap.py
rixx: I was aiming for the format i3status uses on its own, but I can change it to format() if you want to. (i3status uses things like %speed normally). Your choice :)

ultrabug: This is a very good point @rixx indeed. It's just that it seems more 'pythonist' to use format and quite a lot of other modules now use the {placeholder} notation so I guess it's better to stick with it. (maybe it has the also good effect on differencing py3status modules from i3status' ones for users too). Cheers !

rixx:
Done :)

py3status/modules/imap.py
```diff
@@ -27,17 +29,24 @@ def check_mail(self, i3s_output_list, i3s_config):
...
         elif mail_count != 0:
-            response['color'] = new_mail_color
+            response['color'] = self.new_mail_color
+            response['full_text'] = self.name.format(unseen=mail_count)
+        else:
+            response['color'] = ''
+            if self.hide_if_zero:
+                response['full_text'] = ''
+            else:
+                response['full_text'] = self.name.replace('%unseen', '0')
```
ultrabug: little forget :)

rixx:
>.<

I'm sorry, it's late over here and I got sloppy.

ultrabug:
Thx a lot Rike !

py3status/modules/imap.py
```diff
@@ -27,17 +29,24 @@ def check_mail(self, i3s_output_list, i3s_config):
...
         response = {
-            'cached_until': time() + self.cache_timeout,
-            'full_text': '{}: {}'.format(self.name, mail_count)
+            'cached_until': time() + self.cache_timeout
         }
 
-        new_mail_color = i3s_config['color_good']
-        check_failed_color = i3s_config['color_bad']
+        if not self.new_mail_color:
+            self.new_mail_color = i3s_config['color_bad']
```
ultrabug: Oh btw @rixx , am I wrong or should this be **color_good** instead ?

rixx: Feel free to change it -- I used it this way because I like to be highlighted when there are new mails, and I only actively notice things when they indicate a problem :D

ultrabug: Thanks for your quick answer @rixx :) I've restored the original color.

PR #79, opened 2015-03-18
Pomodoro module: display time as a progress bar

FedericoCeratto:
Display remaining time as a unicode progress bar.
2 alternative styles are supported and the number of bars can be configured by editing the script.
The braces around the timer change to show if the timer is running.

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -13,21 +13,21 @@ PROGRESS_BAR_ITEMS = u"▏▎▍▌▋▊▉"
 class Py3status:
     """
     Configuration parameters:
+        - display_bar: display time in bars when True, otherwise in seconds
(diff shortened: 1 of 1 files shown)


py3status/modules/pomodoro.py
```diff
@@ -6,14 +6,28 @@
...
+        - num_progress_bars: number of progress bars
+    """
 
     # available configuration parameters
     max_breaks = 4
     timer_break = 5 * 60
     timer_long_break = 15 * 60
     timer_pomodoro = 25 * 60
+    display_bar = True
+    num_progress_bars = 5
```
ultrabug: plz order variables alphabetically here and in the docstring, thx

ultrabug:
Great addition, thanks @FedericoCeratto !

PR #80, opened 2015-03-19
Pomodoro module: add sound support

FedericoCeratto:
Add sounds for three events. Pygame is imported tentatively in order not to introduce any hard dependency.


ultrabug:
neat :)

PR #81, opened 2015-03-19
More capabilities and documentation for Pomodoro module

rixx:
In the Pomodoro module, you now can end breaks earlier and pause an ongoing Pomodoro. Those actions are also documented so that one has not to guess how to use the module :)

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -34,7 +34,10 @@ class Py3status:
             self.run = False
 
         elif event['button'] == 3:
-            self.__setup('break')
+            if self.status == 'break':
+                self.__setup('start')
+            else:
+                self.__setup('break')
             self.run = False
 
     @property
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -22,7 +22,12 @@ class Py3status:
 
     def on_click(self, i3s_output_list, i3s_config, event):
         """
-        Handles click events
+        Handles click events:
+         - left click starts an inactive counter and pauses a running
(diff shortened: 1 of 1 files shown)


rixx:
Oh, I only just saw the merge conflicts and the updates to the pomodoro module. I'll merge it locally and update my changes accordingly

rixx:
Merged and works with sound and everything, although I have to say that the upstream variable `display_bar` makes gibberish appear for me although my font should be able to display those kinds of unicode. Not my department though :)

FedericoCeratto:
Am I missing something ? It was already possible to end a break at any time with the central mouse key.
Regarding pausing a Pomodoro... well, in theory it goes against the idea of pomodoros, but that's arguable :)

rixx:
But the central mouse button would reset everything, if you'd choose to end your third break early, you'd start with your first break after the next Pomodoro.

About pausing … dunno, I sometimes like to use it for bathroom breaks but if you don't want to include it, I understand :)

ultrabug:
@rixx : I don't use it so my opinion is of little value that's why I took the liberty to ping @FedericoCeratto 

I don't see a problem in allowing a left click to pause a running pomodoro even tho it's not really meant to be... after all, rules are meant to be bent ;)

rixx:
Both of you just decide, I'm totally finde with having my own fork there, too ;)

FedericoCeratto:
@rixx: oh I get it, +1 for merging then.
Maybe later that self.run could be removed to have a simple state machine based on self.status

ultrabug:
Thanks a lot mates :)

PR #83, opened 2015-03-25
Improve documentation in the example module

FedericoCeratto:
Maybe the example module should be moved to doc/examples/ to move it out the real modules directory and also make it easier to find for people looking at the docs.

### DIFF
--- a/py3status/modules/empty_class.py
+++ b/py3status/modules/empty_class.py
@@ -1,15 +1,35 @@
 # -*- coding: utf-8 -*-
-"""
-Empty and basic py3status class.
 
-NOTE: py3status will NOT execute:
-    - methods starting with '_'
-    - methods decorated by @property and @staticmethod
-
-NOTE: reserved method names:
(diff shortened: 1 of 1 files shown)


py3status/modules/empty_class.py
```diff
@@ -1,15 +1,35 @@
...
-NOTE: py3status will NOT execute:
-    - methods starting with '_'
-    - methods decorated by @property and @staticmethod
+Multi-line description followed by an empty line.
 
-NOTE: reserved method names:
-    - 'kill' method for py3status exit notification
-    - 'on_click' method for click events from i3bar (read below please)
+@author <your full name> <your email address>
+@license <ancronym>     (please favor BSD or GPLv3)
```
ultrabug: I would just replace this and enforce a BSD license just like py3status so packagers don't have to loose time with this from now on

ultrabug:
It's way more clearer you're right mate, thx a lot! just a little comment above, no biggie

PR #89, opened 2015-03-31
Add profiling support and setproctitle

FedericoCeratto:
setproctitle is an optional dependency to set the process name.
The profiler can generate metrics for different threads and can be enabled using the enable_profiling  variable.

### DIFF
--- a/py3status/__init__.py
+++ b/py3status/__init__.py
@@ -23,6 +23,11 @@ from threading import Event, Thread
 from time import sleep, time
 from syslog import syslog, LOG_ERR, LOG_INFO, LOG_WARNING
 
+try:
+    from setproctitle import setproctitle
+    setproctitle('py3status')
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks @FedericoCeratto, before I merge I'd like to test this but I'm now sure how. Care to explain how it is done/read ? cheers

FedericoCeratto:
@ultrabug: the output files are in the usual cProfile format and named py3status-<thread-id>.profile + py3status-core.profile
They can be inspected with profile analyzers including http://www.vrplumber.com/programming/runsnakerun/

ultrabug:
Neat stuff, I'm eager to have some feedback on what it shows as I'm not sure I'm good enough to grasp it yet but I'll surely give it a try.

Thanks a lot Federico !

PR #91, opened 2015-04-01
External processes might change DPMS

jmdana:
External processes might change DPMS so the module should check in the main loop

### DIFF
--- a/py3status/modules/dpms.py
+++ b/py3status/modules/dpms.py
@@ -25,9 +25,13 @@ class Py3status:
         """
         Display a colorful state of DPMS.
         """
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @jmdana, you're indeed right about this. I think that if you run the detection from the `dpms` method like your commit is proposing, you should then just drop the `__init__()` altogether right ?

Thanks

jmdana:
Sure. I didn't know what was the QA.

ultrabug:
Yeah I need to write a doc about this badly :( Thanks mate !

PR #95, opened 2015-04-29
Add aws_bill module

nawadanp:
Add a new module which display the current bill for an AWS account.
You must enable the billing export into an S3 bucket to use this module : http://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/detailed-billing-reports.html


ultrabug:
Great idea and neat implementation @nawadanp thanks !

PR #96, opened 2015-05-07
imap module: fix empty color that made text invisible on i3bar when no unread email

Horgix:
When the **imap module** isn't able to connect to the mailbox through imap,
or when there is no unread email, the color is set like that :

    response['color'] = ''

That makes the color black on i3bar, and thus invisible on most
backgrounds.

Removing these 2 lines makes it use the default color, which is white
and visible in most of the case. That also allow the "hide_if_zero"
module option to be useful !


ultrabug:
It's perfect, thanks a lot mate !

PR #102, opened 2015-06-10
pomodoro: fix display_bar in python3

mmlb:
In python3 calling `encode` on a string will yield a `bytes` obj which
then leads to a status bar with escaped hex chars instead of the bars, i.e.
`Pomodoro [b'\xe2\x96\x89\xe2\x96\x89\xe2\x96\x8d ']` vs `Pomodoro [b'▉▉▍ ']`.

The fix in python2 is to call format on a `unicode` object vs a `str` (ansii)
object. This will work fine in python3 > 3.3 where the `u` prefix was brought
into python3.

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -106,14 +106,14 @@ class Py3status:
                 bar += PROGRESS_BAR_ITEMS[selector]
                 bar_val -= 1
 
-            bar = bar.ljust(self.num_progress_bars).encode('utf_8')
+            bar = bar.ljust(self.num_progress_bars)
         else:
             bar = self.timer
 
         if self.run:
-            text = '{} [{}]'.format(self.prefix, bar)
+            text = u'{} [{}]'.format(self.prefix, bar)
         else:
-            text = '{} ({})'.format(self.prefix, bar)
+            text = u'{} ({})'.format(self.prefix, bar)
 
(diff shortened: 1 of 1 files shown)


ultrabug:
ping @FedericoCeratto and @rixx, you guys run python 3 AFAIR

@mmlb : I'd like to support the most versions of python available, care to have it working for all py3 ?

mmlb:
@ultrabug The code ASIS won't work on py3 < 3.3 since other strings have the `u` prefix. If no one has complained about that already I am assume that all your users are >= 3.3.

If @FedericoCeratto or @rixx say otherwise I will prepare fix for <3.3, but it'll be uglier :(

ultrabug:
Yes maybe but not all py3 users are using the pomodoro module either. So let us give some time to our mates here to respond, I'd love their insight.

mmlb:
Sounds good to me, just fyi player_control.py has unicode prefixs also, and depending on action here may warrant fixing there too.

rixx:
I think folks using Python3 will normally use 3.3 or even 3.4. Unless somebody runs into problems, I'd put in support for 3.3+, but leave older 3.x versions alone for now.

mmlb:
@ultrabug ping

ultrabug:
Great addition and fix, thanks a lot @mmlb and @rixx !

Issue #98, opened 2015-05-12
Uniformize and standardize 'format' option


Horgix:
Looking at the current modules, I found that the option allowing users
to define a custom output format is not consistent or not present. One
thing that made me like i3status was the idea that no matter the module,
it is the `format` option that is used to customize its output. I
think it would be nice to uniformize existant modules around this idea
and to document this as the way to go for future modules.

Beside the uniformization, it offers more flexibility for users : take
for example a simple module like
[whoami](https://github.com/Ultrabug/py3status/blob/master/py3status/modules/whoami.py).
It actually only offers the possibility to display the username of the
user running py3status. No formatting of the output is possible.
Wouldn't it be nice if it defined a `format = "{username}"` in the
configuration ? It would allow the following :
- Still the same default behavior
- Definable by users in i3status.conf, to add simple things sur as
  prefix or suffix (which are definitely handy and pretty when used with
  UTF8 icons)
- By using directly `{username}` and making it available to the user,
  there isn't more parsing to do on the module side. Just return
  `format.format(username = username)` as response and it's done and
  working for both default value and possible user custom values

Details for existing modules (OK means it does it the way I described as
ideal; KO means the output is absolutely not configurable atm; links are
on the concerned lines) :
- [aws_bill](https://github.com/Horgix/py3status/blob/master/py3status/modules/aws_bill.py#L38)
  : KO; no output customization available
- [battery_level](https://github.com/Horgix/py3status/blob/master/py3status/modules/battery_level.py#L40)
  has a `format` option but with unamed parameters, which is probably
  not intuitive for users that would want to redefine it themselves
- [bitcoin_price](https://github.com/Horgix/py3status/blob/master/py3status/modules/bitcoin_price.py#L38)
  KO; doesn't have any output configuration
- [bluetooth](https://github.com/Horgix/py3status/blob/master/py3status/modules/bluetooth.py#L38)
  **OK**
- [clementine](https://github.com/Horgix/py3status/blob/master/py3status/modules/clementine.py#L16)
  KO; offers absolutely no output costomization, would probably greatly
  benefit from it and would be easily adaptable
- [dpms](https://github.com/Horgix/py3status/blob/master/py3status/modules/dpms.py#L16)
  : KO; no customization, easy to add
- [glpi](https://github.com/Horgix/py3status/blob/master/py3status/modules/glpi.py#L24)
  KO; would benefit from it
[...]

PR #109, opened 2015-07-20
add format string support for dpms module, and add some options

shrimpza:
In response to issue #98 - add format string support for the `dpms` module.

Since it's going to be a bit more configurable with a `format` string, I added the placeholder, `{status}`, which allows customising the textual output based on DPMS state. I have not added this new parameter to the default string, to not confuse or surprise existing users.

Also added a little test which this module was lacking.

### DIFF
--- a/py3status/modules/dpms.py
+++ b/py3status/modules/dpms.py
@@ -7,12 +7,8 @@ of DPMS (Display Power Management Signaling)
 by clicking on 'DPMS' in the status bar.
 
 Configuration parameters:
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks a lot for your contribution @shrimpza. If I may, I think we should use `format_on` and `format_off` options instead if that's okay with @Horgix 's RFC ?

ultrabug:
Perfect, thanks @shrimpza !

PR #119, opened 2015-09-16
added module xsel to display xsel buffers in py3status

UmBsublime:
As discussed in #py3status on freenode here is my PR for xsel.py


ultrabug:
Thanks a lot @UmBsublime !

PR #120, opened 2015-09-23
Module update: Additional display options for pomodoro.py timer

schober-ch:
Added additional display features to pomodoro.py. It is now possible to display the bar and the time as well as choosing a hh:mm:ss format instead of just plain seconds.

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -3,9 +3,17 @@
 Display and control a Pomodoro countdown.
 
 Configuration parameters:
-    - display_bar: display time in bars when True, otherwise in seconds
-    - display_bar_and_time: display time in bar and seconds when True
-    - display_mmss: display time in hh:mm:ss format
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @schober-ch, thanks for this contribution !

Before merging it, I'd like your point of view about implementing a `format` parameter instead with a selection of all those formats instead ? So would allow for the maximal flexibility of rendering don't you think ?

We could get then get something like : `format = "{bar} {time_mmss}"`

Sounds fun ?

schober-ch:
I think this is a good idea. This would replace the `display_bar` config option, right? 

So we could for example define as possible elements: `{bar} {time_mmss} {time_seconds}` and have `{time_seconds}` as default. This would keep the actual behaviour.

ultrabug:
That's right, basically replace all those "one parameter per feature" by format placeholders !

schober-ch:
Sounds like a plan. I can do these changes and submit another pull request.

PR #121, opened 2015-09-24
Module update: Additional display options for pomodoro.py timer

schober-ch:
Implemented the display_format option discussed in https://github.com/ultrabug/py3status/pull/120 

Allows any combination of bar, seconds, minutes or (hh:)mm:ss as display for the remaining time.

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -3,9 +3,17 @@
 Display and control a Pomodoro countdown.
 
 Configuration parameters:
-    - display_bar: display time in bars when True, otherwise in seconds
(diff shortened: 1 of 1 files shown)


ultrabug:
I was more willing to be in line with issue #98 with a `format` parameter plz ?

Also, see the spotify module sources as to how to format the docstring : https://github.com/ultrabug/py3status/blob/master/py3status/modules/spotify.py

Thanks !

schober-ch:
Ah, you should have posted that issue before, I was not aware of it. Will add the changes today.

ultrabug:
I love it @schober-ch thank you so much, and sorry for not saying it earlier :)

Issue #110, opened 2015-07-23
timezone option in tztime ignored


olhotak:
My config file is the following:
```
order += "tztime local"
order += "tztime toronto"

tztime local {
        format = "%Y-%m-%d %H:%M"
}

tztime toronto {
        format = "%H:%M %Z"
        timezone = "America/Toronto"
}
```

With this config file, `i3status` displays:
`2015-07-23 22:02 | 16:02 EDT`

However, `i3bar` configured with `py3status` displays (in the status bar):
`2015-07-23 22:02 | 22:02`

Notice that:
1. The time is in the wrong timezone (22:02 instead of 16:02).
2. The timezone identifier (EDT) is missing (yet it is supposed to be there due to the `%Z` in the format string).

I would expect `py3status` to match `i3status`.

The JSON output of `py3status` with this configuration file is:
```
{"click_events": true, "version": 1}
[
[{"name":"tztime","full_text":"2015-07-23 22:06"},{"name":"tztime","full_text":"16:06 EDT"}]
,[{"name": "tztime", "full_text": "2015-07-23 22:06"}, {"name": "tztime", "full_text": "22:06 "}]
```

My versions are:
```
py3status version 2.4 (python 2.7.10)
i3status 2.9 (2015-03-22) © 2008-2012 Michael Stapelberg and contributors
i3bar version 4.10.2 (2015-04-16, branch "4.10.2") © 2010-2014 Axel Wagner and contributors
```

ultrabug:
Hi @olhotak 

The problem here is the "%Z" I think and it's a tricky one because the C and python implementation of it is different.

First of all, can you confirm that removing `%Z` fixes the issue ?

olhotak:
Thank you for replying so quickly.

Yes, removing `%Z` fixes the first issue: the timezone is now correct.

ultrabug:
FYI as you can see, I've coded a workaround which allows %z to be used now. This will be available in the 2.7 release :)

PR #123, opened 2015-10-11
Fix crash when using %z for tz_time

derekdreery:
tl;dr python's support for datetime leaves a bit to be desired.

The system can store 2 different objects under the same banner (datetime). These can either have tz info (including UTC) or be "naive" datetimes (no tzinfo). Stupidly, utcnow creates a naive datetime that cannot be used with other proper datetimes.

This was causing a crash when %z was included. The patch fixes this. I don't yet know why the timezone (%z) won't show up in i3bar yet though.

### DIFF
--- a/py3status/__init__.py
+++ b/py3status/__init__.py
@@ -13,7 +13,7 @@ import sys
 from collections import OrderedDict
 from contextlib import contextmanager
 from copy import deepcopy
-from datetime import datetime
+from datetime import datetime, timezone
 from json import dumps, loads
 from py3status import modules as sitepkg_modules
 from signal import signal
@@ -548,7 +548,7 @@ class I3status(Thread):
                                 print_line(line)
                                 with jsonify(line) as (prefix, json_list):
                                     self.last_output = json_list
-                                    self.last_output_ts = datetime.utcnow()
+                                    self.last_output_ts = datetime.now(timezone.utc)
                                     self.last_prefix = ','
                                     self.update_json_list()
                                     self.set_responses(json_list)


ultrabug:
Wow @derekdreery such a simple fix for one of the oldest problem on py3status ? I'll try asap but I wanted to already thank you a lot.

Ping issue #110 and @olhotak !

derekdreery:
Thanks :)

It's still not showing the timezone but at least it doesn't crash.

ultrabug:
Hi again @derekdreery, after testing and wondering why I didn't have the crash issue I finally found out that your PR is for **python3** only !

py3status must be also 2.x compliant, so I cannot merge this as is because the 2.x `datetime` module has no `timezone` module !

Can you include python2 support in your patch please ? (FYI, %z doesn't crash python2 py3status, it is just ignored and replaced by an empty space).

pydsigner:
Is http://stackoverflow.com/a/2331635/1795505 helpful?

derekdreery:
OK n.p. I'll take a look

ultrabug:
bump @derekdreery :)

ultrabug:
Okay, after giving it some thought I've coded a complete workaround which now allows to use %z on time and tztime modules properly :) Give it a try and reopen if needed, thanks mates !

pydsigner:
Closing commit: https://github.com/ultrabug/py3status/commit/870491ab89413d0395358c13fde99e7e95abcda8

derekdreery:
Hi sorry life distracted me. I hope I was of some use :)

ultrabug:
np @derekdreery , this gave me the will to fix this too old limitation, thanks again !

PR #125, opened 2015-10-13
Fix poll() calls to convert from seconds

pydsigner:
Fixes bug where click events are never executes.


ultrabug:
Awesome bugfix and impressive optimization, thanks a bunch @pydsigner !

PR #126, opened 2015-10-14
trim spaces in `cpu_temp`

beetleman:
Removing annoying spaces in `cpu_temp`


ultrabug:
Good catch, thanks a lot @beetleman ! :)

PR #127, opened 2015-10-14
Fix mpd connection leak

daedric:


py3status/modules/mpd_status.py
```diff
@@ -213,7 +213,8 @@ def attr_getter(attr):
             text = "Failed to connect to mpd!"
         except CommandError:
             text = "Failed to authenticate to mpd!"
-            c.disconnect()
+
+        c.disconnect()
```
ultrabug: maybe it would be best in a `else:` AND  in the `except CommandError` ? because if the connection fails the disconnect() will raise another Exception right ?

ultrabug:
Thanks @daedric, commented as I think it could benefit from a finer control.

py3status/modules/mpd_status.py
daedric: Unfortunately you're right and there is no way to check whether we're connected or not with the mpd module.

Fixed.

ultrabug:
Cheers mate :)

PR #128, opened 2015-10-14
Call the method disconnect whatever the outcome of block try/except, …

btall:
…or being ensure to avoid the connection leak. :)

In seeing this #127 , I thought it would be better to make the call to the disconnect method in a finally block. :)

But not possible to call this method in this condition, which is why I have proposed an enhancement on python-mpd2 (https://github.com/Mic92/python-mpd2/pull/59).

### DIFF
--- a/py3status/modules/mpd_status.py
+++ b/py3status/modules/mpd_status.py
@@ -213,8 +213,7 @@ class Py3status:
             text = "Failed to connect to mpd!"
         except CommandError:
             text = "Failed to authenticate to mpd!"
-            c.disconnect()
-        else:
+        finally:
             c.disconnect()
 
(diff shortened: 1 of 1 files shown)


pydsigner:
I honestly think that having disconnect always work is not a good thing, but regardless, this PR improves the codebase slightly while requiring a 3-hour-old revision of a widely used stable package. I think this PR should be stashed at least until your mpd2 revision is released on PyPI.

ultrabug:
I agree, it's a bit too new to be merged here, the current implementation works good enough. We'll keep this in mind tho @btall thank you very much !

Closing for a while, will reopen/merge some other day.

Issue #124, opened 2015-10-13
Spotify modules doesn't work with latest version of Spotify


epaaj:
The Spotify modules doesn't work with the latest version of Spotify.
There seems to have been som changes to the data in dbus.

ultrabug:
ping @guilbep, mind looking into it mate ?

jakubjedelsky:
fyi, I just found some release notes for [1.0.13](https://community.spotify.com/t5/Spotify-Community-Blog/Spotify-beta-1-0-13-for-Linux-released/ba-p/1197675):

- The application identifier is com.spotify.music, not com.spotify.qt
- The album artist field (xesam:albumArtist) is now respected and filled in if available
- The xesam:url property is now a proper URL (it was previously a Spotify URI)
- We no longer support the xesam:contentCreated field, sorry!
- The client does not yet send the PropertiesChanged signal to alert other applications that playback state has changed. This should be fixed in the next version.

ultrabug:
@jakubjedelsky & @epaaj, mind trying @pydsigner's code and reporting if this fixes the issue ? I'm no spotify user so this would help a lot. Thx

epaaj:
Those changes does not work for me. It still does not recognize that Spotify is running.
I got it to work with some changes of my own. But I'm a newbie when it comes to DBus.
See #130

pydsigner:
@epaaj I'm not a DBus expert either (and don't plan on becoming one.....) but if that works for you then that's probably a good thing.

ultrabug:
Fixed by @epaaj on MR #130 , thanks to everyone !

PR #130, opened 2015-10-21
Update of spotify module for Spotify 1.0.16

epaaj:
This works for me for Spotify 1.0.16. But I'm a newbie when it comes to getting data from DBus so I'm not sure this is the right way. The DBus path shouldn't have changed this much, should it?

### DIFF
--- a/py3status/modules/spotify.py
+++ b/py3status/modules/spotify.py
@@ -39,15 +39,18 @@ class Py3status:
         """
         bus = dbus.SessionBus()
         try:
-            self.__bus = bus.get_object('com.spotify.qt', '/')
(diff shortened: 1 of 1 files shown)


pydsigner:
I've seen that DBus path in relation to both 0.9 and 1.0, I think it's just an alternative path.

Proper ping back to #124

epaaj:
Yeah. On an Ubuntu machine I have, which is still using Spotify 0.9, I see both the com.spotify.qt and the org.mpris.MediaPlayer2.spotify paths.
But on Arch Linux with Spotify 1.0.16 I only have the latter. I guess com.spotify.music or something similar should be there as well. Strange.

ultrabug:
Well if you got it working like this @epaaj then I guess it's okay :) thank you !

PR #131, opened 2015-10-25
[battery_level] Allow user-defined blocks with percentage

maximbaz:
Hello, I'd like to propose extending the `battery_level` module to allow using user-defined characters as blocks. This is especially beneficial when using icon font:

selection_008
selection_003
selection_004
selection_005
selection_006
selection_007

As you can see, I also made it possible to show percentage next to the icon.

Both settings are optional, and their default values are set to preserve the existing behavior -- so current users shouldn't notice any difference. 

This is the configuration that produces the screenshots above (it assumes the icon font to be installed and configured in `.i3/config`):

```
battery_level {
  blocks = ""
  charging_character = ""
  show_percent_with_blocks = true
}
```

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -3,12 +3,12 @@
 Display the battery level.
 
 Configuration parameters:
+    - blocks : a string, where each character represents a battery level
+    - charging_character : a character to represent charging battery
     - color_* : None means - get it from i3status config
     - format : text with "text" mode. percentage with % replaces {}
     - hide_when_full : hide any information when battery is fully charged
     - mode : for primitive-one-char bar, or "text" for text percentage ouput
-    - blocks : a string, where each character represents a battery level
-    - charging_character : a character to represent charging battery
     - show_percent_with_blocks : show battery level percentage in blocks mode
 
 Requires:
@@ -35,7 +35,9 @@ class Py3status:
     """
     """
     # available configuration parameters
+    blocks = BLOCKS
     cache_timeout = 30
+    charging_character = CHARGING_CHARACTER
     color_bad = None
     color_charging = "#FCE94F"
     color_degraded = None
@@ -44,8 +46,6 @@ class Py3status:
     hide_when_full = False
     mode = "bar"
     notification = False
-    blocks = BLOCKS
-    charging_character = CHARGING_CHARACTER
     show_percent_with_blocks = False
 
     def battery_level(self, i3s_output_list, i3s_config):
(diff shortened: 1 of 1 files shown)


ultrabug:
Very nice proposal @maximbaz, thanks a lot.

A few questions/remarks before I merge :
- what icon font are you talking about ? please add a quick howto in the docstring of the module
- please order alphabetically the variables you add with the others
- do you think you could implement format string placeholders as described in issue #98 ?

Thanks !

maximbaz:
Good remarks, @ultrabug. 

### Icon fonts

I've added a hint to _look into icon fonts_ in the docstring, and I don't think we should put more details, at least not in there. Ideally it could be a wiki page with the full tutorial, and telling from experience, it's going to be neither small nor easy to write :smiley:.

To clarify this particular change, previously the `blocks` and `charging_character` variables had _hardcoded_ values `"_▁▂▃▄▅▆▇█"` and `"⚡"` respectively. I've only made them configurable, preserving the those hardcoded values as defaults. 

Now if people want, they can for example set `blocks` variable to be `"LMH"`, and `L` letter would be shown when the battery level is low, `M` when it's medium and `H` when it's high. 

If a person happens to use an icon font (let's take FontAwesome as an example), they will have to check the unicode codes of icons in that particular font and compose the `blocks` and `charging_character` variables out of them. 
> Since we took FontAwesome as an example, [the icon representing half of the battery ~> "" <~](https://fortawesome.github.io/Font-Awesome/icon/battery-half/) has a unicode code of `f242` - you would see it here if you have the font installed.

In my particular case, I have created a font on my own, thus I chose which icons correspond to which unicode codes, and based on that I configured my `blocks` and `charging_character` variables. Therefore this part of my config is useless to anyone else, unless they also install my font.

------

As you can see, this very quickly becomes complicated to explain and to provide a single set of steps, that would suit everyone. That's why I want to keep it simple, as it is today.

Please have a look, do you agree that this is sufficient?

### Implement the proper format string

I can certainly help refactoring the `battery_level` module, but I would like to make it in a separate pull request. I anticipate that to be a bigger change, with more questions to be resolved -- while this pull request on contrast is a small improvement, that doesn't affect current users of the module.

The primary question that I have is about backward compatibility, what is your strategy in this project? 

> 1. I think the whole concept of `modes` in this plugin becomes obsolete, if there is a proper support for `format` parameter.
> 2. The parameter `show_percent_with_blocks` will also be useless.
> 3. I would give the `percentage` a more descriptive format than the current `{}`.

The question is, do we want to keep backward compatibility, or rather get rid of everything obsolete completely during the refactoring?

ultrabug:
Thanks for the detailed response @maximbaz. This icon font stuff is indeed too complex to fit into the docstring. I think that other py3status users could benefit from a wiki page tho which we could refer to in the README if you can spare the time.

I agree with you that we can implement the format placeholders in another PR. Backwards compatibility is always important so maybe we could make a sane default out of the placeholders while keeping the old behaviour as well ? so that is say `mode` was not used, we can assume the default format is okay etc... I'm clearly open to suggestions here.

Thanks a lot for your time and commitment !

PR #132, opened 2015-10-27
Enhanced version of keyboard_layout module.

tuxitop:
I have made a few enhancements for the keyboard_layout module. I used xset to get the currently active layout. It doesn't work correctly if more than two layouts are configured, but since it doesn't happen in most cases, It almost resolves #33 and doesn't always require installing xkblayout-state.
I also added a better way of customizing colors, and made a few code optimizations.

### DIFF
--- a/py3status/modules/keyboard_layout.py
+++ b/py3status/modules/keyboard_layout.py
@@ -4,14 +4,15 @@ Display the current keyboard layout.
 
 This module displays the current active keyboard layout.
 Requires:
+    - setxkbmap
     - xkblayout-state
     or
-    - setxkbmap and xset (Works for the first two predefined layouts.)
+    - xset (Works for the first two predefined layouts.)
 
 Configuration parameters:
     - cache_timeout: check for keyboard layout change every seconds
     - colors: a comma separated string of color values for each layout,
-              eg: "#FCE94F, #729FCF".
+              eg: "fr=#FCE94F, en=#729FCF".
 
 @author shadowprince, tuxitop
 @license Eclipse Public License
@@ -29,6 +30,16 @@ class Py3status:
     cache_timeout = 1
     colors = '#729FCF, #FCE94F'
 
+    # The following options has been placed as an alternative, for backward
+    # compatibility
+    LANG_COLORS = {
+        # 'fr': '#268BD2',  # solarized blue
+        # 'ru': '#F75252',  # red
+        # 'ua': '#FCE94F',  # yellow
+        # 'us': '#729FCF',  # light blue
+    }
+    color = ""
+
     def __init__(self):
         """
         find the best implementation to get the keyboard's layout
@@ -39,8 +50,8 @@ class Py3status:
             self.command = self._xset
         else:
             self.command = self._xkblayout
-        self.colors_lst = self.colors.split(",")
         self.layouts = self._get_layouts()
+        self.colors_lst = self._get_colors_lst()
 
     def keyboard_layout(self, i3s_output_list, i3s_config):
         response = {
@@ -59,6 +70,17 @@ class Py3status:
         response['full_text'] = lang or '??'
         return response
 
+    def _get_colors_lst(self):
+        """
+        Returns a list of colors for each layout
+        """
+        if self.color:
+            return [self.color for lang in self.layouts]
+        if self.LANG_COLORS:
+            return [self.LANG_COLORS.get(lang) for lang in self.layouts]
+
+        return self.colors.split(",")
+
     def _get_layouts(self):
         """
         Returns a list of predefined keyboard layouts
@@ -70,7 +92,7 @@ class Py3status:
 
     def _xkblayout(self):
         """
-        check using xkblayout-state (preferred method)
+        check using xkblayout-state
         """
         return check_output(
             ["xkblayout-state", "print", "%s"]
--- a/py3status/modules/keyboard_layout.py
+++ b/py3status/modules/keyboard_layout.py
@@ -10,7 +10,8 @@ Requires:
 
 Configuration parameters:
     - cache_timeout: check for keyboard layout change every seconds
-    - colors: "comma separated list of color values for each layout"
+    - colors: a comma separated string of color values for each layout,
+              eg: "#FCE94F, #729FCF".
 
 @author shadowprince, tuxitop
 @license Eclipse Public License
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @tuxitop, thanks a lot for your contribution !

I will check it out ASAP and get back to you.

Merci ;)

ultrabug:
Hi @tuxitop, I wanted to try your module and immediately got an error. L88 should be `return self.layouts[0]` right ?

Also, I don't understand how you map a layout name to a colour now ? It looks to me that your PR is not backwards compatible with current users' setups right ?

tuxitop:
Hi @ultrabug 
I'm sorry for the embarressing bug, I fixed that.
About the colors I thought about backward compatibility but it would just make everything more complicated, so I decided to keep it simple. The new configuration is as simple as putting 'colors = "#729FCF #FCE94F"' in the config file and it assigns the proper color to each layout in the right order.
If backward compatibility is required, I can change it back to the more complicated way.

ultrabug:
@tuxitop yes, it needs to be backward compatible plz. key / value assignment seems very handy for this please.

For example something like : `colors="fr=#xxxx,en=#yyyy,ir=#ooooo"` that would be parsed to a dict replacing LANG_COLOR would be great no ?

tuxitop:
@ultrabug 
my whole point was to make it colorful by default for everyone, even if their layout was not specifically mentioned in LANG_COLOR. 
So here's my backward compatible suggestion. the `colors=(#xxx, #yyy)` remains as is, but we will first check for the `color` variable, after that for `LANG_COLOR` (as it was before), and if none of them was availble, we will check for the new `colors` variable.

ultrabug:
@tuxitop I get what you mean but tbh this sounds complicated to me. Why not get rid of `LANG_COLOR` with the same key/values set in the default `colors` of yours which could then be overriden by users ?

My point is that your proposal lacks a clear mapping of `keymap name` - `color`, the `colors=(#xxx, #yyy)` does not make it clear what color goes for what keymap. It may looks simple to you because you **know** about the parsed output but common users don't.

So I still think it's better to *see* what keymap equals what color code somewhere and provide a simple and unique way to override the current default.

tuxitop:
@ultrabug I get your point now, and I think that's a fair point. but I'm not sure what you mean by getting rid of `LANG_COLOR`. If we get rid of LANG_COLOR by default values in `colors`, that would make it backward incompatible, and if someone has made customizations in `LANG_COLOR`, it would be overwritten by the default values.

ultrabug:
@tuxitop well actually LANG_COLOR is a global variable which is not configurable by users from i3status.conf. Configurable variables are the ones defined in the Py3status class declaration under the `# available configuration parameters` comment.

The idea here is that `colors` would get there and get the required default values from LANG_COLOR. To quote again my previous example this would show as something like :

```
    cache_timeout = 10
    color = ''
    colors = 'fr=#268BD2,ru=#F75252,ua=#FCE94F,us=#729FCF'
```

See what I mean ? :)

tuxitop:
@ultrabug I think I got what you mean now. This one is hopefully what you have in mind. :)

py3status/modules/keyboard_layout.py
```diff
@@ -63,11 +36,16 @@ def __init__(self):
         find the best implementation to get the keyboard's layout
         """
         try:
-            xkblayout()
+            self._xkblayout()
+            self.command = self._xkblayout
```
ultrabug: setting `self.command` in the __init__ will make this a class method and will thus have py3status try to execute it directly like the `keyboard_layout` one. You can see this on the logs when running with -d flag :

`user method command failed (__xset() takes exactly 1 argument (3 given))`

To avoid this, just rename self.command to **self._command** please

py3status/modules/keyboard_layout.py
```diff
@@ -2,60 +2,33 @@
...
-    return re.match(LAYOUT_RE, out).group(1)
-
 
 class Py3status:
     """
     """
     # available configuration parameters
     cache_timeout = 10
+    colors = 'us=#729FCF, fr=#268BD2, ua=#FCE94F, ru=#F75252'
     color = ''
```
ultrabug: maybe this would be better to set it to None by default ?

py3status/modules/keyboard_layout.py
```diff
@@ -63,11 +36,16 @@ def __init__(self):
...
+            self._xkblayout()
+            self.command = self._xkblayout
         except:
-            self.command = setxkbmap
-        else:
-            self.command = xkblayout
+            self.command = self._xset
+            self.layouts = self._get_layouts()
+
+        if not self.color:
```
ultrabug: this whole block and the `self.colors_dict` generation should be put on the `keyboard_layout` method because the class is instance is created by py3status **before** it can override the users' config values. This means that for now the users' config is not taken into account in your module.

py3status/modules/keyboard_layout.py
```diff
@@ -76,14 +54,49 @@ def keyboard_layout(self, i3s_output_list, i3s_config):
         }
```
ultrabug: move the color condition + `self.colors_dict` generation here

ultrabug:
Great @tuxitop, almost there thank you ! :) I've made a few comments in the code please.

Oh and btw, testing your module got me find and fix a bug in my i3status.conf parser ;) cheers

tuxitop:
@ultrabug I fixed all the things you mentioned and now I have a better understanding of py3status. But I'm still getting the _xset() error. why is py3status still trying to run it each time?

ultrabug:
@tuxitop thanks mate.

As for the _xset() it is just because you didn't use a proper **underscore** character ! Check it out, it must be a special unicode character you've used :)

py3status/modules/keyboard_layout.py
```diff
@@ -2,88 +2,97 @@
...
-    return re.match(LAYOUT_RE, out).group(1)
-
 
 class Py3status:
     """
     """
     # available configuration parameters
-    cache_timeout = 10
-    color = ''
+    cache_timeout = 1
```
ultrabug: please do not change default cache timeouts, running a command every second is kinda overkill, leave it to 10s

tuxitop:
@ultrabug I'm really sorry for that wrong unicode character. I must have been on the wrong keyboard layout. 
About the `cache_timeout`, I figured it doesn't make sence to update the layout each 10 secs. I mean it is ususally expected that the the text gets updated once I change the layout. 
Anyway, Thank you for your comments and your patience. py3status is indeed a great tool.

ultrabug:
Merci Ali !

PR #134, opened 2015-10-28
Proper parsing for boolean values in config

Anon1234:
Hi! 
At this moment such config parameters as ```separator = false``` converts to ```"separator": "false"``` when parsed, that is not a proper representation for a boolean variable in Python or in JSON.
This PR must fix it.


ultrabug:
This is a good point @Anon1234 thanks.

I think you also need to take into account the `separator = "false"` configurations, atm your code doesn't support it. Do you think you can adapt this PR for this too ?

Anon1234:
Done.

ultrabug:
Thanks a lot @Anon1234 !

PR #135, opened 2015-10-29
Add window_title_async.py module

Anon1234:
We already have the window_title.py module, but this version provides async update for the title and uses well maintained i3 IPC library.

### DIFF
--- a/py3status/modules/window_title_async.py
+++ b/py3status/modules/window_title_async.py
@@ -7,7 +7,7 @@ Provides instant title update only when it required.
 Configuration parameters:
     - format : format of the title, default: "{title}".
     - empty_title : string that will be shown instead of the title
-                    when the title is hidden, default: 500 spaces.
+                    when the title is hidden, default: " ".
     - always_show : do not hide the title when it can be already
                     visible (e.g. in tabbed layout), default: False.
 
@@ -26,7 +26,7 @@ import i3ipc
 class Py3status:
 
     format = "{title}"
-    empty_title = " " * 500
+    empty_title = " "
     always_show = False
 
     def __init__(self):
--- a/py3status/modules/window_title_async.py
+++ b/py3status/modules/window_title_async.py
@@ -47,6 +47,7 @@ class Py3status:
             # dont show window title when the window already has means
             # to display it
             if not self.always_show and (w.border == "normal" or
+                    w.type == "workspace" or
                     (p.layout in ("stacked", "tabbed") and len(p.nodes) > 1)):
                 return self.empty_title
             else:
--- a/py3status/modules/window_title_async.py
+++ b/py3status/modules/window_title_async.py
@@ -73,7 +73,8 @@ class Py3status:
             # in the workspace layout
             layout_changed = (hasattr(e, "binding") and
                               (e.binding.command.startswith("layout") or
-                               e.binding.command.startswith("border")))
+                               e.binding.command.startswith("move container")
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @Anon1234 

Just to understand, do you think it's impossible to enhance the current module with this async updating mechanism instead ? It would allow to enhance also existing users experience in a single commit :)

Anon1234:
Current window_title module uses [not maintained ipc library](https://github.com/ziberna/i3-py) which is outdated (last commit was on 20 Jun 2012), buggy ([1](https://github.com/ultrabug/py3status/blob/master/py3status/modules/window_title.py#L12), [2](https://github.com/ziberna/i3-py/issues/19)) and it doesn't support all the events I need ([i3wm events](http://i3wm.org/docs/ipc.html#_available_events), [i3-py events](https://github.com/ziberna/i3-py/blob/master/i3.py#L47)).
So unfortunately my answer is no.

ultrabug:
ok @Anon1234 . I only have two comments then :
* empty_title to 500 white spaces ? I don't get why you do that tbh, an empty string output would sound more reasonable and more in line with the other modules
* it is strongly recommended to use the `kill` magic method when using threads on modules and properly stop the thread there. The kill method will be called on i3 reloads and this ensures proper thread management on modules (+ avoids zombie theads if something behaves badly).

PS : I will have to do some QA on your module, you removed some basic comments and the basic config on the standalone test run. Also, please try to apply a `yapf` formatting on it, else I'll do it np.

ultrabug:
Friendly ping.

If you can't do it this week, I'll permit myself to merge & do the work in order for your module to hit the very upcoming 2.7 release which I plan to release as soon as this PR is taken care of.

Anon1234:
Hi @ultrabug! 
Sorry for the long silence, I was busy with uni stuff. 
I fixed the 500 spaces :smile:

Could you give me an advice on how to implement the kill() method?
I don't have much experience in threading so I thought it's save to have a deamon thread in program. From python docs:
> The entire Python program exits when no alive **non-daemon** threads are left.

 AFAIK [there are no an easy way to kill a thread in Python](http://stackoverflow.com/questions/323972/is-there-any-way-to-kill-a-thread-in-python). Please remember that there are [a blocking event loop running in the thread](https://github.com/Anon1234/py3status/blob/patch-2/py3status/modules/window_title_async.py#L89).

py3status/modules/window_title_async.py
```diff
@@ -0,0 +1,114 @@
...
+
+from threading import Thread
+
+import i3ipc
+
+
+class Py3status:
+
+    format = "{title}"
+    empty_title = " "
```
ultrabug: please use "" instead, this will make i3bar not to show the module completely instead of having to separators separated by an empty space ;)

ultrabug:
Cheers @Anon1234 . I've made a quick comment.

As for the kill, looking at your code indeed you don't need to worry about it sorry. It should be fine as-is :) so only my little comment and we're good to go !

Anon1234:
todo: 
- fix when the title not being cleared after moving the last window to another workspace
- add basic config for standalone run

ultrabug:
cool @Anon1234 thx, waiting for the final todo then !

Anon1234:
Done.

ultrabug:
Thanks a lot @Anon1234 ! please see referenced commit as to what I quickly QA and fixed on the module :) cheers

PR #136, opened 2015-10-29
[battery_level] Uniformize and standardize 'format' option

maximbaz:
@ultrabug, as promised, here's an implementation for the uniform 'format' option in battery_level module. I've also refactored quite a bit and provided previously missing documentation for several options.

It should be fully backwards compatible, and all those compatibility rules are collected in one place for easier future reference.

Feedback is welcome :smiley: 

/link: #98

**UPDATE:** Forgot to mention, there's also one bug fixed while refactoring, so the behavior is *not exactly the same* as before. Previously `battery_full` variable was set to `False` and never updated, which means that `color_good` and `hide_when_full` parameters did nothing. Now this is fixed.

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -92,6 +92,7 @@ class Py3status:
 
         self.refresh_battery_info()
 
+        self.provide_backwards_compatibility()
         self.update_icon()
         self.update_ascii_bar()
         self.update_full_text()
@@ -108,6 +109,18 @@ class Py3status:
                 stdout=open('/dev/null', 'w'),
                 stderr=open('/dev/null', 'w')
             )
+
+    def provide_backwards_compatibility(self):
+        # Backwards compatibility for 'mode' parameter
+        if self.format == FORMAT and self.mode == 'ascii_bar':
+            self.format = "{ascii_bar}"
+
+        # Backwards compatibility for 'show_percent_with_blocks' parameter
+        if self.format == FORMAT and self.show_percent_with_blocks:
+            self.format = "{icon} {percent}%"
+
+        # Backwards compatibility for '{}' option in format string
+        self.format = self.format.replace('{}', '{percent}')
 
     def refresh_battery_info(self):
         # Example acpi raw output: "Battery 0: Discharging, 43%, 00:59:20 remaining"
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -3,17 +3,32 @@
 Display the battery level.
 
 Configuration parameters:
-    - blocks : a string, where each character represents battery level in bar mode
+    - blocks : a string, where each character represents battery level
       especially useful when using icon fonts (e.g. FontAwesome)
       default is "_▁▂▃▄▅▆▇█"
-    - charging_character : a character to represent charging battery in bar mode
+    - charging_character : a character to represent charging battery
       especially useful when using icon fonts (e.g. FontAwesome)
       default is "⚡"
     - color_* : None means - get it from i3status config
-    - format : text with "text" mode. percentage with % replaces {}
+    - format : string that formats the output. Supported options:
+      - '{percent}' : the remaining battery percentage (previously '{}')
+      - '{icon}' : a character representing the battery level,
+                   as defined by the 'blocks' and 'charging_character' parameters
+      - '{ascii_bar}' : a string of ascii characters representing the battery level,
+                        an alternative visualization to '{icon}' option
+      default is "Battery: {percent}"
     - hide_when_full : hide any information when battery is fully charged
-    - mode : for primitive-one-char bar, or "text" for text percentage ouput
-    - show_percent_with_blocks : show battery level percentage in bar mode
+
+Obsolete configuration parameters:
+    - mode : an old way to define 'format' parameter. The current behavior is:
+      - if 'format' is specified, this parameter is completely ignored
+      - if the value is 'ascii_bar', the 'format' is set to "{ascii_bar}"
+      - all other values are ignored
+      - there is no default value for this parameter
+    - show_percent_with_blocks : an old way to define 'format' parameter:
+      - if 'format' is specified, this parameter is completely ignored
+      - if the value is True, the 'format' is set to "{icon} {percent}%"
+      - there is no default value for this parameter
 
 Requires:
     - the 'acpi' command line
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks a lot for this impressive and nice work @maximbaz !

Few remarks :
* all **internal** methods should be prefixed by **_** otherwise py3status will try to execute them all and spam your logs (run with -d, you'll see). This means that all your methods should be prefixed by an underscore except `battery_level` and `on_click`
* when I run the current module by hand, I get `{'cached_until': 1446215804.6845903, 'full_text': '█'}` so are you sure the default output really is `Battery: 100` ? The way I see it, the default mode is `bar`
* please add a default `color_degraded` key on L188 (see on another module for the colour)

maximbaz:
Thanks @ultrabug, nice catches. I have a question though on the second point, could you please elaborate how exactly are you reproducing this?

I've tested the following ways:

1 - Just running the battery_level.py directly: `python py3status/modules/battery_level.py`

The output I get is:
```
{'cached_until': 1446230631.3998244, 'full_text': 'Battery: 72', 'color': '#FCE94F'}
{'cached_until': 1446230632.4028993, 'full_text': 'Battery: 72', 'color': '#FCE94F'}
{'cached_until': 1446230633.4069226, 'full_text': 'Battery: 72', 'color': '#FCE94F'}
{'cached_until': 1446230634.4107358, 'full_text': 'Battery: 72', 'color': '#FCE94F'}
```

2 - Installing it and running `py3status`. My `.i3status.conf` file has no configuration for "battery_level" module, the only occurrence of "battery_level" word is in line:
```
order += "battery_level"
```

I execute `py3status` in console, and here's the output that I get:
```
{"color": "#FCE94F", "instance": "", "name": "battery_level", "cached_until": 1446230150.5564356, "full_text": "Battery: 61"}
{"color": "#FCE94F", "instance": "", "name": "battery_level", "cached_until": 1446230150.5564356, "full_text": "Battery: 61"}
```


Could you maybe provide the related parts of your `.i3config.conf` file?

ultrabug:
@maximbaz I'm not sure about your current py3status version but I just use my local copy of the sources and run `python py3status/modules/battery_level.py`.

Maybe the master version of the battery_level module changed behaviour then ? what version are you running ?

maximbaz:
I'm really surprised. I tried this with both python3 and python2 (see below). This branch is based on the latest `battery_level.py` from the master, so there's no difference since then.

@ultrabug could I please ask you to try again and confirm that it is reproducible on commit `8d72d36` (the latest commit in this PR)? Just want to be sure that we accidentally didn't end up testing different commits.

Any ideas what else we could check? I don't think the `.i3status.conf` file plays any difference then, since you see the wrong behavior when running the module directly.

```
$ git rev-parse HEAD
8d72d363ea72afef3200182087904b30e40bc575

#-----------------------------------

$ python --version
Python 3.5.0

$ python py3status/modules/battery_level.py 
{'full_text': 'Battery: 100', 'color': '#00FF00', 'cached_until': 1446243464.8753762}
{'full_text': 'Battery: 100', 'color': '#00FF00', 'cached_until': 1446243465.879005}
{'full_text': 'Battery: 100', 'color': '#00FF00', 'cached_until': 1446243466.8826337}
{'full_text': 'Battery: 100', 'color': '#00FF00', 'cached_until': 1446243467.8864028}

#-----------------------------------

$ python2 --version
Python 2.7.10

$ python2 py3status/modules/battery_level.py 
{'color': '#00FF00', 'full_text': 'Battery: 100', 'cached_until': 1446243605.292016}
{'color': '#00FF00', 'full_text': 'Battery: 100', 'cached_until': 1446243606.295984}
{'color': '#00FF00', 'full_text': 'Battery: 100', 'cached_until': 1446243607.300028}

#-----------------------------------

$ py3status --version
py3status version 2.6 (python 3.5.0)

$ i3status --version
i3status 2.9 (2015-03-22) © 2008-2012 Michael Stapelberg and contributors
```

ultrabug:
@maximbaz I think we're talking about different repositories here and we both get confused somehow :)

Using **your** fork and the commit you list above, I get the same output as you.

On **your** fork, if you run `git diff 2.6.. py3status/modules/battery_level.py` you'll see that your modifications changed the default behaviour because you now set `mode = None` whereas it was `mode = 'bar'` in the 2.6 release. This explains the change :)

I think we ought to keep `mode = 'bar'` as a default please.

maximbaz:
Of course, you are totally right! Now I understand, we indeed tested different branches :)

Just to be clear, with this change the `mode` parameter becomes obsolete, so I believe it's fine to keep it's default value to `None`. Perhaps in future you will want to do a cleanup and remove deprecated parameters, and I want that cleanup to be harmless. 

But, what's important for us today is to match the default value of the parameter `format` to replicate the output, previously achieved by having default `mode = 'bar'`. What I did wrong is I set default value of `format` as if the default value of `mode` was `'text'`.

I'll ping you when the updated code is pushed.

maximbaz:
@ultrabug please have another look. As I said before, I still kept the `mode = None` as a default in order to avoid dependency on obsolete parameters, but the default behavior now should be exactly the same as in master -- when no parameters are defined, it behaves as `mode = 'bar'`, and it's also possible to set `mode = 'ascii_bar'` or `mode = 'text'`.

py3status/modules/battery_level.py
```diff
@@ -125,13 +111,88 @@ def on_click(self, i3s_output_list, i3s_config, event):
...
+            self.ascii_bar += EMPTY_BLOCK_DISCHARGING * (10 - int(self.percent_charged/10))
+
+    def _update_icon(self):
+        if self.charging:
+            self.icon = self.charging_character
+        else:
+            self.icon = self.blocks[int(math.ceil(self.percent_charged/100*(len(self.blocks) - 1)))]
+
+    def _update_full_text(self):
+        self.full_text = self.format                                          \
```
ultrabug: let's be pythonic till the end, please use `self.full_text = self.format.format(ascii_bar=self.ascii_bar, icon=self.icon, percent=self.percent_charged)` ? :)

ultrabug:
Looks awesome @maximbaz thanks ! just a last *pythonic* comment and we're good to merge :)

maximbaz:
Wow I didn't know (or forgot) that it's possible to use named parameters in `string.format` in python. Thanks for the hint, @ultrabug, it is pushed :+1:

ultrabug:
Awesome thanks @maximbaz !

PR #140, opened 2015-11-18
replace pygame with pyglet

beetleman:
It is easer to install pyglet(pip) than pygame(sources or packages).

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -73,11 +73,11 @@ class Player(object):
 
     def _pyglet(self, sound_fname):
         res_dir, f = os.path.split(
-           sound_fname
+            sound_fname
         )
 
         if res_dir not in pyglet.resource.path:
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @beetleman, thanks for this proposal.

In order to not force current users to switch, would you please rather implement pygame as a fallback from pyglet ? This would allow new users to choose pyglet over pygame while keeping maximum compatibility.

beetleman:
Ok. I'll do it after work.

beetleman:
II have some problem with install pygame. Can You test it with pygame?

ultrabug:
@beetleman unfortunately I don't run pygame either, I'll have a look at how you implemented this, if it looks safe enough, I'll merge ;)

ultrabug:
@beetleman thanks for your patience, I tested it with both and it works great ! There's a little typo on method names which I'll correct after merging.

Cheers !

ultrabug:
All good, thx again !

PR #144, opened 2015-12-03
Fix parsing of taskWarrior's output

Shir0kamii:
I don't know for older versions, but the latest stable release (2.5.1) already
have '[' and ']' when doing an export. This lead to raising an exception
because it makes a nested list and you try to access it using a string
rather than an integer


ultrabug:
Thanks @Shir0kamii ! I trust you tested it alright.

jazmit:
Thanks @Shir0kamii!  I'm the original author of this module, I just upgraded to the latest taskwarrior & really happy that someone already fixed this bug!

jazmit:
Older versions of taskwarrior didn't have the [] around the JSON, really weird.

PR #147, opened 2015-12-08
Eliminate intrusive log messages from pomodoro's module

Shir0kamii:
Hi,

Py3status imports all methods of a module except those starting with an
underscore and special methods. It then call these methods with self,
i3s_output_list and i3s_config.

Because of these rules, py3status call setup_mmss_time and setup_bar
with too many arguments, causing intrusive messages in logs. This PR 
fix this behaviour by adding an underscore before these methods' name.


ultrabug:
Good catch thx !

PR #150, opened 2015-12-16
wwan_status: A network connectivity status indicator for newer Huawei modems

hburg1234:
Hello,

here's my first pull request. I hope you consider it useful because as of now only Huawei modems which understand the AT command I am using are supported. I tried to extend it to the Ericsson F3507g modem, but I could not verify that the AT commands I found actually return what I expect.

When reviewing, please keep an eye on the following things:
 * I am new to Python. If I violated any conventions, just point it out and I'll fix it.
 * I moved the main part of the documentation to the README.md file. Is this what it should be like?

### DIFF
--- a/py3status/modules/wwan_status.py
+++ b/py3status/modules/wwan_status.py
@@ -49,24 +49,24 @@ class Py3status:
 
     def wwan_status(self, i3s_output_list, i3s_config):
 
-        query = "AT^SYSINFOEX" 
-        target_line = "^SYSINFOEX" 
+        query = "AT^SYSINFOEX"
+        target_line = "^SYSINFOEX"
         noipstring = "no ip"
 
         response = {}
         # Check if path exists and is a character device
-        if os.path.exists(self.modem) and stat.S_ISCHR(os.stat(self.modem).st_mode):
+        if os.path.exists(self.modem) and stat.S_ISCHR(os.stat(
+                self.modem).st_mode):
             print("Found modem " + self.modem)
             try:
-                ser = serial.Serial (
-                    port = self.modem,
-                    baudrate = self.baudrate,
+                ser = serial.Serial(
+                    port=self.modem,
+                    baudrate=self.baudrate,
                     # Values below work for my modem. Not sure if
                     # they neccessarily work for all modems
-                    parity = serial.PARITY_ODD,
-                    stopbits = serial.STOPBITS_ONE,
-                    bytesize = serial.EIGHTBITS
-                )
+                    parity=serial.PARITY_ODD,
+                    stopbits=serial.STOPBITS_ONE,
+                    bytesize=serial.EIGHTBITS)
                 if ser.isOpen():
                     ser.close()
                 ser.open()
@@ -85,7 +85,8 @@ class Py3status:
                 PermissionError
                 print("Permission error")
                 response['color'] = i3s_config['color_bad']
-                response['full_text'] = self.prefix + "no access to " + self.modem
+                response[
+                    'full_text'] = self.prefix + "no access to " + self.modem
                 return response
             # Dissect response
             for line in modem_response.decode("utf-8").split('\n'):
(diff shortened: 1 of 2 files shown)


ultrabug:
Wow @hburg1234 what a module and a first contribution, congratz :)

Please bear with me while I take the time to read it. Meanwhile, please make sure :
* that it works for both python2 AND python3
* that you make sure your code follows the basic python syntax conventions, for this I recommend you install and run it through a tool named `yapf`

Thanks again !

hburg1234:
Hey @ultrabug and thank you for your nice response!

I put some more work into the module (see the above commits) and tested it to work both with python2 and python3. It runs fine, provided the dependencies are installed for the respective python version. Therefore I wonder: How do you do dependency management after all? My module requires netifaces and pyserial.

ultrabug:
@hburg1234 thanks.

As for the requirements, they should be listed in the docstring of your module (see the mpd_status one for example)

Also please order alphabetically all your configurable variables, both in code and in doc, thanks !

hburg1234:
@ultrabug Done!

I also made sure that a user who wishes not to display an IP address isn't bothered with the "no ip"-String.

py3status/modules/wwan_status.py
```diff
@@ -0,0 +1,146 @@
+# -*- coding: utf-8 -*-
+"""
+Display current network and ip address for newer Huwei modems. It
```
ultrabug: We miss a one line docstring for the short description please (see the other modules for example)

ultrabug:
Ok that looks great mate, I've added a comment.

The last thing is to change the no-ip parameter with a `format` one where users can choose how to format the output themselves (see lots of modules, example : keyboard_layout).

If this sounds too complex or obscure, I'll do it after merging so you can see, tell me what you think.

hburg1234:
Hey,

as you suggested, I now work with string formatters and let the users choose what to display and how. On top of that, I added the ability to display the network generation (2G/3G/4G) and included a setting to choose the degradation color based on that. See 5c8653107af4cfb039bbcbcc9476574d538a3b01 for details.

The only thing that I am not really comfortable with is the fact that a user must/can define four independent formats for the different situations that might arise.

Other than that, I'd say it works fine. Please let me know if you think I should amend other things.

py3status/modules/wwan_status.py
```diff
@@ -0,0 +1,178 @@
...
+                              show_ip is false, then this settings has no
+                              effect
+    - modem                 : The device to send commands to. Default is
+    - modem_timeout         : The timespan betwenn querying the modem and
+                              collecting the response.
+                              Default is 0.4 (which should be sufficient)
+
+@author Timo Kohorst timo@kohorst-online.com
+PGP: B383 6AE6 6B46 5C45 E594 96AB 89D2 209D DBF3 2BB5
+@license GPLv3
```
ultrabug: whole py3status and modules are BSD, please accept to use it too ?

py3status/modules/wwan_status.py
```diff
@@ -0,0 +1,178 @@
...
+                else:
+                    # Outputs can be multiline, so just try the next one
+                    pass
+        else:
+            print(self.modem + " not found")
+            response['color'] = i3s_config['color_bad']
+            response['full_text'] = self.format_down
+        return response
+
+    def get_ip(self, interface):
```
ultrabug: this should be a private method `_get_ip` else it will be executed by py3status directly too

ultrabug:
I love it, almost there ! see comments, thx

hburg1234:
Hey,

fixed the method.

About the license: Despite my minor contribution to your previous efforts, I consider myself a strong proponent of copyleft. Grepping the README indicates there have been other GPL'ed contributions. Do I and do these other modules "break" something by using GPL, like is there something in particular that you plan to do with the BSD license that cannot be done with GPL?

> $ grep @license README.md | sort | uniq
                         @license BSD
                         @license Eclipse Public License
                         @license GNU GPL http://www.gnu.org/licenses/gpl.html
                         @license GPLv3
                         @license GPLv3 <http://www.gnu.org/licenses/gpl-3.0.txt>
                         @license WTFPL <http://www.wtfpl.net/txt/copying/>

ultrabug:
Those are indeed old modules which needs "fixing". I've been brought to this subject by a debian packager because adding all those LICENSES is quite a pain.

Since I released py3status as BSD initially, I'm enclined to stick with it because changing would mean a lot of work and pings.

That's why from now on I try to enforce a whole BSD license motto. It's open enough for me. Does that bother you much ?

hburg1234:
OK, that should be it. Did some more fine tuning and removed the license so that BSD applies.

ultrabug:
Thanks a lot for this amazing work @hburg1234 ! I wish you some very happy holidays :)

PR #137, opened 2015-11-06
Tweak battery_level

ben-e-whitney:
Changes explained in the commit messages. They fall into the following categories:
* add new parameters for controlling how the battery level is decided to be bad, degraded, or good
* tweak existing code to make it clearer (to me, at least) in ways that do not change functionality
* tweak existing code in ways that (slightly) change functionality!

PR #152, opened 2015-12-20
Notify when battery is running low

maximbaz:
With this simple change it is now possible to receive notifications when battery level becomes low (<30%) or critically low (<10%).

To achieve that, the following configuration parameter must be set in `.i3status.conf`:
```
battery_level {
  ...
  notify_low_level = true
}
```
Notifications have different urgency levels, which can be used by notifications daemon to choose colors, timeouts, etc. For example, my notification daemon is configured to display "urgent" notifications in red, and they will not hide automatically, comparing to "normal" notifications.

selection_013
selection_016

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -29,6 +29,8 @@ Configuration parameters:
       default is False
     - notification : show current battery state as notification on click
       default is False
+    - notify_low_level : display notification when battery is running low.
+      default is False
 
 Format of status string placeholders:
     {ascii_bar} - a string of ascii characters representing the battery level,
@@ -84,6 +86,7 @@ class Py3status:
     format = FORMAT
     hide_when_full = False
     notification = False
+    notify_low_level = False
     # obsolete configuration parameters
     mode = None
     show_percent_with_blocks = None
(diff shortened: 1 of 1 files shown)


ultrabug:
thanks @maximbaz that's a good idea indeed. my only concern is that I found the work of @ben-e-whitney very interesting on PR #137 who introduced thresholds

would you mind having a look at this too ?

maximbaz:
Had a quick look, I think this change won't conflict with the idea of custom thresholds introduced in #137. It is still the same number of "slices" (statuses that battery can be): [0..9% -> bad, 10..29% -> degraded, 30..99% -> good, 100% -> full]. The only difference is that the thresholds become configurable, rather than hardcoded values of 10%, 30% and 100%.

I can see from the discussions that the final outcome/expectation of #137 is finalized yet. I think it is safe to merge this change, but warn @ben-e-whitney to rebase and consider notifications for `degraded` and `bad` states once they find time to finish their PR.

ben-e-whitney:
I will gladly rebase and look through these changes before resubmitting my pull request. Thanks for the heads up.

maximbaz:
So what do you say @ultrabug, should we merge this PR?

ultrabug:
@maximbaz yes, done thanks !

PR #155, opened 2016-01-04
wifi module

tjaartvdwalt:
This module displays some details on your wifi connection. It has 2 modes: ESSID and IP. You can toggle which one is displayed with a mouse click.

Toggling between the modes is very useful for saving space in the status bar, and is not possible (afaik) with the default i3status wifi module.

### DIFF
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -1,0 +1,82 @@
+"""
+Display your wifi essid and IP address.
+This module improves the default wifi module by having two display modes
+that can toggled with a mouse click. This saves space in the status bar.
+
+This module is inspired by the "whatismyip" module.
+
+Requires:
+    - iwlib (https://pypi.python.org/pypi/iwlib)
+
+Configuration parameters:
+    - mode: default mode to display. Can be '{essid}' or '{ip}'
+    - interface: the wireless interface to query (default wlan0)
+
+@author Tjaart van der Walt (github:tjaartvdwalt)
+@license BSD
+"""
+
+import iwlib
+import socket
+from time import time
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 30
+    mode = 'essid'
+    negative_cache_timeout = 2
+    timeout = 5
+
+    def on_click(self, i3s_output_list, i3s_config, event):
+        """
+        Toggle between display modes 'essid' and 'ip'
+        """
+        if self.mode == 'essid':
+            self.mode = 'ip'
+        else:
+            self.mode = 'essid'
+
+    def _get_status(self, interface):
+        interface = iwlib.get_iwconfig(interface)
+        quality = interface['stats']['quality']
+        essid = bytes(interface['ESSID']).decode()
+        return (essid, quality, socket.gethostbyname(socket.gethostname()))
+
+    def wifi(self, i3s_output_list, i3s_config):
+        """
+        """
+        (essid, quality, ip) = self._get_status(self.interface)
+        response = {'cached_until': time() + self.negative_cache_timeout}
+
+        if(self.mode == 'essid'):
+            response['full_text'] = essid
+        else:
+            response['full_text'] = ip
+
+        if(quality < 33):
+            response['color'] = i3s_config['color_bad']
+            
+        if(quality >= 66):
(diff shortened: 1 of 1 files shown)


py3status/modules/wifi.py
```diff
@@ -0,0 +1,84 @@
+"""
+Display the current wifi ESSID and IP address.
+It improves on the default i3status wifi module by allowing the user
```
ultrabug: Add a blank line here

py3status/modules/wifi.py
```diff
@@ -0,0 +1,84 @@
...
+import iwlib
+import socket
+from time import time
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 30
```
ultrabug: cache_timeout is not used in your code

py3status/modules/wifi.py
```diff
@@ -0,0 +1,84 @@
...
+    mode = 'essid'
+    negative_cache_timeout = 2
```
ultrabug: shouldn't it be cache_timeout instead to be in line with the other modules ? I don't see why the negative part

py3status/modules/wifi.py
```diff
@@ -0,0 +1,84 @@
...
+        return response
+
+if __name__ == "__main__":
+    """
+    Test this module by calling it directly.
+    """
+    from time import sleep
+    x = Py3status()
+    config = {
+        'interface': 'wlan0',
```
ultrabug: please, set it as a default with the other configuration parameters under the class definition instead

ultrabug:
Interesting idea @tjaartvdwalt ! I have also the same need :) please see my comments

py3status/modules/wifi.py
tjaartvdwalt: This is sloppy code on my part. I copied the skeleton code from the `whatismyip` module, but I did not pay enough attention.

py3status/modules/wifi.py
tjaartvdwalt: I will fix this in my next commit. 
Is there any guide on what the default `cache_timeout` value should be?
Looking at other modules it seems to range anywhere from 5-30.

py3status/modules/wifi.py
tjaartvdwalt: Will do

py3status/modules/wifi.py
tjaartvdwalt: Will do

pydsigner:
As far as cache_timeout defaults go, that depends greatly on how forgiveable stale information is versus how cheap it is to update. 10-15 seconds is quite fine here I think since getting the IP is cheap and unlikely to change often, though loss of wifi connectivity is something the user will want to know about sooner rather than later.

ultrabug:
Ok, one thing about the code itself :
- maybe we need `format_ip` and `format_essid` parameters, where we have placeholders for users to decide what to show on the toggle ?

ultrabug:
Other than that I gave some more thought about this because I always find it a pity to have to recode an i3status module because it lacks features...

Maybe the right answer to this PR and your feature @tjaartvdwalt is to actually develop a `toggle_group` or `scroll_group` or `module_group` whatever the name on py3status itself where users could **group multiple modules together** and switch between them using the mouse scroll buttons (4 and 5) !

This way users could still benefit from click events on mouse buttons 1, 2 and 3 on each module. My first idea would give something like this on i3status.conf for your use case (I need to use two different module instance _first_ and wlan0 to get them work on the same i3status output) :+1: 

````
wireless _first_ {
	format_down = "W"
	format_up = "%ip"
	scroll_group = 1
}

wireless wlan0 {
	format_down = "W"
	format_up = "%essid"
	scroll_group = 1
}
````

This would make py3status group the two outputs together and display one or the other on mouse scroll automatically.

What's interesting with this is that this would work with every module available and users could then group and *toggle* whatever modules they want together with a single line config.

If you and @pydsigner find this idea worth developing, I'll open a special RFC issue for it.

tjaartvdwalt:
I agree that this is a better solution!

I will try to find some time over the weekend to look into the py3status code.

ultrabug:
opened #178, closing here if that's ok. cheers

PR #156, opened 2016-01-10
Support woeid, add more configuration parameters

rail:
* add woeid support, see https://developer.yahoo.com/geo/geoplanet/guide/concepts.html#woeids
  * configurable temperature units
  * configurable icons
  * add formats for today condition and forecasts
  * use raise_for_status() instead of raising a generic exception


rail:
My config:
```
weather_yahoo toronto {
    woeid = "91978064"
    forecast_days = 2
    icon_snow = "❄"
    icon_rain = "☔"
    today_format = "{icon}{temp}{units}°  "
    forecast_format = "{day}: {icon}{high}/{low} "
}
```
which gives me something like http://i.imgur.com/snagJHr.png

py3status/modules/weather_yahoo.py
```diff
@@ -8,11 +8,29 @@
...
+
 Configuration parameters:
     - cache_timeout : how often to check for new forecasts
     - city_code : city code to use
+    - woeid : use Yahoo woeid (extended location) instead of city_code
     - forecast_days : how many forecast days you want shown
     - request_timeout : check timeout
+    - units : Celsius (C) of Fahrenheit (F)
+    - today_format : possibple placeholders: {icon}, {temp}, {units}, {text}
+        example:
```
pydsigner: I would suggest changing this example to °{units} and °C so that it is correct. Just a minor nitpick.

py3status/modules/weather_yahoo.py
```diff
@@ -8,11 +8,29 @@
...
+Find your woeid using:
+    http://woeid.rosselliot.co.nz/
+
 Configuration parameters:
     - cache_timeout : how often to check for new forecasts
     - city_code : city code to use
+    - woeid : use Yahoo woeid (extended location) instead of city_code
     - forecast_days : how many forecast days you want shown
     - request_timeout : check timeout
+    - units : Celsius (C) of Fahrenheit (F)
```
pydsigner: "of" → "or"

py3status/modules/weather_yahoo.py
```diff
@@ -8,11 +8,29 @@
...
+    http://woeid.rosselliot.co.nz/
+
 Configuration parameters:
     - cache_timeout : how often to check for new forecasts
     - city_code : city code to use
+    - woeid : use Yahoo woeid (extended location) instead of city_code
     - forecast_days : how many forecast days you want shown
     - request_timeout : check timeout
+    - units : Celsius (C) or Fahrenheit (F)
+    - today_format : possibple placeholders: {icon}, {temp}, {units}, {text}
```
ultrabug: possible ;)

ultrabug:
Thanks for this awesome PR @rail ! I didn't know about woeid.

If I'm not mistaken, your modification breaks the default output from the module right ? please make it so it doesn't.

Also, please sort your config variables declaration and doc alphabetically.

Last but not least, I think it may be good to implement a `format` option maybe defaulting to `{today_format} {forecast_format}` as a compliance to #98, maybe you'll have an implementation for it ?

Thanks again

rail:
No problem.

> If I'm not mistaken, your modification breaks the default output from the module right ? please make it so it doesn't.

I think I tried to keep it the same, but I'm not sure now. :) Let me double check that.

> Also, please sort your config variables declaration and doc alphabetically.

Roger.

> Last but not least, I think it may be good to implement a format option maybe defaulting to {today_format} {forecast_format} as a compliance to #98, maybe you'll have an implementation for it ?

Yeah, I thought about that to get rid of `strip()`, good idea.

pydsigner:
@ultrabug I saw the dual format part and paused for a bit but didn't think of any good way to make it happen. Your solution sounds pretty good to me.

rail:
Something like this probably...
I thought about simplifying configs, but don't see anything that I can drop without introducing some kind of template engine. :P

pydsigner:
In the end, weather is a really complex thing to be sticking in a simple format like i3bar's, and this PR adds a lot of power to manage that complexity even if it isn't as neat as we might wish as programmers and users.

py3status/modules/weather_yahoo.py
```diff
@@ -8,11 +8,34 @@
...
+Find your woeid using:
+    http://woeid.rosselliot.co.nz/
+
 Configuration parameters:
     - cache_timeout : how often to check for new forecasts
     - city_code : city code to use
     - forecast_days : how many forecast days you want shown
+    - forecast_format : possible placeholders: {icon}, {low}, {high}, {units},
+        {text}
+    - forecast_include_today: show today forecast as well, default false
```
ultrabug: since users can decide on `format` now, I think this parameter is useless right ?

ultrabug:
I agree with you both and I like very much this PR and the flexibility it offers. I just made a last little comment for a (I think) useless parameter.

Thanks again

py3status/modules/weather_yahoo.py
rail: Not really... If `forecast_include_today` is `True`, you get today in the list of forecast, so you still can have in the both today and forecast. In other words you have your current conditions and forecast for today.

rail: Maybe this option deserves better description...

ultrabug: well I don't have the API JSON in mind so I guess I'll trust you on the fact that this makes sense :) mind updating the description about it a little more plz ?

rail:
Something like this maybe?

ultrabug:
Yes @rail the description makes sense to me now thanks.

As for code style QA, I always pass the code through `yapf` so maybe you can do it or I will after the merge.

I just need a little time to test the module now and will merge :) plz bear with me

ultrabug:
Ok @rail I tested the module successfully ! One last remark I promise : I think we should rename "today_format" and "forecasts_format" to `format_` prefixed variables since it's the standard i3status way of doing (and I did this exact mistaken expectation when testing the module)

So please make them `format_today` and `format_forecasts`, then we're good to go :)

rail:
The easiest thing ever! :)

ultrabug:
Thanks a lot for this awesome enhancement @rail and @pydsigner for your help & review !

rail:
No problem. Thank you both for great reviews!

PR #157, opened 2016-01-10
Add dual battery support to battery_level

4iar:
I've added dual battery support to the battery_level module.
The batteries can now either be displayed individually and selected through the battery_id parameter or combined into a single display (so % charge represents all batteries).

The on_click function didn't seem to work because self.time_remaining wasn't provided anywhere, so I wrote some code to extract time remaining from the acpi output.

This is my first code pull request and any suggestions on how to improve or simplify the code are super welcome.

Cheers.

### DIFF
--- a/py3status/modules/README.md
+++ b/py3status/modules/README.md
@@ -29,9 +29,9 @@ Available modules:
   battery_level          Display the battery level.
                          
                          Configuration parameters:
-                             - battery_id : id of the battery to be displayed which
-                               can be found from acpi (e.g. "Battery 0: ..." where 0 is the id).
-                               Default is 0.
+                             - battery_id : id of the battery to be displayed
+                               set to 'all' for combined display of all batteries
+                               default is 0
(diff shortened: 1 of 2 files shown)


ultrabug:
Wow that's a tough first PR @4iar congratz :)

Please bear with me while I take the correct time to review and test it.

Quick question : did you test it with both python2 and python3 ?

Cheers

4iar:
Hey,

Yeah it works fine under both Python 2 and 3.

Thanks for taking a look

ultrabug:
Neat and clean work @4iar thanks ! All tested OK for me.

Friendly ping @maximbaz to test it too if you can spare the time.

maximbaz:
Yup, works for me as well, thanks again @4iar! 

I have only one battery, so my tests are pretty limited, but for me setting both `battery_id = 0` and `battery_id = 'all'` behaves as expected (as previously).

Issue #159, opened 2016-01-12
icinga2 module


nazco:
I've written an icinga2 module for py3status. Would you like a PR? https://github.com/nazco/i3status-modules

pydsigner:
I think it looks really useful, although I don't use icinga myself.

ultrabug:
sure @nazco ! any added module will help all the users :) thanks !

closing here, FU on #160

PR #160, opened 2016-01-13
add icinga2 status module to the default module folder

nazco:
my PR like referenced here #159


py3status/modules/icinga.py
```diff
@@ -0,0 +1,85 @@
...
+    def ok(self, i3s_output_list, i3s_config):
+        format = 'OK: {ok}'
+        response = {
+            'color': '#44bb77',
+            'cached_until': time() + self.cache_timeout,
+            'full_text': format.format(ok=self._queryServiceCount(Status.OK))
+        }
+        return response
+
+    def _queryServiceCount(self, state):
```
pydsigner: Should be named _query_service_count() for PEP8 compliance.

py3status/modules/icinga.py
```diff
@@ -0,0 +1,85 @@
...
+        if self.disable_acknowledge:
+            self.url_parameters = self.url_parameters + "&service_handled=0"
+        result = requests.get(
+            self.base_url + self.url_parameters.format(service_state=state.value),
+            auth=(self.user, self.password), verify=self.ca)
+        return len(result.json())
+
+if __name__ == "__main__":
+    pass
```
pydsigner: You can look at other existing modules to see what sort of test is supposed to be here for inclusion as an official module.

pydsigner:
How this module is structured, with the four functions returning the various statuses, is a nice concept for reducing configuration file bloat. But it does add redundancy and reduce how configurable it can be. If the status type to be used was set in the config rather than being hard-coded in, a user would immediately be able to reorder the statuses. Plus, format and coloring could be made sanely customizable.

py3status/modules/icinga.py
nazco: Sorry, but such a test wouldn't work here because you need a running icinga2 instance with the correct credentials.

pydsigner: Note that these aren't run as unit tests; compare with the aws_bill module.

nazco: My problem with this is that I see no sense in a test that never works because of missing user specific variables.

nazco:
I've updated the module accordingly to your comments. The test at the end is missing because I can't see any sense in a test that is guaranteed to fail. The problem are the user specific icinga installation and the also specific authentication data. If you tell me how to implement that test to your contentment I'll do so.

pydsigner:
This looks a lot better. I will point out that `format` should 1) have a reasonable default and 2) be processed with `.format(key=val)` rather than `% (val)`. In order to have a reasonable default, I'd suggest having a constant dictionary that maps the status ints to strings and adding something like `status_name=STATUS_NAMES.get(status, status)` to the `format.format()` call. You could then have a default `format = '{status_name}: {count}'` or something similar.

nazco:
Well I know that % (val) is legacy and  first I had implemented it with .format(key=val) but than I though it would be great if there is only one type of format strings in the i3status.conf.

pydsigner:
Refer to https://github.com/ultrabug/py3status/issues/98 for the discussion/accepted policy on format strings. I know that there are some modules that have the mod strings but that behavior is intended to change (in a backwards-compatible way).

nazco:
Oh ok, great! I've fixed the formatting stuff as I had it before and added your suggestion for the format string.

ultrabug:
I'll do my best to review it up soon, but I already can ask : are you forced to specify the type of alert on the API call ? Isn't there a way to call the API once and get all the statuses count directly ? Then we would only need to filter it out based on the user's chosen format.

nazco:
Well, at the beginning it was the easiest way but as I now learn how to use py3status I can update it again to support a better experience as you mention above.

nazco:
Yes I could query all services together and iterate over them to read there status. But the problem is the output. As far as I know you can set the color only on the whole output not on basis of a specific part of the output and this is a important feature for this plugin. That's the reason why I query every status on it's own (current state of the PR).

 I've also build a version (First PR try or [here](https://github.com/nazco/i3status-modules/blob/master/icinga.py)) which queries the status of all services in one plugin, but this leads to a bad formatting experience for the user.

pydsigner:
@nazco By the way, if you want to attribute a commit to someone else, you can specify an author for commits other than the committer. It's a handy thing every once in a while.

py3status/modules/icinga_simple.py
```diff
@@ -0,0 +1,64 @@
...
+    - base_url: the base url to the icinga-web2 services list
+    - disable_acknowledge: enable or disable counting of acknowledged service problems
+    - user: username to authenticate against the icinga-web2 interface
+    - password: password to authenticate against the icinga-web2 interface
+    - format: define a format string like "CRITICAL: %d"
+    - color: define a color for the output
+    - status: set the status you want to optain (0=OK,1=WARNING,2=CRITICAL,3=UNKNOWN)
+
+@author Ben Oswald <ben.oswald@root-space.de>
+@license MIT License <https://opensource.org/licenses/MIT>
```
ultrabug: please, accept to release your code as BSD license ? (like other modules)

ultrabug:
sorry it took so long @nazco , the module is great as is. Would you just plz accept to release it under BSD license ? thx

nazco:
releasing it under the BSD license is ok

ultrabug:
@nazco then plz update ur PR ;) thx

nazco:
done :)

ultrabug:
Thanks a lot for your contribution and patience @nazco !

Issue #166, opened 2016-01-20
spotify module generates errors in /var/log/messages


UmBsublime:
Errors in my /var/log/messages: py3status: 
```
user method getText failed (getText() takes exactly 2 arguments (3 given))
```
Here is my config for the module

```
order += "spotify"

spotify {
  format = "{title} by {artist} -> {time}"
}

```

And my py3status version

```
py3status --version
py3status version 2.8 (python 2.7.9)
```

I'm not sure, but I think this is because the method getText should be renamed _getText. 

When doing this locally the error dispersals.

pydsigner:
Good catch.

UmBsublime:
Awesome !

PR #167, opened 2016-01-20
Fix issue #166

pydsigner:


ultrabug:
Thanks @pydsigner and @UmBsublime

PR #170, opened 2016-01-25
Workaround exception handling for PlaybackStatus

AnwariasEu:
Return artist and song despite timeout while retrieving PlaybackStatus,
to work around a bug know in the spotify linux client.
Song information is delayed by the dbus timeout (some seconds).

### DIFF
--- a/py3status/modules/spotify.py
+++ b/py3status/modules/spotify.py
@@ -59,23 +59,32 @@ class Py3status:
             microtime = metadata.get('mpris:length')
             rtime = str(timedelta(microseconds=microtime))
             title = metadata.get('xesam:title')
-
+        except Exception:
+            return (
+                'Spotify not running',
+                self.color_offline or i3s_config['color_bad'])
+        try:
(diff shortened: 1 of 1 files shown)


py3status/modules/spotify.py
```diff
@@ -59,23 +59,32 @@ def _get_text(self, i3s_config):
...
                                    album=album,
-                                   time=rtime), color)
-        except Exception:
-            return (
-                'Spotify not running',
-                self.color_offline or i3s_config['color_bad'])
+                                   time=rtime),
+                                   self.color_playing or i3s_config['color_good'])
+        else:
+            return ('Unhandled Error while retrieving PlaybackStatus', self.color_degraded)
```
pydsigner: Try to only use parentheses around return values when they're needed.

ultrabug:
I'm confused mates as to which workaround PR is the best and won't affect existing users ?

ultrabug:
Addressed by #170

PR #171, opened 2016-01-25
Workaround, removed PlaybackStatus query

AnwariasEu:
A bug in the spotify linux client makes dbus run into a timout while
retrieving the PlaybackStatus. Song information are updated immediately.


ultrabug:
dunno which to merge, this or #170 ?

AnwariasEu:
The other workaround keeps the functionality in case playbackstatus is fixed by spotify but until then the app information is delayed by ~3 seconds. My suggestions would be to merge #170 and revert it/reiimplement when the spotify bug is fixed.

ultrabug:
As @Andrwe is also proposing this on #184 , I'll merge yours to retain the credits, thanks @AnwariasEu

PR #176, opened 2016-01-30
Universal module options

cornerman:
Introduce universal module options like [i3status](https://i3wm.org/i3status/manpage.html#_universal_module_options), which can be used in each module block. Some user modules already implemented this functionality themselves, but i think it is preferable to handle this once for all modules.

The first commits adds the two options ```min_width``` and ```align```, like with i3status. The second commit additionally adds ```separator``` and ```separator_block_width```, which are defined in the [i3bar protocol](https://i3wm.org/docs/i3bar-protocol.html#_blocks_in_detail).

I have updated the user modules accordingly, but two modules (bluetooth and nvidia-temp) already made use of the ```separator``` property for another purpose. So I renamed these two properties, which might break backwards compatibility. What do you think?

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -106,6 +106,20 @@ class Module(Thread):
         if 'min_width' in mod_config:
             self.module_options['min_width'] = mod_config['min_width']
 
+        if 'separator' in mod_config:
+            separator = mod_config['separator']
+            if not isinstance(separator, bool):
+                raise TypeError("invalid 'separator' attribute, should be a bool")
+
+            self.module_options['separator'] = separator
+
+        if 'separator_block_width' in mod_config:
+            separator_block_width = mod_config['separator_block_width']
+            if not isinstance(separator_block_width, int):
+                raise TypeError("invalid 'separator_block_width' attribute, should be an int")
+
+            self.module_options['separator_block_width'] = separator_block_width
+
         if 'align' in mod_config:
             align = mod_config['align']
             if not (isinstance(align, str) and align.lower() in ("left", "center", "right")):
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -43,6 +43,7 @@ class Module(Thread):
         # if the module is part of a group then we want to store it's name
         self.group = self.i3status_thread.config.get(module, {}).get('.group')
         #
+        self.set_module_options(module)
         self.load_methods(module, user_modules)
 
     @staticmethod
@@ -93,6 +94,24 @@ class Module(Thread):
         for method in self.methods.values():
             output.append(method['last_output'])
         return output
+
+    def set_module_options(self, module):
+        """
+        Set universal module options to be interpreted by i3bar
+        https://i3wm.org/i3status/manpage.html#_universal_module_options
+        """
+        self.module_options = {}
+        mod_config = self.i3status_thread.config.get(module, {})
+
+        if 'min_width' in mod_config:
+            self.module_options['min_width'] = mod_config['min_width']
+
+        if 'align' in mod_config:
+            align = mod_config['align']
+            if not (isinstance(align, str) and align.lower() in ("left", "center", "right")):
+                raise ValueError("invalid 'align' attribute, valid values are: left, center, right")
+
+            self.module_options['align'] = align
 
     def load_methods(self, module, user_modules):
         """
@@ -233,6 +252,9 @@ class Module(Thread):
                     else:
                         result['instance'] = self.module_inst
                         result['name'] = self.module_name
+
+                    # set universal module options in result
+                    result.update(self.module_options)
 
                     # initialize method object
                     if my_method['name'] is None:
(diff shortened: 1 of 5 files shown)


pydsigner:
Backwards incompatibility tends to be bad, but I'll take a look at the whole of it.

cornerman:
The ```separator``` and ```separator_block_width``` properties are now also supported by i3status as universal module options (https://github.com/i3/i3status/pull/105).

ultrabug:
I like the idea very much, please bear with me while I take the time to read through your code PR.

cornerman:
Sure, take your time.

Right now the four module options will also be set on the module instances (in [load_methods](https://github.com/ultrabug/py3status/blob/master/py3status/__init__.py#L1005)). I am not sure whether this is a good idea, because it makes it seem like they were to be interpreted by the module.

ultrabug:
ping @cornerman @tobes : I think it's time this one goes in as it fits well with the recent developments, don't you agree ?

tobes:
I think it'd be good idea, definitely better than being added to every module.

As far as the backwards compatibility issue with `separator` goes I think in this instance I'd be inclined to break things.  The worst case scenario is that someones bar would look a little freaky.  But otherwise we have to use a different name for all and that breaks compatibility with i3status.

Saying that, I don't like breaking things but I'm not sure how we can avoid this.  The only option I can see would be to add this as a general option defaulted off.  And warn any users using separator in their config if not switched on.  Then later defaulting to on.  But that just makes it less usable for all users.  It also just feels too messy.

We could also add `background` too now

cornerman:
Sounds good to me and I also think that backwards incompatibility is inevitable in this case.

I can rebase it to the current master and add the background property tomorrow :)

ultrabug:
I agree that this is a reasonable and acceptable *breakage*, let's do this :)

cornerman:
Ok, I rebased this branch, it should be ready to be merged.

Though, I did not add the ```background``` property yet, because i3status does not support it as a universal module option, as of now. Probably would be best to keep these options in sync with i3status.

py3status/module.py
```diff
@@ -219,6 +252,9 @@ def run(self):
                         result['instance'] = self.module_inst
                         result['name'] = self.module_name
 
+                    # set universal module options in result
+                    result.update(self.module_options)
```
tobes: So we just clobber anything set by the module?

I'm not necessarily opposed to this but good to be sure this is what we want.

cornerman: Imho, the universal options should override anything set by the module, because this is what the user specified and therefore is what he requested. Can you think of a usecase where this might be undesirable?

It might be worth to mention it in the docs!

ultrabug: I'm okay with this too.

ultrabug: > It might be worth to mention it in the docs! 

please, add it in your MR ?

ultrabug: > please, add it in your MR ?

scratch that, I guess it should be on the wiki right...

cornerman: Can you add this to the wiki or should I make a PR on the wiki (is this even possible)?

pydsigner: How then do you override your global options for an individual module?

cornerman: They are not global, you can specify them for each module individually. For example:
```
time {
    ...
    separator = false
}
```

pydsigner: @cornerman Ok, I'm clearly not paying enough attention. Carry on.

tobes:
> I did not add the background property yet...

Ok, that sounds reasonable.  We can add it and also `markup` in a later patch.  It needs a bit more work as we'd have to remove the options to pass it to i3status and then add them to the output.

This PR is quite clean as it is so let's keep it simple.

tobes:
`separator` and `separator_block_width` do seem to be in i3status master but not in the current 2.10 release and I'm still on 2.9.

So maybe we still need to be a little clever with this branch.

cornerman:
@tobes yeah, I opened the PR after the release of 2.10, so they are not in the official release yet. I do not think that this is a big problem. The options work in all py3modules and will start working in i3status modules in the hopefully not so far future - or if you checkout the git version.

About the ```markup``` option, isn't that handled as a global option in the ```general``` section?

tobes:
> About the markup option, isn't that handled as a global option in the general section?

The documentation at https://i3wm.org/i3status/manpage.html isn't that clear on the issue, but seeing as you've added `separator_block_width` you are much better informed than I am.

I might have a play at some point

cornerman:
You are right, this is not entirely clear in the docs. In i3status you set the markup globally and not per module. Though, the option will be send in each block to i3bar. For example for pango:

```
general {
    ...
    markup = "pango"
}
```

ultrabug:
Well I'm sorry again @cornerman but you need to rebase, then this PR is the next in line to get in !

cornerman:
Done!

But the current master has a regression of https://github.com/ultrabug/py3status/issues/177

ultrabug:
This can be worked out later ok, thanks @cornerman !

PS: would you please try to see if the wiki can be updated wrt these universal options ? maybe it works like PRs

cornerman:
Added some documentation for the universal module options in this wiki fork: https://github.com/cornerman/py3status-wiki

PR #180, opened 2016-02-16
Added module to display WiFi status using iw.

Markus00000:
The reason for this module is that it works for some cards (at least `iwlwifi`) that do not work with the default wireless status in i3status. In addition, this module is more configurable.

i3status issue: https://github.com/i3/i3status/issues/106

### DIFF
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -44,8 +44,8 @@ class Py3status:
     cache_timeout = 5
     device = 'wlan0'
     down_color = 'bad'
-    signal_percent_bad = 29
-    signal_percent_degraded = 49
+    signal_bad = 29
+    signal_degraded = 49
     bitrate_bad = 26
     bitrate_degraded = 53
     round_bitrate = True
@@ -56,9 +56,8 @@ class Py3status:
         """
         Get WiFi status using iw.
         """
-        self.signal_dbm_bad = self._percent_to_dbm(self.signal_percent_bad)
-        self.signal_dbm_degraded = \
-            self._percent_to_dbm(self.signal_percent_degraded)
+        self.signal_dbm_bad = self._percent_to_dbm(self.signal_bad)
+        self.signal_dbm_degraded = self._percent_to_dbm(self.signal_degraded)
 
         iw = subprocess.check_output(['iw', self.device, 'link']).decode(
             'utf-8')
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -5,7 +5,7 @@ Display WiFi bit rate, quality, signal and SSID using iw.
 Configuration parameters:
     - bitrate_bad : Bad bit rate in Mbit/s (default: 26)
     - bitrate_degraded : Degraded bit rate in Mbit/s (default: 53)
-    - cache_timeout : Update interval in seconds (default: 5)
+    - cache_timeout : Update interval in seconds (default: 10)
     - device : Wireless device name (default: "wlan0")
     - down_color : Output color when disconnected, possible values:
       "good", "degraded", "bad" (default: "bad")
@@ -41,7 +41,7 @@ class Py3status:
     # available configuration parameters
     bitrate_bad = 26
     bitrate_degraded = 53
-    cache_timeout = 5
+    cache_timeout = 10
     device = 'wlan0'
     down_color = 'bad'
     format_down = 'W: down'
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -3,26 +3,24 @@
 Display WiFi bit rate, quality, signal and SSID using iw.
 
 Configuration parameters:
+    - bitrate_bad : Bad bit rate in Mbit/s (default: 26)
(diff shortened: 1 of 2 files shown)


ultrabug:
This reminds me of PR #155 ! Mind you read it and tell me what you feel about it @Markus00000 ?

Friendly ping @tjaartvdwalt

Markus00000:
Regarding the other WiFi module, I had actually seen it but decided to write another one for several reasons:

* It seems to use `iwconfig`, which shows a stuck, wrong bit rate in some cases ([as described here](https://bugs.launchpad.net/ubuntu/+source/wireless-tools/+bug/1492180)) just like the module from i3status. As far as I understand `iwconfig` [is deprecated](https://wireless.wiki.kernel.org/en/users/Documentation/iw) anyway.
* I think showing the IP would better fit into a dedicated module that could support more than just the wireless device.

Showing the SSID makes sense and as it is already in the output of `iw`, I will add it as an option.

I have no opinion on the grouping features as I have no need to group modules.

py3status/modules/wifi.py
```diff
@@ -0,0 +1,140 @@
...
+
+import re
+import subprocess
+from time import time
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
```
ultrabug: please sort the config params alphabetically (and please update the docstring respectfully)

py3status/modules/wifi.py
```diff
@@ -0,0 +1,140 @@
...
+    cache_timeout = 5
```
ultrabug: isn't running this command every 5 seconds a bit overkill ? wouldn't say 10s be satisfying enough ?

ultrabug:
Also, you may want to propose a sudo option. On some systems, the iw command is only accessible to root !

Markus00000:
Thanks for your comments. I have addressed them all.

pydsigner:
That sudo call is just going to hang I think. You'll need a dialog probably to get the password.

Markus00000:
- use_sudo : Use sudo to run iw, make sure it requires no password

It is not going to hang if `iw` requires no password.

Is typing their password twice, once to log in to their computer, once to authenticate their status bar, something users want to do? If so, it might be possible to only ask for a password when required.

Can anyone name any drawbacks of making link status accessible without a password?

Either way the sudoers entry should be documented:

    - use_sudo : Use sudo to run iw, make sure iw requires no password by
      adding a sudoers entry like
      "<username> ALL=(ALL) NOPASSWD: /usr/bin/iw dev wl* link"

ultrabug:
OR rename the current use_sudo to simple `sudo` true/false and add another option `sudo_cmd` defaulting to `sudo` which could be overwritten by `gksudo` for example (which would prompt for password on a gtk popup) ? :)

pydsigner:
@ultrabug That's more of what I was thinking.

Markus00000:
I am not familiar with gksudo. Any suggestions how to implement this?

When I run `gksudo` instead of `sudo`, an invisible password prompt will block any user interaction with i3.

If you want to give it a try, change line 67 to:

    cmd.insert(0, 'gksudo')

Make sure you are able to kill the prompt if necessary.

pydsigner:
@Markus00000 Invisible?

Markus00000:
@pydsigner Yes, invisible. I see the gksudo process (in a TTY) but cannot see the prompt and cannot do anything in i3. This only happens after invoking gksudo from the module running in py3status.

Does it work for you?

ultrabug:
@Markus00000 it works for me, but it keeps on asking me the password at each cache_timeout rate so this is not usable :( You might want to add the **-g** flag for testing.

So I guess this only should be done using sudo as you proposed, thanks !

ultrabug:
Also, FYI my testing shows no signal_percent and a quite not so good looking output : `W: 64.0 MBit/s None% 1M PV`

I guess you might want to strip the whole placeholder in this case and only output `W: 64.0 MBit/s 1M PV` ?

FYI, my command output is
```
Connected to 00:00:00:00:00:00 (on wlp58s0)
	SSID: 1M PV
	freq: 2412
	RX: 2777669034 bytes (4642109 packets)
	TX: 1444261493 bytes (3674469 packets)
	tx bitrate: 64.0 MBit/s

	bss flags:	short-preamble short-slot-time
	dtim period:	1
	beacon int:	100
```

Markus00000:
@ultrabug Thanks for testing and bringing this issue up!

Do we really want to silently strip placeholders that the user set either directly in their configuration or indirectly by accepting the defaults? If we did, I would expect to see an issue sooner or later: “Signal placeholder does not work”

In addition:

    # The format
    'W: {bitrate} {signal_percent} {ssid}'
    # would become
    'W: {bitrate}  {ssid}'

Should we just remove one space on the right of an unknown value? Would this reliably lead to the format the user wants?

Alternatively, the output could inform the user that a requested value is unknown. For example:

    W: 64.0 MBit/s [?%] 1M PV
    W: 64.0 MBit/s [? dBm] 1M PV

What do you think?

ultrabug:
The latter with interrogation mark looks the best solution you're right

ultrabug:
Awesome thanks @Markus00000 !

PR #181, opened 2016-02-20
Adds 'arch_updates' module

lost-tourist:
This module will display a simple count of how many official Arch Linux
updates are waiting to be installed via the 'pacman' utility.  It can
optionally also display how many unofficial (AUR) updates are waiting to
be installed, this functionality requires the 'cower' command line tool
to be installed on the user's system.


lost-tourist:
Sorry, this needs some more work. Will update and resubmit the PR!

PR #182, opened 2016-02-20
added format_down option

Andrwe:
Added option format_down which will be shown when spotify is not running.
This enables the user to define an empty string thus the element won't be shown in i3bar.

Currently there are not format strings supported as there are no useful data.


ultrabug:
Thanks a lot @Andrwe !

PR #185, opened 2016-02-22
added option hide_seconds

Andrwe:
For me and maybe some other users the seconds a battery has left do not matter thus they can be removed.
Thus I added the option hide_seconds to control the visibility of seconds.
It defaults to False. If set to True seconds won't be shown.


ultrabug:
Sane idea @Andrwe thanks !

PR #187, opened 2016-02-23
Arch status module

lost-tourist:
Adds a module which displays the number of pending updates waiting to be applied/installed on an Arch Linux installation.  As Arch is a rolling-update distribution there are frequently many updates released over any 24-hour period so it's convenient for Arch i3-users (and there are many of them!) to have an easy way to see if they should be downloading and installing updates.

### DIFF
--- a/py3status/modules/arch_updates.py
+++ b/py3status/modules/arch_updates.py
@@ -18,7 +18,7 @@ Configuration parameters:
 """
 
 from time import time
-import sh
+import subprocess
 
 
 class Py3status:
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks a lot @lost-tourist, can you confirm you tested your module with both python 2 and 3 ?

lost-tourist:
Good call.  Arch Linux defaults to Python 3 but it's possible there could be people out there who have stuck with 2 for some reason.  Anyway it wasn't quite working with Python 2 after you prompted me to test it, so there's a new commit about to come in!

ultrabug:
okay @lost-tourist thanks, would you mind trying to make sure your code is flake8 compliant ? (see travis-ci failing there). The `yapf` tool can help you if needed.

Andrwe:
IMHO it would be nice if this module implemented the format option and provide a placeholder for the output of pacman and AUR as users might want to change the text shown in the bar.

This way it would already comply with #98.

ultrabug:
Thanks @Andrwe for pointing #98 out, good call.

And many thanks @lost-tourist for your module contribution :)

Friendly ping to @btall, you might be interested in this (for your home machine ofc) ;)

btall:
Interesting, thx :)

PR #188, opened 2016-02-27
deadbeef module

mrt-prodz:
Show current track playing in deadbeef, available parameters are:
    {artist} : artist
    {title} : title
    {elapsed} : elapsed time
    {length} : total length
    {year} : year
    {tracknum} : track number

### DIFF
--- a/py3status/modules/deadbeef.py
+++ b/py3status/modules/deadbeef.py
@@ -58,23 +58,18 @@ class Py3status:
             # split properties using special delimiter
             parts = status.split(self.delimiter)
             if len(parts) == 6:
-                self.artist = parts[0]
-                self.title = parts[1]
-                self.length = parts[2]
-                self.elapsed = parts[3]
-                self.year = parts[4]
-                self.tracknum = parts[5]
+                artist, title, length, elapsed, year, tracknum = parts
             else:
                 return self._error_response(i3s_config['color_bad'])
 
             response = {
                 'cached_until': time() + self.cache_timeout,
-                'full_text': self.format.format(artist=self.artist,
-                                                title=self.title,
-                                                length=self.length,
-                                                elapsed=self.elapsed,
-                                                year=self.year,
-                                                tracknum=self.tracknum)
+                'full_text': self.format.format(artist=artist,
+                                                title=title,
+                                                length=length,
+                                                elapsed=elapsed,
+                                                year=year,
+                                                tracknum=tracknum)
             }
             return response
         except:
--- a/py3status/modules/deadbeef.py
+++ b/py3status/modules/deadbeef.py
@@ -38,21 +38,16 @@ class Py3status:
 
     # return track currently playing in deadbeef
     def get_status(self, i3s_output_list, i3s_config):
-        self.artist = ''
-        self.title = ''
-        self.length = ''
-        self.elapsed = ''
-        self.year = ''
         try:
             # get all properties using ¥ as delimiter
             status = check_output(['deadbeef',
                                    '--nowplaying',
-                                   '%a' + self.delimiter +
-                                   '%t' + self.delimiter +
-                                   '%l' + self.delimiter +
-                                   '%e' + self.delimiter +
-                                   '%y' + self.delimiter +
-                                   '%n'])
+                                   self.delimiter.join(['%a',
+                                                        '%t',
+                                                        '%l',
+                                                        '%e',
+                                                        '%y',
+                                                        '%n'])])
             # check if we have music currently  playing
             if 'nothing' in status:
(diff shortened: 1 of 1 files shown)


pydsigner:
Should use `{'key': 'value'}` style whitespace in dicts in accordance with PEP 8.

mrt-prodz:
8ae90f6 should fix syntax/coding style issue. (sorry about that)

pydsigner:
Only other suggestion would be to use \0 (Null byte) as the delimiter instead of a random uncommon character.

mrt-prodz:
Null byte is messing with the command/parsing, I could change the symbol to another character if needed. I want to avoid having issues with the parsing if the delimiter is in the title/artist. Any suggestion?

py3status/modules/deadbeef.py
```diff
@@ -0,0 +1,98 @@
...
+            # check if we have music currently  playing
+            if 'nothing' in status:
+                response = {
+                    'cached_until': time() + self.cache_timeout,
+                    'full_text': ''
+                }
+                return response
+            # split properties using special delimiter
+            parts = status.split(self.delimiter)
+            if len(parts) == 6:
```
pydsigner: try
```python
artist, title, length, elapsed, year, tracknum = parts
```
instead

mrt-prodz:
Made the variables assignation change in ae7da56 , it's much cleaner that way. Thanks. (sorry about the multiple commits)

pydsigner:
Looks good now, just need @ultrabug to sign off.

py3status/modules/deadbeef.py
```diff
@@ -0,0 +1,88 @@
...
+            status = check_output(['deadbeef',
+                                   '--nowplaying',
+                                   self.delimiter.join(['%a',
+                                                        '%t',
+                                                        '%l',
+                                                        '%e',
+                                                        '%y',
+                                                        '%n'])])
+            # check if we have music currently  playing
+            if 'nothing' in status:
```
Anon1234: What about tracks with "nothing" in their properties like "metallica — **nothing** else matters"? :disappointed:

py3status/modules/deadbeef.py
```diff
@@ -0,0 +1,88 @@
...
+@author mrt-prodz
+"""
+from subprocess import check_output
+from time import time
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 1
+    delimiter = '¥'
```
Anon1234: Will fail on `Maximum the Hormone — Black ¥ Power G-MEN Spy` etc, We could use some long random string instead, e.g. `ooxohF2uDief8DooAhcaixu3Ju7doo2i`

py3status/modules/deadbeef.py
```diff
@@ -0,0 +1,88 @@
...
+    def _error_response(self, color):
+        response = {
+            'cached_until': time() + self.cache_timeout,
+            'full_text': 'deadbeef: error',
+            'color': color
+        }
+        return response
+
+    # return track currently playing in deadbeef
+    def get_status(self, i3s_output_list, i3s_config):
```
Anon1234: It should not start Deadbeef if it is not already running. Deadbeef start up is very expensive:
```
$ time deadbeef --nowplaying-tf "%artist% %title%" &> /dev/null

real	0m2.258s
user	0m2.093s
sys	0m0.097s
```

py3status/modules/deadbeef.py
```diff
@@ -0,0 +1,88 @@
...
+            if len(parts) == 6:
+                artist, title, length, elapsed, year, tracknum = parts
+            else:
+                return self._error_response(i3s_config['color_bad'])
+
+            response = {
+                'cached_until': time() + self.cache_timeout,
+                'full_text': self.format.format(artist=artist,
+                                                title=title,
+                                                length=length,
```
Anon1234: Could be reduced to `**locals()`.

pydsigner: @Anon1234 It could, yes, but that's definitely not good practice.

py3status/modules/deadbeef.py
mrt-prodz: This is correct, I could switch to a try/except with pidof instead to verify that deadbeef is running.

py3status/modules/deadbeef.py
mrt-prodz: Checking for deadbeef presence using pidof would solve that issue as well.

ultrabug:
My my, that's a collective effort, thanks everyone :)

Never heard of deadbeef before (silly me ?) so thanks for this as well !

PR #190, opened 2016-02-29
deadbeef: fix permissions, remove execution mode

Horgix:
Hi,

Mostly CF title. The deadbeef module has execution permissions, and I didn't see any reason for it (plus, there is no shebang in it).

This commit is just a chmod -x on py3status/modules/deadbeef.py


ultrabug:
<3

PR #191, opened 2016-03-01
Logging: entire rework, now uses standard log facilities

Horgix:
Good evening,

## Why

As announced on IRC, I'm trying to take a look again at how to split up this big `__init__.py` file.
Doing so, and making some tests in a separate env in Docker (yup, I know about virtualenv but eh), I noticed that everything was logged directly through `syslog`.

This commit is an attempt at improving that; it now uses the [standard logging facility](https://docs.python.org/3/library/logging.html).

## Benefits are the following

- **Cleaner logging levels** (debug, warning, etc) handling: no more `if config['debug'] then log` or stuff like that
- **Flexibility**: it's now possible to debug on stdout (which it what I was originally looking for) even though the corresponding handler is commented out in this commit
- **Cleaner nagbar notifications handling**: the i3_nagbar warning/error part doesn't have to have its own function anymore. It's handled directly by a custom StreamHandler, and a logging level set to `WARNING` so only Warnings/Errors/Criticals are displayed in the i3 nagbar

Also, it's a step in the direction of putting some stuff away from the main class, to allow splitting it up later (this commit still puts it in __init__.py to allow an easier review, it will splitted up later).

## Tests

I tested this in place of the current py3status in my setup : syslog and i3_nagbar logs still work as expected, and I only have to enable the stdoutHandler to debug if I want to (will probably later be implemented as an option to avoid modifying code just for that).

## Last word

Comments are ofc welcome :)

### DIFF
--- a/py3status/__init__.py
+++ b/py3status/__init__.py
@@ -9,6 +9,8 @@ import os
 import pkgutil
 import select
 import sys
+import logging
+import logging.handlers
 
 from collections import OrderedDict
 from contextlib import contextmanager
@@ -25,7 +27,6 @@ from subprocess import call
 from tempfile import NamedTemporaryFile
 from threading import Event, Thread
 from time import sleep, time
-from syslog import syslog, LOG_ERR, LOG_INFO, LOG_WARNING
 
 try:
     from setproctitle import setproctitle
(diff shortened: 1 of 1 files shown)


tobes:
Not commenting on the actual code, but I think that the idea behind this is great and would make py3status easier to hack on.

One thought I do have is to maybe rate limit the i3nag bar (I have managed to cover my screen in them before, using as a cheap logging substitute - yeah I'm crazy sometimes)

Horgix:
@tobes : I'll give it a try again now that the main `__init__.py` file is splitted. The rate limit on the nagbar is a pretty good idea, I actually found myself filling up my screen with warnings too.

We discussed this PR on IRC when it was created, and while @ultrabug was for this change, @pydsigner didn't seem to agree. Maybe we could discuss it a bit more here ? I really think that the actual logging could be improved.

tobes:
So when do hangout on irc? I might pop by

pydsigner:
@Horgix I didn't disagree with fixing logging, I just don't want to do it with logging.py; also @ultrabug said he wasn't a fan of the library either.

Horgix:
@pydsigner yeah I kind of shortened that, sorry. Anyway, the decision is up to you guys; I already presented most of the reasons I'm for the standard Logging lib in the original description of this PR.

tobes:
What have people got against `logging`?  It's part of core python and pretty simple to use.

I'd like to see some easier logging in py3status, what other way would people prefer?

tobes:
Should we close this PR?

We now have logging to file via `py3status -l <logfile>` and logging can be done from modules via `self.py3.log()`.  Also the code base has moved on quite a bit since March.

ultrabug:
I'm inclined to, @Horgix ?

ultrabug:
Closing, @Horgix if you feed I shouldn't have, ping me thanks a lot !

PR #192, opened 2016-03-01
[WIP] IO Refactor

pydsigner:
Note that this is not ready to merge. I just wanted to put this work up so that it could be discussed; I'll check to see if it even runs correctly when I'm back at my i3wm machine.


ultrabug:
merged on #193 thx

PR #193, opened 2016-03-01
Split up: take away profiling, events, helpers and module from __init__.py

Horgix:
Hi,

Here is my second attempt at splitting up this big `__init__.py` file. It's absolutely not what the end result should be, many things remain to be changed imho, but it's a first step in the right direction.

Anyway, these commits are as minimal as possible : only taking things away from the main file and fixing imports. Tested with my current setup and it works fine, log fine, etc.

Any comment is, as always, warmly welcome.

PS : don't just look at the total changes : take a look at each individual commit, making a pull request for each would honnestly be ridiculous.


Horgix:
[CI seems to be failing](https://travis-ci.org/ultrabug/py3status/builds/112820743#L176) with an `invalid syntax`. After testing, it seems to only fail in python2 :

```
$ python3
Python 3.5.1 (default, Dec  7 2015, 12:58:09)
[GCC 5.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information
>>> import helpers
>>> helpers.print_stderr('Test')
Test
```

```
$ python2
Python 2.7.11 (default, Dec  6 2015, 15:43:46)
[GCC 5.2.0] on linux2
>>> import helpers
  File "helpers.py", line 29
    print(line, file=sys.stderr)
                    ^
SyntaxError: invalid syntax
```

Since it's only syntax, how wasn't it failing already when it was in the `__init__.py` file ?

ultrabug:
I guess you've missed the `from __future__ import print_function` on the helper module :)

Horgix:
> I guess you've missed the from __future__ import print_function on the helper module :)

Indeed ! Fixed and rebased, CI now passing :)
Thanks for the pointer.

pydsigner:
You might want to change your implicit relative imports into explicit relative or absolute imports. I.e. `from pystatus.event import x` or `from .event import x` instead of `from event import x`.

Horgix:
@pydsigner  :

> You might want to change your implicit relative imports into explicit relative or absolute imports. I.e. from pystatus.event import x or from .event import x instead of from event import x.

Done ! Changed to py3status.<name>

ultrabug:
Now now, two PR for the same split, what do you think @pydsigner ? Shall we proceed with this one or you see a blocker in favour of yours ?

pydsigner:
This one is fine. I'll just refocus on the IO fixes.

ultrabug:
Alright, let's do this :)

PR #195, opened 2016-03-02
Doc: add basic contribution guidelines

Horgix:
Hi,

There is currently no guidelines for "how to contribute to py3status" and I think I saw something about it on the IRC channel but can't find it anymore :(

Anyway, I decided to write some basic informations down so we have the beginning of something :)
I think having a `CONTRIBUTING` file makes more sense than a Github Wiki page but I might be wrong.

Since English is not my native language, there may be some (a lot ?) mistakes; feel free to point it out so I can fix it !

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,58 @@
+# What you will need
+
+- python3
+- i3status
+    - http://i3wm.org/i3status/
+    - https://github.com/i3/i3status
+- flake8
+    - https://pypi.python.org/pypi/flake8
+    - https://github.com/PyCQA/flake8
+
+# Flake 8
+
+Before making any Pull Request, make sure that flake8 tests pass without any
+error/warning:
+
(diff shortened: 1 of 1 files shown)


ultrabug:
If think you can mention

- the contributor accepts the BSD license for his code
- the usage of yapf is recommended (for easy flake8 compliance)
- all the variables in the docs AND on the code should be ordered alphabetically
- the standalone test function at the end of the module should always be the same as the one on the example module in the doc (which you can link to) except for the function call obviously
- overall should "look like" the example module, especially the docstring on top of the module (short description, blank line, details)

ultrabug:
I think this has been done by @tobes doc/REAME right ?

tobes:
> I think this has been done by @tobes doc/REAME right ?

No but I'll look at adding this

ultrabug:
my bad, thanks to you both :)

PR #196, opened 2016-03-02
Doc: update Arch Linux AUR packaging infos

Horgix:
- README.rst: mention py3status and py3status-git packages

Only one package on the 2 was mentionned, and there wasn't any mention of initial packagers for Arch Linux AUR. rst format is a pain btw :(

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,10 +88,14 @@ Using emerge:
 
 Arch Linux
 ----------
-Thanks to @Horgix, py3status is present in the Arch User Repository using this URL:
-::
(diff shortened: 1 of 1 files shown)


pydsigner:
Agree that Markdown is miles better than reStructuredText.

ultrabug:
no sorry, because of pypi ! use markdown and break https://pypi.python.org/pypi/py3status/2.9 :(

pydsigner:
Looks like the fight to get markdown supported on PyPI/Warehouse has been on for several years but has stalled out because no-one can decide how to detect filetype between the two -_-

PR #197, opened 2016-03-02
Split up: take away I3status and Py3statusWrapper

Horgix:
Hi !

In the serie "split up this big `__init__.py` file", here is the second movie.

These 2 commits move `I3status` and `Py3statusWrapper` classes from `__init__.py` to dedicated files, (respectively `i3status.py` and `wrapper.py`) and update imports accordingly.

We're finally seeing a bit more clearly in this code base :)

Comments welcome. I'm not sure the "wrapper" name fit perfectly, but naming the file `py3status.py` would be painful when importing stuff.


pydsigner:
Maybe call it `core.py`?

Horgix:
> Maybe call it `core.py`?

Thx for the idea @pydsigner ! Renamed and rebased. The rest is up to you @ultrabug :)

PR #198, opened 2016-03-03
Add dnf module

tobes:
Module for checking updates on a Fedora distribution, similar to the arch_updates module.

### DIFF
--- a/py3status/modules/dnf.py
+++ b/py3status/modules/dnf.py
@@ -1,0 +1,106 @@
+# -*- coding: utf-8 -*-
+"""
+Displays the number of package updates pending for a Fedora Linux installation.
+This will display a count of how many `dnf` updates are waiting
+to be installed.
+Additionally check if any update security notices.
+
+Configuration parameters:
+    - cache_timeout : how often we refresh this module in seconds
+      default is 600
+    - format_updates : display format when updates needed
+      default is 'DNF: {updates}'
+    - format_uptodate : display format when no updates needed
+      default is 'DNF:'
+    - format_waiting : display format when first started as initial update
+      can take time
+      default is 'DNF:'
+    - color_good : color when no upgrades needed
+      default is None
+    - color_degraded : color when upgrade available
+      default is None
+    - color_bad : color when security notice
+      default is None
+
+Format status string parameters:
+    - updates : number of pending dnf updates
+
+@author Toby Dacre <https://github.com/tobes>
+@license BSD
+"""
+
+from time import time
+import subprocess
+import re
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 600
+    format_updates = 'DNF: {updates}'
+    format_uptodate = 'DNF:'
+    format_waiting = 'DNF:'
+    color_good = None
+    color_bad = None
+    color_degraded = None
+
+    _reg_ex_sec = re.compile('\d+(?=\s+Security)')
+    _reg_ex_pkg = re.compile(b'^\S+\.', re.M)
+    _first = True
+    _updates = None
+    _security_notice = False
+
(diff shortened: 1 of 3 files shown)


py3status/modules/dnf.py
```diff
@@ -0,0 +1,99 @@
...
+        output, error = subprocess.Popen(
+            ['dnf', 'check-update'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE).communicate()
+
+        updates = len(self._reg_ex_pkg.findall(output))
+
+        if updates == 0:
+            color = self.color_good or i3s_config['color_good']
+            results = self.format.format(updates='')
```
pydsigner: Why `updates=''` instead of passing in the `0`?

py3status/modules/dnf.py
```diff
@@ -0,0 +1,99 @@
...
+to be installed.
+Additionally check if any update security notices.
+
+Configuration parameters:
+    - cache_timeout : how often we refresh this module in seconds
+      default is 600
+    - format : display format
+      default is 'DNF: {updates}'
+    - color_good : color when no upgrades needed
+      default is None
```
pydsigner: I'd suggest changing "default is None" to "defaults to global color_good" and so on for the other two colors.

ultrabug:
Thanks for your contribution @tobes !

To add to what @pydsigner said, I'd like to suggest to rename this module to something more understandable for non-fedora users too, like `fedora_dnf` maybe ? OR to stay in line with the arch module, why not `fedora_updates` ?

pydsigner:
I'll agree with fedora_updates. But fedora_dnf for people who don't know what dnf is seems like putting mediaplayer_deadbeef or _spotify for people who don't know what those are. A user looking through the module listings will get more information anyways.

py3status/modules/dnf.py
tobes: I've changed this so now 0 if no updates and ? when unknown

py3status/modules/dnf.py
tobes: Done.  I copied that from another module but yes this is better

tobes:
Renamed to `fedora_updates` and other suggested changes.  Let me know if it needs and other work.  or if you want a fresh pull request as a single commit?

py3status/modules/fedora_updates.py
```diff
@@ -0,0 +1,97 @@
+# -*- coding: utf-8 -*-
+"""
+Displays the number of package updates pending for a Fedora Linux installation.
+This will display a count of how many `dnf` updates are waiting
+to be installed.
+Additionally check if any update security notices.
+
+Configuration parameters:
+    - cache_timeout : how often we refresh this module in seconds
```
ultrabug: please order the parameters alphabetically

py3status/modules/fedora_updates.py
```diff
@@ -0,0 +1,97 @@
...
+from time import time
+import subprocess
+import re
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 600
+    format = 'DNF: {updates}'
+    color_good = None
```
ultrabug: please order the parameters alphabetically

py3status/modules/fedora_updates.py
```diff
@@ -0,0 +1,97 @@
...
+    color_bad = None
+    color_degraded = None
+
+    _reg_ex_sec = re.compile('\d+(?=\s+Security)')
```
ultrabug: those non configuration parameters may be better in a proper __init__(self) function please

ultrabug:
Last comments in ! thx :)

tobes:
ok, I've made the changes

PR #200, opened 2016-03-03
Battery time remaining '' when unknown

tobes:
```
acpi -b -i
Battery 0: Unknown, 99%
```
if `format = "bat: {time_remaining}"` 
with master we see `bat: None`
this changes it to show `bat: ` which feels better to me.


pydsigner:
Why not change it to "unknown"?

ultrabug:
or even simpler '?' ?

tobes:
Changed to `?`

ultrabug:
Thanks @tobes !

Issue #178, opened 2016-02-02
RFC : module_group / scroll_group


ultrabug:
**Dear py3status users, tell us what you think !**

As discussed in PR #155 with @tjaartvdwalt and @pydsigner.

A `scroll_group` or `module_group` whatever the name handled by py3status would allow users to **group multiple modules together** and switch between them using the mouse scroll buttons (4 and 5) !

This way users could still benefit from click events on mouse buttons 1, 2 and 3 on each module. My first idea would give something like this on i3status.conf for your use case (I need to use two different module instance _first_ and wlan0 to get them work on the same i3status output) :

````
wireless _first_ {
	format_down = "W"
	format_up = "%ip"
	scroll_group = 1
}

wireless wlan0 {
	format_down = "W"
	format_up = "%essid"
	scroll_group = 1
}
````

This would make py3status group the two outputs together and display one or the other on mouse scroll automatically.

What's interesting with this is that this would work with every module available and users could then group and *toggle* whatever modules they want together with a single line config.

Of course, the scroll_group itself would be configurable (we could even want it to scroll each module automatically at a given interval !)

**Comment by shrimpza** [comment]
The only thing I think is weird, might be the requirement for unique module names, and the `_first_` work-around for that (I presume there must be unique `<module> [param]` combinations within a single config). If a user is at this point of customisation of their configuration, is backward compatibility with a default i3status config still a requirement?

Could a configuration something like this perhaps work better:

```
...
order += "group wifi"
...

group wifi {
    scroll_interval = "30s"

    wireless wlan0 {
        format_down = "W"
        format_up = "%ip"
    }

    wireless wlan0 {
        format_down = "W"
        format_up = "%essid"
    }
}
```

py3status could then "flatten" the group internally and orchestrate scroll events, intervals, clicking, etc.

It's more work, but I think the above may be more flexible and readable, plus allows for the mixing of different modules within a group (for example, combine `disk "/var"` and `disk "/home"`, or combine `sysdata` and `load` into one group, etc.).

UmBsublime:
Hi,

Kenneth's config solutions would be really nice to make this really simple
to setup !!!

ultrabug:
Okay guys, but I really don't get how py3status is supposed to *flatten* those two configurations together ? Because one way or another, it has to execute i3status to get its output and i3status doesn't accept such dual module definition :(

An idea @shrimpza ?

pydsigner:
[...]

PR #203, opened 2016-03-06
Add Groups module

tobes:
From #178 

I thought that this would be a fun way to get to know the codebase.

It allows a group of modules that can be cycled between

Configuration parameters:
- `button_next` Button that when clicked will switch to display next module.
  Setting to `0` will disable this action. *(default 4)*
- `button_prev` Button that when clicked will switch to display previous
  module.  Setting to `0` will disable this action. *(default 5)*
- `color` If the active module does not supply a color use this if set.
  Format as hex value eg `'#0000FF'` *(default None)*
- `cycle` Time in seconds till changing to next module to display. Setting to
  `0` will disable cycling. *(default 0)*
- `format` Format for module output. *(default "GROUP: {output}")*

- `{output}` Output of current active module

Example config:
```
# Create a disks group that will show space on '/' and '/home'
# Change between disk modules every 30 seconds
...
order += "group disks"
...

group disks {
    cycle = 30
    format = "Disks: {output}"

    disk "/" {
        format = "/ %avail"
    }

    disk "/home" {
        format = "/home %avail"
    }
}
```

The changes are hopefully not as large as they seem.

`I3status.i3status_config_reader()` has been updated to allow the new formatting.  We also need to know about modules used in groups but not in config `order` and allow modules in groups to know what group they are in.

`Module` gets a reference to `Py3statusWrapper` that gets passed to the group module so that it can read the output of its modules etc.  This also reduces the number of parameters that need passing to it.

I moved the execute each method of `Module.run()` into `Module.update_module()` so it can be called outside of run. Additionally it now updates the group it is a part of when it is updated.

Uncaught button events are dispatched to the active module.


py3status/i3status.py
```diff
@@ -143,6 +145,25 @@ def i3status_config_reader(self, i3status_config_path):
                 in_section = True
                 section_name = 'order'
 
+            if (not in_section) and line.startswith('group'):
```
pydsigner: `not` comes before `and` in the order of operations, so these parentheses are unneeded. The same goes for several other places in this function.

tobes: True, but sometimes it's nice to make things explicit

I'm happy to change this if there is a strong feeling

tobes: but `if line.startswith('group') and not in_section:` is much nicer

pydsigner: Leave the order as it is so that it makes sense (and saves cycles).

tobes: Sure

Really I'd like to change to pyparsing or something but that would be for another day

ultrabug:
please rebase

tobes:
Updated.

Not sure if you'll accept PR #212  but If you accept that, I'd like to fix this to work with those changes, rather than the other way.  It would also create less noise that way.

ultrabug:
done merged #212 thanks !

tobes:
OK,

updated for changes in master.

**The main changes are:**

* It now gets `Module` rather than `Py3statusWrapper`.

* group `Py3status` class now has `class Meta:` and a setting in that of `include_py3_module = True` which is used to include the `Module` class needed to access `Py3statusWrapper` etc.

* has a new `fixed_width` config option.

py3status/module.py
```diff
@@ -102,7 +107,17 @@ def load_methods(self, module, user_modules):
             # apply module configuration from i3status config
             mod_config = self.i3status_thread.config.get(module, {})
             for config, value in mod_config.items():
-                setattr(self.module_class, config, value)
+                # names starting with '.' are private
```
ultrabug: wouldn't `_` or `__` be a more pythonic way of defining this private state ?

pydsigner: Being linuxy works here too I think.

tobes: The reason I've used `.` is to keep the namespace clean.  It is possible that some user module uses a `_variable` in it's config and I didn't want to break such modules.

I can change to the underscore/double if you prefer but that was my reasoning.

ultrabug: okay, it sounds reasonable, thanks

py3status/modules/group.py
```diff
@@ -0,0 +1,196 @@
...
+    fixed_width = True
+    format = "GROUP: {output}"
+
+    class Meta:
+        include_py3_module = True
+
+    def __init__(self):
+        self.items = []
+        self.active = 0
+        self._cycle_time = time() + self.cycle
```
ultrabug: this is not a good place to do such a thing because the module is instantiated **before** the user's configuration parameters are applied, this means that the self.cycle will always be 0.

the result here is that the first displayed value will actually be the second module listed in the group :(

tobes: good spot should be in `_init()`

ultrabug: yes please !

tobes: commit  ad58253 fixed this issue

ultrabug:
I'd **love** to merge this but my testing shows that some modules are not displayed on a random fashion.

I can't find the culprit yet but it seems related to modules which can take time to process and output something and then you have to wait for the next update for them to show !

py3status/modules/group.py
```diff
@@ -0,0 +1,193 @@
...
+        else:
+            return py3_wrapper.config.get(current)
+
+    def _next(self):
+        self.active = (self.active + 1) % len(self.items)
+
+    def _prev(self):
+        self.active = (self.active - 1) % len(self.items)
+
+    def output(self, i3s_output_list, i3s_config):
```
ultrabug: I think we should rename it to `group` or `group_output` so it shows correctly on the logs.

`method output returned {'instance': 'tz', 'name': 'group', 'full_text': 'LA 01:56', 'cached_until': 1458550600.423212}`

tobes: I've changed this

tobes:
> some modules are not displayed on a random fashion.

Do you know any modules that seem to be affected and I'll look at them too.

ultrabug:
> Do you know any modules that seem to be affected and I'll look at them too.

I think it's related to #212 because if the module.run() raises an exception, it is never rescheduled for running and becomes stale on the bar ! (so if it fails on the first run, it is never displayed)

Would you please take a look at it and confirm ?

tobes:
yeah if the module has no output eg raises exception then it is not scheduled.

commit 2aa9a3 forces an update to be set if the module does not set one.

config['cache_timeout'] is used as the interval time

tobes:
I've also created a debug module that is based off this branch https://github.com/tobes/py3status/blob/debug/py3status/modules/debug.py

It gives info on py3status modules such as number of times updated, when next update due etc that can be helpful in checking the health of modules.

ultrabug:
Huge thanks @tobes, I think this is an insane feature and smart implementation :+1:

As a side note, I think this feature should be present in the main README because it's very attractive :)

Issue #201, opened 2016-03-04
Making modules/README.md more friendly


tobes:
the current README.md https://github.com/ultrabug/py3status/tree/master/py3status/modules is hard to navigate and read.

I've had a go at making it more readable see  https://github.com/tobes/py3status/tree/modules_readme/py3status/modules#top

It's a work in progress I've done what I can using my vim skills without having to do much manual intervention.  If you are interested I'll manually fix it up.

I'm quite happy to change how things like defaults are shown etc

general layout

### module

description

Configuration parameters:
- `param` Param description *(default value)*

Format status string parameters:
- `{item}` Description

requires:
- `{requirement}` extra info

**author** Author  **license** License

pydsigner:
I like it, other than the `^top` links. I'd rather just hit Back in my web browser.

tobes:
closing this as there is now a pull request for this feature

PR #206, opened 2016-03-07
modules/README.md reformat

tobes:
Following from #201 This is an initial cleanup of `modules/README.md`

The modules have also had their docstrings updated to match the README there is a new cli command `py3status docs update modules` to update the modules docstrings from the README

If you like the idea I'll do some more cleanups of the docstrings module by module and add a cli script to build the README from the module docstrings as it seems better to update from the modules.

A simple view of the rendered markdown is here
https://github.com/tobes/py3status/tree/modules_readme/py3status/modules#top


tobes:
> If you like the idea I'll do some more cleanups of the docstrings module by module 

By this I mean it would be good to merge the basic changes and then fix the modules in other pull requests so I don't have to keep this in sync with documentation changes in the modules as that will get painful as time goes on

but open to suggestions till then

tobes:
OK,

I added cli commands

`py3status docstring update modules`  update module docstring from README.md

`py3status docstring update xrandr xsel` update README.md from modules given

`py3status docstring check` check that README.md matches module docstrings

`py3status docstring check diff` as abouve but show a diff

ultrabug:
One of the biggest (if not THE biggest) MR ever submitted :open_mouth: thanks @tobes !

First question, do you think the `print_module_description` on the `core` module could be merged into your `docstrings` module ? maybe with a sort of `output` option so that we can avoid code duplication while retaining markdown and console outputs ?

tobes:
> First question, do you think the print_module_description on the core module could be merged into your docstrings module ? 

This would seem sensible

> maybe with a sort of output option so that we can avoid code duplication while retaining markdown and console outputs ?

yeah we can strip some markdown out

ultrabug:
> yeah we can strip some markdown out

That's not what I meant, I think your MD styling is cool actually. Just meant that a single function could maybe handle both CLI output and MD output at once.

tobes:
I'll have a play and we can then see if it meets your needs.

I'll leave the cli output as similar to how it is now

tobes:
OK ready for testing and review

tobes:
I updated the commands

`modules list`  as before 

`modules list user` just shows user modules

`modules list core` just core ones

`modules details` as before

`modules details wifi xsel` details for the selected module

tobes:
commit e7c64cd changes the details output I'll revert if you don't like it

old
```
Module details:
  wifi                   Display WiFi bit rate, quality, signal and SSID using iw.
                         
                         Configuration parameters:
                           - `bitrate_bad` Bad bit rate in Mbit/s (default: 26)
                           - `bitrate_degraded` Degraded bit rate in Mbit/s (default: 53)
                           - `cache_timeout` Update interval in seconds (default: 10)
                           - `device` Wireless device name (default: "wlan0")
                           - `down_color` Output color when disconnected, possible values:
                         "good", "degraded", "bad" (default: "bad")
                           - `format_down` Output when disconnected (default: "down")
                           - `format_up` See placeholders below
                         (default: "W: {bitrate} {signal_percent} {ssid}")
                           - `round_bitrate` If true, bit rate is rounded to the nearest whole number
                         (default: true)
                           - `signal_bad` Bad signal strength in percent (default: 29)
                           - `signal_degraded` Degraded signal strength in percent (default: 49)
                           - `use_sudo` Use sudo to run iw, make sure iw requires no password by
                         adding a sudoers entry like
                         "<username> ALL=(ALL) NOPASSWD: /usr/bin/iw dev wl* link"
                         (default: false)
                         
                           - `{bitrate}` Display bit rate
                           - `{signal_dbm}` Display signal in dBm
                           - `{signal_percent}` Display signal in percent
                           - `{ssid}` Display SSID
                         
                         Requires:
                           - `iw`
                         
                         **author** Markus Weimar <mail@markusweimar.de>
                         **license** BSD
                         
                         ---
```
new
```
Module details:
--------
  wifi
--------
Display WiFi bit rate, quality, signal and SSID using iw.

Configuration parameters:
  - `bitrate_bad` Bad bit rate in Mbit/s (default: 26)
  - `bitrate_degraded` Degraded bit rate in Mbit/s (default: 53)
  - `cache_timeout` Update interval in seconds (default: 10)
  - `device` Wireless device name (default: "wlan0")
  - `down_color` Output color when disconnected, possible values:
"good", "degraded", "bad" (default: "bad")
  - `format_down` Output when disconnected (default: "down")
  - `format_up` See placeholders below
(default: "W: {bitrate} {signal_percent} {ssid}")
  - `round_bitrate` If true, bit rate is rounded to the nearest whole number
(default: true)
  - `signal_bad` Bad signal strength in percent (default: 29)
  - `signal_degraded` Degraded signal strength in percent (default: 49)
  - `use_sudo` Use sudo to run iw, make sure iw requires no password by
adding a sudoers entry like
"<username> ALL=(ALL) NOPASSWD: /usr/bin/iw dev wl* link"
(default: false)

  - `{bitrate}` Display bit rate
  - `{signal_dbm}` Display signal in dBm
  - `{signal_percent}` Display signal in percent
  - `{ssid}` Display SSID

Requires:
  - `iw`

**author** Markus Weimar <mail@markusweimar.de>
**license** BSD
```

ultrabug:
I think the new output is okay (tho I don't quite like the ----) but I get a nagbar errorr when running `python py3status/__init__.py modules list xsel`

`local variable 'core_mods' referenced before assignment`

tobes:
I've fixed that error

tobes:
anything you'd prefer to the `---` in the output?

ultrabug:
> anything you'd prefer to the --- in the output?

would "===" satisfy both MD and plain outputs ? If so, plz also don't "center" the module name plz

tobes:
like so

pydsigner:
Colons or em-dashes after the configuration/format parameter names would help in plain-text mode. For example:
```
- `{bitrate}` — Display bit rate
```
Instead of
```
- `{bitrate}` Display bit rate
```

ultrabug:
> - `{bitrate}` — Display bit rate

Agreed, and I'd also do the same on the markdown version on the main module list (when the anchors), seeing `rt Display the number of ongoing tickets from selected RT queues.` is not very readable.

pydsigner:
@ultrabug That was what I was saying, sorry for the confusing phrasing.

tobes:
Ok,

I've gone with the suggestions. I've used a colon as the em-dash would need to be in all module docstrings and might be hard for peoploe to work with `:` is just easier to type etc.

Example

### <a name="bluetooth"></a>bluetooth

Display bluetooth status.

Confiuration parameters:
  - `format` : format when there is a connected device
  - `format_no_conn` : format when there is no connected device
  - `format_no_conn_prefix` : prefix when there is no connected device
  - `format_prefix` : prefix when there is a connected device

Format of status string placeholders
  - `{name}` :  device name
  - `{mac}` :  device MAC address

Requires:
  - `hcitool`

**author** jmdana <https://github.com/jmdana>
**license** GPLv3 <http://www.gnu.org/licenses/gpl-3.0.txt>

---

The modules list should also be more readable has em-dash and module name now bold too

pydsigner:
Adding a colon like that makes it worse rather than better IMO. It's misaligned. I don't think that em-dashes would be too big of a deal to work with, but I could be wrong on that.

`{name}`: device name
Is better but not by much :-/

tobes:
> {name}: device name

can we discuss this using code blocks and as markdown so we can know what we are actually writing

from my point of view for say config options I'd like
```
  - `option`: description
```
  - `option`: description

or
```
  - `option` - description
```
  - `option` - description

this is so that we get markdown rendering nicely and is readable in the code.  My preference is the hyphen

I am totally against the em-dash as I think it will cause work when people submit modules etc and it can render too similarly to a hyphen.  Also It's a pain to type at least for me.

There is always the option of different formatting in the README.md and the modules docstring but I'd rather avoid that for now.

Additionally I'm not sure why the flake8 is failing, locally I have no problem anyhow getting the format agreed is the priority and it should be fixed in the next push

pydsigner:
I dislike the colon and hyphen because the inline code tag has padding that makes the rendered markdown unbalanced.

```
- `option` - description
- `option`- description
- `option` — description
- `option`— description
- `option` : description
- `option`: description
```

- `option` - description
- `option`- description
- `option` — description
- `option`— description
- `option` : description
- `option`: description

I type em-dashes using the compose key which makes it quite a bit easier, but copy-pasting the general formatting isn't too hard either IMO.

tobes:
with colons it'd be like this.
the continuation lines need fixing but this is one option for them
```
"""
Display the X selection.

Configuration parameters:
        (default is at every py3status configured interval)
    command: the xsel command to run (default 'xsel')
    max_size: stip the selection to this value (default 15)
    symmetric: show the beginning and the end of the selection string
        with respect to configured max_size.

Requires:
    xsel: command line tool

**author** Sublim3 umbsublime@gamil.com
**license** BSD
"""
```

pydsigner:
That'd be better with the colons, otherwise we're literally back where we originally started.

ultrabug:
tbh I like it like this very much. The only point is that some license are not on their own line on the markdown version, is it inteded ?

**author** bstinsonmhk **license** BSD

tobes:
> The only point is that some license are not on their own line on the markdown version, is it inteded ?
No real reason for it now.

I'll put them on their own lines when I get the time, probably tomorrow

ultrabug:
@tobes also, there are some modules' PR waiting, would you care if I merge them and you rebase from it ?

tobes:
No problem

tobes:
Ok, I think we are done.

The last commit probably could be improved but I think it's fine for now.  I'd rather add newlines than the `<br/>` but that needs a bit of a code re-factor.  I might add it to this or if you've merged I'll do in a separate PR.

I'll keep this updated for any module changes you make.  Merge when you want.

tobes:
Drunken coding complete.  I removed the`<br/>` tags and made it nice

ultrabug:
Awesome thanks @tobes ! Now we need to enforce the formatting to all the (waiting) modules !

tobes:
if we run `py3status docstring check` in our tests it creates output if there are issues around non-matching docstring and modules missing from README.md all good is no output.

That would be a start on module docstrings etc.  I was thinking of creating a document on how to write a modules docstring.

PR #210, opened 2016-03-14
keyboard_layout module and static LED mask

mrt-prodz:
As mentioned on IRC the module to report current keyboard layout wasn't working properly with my specific setup. The static LED mask values were not meeting any of my layouts and while it's not taking much to edit these I'm trying to automate the process but I'm not sure if it would work for everybody.

In my specific case it works fine but I would like your input.

### DIFF
--- a/py3status/modules/keyboard_layout.py
+++ b/py3status/modules/keyboard_layout.py
@@ -87,10 +87,10 @@ class Py3status:
             return layouts[0]
         xset_output = check_output(["xset", "-q"]).decode("utf-8")
         led_mask = re.match(ledmask_re, xset_output).group(1)
-        if led_mask == "00000000":
-            return layouts[0]
-        elif led_mask == "00001000":
-            return layouts[1]
+        if len(led_mask) == 8:
+            lang = int(led_mask[4])
+            if layouts[lang] is not None:
+                return layouts[lang]
         return "Err"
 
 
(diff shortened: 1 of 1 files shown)


py3status/modules/keyboard_layout.py
```diff
@@ -87,10 +87,10 @@ def _xset(self):
             return layouts[0]
         xset_output = check_output(["xset", "-q"]).decode("utf-8")
         led_mask = re.match(ledmask_re, xset_output).group(1)
-        if led_mask == "00000000":
-            return layouts[0]
-        elif led_mask == "00001000":
-            return layouts[1]
+        if len(led_mask) == 8:
+            lang = int(led_mask[4], 16)
+            if layouts[lang] is not None:
```
ultrabug: better to use `if layouts.get(lang):` as it won't throw a KeyIndex error if the lang is not present please.

ultrabug:
ping @tuxitop if you may test & review please ?

ultrabug:
tested, works for me.

ultrabug:
So, @mrt-prodz, what should we do here since you pushed the thunderbird commit ? force reset your branch or make a new PR ?

mrt-prodz:
Sorry about that @ultrabug , should have used a different branch before pushing new commit. It should be ok now.

tuxitop:
Thank you for bringing this up. I noticed that even turning the caps lock on would have reproduced this bug.

mrt-prodz:
@tuxitop your patch is working fine on my end, made a commit with your changes.

ultrabug:
Thanks guys, nice collaboration :+1:

PR #211, opened 2016-03-15
Create xscreensaver.py

neutronst4r:
I have addapted the dpms.py module to work with xscreensaver instead. I changed a lot of the original system calls. It would have been very painful to add it as an option in dpms.py. I also switched from using the system package to subprocess, which is in my opinion superior. Calling xscreensaver without messing up the standard output in the py3status did not seem possible with "system". Using Popen and sending its output to DEVNULL, however, works nicely. Furthermore, the new xscreensaver needs to be forked by assigning it its own process group.

### DIFF
--- a/py3status/modules/xscreensaver.py
+++ b/py3status/modules/xscreensaver.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 """
+Indicator for Xscreensaver, can be toggled by left clicking.
+
 This is a partial rewrite of dpms.py. Most credit
 goes to the original author Andre Doser. This module shows the status
 of Xscreensaver and activates or deactivates it upon left click.
@@ -11,18 +13,19 @@ Settings can be managed using "xscreensaver-demo".
 
 Configuration parameters:
     - format_off: string to display when Xscreensaver is disabled
-    - format_on: string to display when Xscreensaver is enabled
+    - format_on : string to display when Xscreensaver is enabled
 
 Example configuration in py3status.conf:
 
+```
 order += "xscreensaver"
 xscreensaver {
     format_off = "xscreensaver off"
     format_on = "xscreensaver on"
 }
+```
 
-@author Andre Doser <dosera AT tf.uni-freiburg.de> (dpms.py)
-@author neutronst4r <c7420{at}gmx{dot}net> (Xscreensaver rewrite)
+@author neutronst4r <c7420{at}gmx{dot}net>
 """
 from os import setpgrp
 from subprocess import call, Popen, DEVNULL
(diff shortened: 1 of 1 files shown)


py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def xscreensaver(self, i3s_output_list, i3s_config):
+        """
+        Display a colorful state of xscreensaver.
+        """
+        self.run = call(["pidof", "xscreensaver"], stdout=DEVNULL,
+                        stderr=DEVNULL, preexec_fn=setpgrp) == 0
+        return {
+            'full_text': self.format_on if self.run else self.format_off,
+            'color': self.color_on or i3s_config['color_good']
```
pydsigner: Could you reindent lines 48-49? I.e.
```python
            'color': self.color_on or i3s_config['color_good'] if self.run
                     else self.color_off or i3s_config['color_bad']
```

py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,78 @@
...
+
+order += "xscreensaver"
+xscreensaver {
+    format_off = "xscreensaver off"
+    format_on = "xscreensaver on"
+}
+
+@author Andre Doser <dosera AT tf.uni-freiburg.de> (dpms.py)
+@author neutronst4r <c7420{at}gmx{dot}net> (Xscreensaver rewrite)
+"""
```
tobes: only one @author please just separate authors by a comma

py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,78 @@
...
+    - format_off: string to display when Xscreensaver is disabled
+    - format_on: string to display when Xscreensaver is enabled
+
+Example configuration in py3status.conf:
+
+order += "xscreensaver"
+xscreensaver {
+    format_off = "xscreensaver off"
+    format_on = "xscreensaver on"
+}
```
tobes: please surround example code with \`\`\` for documentation purposes.

py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+"""
```
tobes: There should be a single line at the top of the docstring briefly explaining the modules purpose followed by a blank line

py3status/modules/xscreensaver.py
neutronst4r: If I do that, flake8 tells me "continuation line unaligned for hanging indent"

ultrabug: @neutronst4r if in doubt, just install and use `yapf`, it's very handy :)

py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,86 @@
...
+
+This script is useful for people who let Xscreensaver manage dpms
+settings. Xscreensaver has its own dpms variables separate from xset.
+Dpms can be savely turned of in xset as long as xscreensaver is running.
+Settings can be managed using "xscreensaver-demo".
+
+Configuration parameters:
+    - format_off: string to display when Xscreensaver is disabled
+    - format_on : string to display when Xscreensaver is enabled
+
```
tobes: Being picky can you remove the `- ` in these config parameters and follow them with a colon without a space
like
```
Configuration parameters:
    format_off: string to display when Xscreensaver is disabled
    format_on: string to display when Xscreensaver is enabled
```
It helps with creating modules/README.md

PR #212, opened 2016-03-15
Only update modules when they need to be

tobes:
**Issues addressed:**

1) I have a world clock user module.  With the current setup the time displayed is out of sync as it can take 1 second for the updated info to be used due to `sleep(self.config['interval'])`.

2) When clicking on a module it can take up to `config['interval']` for the result of the of the click to update eg `dpms`

3) Modules that have no update for long time periods eg `weather_yahoo` spend lots of time checking if they need to update when they don't need to (minor issue)

**This pull request allows these updates to be shown immediately**

I could change `config['interval']` but then my modules are working mainly doing nothing.

**changes:**

Instead of sleeping we use a timer to wake up and do the update when needed. Modules set their own timers. Triggering an update via a click event, cancels any pending timer and in turn will set a new one.

`config['interval']`  is no longer used.  I've left it for now as it is a cli option but I'd like to depreciate that if you are happy with the idea.

`config['minimum_interval']` has been added to stop module updates happening too frequently (not that I've seen this problem).  Currently this is set at 0.1 (seconds)

`Module.kill()` is now it's own function and is called on termination for all py3status modules.  It clears any timer and calls the modules `kill()`

Modules are no longer 'alive' as such so that test is removed from `Py3statusWrapper.run()`.

From the limited testing I've done py3status does not use any more cpu time than before, slightly less if anything.  Most time is still spent in `Py3statusWrapper.run()` from what I can see.

Non-scientific testing by running on cli and pressing ^C after ~10 seconds
```
git checkout master

time py3status -s -c ~/.i3/i3status.conf
real	0m9.694s
user	0m0.349s
sys	0m0.106s

git checkout deadline

time py3status -s -c ~/.i3/i3status.conf

real	0m10.555s
user	0m0.317s
sys	0m0.084s
```

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -2,10 +2,10 @@ import sys
 import os
 import imp
 
-from threading import Thread
+from threading import Thread, Timer
 from collections import OrderedDict
 from syslog import syslog, LOG_INFO, LOG_WARNING
-from time import sleep, time
+from time import time
 
 from py3status.profiling import profile
 
@@ -32,6 +32,7 @@ class Module(Thread):
         self.module_class = None
         self.module_inst = ''.join(module.split(' ')[1:])
         self.module_name = module.split(' ')[0]
+        self.timer = None
         #
         self.load_methods(module, user_modules)
 
@@ -157,7 +158,12 @@ class Module(Thread):
         didn't already do so.
         We will execute the 'kill' method of the module when we terminate.
         """
-        while self.lock.is_set():
+        # cancel any existing timer
+        if self.timer:
+            self.timer.cancel()
+
+        if self.lock.is_set():
+            cache_time = None
             # execute each method of this module
             for meth, obj in self.methods.items():
                 my_method = self.methods[meth]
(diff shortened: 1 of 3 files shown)


tobes:
Actually seems `config['interval']` is still being used so ignore that comment

pydsigner:
Looks fine to me.

py3status/module.py
```diff
@@ -220,10 +230,19 @@ def run(self):
...
                     syslog(LOG_WARNING,
                            'user method {} failed ({})'.format(meth, err))
-
-            # don't be hasty mate, let's take it easy for now
-            sleep(self.config['interval'])
-
+            # don't be hasty mate
+            # set timer to do update next time one is needed
+            if cache_time:
+                delay = cache_time - time()
```
ultrabug: Just to be picky, wouldn't a `delay = max(cache_time - time(), self.config['minimum_interval'])` be better and avoid the condition below ? ;)

ultrabug:
Impressive PR (again) @tobes, I made a slight comment and I'm ready to merge ! Thank you

py3status/module.py
tobes: I though of that too, but then decided that this makes it more explicit.  I don't mid either way.

tobes:
I made that change

ultrabug:
Thanks a bunch for this neat PR @tobes . If you feel like it, we still have the same kind of problem when interacting on_click with an i3status module (we send a SIGUSR1 in background but we can take time before it being displayed)

ping @cornerman FYI since we discusses this very topic !

PR #216, opened 2016-03-22
Wifi updates

tobes:
This makes 4 changes to the wifi module.

* add `{ip}` format option

* add `{icon}` format option, similar to battery level, showing signal strength

* tries to guess wifi device (has preference for `wlan0` which was old fixed default)

* add `{device}` format option


ultrabug:
LGTM, thanks !

PR #217, opened 2016-03-22
New module: xrandr_rotate

maximbaz:
Hey guys, I've got an idea of a module to rotate screen horizontally/vertically on a single click that I want to share with you. 

My main desire is: I want it to be **simple and intuitive**, which includes:

- As a user I see it behaving just like the `xrandr` module, in terms of selecting the rotation using mouse left click or wheel and applying the selected rotation using mouse right click.
- Even though there are four different rotations possible (normal, left, right, inverted), on a daily basis I want to think of my screen as either in "horizontal" or in "vertical" mode. Therefore the configuration parameters `horizontal_rotation` and `vertical_rotation` to choose the preferred rotation modes (clockwise or counter-clockwise).

-------

*The module looks like this with the default setup (look in the middle)*

DP1 output (provided by `xrandr` module) is in H (horizontal) rotation.

selection_052

Left mouse click, and DP1 output is about to be switched to V (vertical) rotation.

selection_053

Right click to confirm, and DP1 output is now in V (vertical) rotation.

selection_055

-------

If you are a fan of icon fonts, the options `horizontal_icon` and `vertical_icon` are available for you. The module may look like the following:

selection_056

selection_058

-------

*Multi-monitor setup is an open question though.* 

I see this as a non-trivial question: I doubt that it is reasonable to rotate all screens, I don't want to hardcode output name in config, and I don't want to make the selection mechanism complicated. 

As of now, the rotation happens on the *first* detected output. It works fine for the single monitor setup (which I happen to have), but I'm sure people with several screens would want to have some flexibility.

*Let me know what you think.* As for the multi-monitor question, unless some great idea comes during the review, I'm tempted to suggest to keep it as it is and improve in future as more people give their feedback on how they want it to behave.

### DIFF
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -51,7 +51,8 @@ class Py3status:
     vertical_rotation = 'left'
 
     def _call(self, cmd):
-        output = Popen(cmd, stdout=PIPE, shell=True).stdout.read()
+        process = Popen(cmd, stdout=PIPE, shell=True)
+        output = process.communicate()[0] or ""
         try:
             # python3
             output = output.decode()
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -29,6 +29,12 @@ Available placeholders for formatting the output:
     {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
 
 
+Remarks:
+    There have been cases when rotating a screen using this module made i3 unusabe.
+    If you experience a similar behavior, please report as many details as you can:
+    https://github.com/ultrabug/py3status/issues/227
+
+
 @author Maxim Baz (https://github.com/maximbaz)
 @license BSD
 """
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+
+"""
+Switch between horizontal and vertical screen rotation on a single click.
+
+Configuration parameters:
+    - cache_timeout: how often to refresh this module.
+        default is 10
+    - format: a string that formats the output, can include placeholders.
+        default is '{icon}'
+    - horizontal_icon: a character to represent horizontal rotation.
+        default is 'H'
+    - horizontal_rotation: a horizontal rotation for xrandr to use.
+        available options: 'normal' or 'inverted'.
+        default is 'normal'
+    - vertical_icon: a character to represent vertical rotation.
+        default is 'V'
+    - vertical_rotation: a vertical rotation for xrandr to use.
+        available options: 'left' or 'right'.
+        default is 'left'
+
+Available placeholders for formatting the output:
+    - {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
+
+
+@author Maxim Baz (https://github.com/maximbaz)
+@license BSD
+"""
+
+from subprocess import Popen, PIPE
+from time import sleep, time
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    format = '{icon}'
+    horizontal_icon = 'H'
+    horizontal_rotation = 'normal'
+    vertical_icon = 'V'
+    vertical_rotation = 'left'
+
+    def _call(self, cmd):
+        output = Popen(cmd, stdout=PIPE, shell=True).stdout.readline()
+        try:
+            # python3
+            output = output.decode()
+        except:
+            pass
+        return output.strip()
+
+    def _get_first_output(self):
+        cmd = 'xrandr -q --verbose | grep " connected [^(]" | cut -d " " -f1'
+        return self._call(cmd)
+
+    def _get_current_rotation_icon(self):
+        output = self._get_first_output()
+        cmd = 'xrandr -q --verbose | grep "^' + output + '" | cut -d " " -f5'
+        is_horizontal = self._call(cmd) in ['normal', 'inverted']
+        return self.horizontal_icon if is_horizontal else self.vertical_icon
+
+    def _apply(self):
+        rotation = self.horizontal_rotation if self.displayed == self.horizontal_icon else self.vertical_rotation
+        output = self._get_first_output()
+        cmd = 'xrandr --output ' + output + ' --rotate ' + rotation
+        self._call(cmd)
+
+    def _switch_selection(self):
+        self.displayed = self.vertical_icon if self.displayed == self.horizontal_icon else self.horizontal_icon
+
+    def on_click(self, i3s_output_list, i3s_config, event):
+        """
+        Click events
+            - left click & scroll up/down: switch between rotations
+            - right click: apply selected rotation
+        """
+        button = event['button']
+        if button in [1, 4, 5]:
+            self._switch_selection()
+        elif button == 3:
+            self._apply()
+
+    def xrandr_rotate(self, i3s_output_list, i3s_config):
+        if not hasattr(self, 'displayed'):
+            self.displayed = self._get_current_rotation_icon()
+
+        full_text = self.format.format(icon=self.displayed or '?')
+
+        response = {
+            'cached_until': time() + self.cache_timeout,
+            'full_text': full_text
+        }
+
+        # coloration
+        if self.displayed == self._get_current_rotation_icon():
+            response['color'] = i3s_config['color_good']
+
+        return response
+
+
+if __name__ == "__main__":
+    """
+    Test this module by calling it directly.
+    """
+    x = Py3status()
+    config = {
+        'color_bad': '#FF0000',
+        'color_degraded': '#FFFF00',
+        'color_good': '#00FF00'
+    }
+    while True:
+        print(x.xrandr_rotate([], config))
+        sleep(1)
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -27,6 +27,8 @@ Configuration parameters:
 
 Available placeholders for formatting the output:
     {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
+    {screen} a screen name, specified by `screen` option or detected automatically
+        if only one screen is connected, otherwise 'ALL'.
 
 
 Remarks:
@@ -109,7 +111,8 @@ class Py3status:
             if not hasattr(self, 'displayed'):
                 self.displayed = self._get_current_rotation_icon(all_outputs)
 
-            full_text = self.format.format(icon=self.displayed or '?')
+            screen = self.screen or all_outputs[0] if len(all_outputs) == 1 else 'ALL'
+            full_text = self.format.format(icon=self.displayed or '?', screen=screen)
 
         response = {
             'cached_until': time() + self.cache_timeout,
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -4,29 +4,29 @@
 Switch between horizontal and vertical screen rotation on a single click.
 
 Configuration parameters:
-    - cache_timeout: how often to refresh this module.
-        default is 10
-    - format: a string that formats the output, can include placeholders.
-        default is '{icon}'
-    - hide_if_disconnected: a boolean flag to hide icon when `screen` is disconnected.
+    cache_timeout: how often to refresh this module.
+        (default is 10)
+    format: a string that formats the output, can include placeholders.
+        (default is '{icon}')
+    hide_if_disconnected: a boolean flag to hide icon when `screen` is disconnected.
         it has no effect unless `screen` option is also configured.
-        default: None
-    - horizontal_icon: a character to represent horizontal rotation.
-        default is 'H'
-    - horizontal_rotation: a horizontal rotation for xrandr to use.
+        (default: None)
+    horizontal_icon: a character to represent horizontal rotation.
+        (default is 'H')
+    horizontal_rotation: a horizontal rotation for xrandr to use.
         available options: 'normal' or 'inverted'.
-        default is 'normal'
-    - screen: display output name to rotate, as detected by xrandr.
+        (default is 'normal')
+    screen: display output name to rotate, as detected by xrandr.
         if not provided, all enabled screens will be rotated.
-        default: None
-    - vertical_icon: a character to represent vertical rotation.
-        default is 'V'
-    - vertical_rotation: a vertical rotation for xrandr to use.
+        (default: None)
+    vertical_icon: a character to represent vertical rotation.
+        (default is 'V')
+    vertical_rotation: a vertical rotation for xrandr to use.
         available options: 'left' or 'right'.
-        default is 'left'
+        (default is 'left')
 
 Available placeholders for formatting the output:
-    - {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
+    {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
 
 
 @author Maxim Baz (https://github.com/maximbaz)
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -60,13 +60,15 @@ class Py3status:
         return output.strip()
 
     def _get_all_outputs(self):
-        cmd = 'xrandr -q --verbose | grep " connected [^(]" | cut -d " " -f1'
+        cmd = 'xrandr -q | grep " connected [^(]" | cut -d " " -f1'
         return self._call(cmd).split()
 
     def _get_current_rotation_icon(self, all_outputs):
         output = self.screen or all_outputs[0]
-        cmd = 'xrandr -q --verbose | grep "^' + output + '" | cut -d " " -f5'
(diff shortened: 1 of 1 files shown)


tobes:
This looks nice.

It would be fantastic if the functionality could be added into the existing xrandr module but looking at it I think that would be a tight squeeze so maybe it's much better to have them separate

>  As for the multi-monitor question, unless some great idea comes during the review

I'm sure I can 'magically' get you the current display from the `xrandr` module (if it is running) that would solve the multi-monitor issue.

py3status/modules/xrandr_rotate.py
```diff
@@ -0,0 +1,115 @@
...
+        available options: 'normal' or 'inverted'.
+        default is 'normal'
+    - vertical_icon: a character to represent vertical rotation.
+        default is 'V'
+    - vertical_rotation: a vertical rotation for xrandr to use.
+        available options: 'left' or 'right'.
+        default is 'left'
+
+Available placeholders for formatting the output:
+    - {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
```
tobes: Instead of the `{icon}` I'd be inclined to just have a `format_horizontal` and `format_vertical`  or maybe `format_h` and `format_v`.

Although it might be nice to spin the display all the way normal > 90 left > 180 > 90 right is there a reason for limiting this?

maximbaz:
Hey @tobes, thanks for the feedback :)

> It would be fantastic if the functionality could be added into the existing xrandr module but looking at it I think that would be a tight squeeze so maybe it's much better to have them separate

I also initially thought of extending existing xrandr, but I kind of want to control monitor layout and monitor rotation separately, thus I ended up with a small separate module.

> I'm sure I can 'magically' get you the current display from the xrandr module (if it is running) that would solve the multi-monitor issue.

Not really, or I don't fully understand the suggestion. I can retrieve all currently used screens from xrandr (execute `xrandr` and look where `*` sign is), but if people have 2 screen setup where one screen extends another, do you think we should rotate *all of them* at the same time? If I completely misunderstood you, could you give me an example? :)

> Instead of the {icon} I'd be inclined to just have a format_horizontal and format_vertical or maybe format_h and format_v.

I think having `format` option with possible placeholders is kind of standard and obvious way. There is also an attempt to bring old modules to such approach (see #98). That's the reason I created a `format` with the `{icon}` placeholder.

In most cases I think nobody would extend the `format` itself, but will specify `horizontal_icon` and `vertical_icon` - if using FontAwesome, for example.

> Although it might be nice to spin the display all the way normal > 90 left > 180 > 90 right is there a reason for limiting this?

The reason is simplicity, you won't do it every day. As I see the usage scenario, you have a screen that you have normally in horizontal position, and when sometimes you want to read a long text, you rotate it vertically. You won't rotate it every time differently though, it would be a habit - either clockwise, or counter-clockwise. Therefore I think in everyday usage you would require only two options.

I'm very much annoyed by `xrandr` module asking me options that I never use (I always switch between 2 out of 4 available), so I didn't want to be annoyed by "too many options" in this module as well :)



Let me know what you think!

tobes:
> Not really, or I don't fully understand the suggestion. I can retrieve all currently used screens from xrandr (execute xrandr and look where * sign is), but if people have 2 screen setup where one screen extends another, do you think we should rotate all of them at the same time? If I completely misunderstood you, could you give me an example? :)

From what I can tell the `xrandr` module has an active screen (which) you can change (I've not really played with it but looks like it'd be useful)  so I mean your module could just apply to the screen that `xrandr` is showing so `DP1` in your screen shot. 

> The reason is simplicity, you won't do it every day. 

I think that flexibility is good You could easily have default options that have just `normal` and `left` that you want but it would allow others to have the flexability they need. and it's not really much more work than the `left` `right` you already have


With the `{icon}` if there was more info displayed I would agree with you on the icon thing but currently there are only two options `V` or `H` and we have 3 settings.  Now I'd like to have four options but still it would want 5 settings the way you have it.

tobes:
I tried the module out I got it to rotate vertically but rotating back crashed my whole session and I had to kill i3 to get things usable again.

tobes:
very odd
`xrandr --output LVDS1 --rotate left` works fine both via cli and your module

`xrandr --output LVDS1 --rotate normal` works fine via cli but destroys my screen when run via your module

maximbaz:
Thanks for sticking with me @tobes :) Let me give you some examples:

I have two monitors plugged in. When one of them (eDP1) is disabled (so only DP1 is showing picture), `xrandr` outputs:

```
Screen 0: minimum 8 x 8, current 3840 x 2160, maximum 32767 x 32767
eDP1 connected (normal left inverted right x axis y axis)
   2560x1440     60.00 +
DP1 connected 3840x2160+0+0 (normal left inverted right x axis y axis) 527mm x 296mm
   3840x2160     29.98*+
```

You can tell by looking at `*` sign that I use only DP1. The `xrandr_rotate` will therefore rotate the DP1, the first (and only) among active.

Now I make one screen extend the other (so both eDP1 and DP1 show something), `xrandr` outputs:

```
Screen 0: minimum 8 x 8, current 6400 x 2160, maximum 32767 x 32767
eDP1 connected 2560x1440+0+0 (normal left inverted right x axis y axis) 310mm x 174mm
   2560x1440     60.00*+
DP1 connected 3840x2160+2560+0 (normal left inverted right x axis y axis) 527mm x 296mm
   3840x2160     29.98*+
```

Again by looking at `*` sign you can tell that both screens are in use. 
**Question is**: which screens should be rotated? eDP1? DP1? Both?

> I think that flexibility is good You could easily have default options that have just normal and left that you want but it would allow others to have the flexability they need. and it's not really much more work than the left right you already have

Let me know if get it right: you suggest to replace the notation of `horizontal` and `vertical` with four `left`, `right`, `top` and `bottom`. That means creating four options `left_icon`, `right_icon`, `normal_icon`, `inverted_icon` instead of the two `horizontal_icon` and `vertical_icon`, and then creating one extra option to configure the level of details, to choose which of the four options should be  there to choose from (so that I can define `normal` and `left` for myself, but someone else can define three).

It is definitely doable, the thing is.... I'm not convinced that we really need this level of granularity. Can you think of a use-case where you would require all four rotations on frequent basis? 

The primary scenario that I have in mind is: normally the screen is horizontal, I see a long text that I wanna read - I turn the screen vertically and read it, then back again to horizontal.



> very odd

Wow that's odd indeed, but thanks for testing the module! :)

As you can see from the code, the module itself is a bunch of `grep` and `cut` calls over `xrandr` output. Can you please check if this command gives you the LVDS1 output as predicted, both when your screen is in horizontal and in vertical rotation?

```
$ xrandr -q --verbose | grep "^' + output + '" | cut -d " " -f5
```

tobes:
> Question is: which screens should be rotated? eDP1? DP1? Both?

I'm suggesting that if `xrandr` module is running then just rotate the screen that it shows.

if it's not running then maybe just the first unless one is set in config

> I'm not convinced that we really need this level of granularity. 

Well this is only my opinion, maybe others will have opinions and the final decision is with @ultrabug 

I think it might be about the context it's running in it is trying to run the correct command I checked that.  I'll investigate a bit more.

maximbaz:
So you are voting for rotating all active screens that `xrandr` detects. It is fine with me, I don't see any reason why the current implementation of selecting one (almost randomly!) is any better - on contrast, rotating all is actually more consistent and "predictable".

Unless any other suggestions will popup in this thread, I will adapt the code.

Thanks a lot @tobes for the suggestions and for testing the module!

maximbaz:
Didn't mention yesterday, for those of you who prefers icon fonts over characters, `horizontal_icon` and `vertical_icon` options are available. The output may look like the following:

selection_056

selection_058

I'll update the head message with these pics.

ultrabug:
First of all, thanks for this nice idea @maximbaz :)

Long thread, lots of questions and good points made. I think that iteration and usage bring the best enhancements. So as much as I like to cover a wide range of use cases, sometimes **simple** is a good enough start.

Thus:
- magically linking a module to another's output is a neat idea but this would complicate the use of the module for users who just-want-to-rotate-their-screen
- multi-monitor can be worked on later or not (see below)
- I would definitely add a `screen` option so users can choose to target a screen of their choosing (and add it as a placeholder)
- I would then add a `hide_if_disconnected` option so that the module can be masked if the screen is unplugged
- The combination of `screen` and `hide_if_disconnected` would allow to virtually construct a multi-monitor support on the bar by showing/masking available screens and using the module multiple times !

As for the `format` options... I'll think it over a bit and report back here :)

tobes:
With my i3 crash issue.

If I attach another physical screen then rotate works as expected.  If I disconnect the screen then i3 becomes frozen

The screen is rotated but the geometry is all wrong and nothing seems to respond.

As before manually running xrandr --rotate on the cli works as expected.  and from my testing the correct command is issue by your module.  

it looks like the xrandr is stalling in some way maybe locking as I see 3 xrandr processes in htop on the console and killing py3stats gives me my screen back

maximbaz:
That's a neat idea to construct multi-monitor support using multiple module instances!

When having two screens extending one another, it can look like this:

selection_059
selection_060

When having multiple screen definitions, the `format` option becomes more useful, for example the pictures above (could be) achieved using the following config:

```
xrandr_rotate eDP1 {
  horizontal_icon = ''
  vertical_icon = ''
  format = 'eDP1: {icon}'
  screen = 'eDP1'
  hide_if_disconnected = true
}

xrandr_rotate DP1 {
  horizontal_icon = ''
  vertical_icon = ''
  format = 'DP1: {icon}'
  screen = 'DP1'
  hide_if_disconnected = true
}
```

---

Guys, do we all agree that when `screen` option is **not** specified, the module should by default rotate **all** available screens?

---

Thanks @tobes for the details, I'll try to see if I can find something that describes the behavior on your system and/or try to repro. If you get more details, please share :+1:

maximbaz:
Had a chance to play with the code, I certainly like the idea of having a way to specify `screen` and `hide_if_disconnected` options as I personally only rotate the external screen (when it is connected). 

I'll show you how it looks, let me know what you think :)

#### Default configuration (except the icons :smile:)

When a single screen is connected, the behavior is intuitive, it is clear what will be rotated:

selection_061

selection_062

Having two screens connected with the default setup means that the icon will represent and configure *both screens at the same time*:

**Both screens** are in horizontal rotation:

selection_063

**Both screens** are about to be rotated vertically:

selection_064


#### Configuring screens separately with `screen` option

Since I have two screens, I configure two `xrandr_rotate` modules, like this:

```
xrandr_rotate eDP1 {
  format = 'eDP1: {icon}'
  screen = 'eDP1'
}
 
xrandr_rotate DP1 {
  format = 'DP1: {icon}'
  screen = 'DP1'
}
```

When two monitors are connected, I see their configuration:

selection_066

With this configuration I can separately configure **one of the two monitors** to be rotated vertically:

selection_065

When I disconnect the external monitor, it is shown as yellow:

selection_067

#### Hiding disconnected screens with `hide_if_disconnected` option:

Extending the configuration with `hide_if_disconnected` allows me not to see the DP1 rotation icon when this display is disconnected:

```
xrandr_rotate eDP1 {
  format = 'eDP1: {icon}'
  hide_if_disconnected = true
  screen = 'eDP1'
}
 
xrandr_rotate DP1 {
  format = 'DP1: {icon}'
  hide_if_disconnected = true
  screen = 'DP1'
}
```

selection_068


-----

In my case, I have a laptop and I sometimes connect an external monitor to it. I only ever want to rotate the external monitor, never the laptop screen. Therefore, my personal configuration would be the following:

```
xrandr_rotate {
  screen = 'DP1'
  hide_if_disconnected = true
}
```

When the external screen is connected, I have the ability to rotate it:

selection_069

When I disconnect the external screen from my laptop, this icon is gone to save the space:

selection_070

tobes:
Looking good.

One thing I have noticed is that the code uses quite a lot of resources (I have a debug module) I think this is due to the multiple calls to `xrandr -q --verbose`. `xrandr_rotate()` calls 4 times this could easily be reduced.

Your latest commit reduces it potentially but only if screen is set

maximbaz:
Valid point, thanks @tobes :+1: 
Should be better with 6301293

py3status/modules/xrandr_rotate.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+
+"""
+Switch between horizontal and vertical screen rotation on a single click.
+
+Configuration parameters:
+    - cache_timeout: how often to refresh this module.
```
tobes: being picky,

can you remove the `- ` from these parameters, and if you are really nice put the default in parenthesis

like this

```
    format: a string that formats the output, can include placeholders.
        (default is '{icon}')
```

This helps with creating the README.md

I need to document this somewhere

this is also the case with the placeholders

eg

```
    {icon} a rotation icon, ...
```

maximbaz: Didn't know that, will adjust :+1:

tobes:
> Should be better with 6301293

Slightly, but can you not just do `xrandr -q --verbose` once per call to  get all the output and then use `re` to get all the data you need.

If not then at least get `_get_all_outputs()` to only use `xrandr -q`

maximbaz:
You are right, there's no need to use `--verbose` flag. Could you check if the performance is acceptable for you now?

I would like to avoid rewriting greps and cuts using `re` unless there is a big need for that :smile:

tobes:
To be honest I'm more keen to fix the crash I get.

I'm off out now, so probably won't get a chance to check things till tomorrow or Friday.  I'll see if your new version helps at all.

I'll let you know how the performance has improve too, really it should be about on par with the xrandr module.

tobes:
The performance is much nicer now uses approx the same amount of time as xrandr.py whereas before it was using a factor more.

tobes:
I've investigated my crash issue some more.

The `xrandr --output LVDS1 --rotate normal` thread ends up in  traced or suspended state `T` in htop.  If I send a `SIGCONT` to it then everything sorts itself out (well it becomes a zombie till output read by your module)

I'm still not sure how/why this happens but at least am getting closer to the issue.

maximbaz:
@tobes could you check if 393740a helps?

tobes:
I think that looks nicer unfortunately it doesn't fix the issue.

I've tried similar things, and sending SIGCONT to the process but so far no luck.  I don't think this is a blocker but might be worth noting that there may be issues for some users in the docstring.

py3status/modules/xrandr_rotate.py
```diff
@@ -29,6 +29,13 @@
     {icon} a rotation icon, specified by `horizontal_icon` or `vertical_icon`.
 
 
+Remarks:
+    We have seen a case when rotating a screen using this module crashed i3 session.
```
tobes: I'd change this to something like "there have been cases where this causes usability issues and make break your i3 session.  But for most users it works great", as trying to explain the issue is hard. i3 doesn't crash but becomes unusable till sending SIGCONT to the xrandr process via a vterm is just too complicated to explain easily. Or maybe add `Experimental` @ultrabug may have a view.

maximbaz: I'm not sure the `Experimental` tag is helpful, because it is not clear at which point we should remove this tag. I don't think we have any statistics on how many people successfully use any specific modules, do we?

In my mind, if we are to write something, it should be a message with very specific instructions, like "please look for *this* and *this*, and if you see *this*, do exactly *this* and *this*". Or don't write anything at all, let people create Github issues with their descriptions, and we will investigate case by case.

maximbaz:
I think we need more people to try out this module. It is sad that we know of an issue, that is not reproducible by everyone and we are not sure how to fix it.

@ultrabug, first of all, could you please check if it works for you?

Perhaps we should merge the module and look for people creating issues on Github?

pydsigner:
@maximbaz Along those lines, how about we open an issue and link to that in the module until it's resolved?

maximbaz:
That's a great idea as it will (hopefully) reduce the number of duplicate bugs and collect all the information in a single Github issue!

ultrabug:
Ok, I tried it out !
- with no configuration AKA "rotate all the screens" mode, it crashed everything (had to hard reboot). Issue #227 is a good idea to start with.
- with a screen configured, it works like a charm :+1: 

One last thing : I think we should have a {screen} placeholder so you don't have to specify it by hand on the `format` option ;)

maximbaz:
> with no configuration AKA "rotate all the screens" mode, it crashed everything

Wow! :eyes:

> I think we should have a {screen} placeholder so you don't have to specify it by hand on the format option ;)

Thought of that as well, but it becomes unclear again what value this option should have in the default "rotate all screens" mode... None? All screens joined by comma? Other character? Need it to be configured? :smile:

ultrabug:
> Thought of that as well, but it becomes unclear again what value this option should have in the default "rotate all screens" mode...

Just `ALL` ?

maximbaz:
> Just `ALL` ?

Or that :) Should it still be `ALL` in case a user has *one* screen, or want to be smart and detect the output name for that specific case?

ultrabug:
> Or that :) Should it still be ALL in case a user has one screen, or want to be smart and detect the output name for that specific case?

If you have a **simple** way of detecting multiple screens vs solo, I'd do it yes.

maximbaz:
@ultrabug: 10d9ed5 :gift:

ultrabug:
Thanks @maximbaz and everyone who helped in refining this module :+1:

PR #218, opened 2016-03-22
Bugfix: modules that set a cache_until of 0

tobes:
It seems some modules set a `cache_until` of `0` eg `window_title_async` 

This causes a delay of `config['cache_timeout']`

This patch makes the module update as soon as possible which is what the module is intending.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -255,7 +255,7 @@ class Module(Thread):
                         obj['cached_until'] = time()
                     group_module.run()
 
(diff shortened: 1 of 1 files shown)


pydsigner:
Sounds good to me.

ultrabug:
@tobes : not related, but it seems that module clicking on a group is broken :(

Anon1234:
@tobes OMG it's so fast now! TY very much.

PR #219, opened 2016-03-22
Bugfix group widths when items not ready.

tobes:
@ultrabug fix for the click issue.

basically if the items weren't initialized we crashed :(

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -161,7 +161,26 @@ class I3status(Thread):
                     key = self.eval_config_parameter(key)
                     value = line.split('=', 1)[1].strip()
                     value = self.eval_config_value(value)
-                    config[group_name][key] = value
+                    if not key.startswith('on_click'):
+                        config[group_name][key] = value
+                    else:
+                        # on_click special parameters
+                        try:
+                            button = int(key.split()[1])
+                            if button not in range(1, 6):
+                                raise ValueError('should be 1, 2, 3, 4 or 5')
+                        except IndexError as e:
+                            raise IndexError(
+                                'missing "button id" for "on_click" '
+                                'parameter in group {}'.format(group_name))
+                        except ValueError as e:
+                            raise ValueError('invalid "button id" '
+                                             'for "on_click" parameter '
+                                             'in group {} ({})'.format(
+                                                 group_name, e))
+                        on_c = config['on_click']
+                        on_c[group_name] = on_c.get(group_name, {})
+                        on_c[group_name][button] = value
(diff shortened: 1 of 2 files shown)


tobes:
I wasn't seeing the error as my group had a clock in it so it refreshed after a second which fixed the issue.  Anyhow I think it'll be all good now.

ultrabug:
Thanks for your quick reply @tobes but it doesn't seem to fix the problem.

log : `on_click failed with ('NoneType' object has no attribute 'click_event') for event ({u'y': 14, u'instance': u'disks', u'button': 1, u'name': u'group', u'x': 5233})`

config : 
```
order += "group disks"

group disks {
	fixed_width = False
	format = "{output}"

	disk "/home" {
		format = "~ %free"
		on_click 1 = "exec thunar ~"
	}

	disk "/" {
		format = "/ %free"
		on_click 1 = "exec thunar /"
	}
}
```

tobes:
I see.  The module.on_click() is covered but not the config based ones.  I should be able to fix this shortly.

tobes:
Ok, this should fix the issue for click events defined in the config.

I am assuming that if an on_click is defined in the config that it over-rules the on_click() of the module

ultrabug:
> I am assuming that if an on_click is defined in the config that it over-rules the on_click() of the module

stupid question: can't they be executed both ? (meta meta meta) ;)

ultrabug:
I can confirm this fixes the click issue, :+1:

tobes:
> stupid question: can't they be executed both ? (meta meta meta) ;)

they can and they now do - why is another question but it's doable.

I also found another issue around initially displaying slow updating py3status modules.  The fix for that is here too.

ultrabug:
Cool thanks, the why is just that it would seem consistent and possibly a cool feature :)

Almost there !
`on_click failed with ('NoneType' object has no attribute 'click_event') for event ({u'y': 7, u'instance': u'disks', u'button': 1, u'name': u'group', u'x': 5195})`

```
group disks {
	fixed_width = False
	format = "{output}"

	on_click 1 = "exec thunar /tmp"

	disk "/home" {
		format = "~ %free"
		on_click 1 = "exec thunar ~"
	}

	disk "/" {
		format = "/ %free"
		on_click 1 = "exec thunar /"
	}
}
```

tobes:
This is now done. It would be nice to unpick the events a bit at some point but this now works as you have commanded

ultrabug:
Oh I don't command anything mate, I was merely asking thank you !

PR #220, opened 2016-03-22
warn on first module Exception

tobes:
This commit causes a nagbar when a module has an exception in a method.  It only does this once so as not to be too annoying.

It made the group issue easy to find and I think is probably good to have so that we are more aware of errors like that.


pydsigner:
The py3_wrapper passed to modules is already proving useful I see :) Looks good to me.

ultrabug:
Very nice yes, one day we should implement an option to choose between nagbar and dbus notifications

PR #224, opened 2016-03-25
Time fix for #221

### LINKED ISSUE #221: Time doesn't update unless using default format '%Y-%m-%d %H:%M:%S'
I updated from py3status 2.6 to 2.9 recently and noticed that the displayed time doesn't update as usual. It seems the problem started with version 2.7 as I have the same problem with this version.

I've been displaying the time formatted as 'w%V %Y-%m-%d %H:%M:%S' to include the week number as well. But this seems to make the displayed time to update only once every minute (if using the computer). When I unlocked i3lock this morning the time was stuck at 15:45, the time I locked the screen yesterday.

I believe the issue started with this commit: https://github.com/ultrabug/py3status/commit/870491ab89413d0395358c13fde99e7e95abcda8

I have not yet been able to fix the problem myself.

tobes:
This fixes #221 

* We now force a format onto i3status so that it can easily be parsed for time offset and timezone.

* User can use any format without affecting the above

* We only update if we need to so only once a min if no %S for example

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -347,6 +347,36 @@ class Py3statusWrapper():
         else:
             self.queue.append(update)
 
+    def create_output_modules(self):
+        """
+        Setup our output modules to allow easy updating of py3modules and
+        i3status modules allows the same module to be used multiple times.
+        """
+        config = self.i3status_thread.config
+        i3modules = self.i3status_thread.i3modules
+        output_modules = {}
+        # position in the bar of the modules
+        positions = {}
+        for index, name in enumerate(config['order']):
+            if name not in positions:
+                positions[name] = []
+            positions[name].append(index)
+
+        # py3status modules
+        for name in self.modules:
+            if name not in output_modules:
+                output_modules[name] = {}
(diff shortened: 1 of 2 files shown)


pydsigner:
I'm not super happy with this. It needs to be done for now, but we're continuing to answer the wrong question; let's make a pure py3status module that uses pytz and let this module just go back to the time cycle of i3status.

tobes:
True but I'd like to clean the config parsing first so we get lists. But
maybe I'll just offer up the simple version first.

I still think this is important for new users as it is the initial wow
after using i3status and good for compatibility.

Also pytz is a dependency then

tobes:
> let's make a pure py3status module

This is the current state of my clock module https://github.com/tobes/py3status/blob/clock-module/py3status/modules/clock.py which I'll probably make a pull request for once the number of open ones have reduced.

ultrabug:
What can I say @tobes... you da master, it's very elegant even if I agree with you guys that we need to find a way to get rid of all this stuff once and for all :(

PR #225, opened 2016-03-25
add screenshot_command configuration option

tjaartvdwalt:
Short version:
Adds the ability to configure what application is used to take the screenshot 

Longer version:
I do not use Gnome, and gnome-screenshot was not working for me when using the plugin. I like the idea behind the plugin, so I thought I would extend it to allow the screenshot command to be configurable.

### DIFF
--- a/py3status/modules/screenshot.py
+++ b/py3status/modules/screenshot.py
@@ -9,6 +9,7 @@ When the screenshot has been taken, 'SHOT' is replaced by the file_name.
 This modules uses the 'gnome-screenshot' program to take the screenshot.
 
 Configuration parameters:
+    screenshot_command: the command used to generate the screenshot
     file_length: generated file_name length
     push: True/False if yo want to push your screenshot to your server
     save_path: Directory where to store your screenshots.
@@ -29,6 +30,7 @@ class Py3status:
     """
     """
     # available configuration parameters
+    screenshot_command = 'gnome-screenshot -f'
     cache_timeout = 5
     save_path = '%s%s' % (os.environ['HOME'], '/Pictures/')
     file_length = 4
@@ -44,7 +46,7 @@ class Py3status:
 
         file_name = self._filename_generator(self.file_length)
 
-        command = '%s %s/%s%s' % ('gnome-screenshot -f', self.save_path,
+        command = '%s %s/%s%s' % (self.screenshot_command, self.save_path,
                                   file_name, '.jpg')
 
         subprocess.Popen(command.split())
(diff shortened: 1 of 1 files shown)


tobes:
Looks good to me.

py3status/modules/screenshot.py
```diff
@@ -9,6 +9,7 @@
 This modules uses the 'gnome-screenshot' program to take the screenshot.
```
tobes: Maybe change this to say other programs can be used too

pydsigner:
Well done, now I might actually use this module. (I have to take screencaps on a regular basis at work.)

tobes:
Thanks for fixing the docstring :)

If you are feeling brave then you could also change things to allow a custom format rather than just 'SHOT' so I could use `📷` although that is a crap icon on my screen.  @Horgix may have thoughts, see issue #98. If not I think this is still a valid improvement on it's own.

tobes:
Also `cache_timeout = 5` makes no sense now.  when clicked the update should take 0.1 second on a bad day, so there is no need for this.  It could never update and be fine but a default of 3600 (1 hour) would be more sensible than every 5 seconds.

pydsigner:
Good points @tobes

tjaartvdwalt:
I will make the change to the format. I was also wondering about 2 other improvements:

**How can the interface allow for taking screenshots of a selected area?**
`gnome-screenshot -a` or `scrot -s` allows for this from the command side, but how do we expose it to the user? Maybe we have a separate right click?

**Copy the file path/url to the clipboard**
The url would be most useful, since I could paste it into a chat window. The ultimate solution for me would be to copy the screenshot into my Dropbox directory, share the file, and copy the url into the clipboard. At this stage I don't know how to achieve that goal.

Any thoughts?

tobes:
> How can the interface allow for taking screenshots of a selected area?

You could add a `right click` option and have a `screenshot_selection_command`.  Personally I like the idea of user settable buttons like https://github.com/ultrabug/py3status/tree/master/py3status/modules#group but I think that's the only module with them.

> Copy the file path/url to the clipboard

Something like `xclip` should work but having a user changeable command would be nice.  maybe you want to copy to  `CLIPBOARD` and I want it in `PRIMARY`

maybe have a `on_screenshot_command` type thing so `echo -n {filename} | xclip  -selection c`

pydsigner:
Hmm, this is beginning to sound like a generic click-to-run-command module with screenshot defaults.

tobes:
> Hmm, this is beginning to sound like a generic click-to-run-command module with screenshot defaults.

True, but giving the user the power to tweak things is good is it not?  As long as the module is easy to use without having to configure. but I take your point.

I suppose the ideal is to sniff for what is available and use that.  So doing `which gnome-screenshot` etc and then running the command if it's there would be nice.

ultrabug:
Thanks @tjaartvdwalt, let's say this is enough as-is as it is a cool improvement for users :+1:

PR #226, opened 2016-03-26
Module helpers, notify-send and cleaner clicks

tobes:
In some ways this should be separate pull requests but there is some overlap.  I've tried to keep the commits independent, but they aren't quite.

What this does

**events**

* As shown by the group click events fun, the events logic in py3status is quite complex.  This patch moves the event processing out of the main loop into it's own function.  This allows events to be triggered on modules using the same code path.  The groups module may be the only user, other than the event loop, but it means that they will stay consistent if the logic is changed.  eg group clicks now use `i3bar_click_events.py` fallback .

**notify-send**

* `i3_nagbar()` is renamed to `notify_user()` all parameters stay as before.

* New `--dbus-notifications` option for `py3status` command if set this switches from using the i3 nagbar to using `notify-send` currently there is a fixed timeout of 10 seconds for these messages.

**module helpers**

* Now that modules can get hold of their instance class, they can access all of py3status.  While this is good on many levels, if users start relying on py3status methods etc then this makes refactoring hard and becomes a potential support nightmare.

* `Module.py3_wrapper` becomes `Module._py3_wrapper` along with comment that its use is not supported.

* some helper functions are added to `Module` to allow useful functionality

  *helper_get_module_info(module_name)* - Get info for named module.

  *helper_trigger_event(module_name, event)* - Trigger the event on named module

  *helper_notification(msg, level)* - Send notification to user.

The names could probably be improved.

### DIFF
--- a/py3status/__init__.py
+++ b/py3status/__init__.py
@@ -17,11 +17,11 @@ def main():
         py3.setup()
     except KeyboardInterrupt:
         err = sys.exc_info()[1]
-        py3.i3_nagbar('setup interrupted (KeyboardInterrupt)')
+        py3.notify_user('setup interrupted (KeyboardInterrupt)')
         sys.exit(0)
     except Exception:
         err = sys.exc_info()[1]
-        py3.i3_nagbar('setup error ({})'.format(err))
+        py3.notify_user('setup error ({})'.format(err))
         py3.stop()
         sys.exit(2)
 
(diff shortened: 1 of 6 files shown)


ultrabug:
neat, please rebase so I can test ;) thx !

tobes:
I've updated this branch

ultrabug:
This is indeed very hard to review seeing all these changes :( so I'm testing it for real instead.

I just think that nobody will use such a long `--dbus-notifications` command line option.
- I'd rename it `--dbus-notify`
- I would add a short option to it like say `-b` ?

tobes:
I've changed the option to `--dbus-notify` or `-b`.

I also changed the internal `Core.config` name to `dbus_notify` to keep them in sync.

Sorry about the size of the PR I'll try to keep them smaller/more specific in future.

ultrabug:
Another big one in ! thanks @tobes !

PR #231, opened 2016-03-29
prevent error when group is empty

tobes:
If a group is empty we get a modulo zero error when it cycles.  This prevents this from happening.

example config to cause issue
```
group bad {
    cycle = 1
}
```

### DIFF
--- a/py3status/modules/group.py
+++ b/py3status/modules/group.py
@@ -137,6 +137,15 @@ class Py3status:
         """
         Display a output of current module
         """
+
+        # hide if no contents
+        # FIXME the module shouldn't even run if no content
+        if not self.items:
+            return {
+                'cached_until': time() + 36000,
+                'full_text': '',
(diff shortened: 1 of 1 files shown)


ultrabug:
If I speculate correctly, that means we would load and update an empty group module.

Shouldn't we ignore it altogether at parsing time to save CPU ?

tobes:
yes that's true but this just removes the error.

Currently we have some issues.  If I have a py3module defined in the config but not in `order +=` and not in a group (special case) then it still runs even though it's output is never used.

So I'd like to do that fix soon.  That would also be a good place to not run empty groups.  As it is the empty group shouldn't update much.  It still gets displayed though :(

tobes:
Ok empty groups hide themselves pending a proper fix

ultrabug:
Thanks, LGTM for now

PR #233, opened 2016-03-29
Combined Improvements

neutronst4r:
Sorry for dumping them all in one pull request, if neccessary I can split them up. I requested theses changes before, they are not new. They have just been rebased on the new master. It took me a while to lern git command line. The changes are:

- added a fix for faulty acpi output in battery_level.py
- added geolocation option in whatismyip.py
- changed dpms.py to work with xscreensaver instead, added new module xscreensaver.py


pydsigner:
They'd be easier to handle broken up. Pretty please? :)

tobes:
Yeah, git can be a bit painful when you're first learning it.  It's fantastic when you get the hang of it.

As @pydsigner said breaking them up into one pull request for each change really helps.  You might also find that it's easiest to start with an up-to-date master and then branch from that. and add the changes you want again.

Looking at `battery_level.py` you have actually reverted the documentation changes that were the cause of the merge conflict.  It also looks like the changes you want are only adding 11 lines, rather than adding 73 and removing 62.  Smaller patches are always nicer.

Also in `whatismyip.py` the blank lines will break the `modules/README.md` updating (not that you'd know that).

Glad to see these patches back though.

PR #236, opened 2016-03-30
Proposed module xscreensaver.py

neutronst4r:
Rewrote the dpms.py module to work with xscreensaver instead.


py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,86 @@
...
+        Display a colorful state of xscreensaver.
+        """
+        self.run = call(["pidof", "xscreensaver"],
+                        stdout=DEVNULL,
+                        stderr=DEVNULL,
+                        preexec_fn=setpgrp) == 0
+        return {
+            'full_text': self.format_on if self.run else self.format_off,
+            'color': self.color_on or i3s_config['color_good']
+            if self.run else self.color_off or i3s_config['color_bad']
```
tobes: I might move the color calculation out of the return dict to make things more readable, maybe combine with the text selection.

Also adding a `cache_until` is nice to control how often we update so not kill the cpu.

`'cached_until': time() + self.cache_timeout` is common practice with `cache_timeout` as a config option. you'd need a `from time import time` too.

tobes:
I really like this!

The `dpms` module is the main reason I started using py3status (I got bored of having my screen blank when watching films)

if you could change 
```
Configuration parameters:
    - format_off: string to display when Xscreensaver is disabled
    - format_on : string to display when Xscreensaver is enabled
```

to 

```
Configuration parameters:
    format_off: string to display when Xscreensaver is disabled
    format_on: string to display when Xscreensaver is enabled
```

it would be great.  It helps generating `modules/README.md`

py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,86 @@
...
+"""
+Indicator for Xscreensaver, can be toggled by left clicking.
+
+This is a partial rewrite of dpms.py. Most credit
+goes to the original author Andre Doser. This module shows the status
+of Xscreensaver and activates or deactivates it upon left click.
+
+This script is useful for people who let Xscreensaver manage dpms
+settings. Xscreensaver has its own dpms variables separate from xset.
+Dpms can be savely turned of in xset as long as xscreensaver is running.
```
tobes: > savely turned of

sa**f**ely turned of**f**

neutronst4r:
- formatted the comment section
- added the cache timeout, please have a look, if it is correctly implemented, I don't know how to test, if it is working
- moved colour calculation for better visibility

py3status/modules/xscreensaver.py
```diff
@@ -8,12 +8,14 @@
...
 settings. Xscreensaver has its own dpms variables separate from xset.
-Dpms can be savely turned of in xset as long as xscreensaver is running.
+Dpms can be safely turned off in xset as long as xscreensaver is running.
 Settings can be managed using "xscreensaver-demo".
 
 Configuration parameters:
-    - format_off: string to display when Xscreensaver is disabled
-    - format_on : string to display when Xscreensaver is enabled
+    format_off   : string to display when Xscreensaver is disabled
+    format_on    : string to display when Xscreensaver is enabled
```
tobes: There cannot be spaces between the config name and the `:` you can have as many spaces after as you want.

We look for `    <name>:`

tobes:
Looking better :+1: 

The `cache_timeout` looks good though you might want to check more often dmps checks every 5 seconds.  But it's only if it gets changed outside py3status that you'd see them out of sync.  And the user can set it so if you like 30 then have 30

py3status/modules/xscreensaver.py
```diff
@@ -0,0 +1,93 @@
...
+        return {'full_text': full_text,
+                'color': color,
+                'cached_until': time() + self.cache_timeout}
+
+    def on_click(self, i3s_output_list, i3s_config, event):
+        """
+        Enable/Disable xscreensaver on left click.
+        """
+        if event['button'] == 1:
+            if self.run:
```
tobes: There is a small chance that the state might of changed since `self.run` was set.  It might be good to check the state when we do the action just to be sure.

maybe move it info it's own function so the code can be shared, just remember to get the function name to start with `_` so it isn't run by py3status.

neutronst4r:
Personally, I would set it even much higher, maybe 5 minutes or so, since I don't turn off the screen saver by other means other than py3status. As you said, the only time it will ever be out of sync is, when the process gets shutdown my other means. And even then... the worst that could happen is that the screensaver starts. A simple reload of I3 will resync the bar. But alright, I will change the default to 5s for consistency.

tobes:
> >  There is a small chance that the state might of changed since self.run was set.

> Could you elaborate on that?

if the cache_timeout is once every 30 seconds (or even 5) and you only check the state of the screensaver then, if it changes between then and the click event, then you will set it when it is already set.

It's not a huge deal here but it is nice to have the module 'correct'

if `on_click()` does
```
        self.run = call(["pidof", "xscreensaver"],
                        stdout=DEVNULL,
                        stderr=DEVNULL,
                        preexec_fn=setpgrp) == 0
```
then the value will be correct

tobes:
@neutronst4r what is happening with this PR?

There are unresolved issues and we would need to update things for the new standards/tests etc

PR #237, opened 2016-03-30
Added geolocation support for whatsmyip.py using "ip-api.com"

neutronst4r:
Added a new format option {country}, that, if present, will show ones outside location. Useful when using VPN.

Removed  the previous empty lines in the comment section, that were supposedly breaking the Readme.

### DIFF
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -57,7 +57,7 @@ class Py3status:
     def _get_my_ip(self):
         """
         """
-        country = None
+        country = ""
         try:
             if '{country}' in self.format:
                 self.url = self.url_geo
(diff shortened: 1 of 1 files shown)


tobes:
`color_good_online_ip_mode` config.

Do we actually need this?  I'm not sure that just showing things as good for all users is a bad thing.  It doesn't break anyone's setup.  I'd keep it for now and see if others have an opinion.

py3status/modules/whatismyip.py
```diff
@@ -49,25 +57,33 @@ def on_click(self, i3s_output_list, i3s_config, event):
     def _get_my_ip(self):
         """
         """
+        country = None
```
tobes: In other modules we have used `?` when things are unknown eg battery_level.

`None` will get shown as `None` so if you just want the county to be blank use `''`

ultrabug:
Please address @tobes comment @neutronst4r and we'll merge.

tobes:
@neutronst4r this PR seems to have stalled, and there are conflicts between it and master.  Is this still a feature that you are still wanting?  If so please resolve the conflicts and we can restart the review process.

PR #239, opened 2016-04-01
fix for xrandr_rotate issue

### LINKED ISSUE #227: Module "xrandr_rotate" makes i3 unusable
We have seen a case (see #217) when i3 becomes unusable after rotating screen using `xrandr_rotate` module. The symptoms are: `xrandr` process is stalled and only sending SIGCONT to it makes i3 usable again.

This issue is not reproducible for everyone. If you experience a similar behavior, please post here as many details as you can to help us investigate and fix this issue.

tobes:
@maximbaz 

I have done some playing around and have managed to fix the issue that I was having with `xrandr_rotate`.  The fix is to run the command via `i3-msg` I'm not quite sure why this works but it does.

Could you see if it works for you too.

This fixes  #227 as far as I'm concerned, unless there have been other problems reported.  And we should remove the warning in the docstring too.

### DIFF
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -84,8 +84,8 @@ class Py3status:
         rotation = self.horizontal_rotation if self.displayed == self.horizontal_icon else self.vertical_rotation
         outputs = [self.screen] if self.screen else self._get_all_outputs()
         for output in outputs:
-            cmd = 'xrandr --output ' + output + ' --rotate ' + rotation
-            self._call(cmd)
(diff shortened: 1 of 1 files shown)


maximbaz:
Just checked, it works for me as well :+1: 

@ultrabug, you had another crash using this module, only in "rotate all the screens" mode. Could you double check if it is still an issue? We don't have any other issues reported apart from that one.

ultrabug:
Will try tomorrow and merge ;) thx

ultrabug:
Didn't crash, worked \o/

PR #240, opened 2016-04-01
Fix not initialized 'displayed' field in xrandr_rotate

maximbaz:
Discovered a couple of small glitches, this is a fix for them.

Repro steps:
- set `screen` property to an output name that is not currently connected.
- set `hide_if_disconnected` to true.
- restart i3

Expected: icon will not show up
Actual: the module crashes, since `self.displayed` is not initialized but later is being accessed.

-----

And another super-tiny one:

- set `screen` property to an output name that is not currently connected.
- set `hide_if_disconnected` to true.
- while in H mode, left click to select V mode, but do not apply it
- disconnect the screen (icon hides)
- connect the screen (icon appears)

Expected: icon shows the current rotation of the screen (H)
Actual: icon remembers your choice and still shows V even though the screen is in H mode.

### DIFF
--- a/py3status/modules/xrandr_rotate.py
+++ b/py3status/modules/xrandr_rotate.py
@@ -58,6 +58,9 @@ class Py3status:
     vertical_icon = 'V'
     vertical_rotation = 'left'
 
+    def __init__(self):
+        self.displayed = ''
+
     def _call(self, cmd):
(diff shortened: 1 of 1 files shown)


ultrabug:
:+1:

PR #242, opened 2016-04-03
Toggle monitor output from the status bar using xrandr.

mrt-prodz:
Hello,

I have three monitors connected to my computer but I only use one of them
from time to time when I want to watch a movie for example. (it's a TV)

I thought it would be really handy to be able to quickly toggle the output for that
monitor from the status bar instead of running any xrandr command manually
each time.

The user can either use the index number of the output from xrandr (starting from 1)
or the source id directly. Positioning the output is also possible.

A configuration example would be as such:

```
xrandr_toggle {
    # we could use the monitor number from xrandr output
    # name = "1"
    # or the monitor id directly
    source = "HDMI-2"
    position = "left-of DVI-I-1"
    format = "[{monitor}:TV32\"]"
}
```

Clicking on it would toggle on/off the output for that source.

Note: I will certainly avoid toggling the source output if there is no other active display.

### DIFF
--- a/py3status/modules/xrandr_toggle.py
+++ b/py3status/modules/xrandr_toggle.py
@@ -1,19 +1,19 @@
 # -*- coding: utf-8 -*-
 """
-Toggle monitor output from the status bar using xrandr.
+Toggles a monitor on or off from the status bar using xrandr.
 
 Configuration parameters:
     cache_timeout: how often we refresh usage in seconds (default: 10s)
-    distext: custom text or symbol for disconnected status
     format: see placeholders below
+    format_disconnected: custom text or symbol for disconnected status
+    format_off: custom text or symbol for off status
+    format_on: custom text or symbol for on status
     number: number of the monitor we want to query
-    offtext: custom text or symbol for off status
-    ontext: custom text or symbol for on status
     position: where the monitor will be toggled (ex: left-of HDMI-1)
     source: name of the source we want to query (ex: HDMI-2)
 
 Format of status string placeholders:
-    {monitor} monitor source name
+    {screen} screen source name
     {status} on/off status
 
 Example config:
@@ -25,7 +25,7 @@ xrandr_toggle {
     # or the monitor id directly
     source = "HDMI-2"
     position = "left-of DVI-I-1"
-    format = "[{monitor}:TV32\"]"
+    format = "[{screen}:TV32\"]"
 }
 ```
 
@@ -39,11 +39,11 @@ from time import time
 class Py3status:
     # available configuration parameters
     cache_timeout = 10
-    distext = 'disconnected'
     format = '{source}:{status}'
+    format_disconnected = 'disconnected'
+    format_on = 'on'
+    format_off = 'off'
     number = '1'
-    offtext = 'off'
-    ontext = 'on'
     position = ''
     source = ''
 
@@ -85,15 +85,15 @@ class Py3status:
                         if details[1] == 'disconnected':
                             source['connected'] = 0
                             source['state'] = 'off'
-                            source['status'] = self.distext
+                            source['status'] = self.format_disconnected
                         else:
                             source['connected'] = 1
                             if search(regex, details[2]):
                                 source['state'] = 'auto'
-                                source['status'] = self.ontext
+                                source['status'] = self.format_on
                             else:
                                 source['state'] = 'off'
-                                source['status'] = self.offtext
+                                source['status'] = self.format_off
                         monitors.append(source)
 
             # index bigger than array of monitors found
@@ -121,7 +121,7 @@ class Py3status:
             response = {
                 'cached_until': time() + self.cache_timeout,
                 'color': color,
-                'full_text': self.format.format(monitor=self.mon['id'],
+                'full_text': self.format.format(screen=self.mon['id'],
                                                 status=self.mon['status'])
             }
             return response
--- a/py3status/modules/xrandr_toggle.py
+++ b/py3status/modules/xrandr_toggle.py
@@ -3,18 +3,18 @@
 Toggle monitor output from the status bar using xrandr.
 
 Configuration parameters:
-    - cache_timeout : how often we refresh usage in seconds (default: 10s)
-    - distext : custom text or symbol for disconnected status
-    - format : see placeholders below
-    - number : number of the monitor we want to query
-    - offtext : custom text or symbol for off status
-    - ontext : custom text or symbol for on status
-    - position : where the monitor will be toggled (ex: left-of HDMI-1)
-    - source : name of the source we want to query (ex: HDMI-2)
+    cache_timeout: how often we refresh usage in seconds (default: 10s)
+    distext: custom text or symbol for disconnected status
+    format: see placeholders below
+    number: number of the monitor we want to query
+    offtext: custom text or symbol for off status
+    ontext: custom text or symbol for on status
+    position: where the monitor will be toggled (ex: left-of HDMI-1)
+    source: name of the source we want to query (ex: HDMI-2)
 
 Format of status string placeholders:
-    {monitor} : monitor source name
-    {status} : on/off status
+    {monitor} monitor source name
+    {status} on/off status
 
 Example config:
 
--- a/py3status/modules/xrandr_toggle.py
+++ b/py3status/modules/xrandr_toggle.py
@@ -1,0 +1,160 @@
+# -*- coding: utf-8 -*-
+"""
+Toggle monitor output from the status bar using xrandr.
+
+Configuration parameters:
+    - cache_timeout : how often we refresh usage in seconds (default: 10s)
+    - distext : custom text or symbol for disconnected status
+    - format : see placeholders below
+    - number : number of the monitor we want to query
+    - offtext : custom text or symbol for off status
+    - ontext : custom text or symbol for on status
+    - position : where the monitor will be toggled (ex: left-of HDMI-1)
+    - source : name of the source we want to query (ex: HDMI-2)
+
+Format of status string placeholders:
+    {monitor} : monitor source name
+    {status} : on/off status
+
+Example config:
+
+```
+xrandr_toggle {
+    # we could use the monitor number from xrandr output
+    # name = "1"
+    # or the monitor id directly
+    source = "HDMI-2"
+    position = "left-of DVI-I-1"
+    format = "[{monitor}:TV32\"]"
+}
+```
+
+@author mrt-prodz
+"""
+from subprocess import call, check_output
+from re import search
+from time import time
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 10
+    distext = 'disconnected'
+    format = '{source}:{status}'
+    number = '1'
+    offtext = 'off'
+    ontext = 'on'
+    position = ''
+    source = ''
+
+    def __init__(self):
+        self.mon = None
+        self.status = self._xrandr()
+
+    # return error occurs
+    def _error_response(self, color, text):
+        response = {
+            'cached_until': time() + self.cache_timeout,
+            'full_text': 'xrandr_toggle: ' + text,
+            'color': color
+        }
+        return response
+
+    # query xrandr for monitors status
+    def _xrandr(self):
+            return check_output(['xrandr']).split('\n')
+
+    # return connected monitor using xrandr and number for index
+    def get_status(self, i3s_output_list, i3s_config):
+        try:
+            # monitor number (index) must be bigger than 0
+            number = int(self.number) - 1
+            if number < 0:
+                return self._error_response(i3s_config['color_bad'],
+                                            'invalid number')
+            monitors = []
+            # width x height + left + top
+            regex = r'\d{1,}x\d{1,}\+\d{1,}\+\d{1,}'
+            # for each line find (dis)connected monitors
+            for line in self.status:
+                if 'connected' in line:
+                    details = line.split(' ')
+                    if len(details) > 2:
+                        source = {}
+                        source['id'] = details[0]
+                        if details[1] == 'disconnected':
+                            source['connected'] = 0
+                            source['state'] = 'off'
+                            source['status'] = self.distext
+                        else:
+                            source['connected'] = 1
+                            if search(regex, details[2]):
+                                source['state'] = 'auto'
+                                source['status'] = self.ontext
+                            else:
+                                source['state'] = 'off'
+                                source['status'] = self.offtext
+                        monitors.append(source)
+
+            # index bigger than array of monitors found
+            if number >= len(monitors):
+                return self._error_response(i3s_config['color_bad'],
+                                            'number is too high')
+            # if source is set, use that instead of number
+            if len(self.source) > 0:
+                self.mon = next((x for x in monitors
+                                if x['id'] == self.source),
+                                None)
+                if self.mon is None:
+                    return self._error_response(i3s_config['color_bad'],
+                                                'monitor not found')
+            else:
+                self.mon = monitors[number]
+
+            if self.mon['connected'] == 0:
+                color = i3s_config['color_bad']
+            elif self.mon['state'] == 'auto':
+                color = i3s_config['color_good']
+            else:
+                color = i3s_config['color_degraded']
+
+            response = {
+                'cached_until': time() + self.cache_timeout,
+                'color': color,
+                'full_text': self.format.format(monitor=self.mon['id'],
+                                                status=self.mon['status'])
+            }
+            return response
+        except:
+            return self._error_response(i3s_config['color_bad'], 'error')
+
+    # toggle source on/off
+    def on_click(self, i3s_output_list, i3s_config, event):
+        if (self.mon is not None) and (self.mon['connected'] == 1):
+            command = ['xrandr', '--output', self.mon['id']]
+            if self.mon['state'] == 'auto':
+                state = 'off'
+            else:
+                state = 'auto'
+                if len(self.position) > 0:
+                    pos = self.position.split(' ')
+                    if len(pos) == 2:
+                        command.append('--' + pos[0])
+                        command.append(pos[1])
+            command.append('--' + state)
+            call(command)
+            # update status
+            self.status = self._xrandr()
+
+
+if __name__ == "__main__":
+    """
+    Test this module by calling it directly.
+    """
+    x = Py3status()
+    config = {
+        'color_good': '#00FF00',
+        'color_degraded': '#00FFFF',
+        'color_bad': '#FF0000'
+    }
+    print(x.get_status([], config))
--- a/py3status/modules/xrandr_toggle.py
+++ b/py3status/modules/xrandr_toggle.py
@@ -88,7 +88,7 @@ class Py3status:
                         monitors.append(source)
             return monitors
         except:
-            return self._error('error while parsing xrandr')
+            return None
 
     # return connected monitor using xrandr and number for index
     def get_status(self, i3s_output_list, i3s_config):
@@ -96,6 +96,9 @@ class Py3status:
             color_bad = i3s_config['color_bad']
             color_degraded = i3s_config['color_degraded']
             color_good = i3s_config['color_good']
+            # make sure we have stored a monitors list
+            if self.monitors is None:
+                return self._error('xrandr output parsing error', color_bad)
             # monitor number (index) must be bigger than 0
             number = int(self.number) - 1
             if number < 0:
--- a/py3status/modules/xrandr_toggle.py
+++ b/py3status/modules/xrandr_toggle.py
@@ -88,7 +88,7 @@ class Py3status:
                         monitors.append(source)
             return monitors
         except:
-            return self._error('error while parsing xrandr')
+            return self._error('error while parsing xrandr', None)
 
     # return connected monitor using xrandr and number for index
     def get_status(self, i3s_output_list, i3s_config):
--- a/py3status/modules/xrandr_toggle.py
+++ b/py3status/modules/xrandr_toggle.py
@@ -48,11 +48,13 @@ class Py3status:
     source = ''
 
     def __init__(self):
-        self.mon = None
-        self.status = self._xrandr()
+        self.monitor = None
+        self.monitors = self._xrandr()
 
     # return error occurs
-    def _error_response(self, color, text):
+    def _error(self, text, color):
+        if color is None:
+            color = '#FF0000'
         response = {
             'cached_until': time() + self.cache_timeout,
             'full_text': 'xrandr_toggle: ' + text,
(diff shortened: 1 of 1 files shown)


py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+"""
+Toggle monitor output from the status bar using xrandr.
+
+Configuration parameters:
+    cache_timeout: how often we refresh usage in seconds (default: 10s)
+    distext: custom text or symbol for disconnected status
```
tobes: `format_disconnected` might be a nicer name and is in keeping with other modules

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+"""
+Toggle monitor output from the status bar using xrandr.
+
```
tobes: To me this isn't very clear.  What are we toggling?  Something like 'Toggles a monitor on or off' feels more explicit.

Also some more information would be nice.

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
...
+    distext: custom text or symbol for disconnected status
+    format: see placeholders below
+    number: number of the monitor we want to query
+    offtext: custom text or symbol for off status
+    ontext: custom text or symbol for on status
+    position: where the monitor will be toggled (ex: left-of HDMI-1)
+    source: name of the source we want to query (ex: HDMI-2)
+
+Format of status string placeholders:
+    {monitor} monitor source name
```
tobes: `xrandr_rotate` uses `{screen}` it would be nice to use the same name here too.

py3status/modules/xrandr_toggle.py
mrt-prodz: Something like using format_on, format_off and format_disconnected instead of the short text variables?

tobes: exactly

mrt-prodz: I have to admit I have often a hard time keeping the 80 characters limit per line to pass flake8. That's why I ended with these odd names.

tobes: > I have often a hard time keeping the 80 characters limit per line

IMHO, generally it's a sign that you need to refactor the code as it's getting too complicated.
Also strictly speaking it's 79 chars.

When I first started using python I found it a pain with all the indentation and lines drifting off to the right.  But it taught me how to write cleaner nicer code after a couple of years.

mrt-prodz:
@tobes made the changes to the short description and variable names.

py3status/modules/xrandr_toggle.py
mrt-prodz: If we were not talking about Python I would have agree with you, but the language itself forces you to indent too much and thus lose quickly characters to match that really old 80 characters limit 'legacy'.

tobes: > 80 characters limit 'legacy'.

79 and it's not legacy in any way.  just because you can have monitors with 200 chars doesn't mean you should use them.  Long lines are hard to scan.  newspapers have columns because of readability.  Also email destroys lines over 80 chars quite a lot and it's nice sometimes to email code.

Also I tend to have several terminals on a screen so actually 79 chars is often more than a line.  Then there are people with accessibility issues screen readers/ people with poor eyesight.  Hey sometimes I look at code on my phone.  And if I need to scroll right to read code here on github I cry inside.

With the indenting issue it is about writing better code, not complaining about the language.  If you are finding it hard then maybe you need to figure out why that's the case.  We learn some very bad habits in our lives.

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
...
+                color = i3s_config['color_degraded']
+
+            response = {
+                'cached_until': time() + self.cache_timeout,
+                'color': color,
+                'full_text': self.format.format(screen=self.mon['id'],
+                                                status=self.mon['status'])
+            }
+            return response
+        except:
```
tobes: Raw exceptions are horrible things.  If things are breaking it's nice to find out and fix them.

If there is an known issue then just catch that.

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
...
+                return self._error_response(i3s_config['color_bad'],
+                                            'invalid number')
+            monitors = []
+            # width x height + left + top
+            regex = r'\d{1,}x\d{1,}\+\d{1,}\+\d{1,}'
+            # for each line find (dis)connected monitors
+            for line in self.status:
+                if 'connected' in line:
+                    details = line.split(' ')
+                    if len(details) > 2:
```
tobes: You see here you could call a function to calculate the `source`.  It'd solve your indentation issue, and probably make the code easier to follow.

py3status/modules/xrandr_toggle.py
mrt-prodz: I am not saying there shouldn't be a 79 characters limit for other languages, just really annoyed with that limit for languages forcing you to indent at every declaration or loop such as Python. It makes it really easy to pass that number of characters if you have explicit variable and function names.

For example with a 4 space characters indentation, just declaring a class, making a function and having a try catch with a loop or condition and you are already left with 63 characters per line. If on top of that you have a couple explicit names you reach quickly that limit.

In other languages where indentation is not forced on you it's much easier to deal with.
This is of course my personal opinion about Python, flake8 and the 79 characters limitation.

mrt-prodz: One example would be this part:

```
            # if source is set, use that instead of number
            if len(self.source) > 0:
                self.mon = next((x for x in monitors                    
                                if x['id'] == self.source),
                                None)
                if self.mon is None:
                    return self._error_response(i3s_config['color_bad'],
                                                'monitor not found')    
            else:
                self.mon = monitors[number]
```

It looks really confusing to me, I also had `self.monitor` but eventually renamed it to `self.mon` and had to split the loop to fit that limit. The easiest way to avoid having that limitation would have been something like this I guess:

```
            # if source is set, use that instead of number
            if len(self.source) > 0:
                self.monitor = None
                for x in monitors:
                    if x['id'] == self.source:
                        self.monitor = x
                        break
                if self.monitor is None:
                    color_bad = i3s_config['color_bad']
                    message = 'monitor not found'
                    return self._error_response(color_bad, message)
            else:
                self.monitor = monitors[number]
```

Or would you have done this in another manner?

pydsigner: @tobes FYI PEP8 has changed to explicitly allow 100 character lines now. I hear you on github scrolling though.

tobes: > FYI PEP8 has changed to explicitly allow 100 character lines now. 

I'd say it hasn't.

> Limit all lines to a maximum of 79 characters. 

there is an option

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters)

This is one of these things, 10 years ago I'd have said it was stupid.  Now I think it is great.  In my current work flow I don't really  like lines over 60 chars.  Also I think sticking to it makes for much better programmers (though it can be frustrating to get there).  

On the subject `yapf` creates some horrible code trying to make things compliant. 

https://www.python.org/dev/peps/pep-0008/#maximum-line-length

tobes:
```
                self.mon = next((x for x in monitors                    
                                if x['id'] == self.source),
                                None)
```

yeah that looks nasty.

some thoughts

a) create `_get_monitor()` that encapsulates it.

b) for the named monitors create a dict `monitor[name] = ...` then easy to access no need to go through the loop.  I'm not sure how the numbers work.  Do they always map to a screen name?  Anyway you could use an orderedDict and then get by index `list(d.items())[i]`

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
...
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 10
+    format = '{source}:{status}'
+    format_disconnected = 'disconnected'
+    format_on = 'on'
+    format_off = 'off'
+    number = '1'
+    position = ''
+    source = ''
```
tobes: again `xrandr_rotate` uses `screen` would be good to be consistent.  Also could you combine this and number?

eg `screen = 'VGA1"` or `screen = 1` the config would parse these as `str` and `int` or you could do some test to see if it's numeric.

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,160 @@
...
+    def _error_response(self, color, text):
+        response = {
+            'cached_until': time() + self.cache_timeout,
+            'full_text': 'xrandr_toggle: ' + text,
+            'color': color
+        }
+        return response
+
+    # query xrandr for monitors status
+    def _xrandr(self):
```
tobes: This function is only called at `__init__()` and `on_click()` you should probably call it whenever `get_status()` is called just in case it has been changed outside of this module.

The other crucial thing to be aware is that `__init__()` is called when the module is instantiated. This is before the config has been set.  So you are always checking `number = "1"`

py3status/modules/xrandr_toggle.py
tobes: looking at the code a bit more `_xrandr()` should parse the output and set self.status to the only monitor we care about.  There isn't any reason to store the state of the other monitors is there?

py3status/modules/xrandr_toggle.py
mrt-prodz: > This function is only called at __init__() and on_click() you should probably call it whenever get_status() is called just in case it has been changed outside of this module.

If I run `xrandr` too often it causes slow down on my configuration, mouse is freezing at each query. That's why I query during instantiation and then only after a user interaction. I could set another timer in `get_status` and refresh from time to time but here at least I cannot refresh the monitors list on each call.

It is an issue if someone unplug/plug a monitor, maybe assigning a right click event to refresh manually?

> The other crucial thing to be aware is that __init__() is called when the module is instantiated. This is before the config has been set. So you are always checking number = "1"

I'm not sure I understand you, xrandr() is not using any variable beside storing `self.status` to refresh the monitors list.

py3status/modules/xrandr_toggle.py
mrt-prodz: I agree I should handle the exception better here.

py3status/modules/xrandr_toggle.py
tobes: > I'm not sure I understand you, xrandr() and __init()__ are not using any variable beside storing self.status to refresh the monitors list.

yeah you're right there

tobes: > If I run xrandr too often it causes slow down on my configuration

Does this just happen via py3status or also if you run it in a terminal?  I had a problem with `xrandr_rotate` causing me problems.

py3status/modules/xrandr_toggle.py
mrt-prodz: I really wanted to avoid calling `xrandr()` too often because of the issue I mentioned in my other note. It's slowing down everything and I have the mouse freezing for a short moment each time it's called. It's even noticeable if the came_timeout is set to 10 seconds, this is why I ended up putting the call into its own method without requiring any arguments or configuration and store all monitors state.

However what I could do is parse the output into this function and return all monitors which would let me just check for the index or name in `get_status()`. What do you think?

tobes: With the slow down issue do you also get this with the existing `xrandr` modules they call it every 10 seconds. maybe this is something to look at first.

py3status/modules/xrandr_toggle.py
mrt-prodz: `while true; do xrandr -q; sleep 2; done`

Would produce the same issue actually.

py3status/modules/xrandr_toggle.py
mrt-prodz: This is happening even in bash if I call `xrandr` too often it will slow down and the mouse will freeze for a couple ms. I was following the xrandr module at first and noticed that issue.

Do you think it takes some time for `xrandr` to query all output and result in a delay to refresh workspaces? I could try in another environment to see if there is a similar issue.

mrt-prodz:
@tobes , I guess changing `_xrandr()` as the following would make more sense:

```python
# query xrandr for monitors status
def _xrandr(self):
        status = check_output(['xrandr']).split('\n')
        monitors = []
        # width x height + left + top
        regex = r'\d{1,}x\d{1,}\+\d{1,}\+\d{1,}'
        # for each line find (dis)connected monitors
        for line in status:
            if 'connected' in line:
                details = line.split(' ')
                if len(details) > 2:
                    source = {}
                    source['id'] = details[0] 
                    if details[1] == 'disconnected':
                        source['connected'] = 0
                        source['state'] = 'off'
                    else:
                        source['connected'] = 1
                        if search(regex, details[2]):
                            source['state'] = 'auto'
                        else:
                            source['state'] = 'off'
                    monitors.append(source)
        return monitors
```

That way I removed completely `xrandr` parsing from `get_status()`.

Thank you for taking the time to comment on this each time, I really appreciate.

mrt-prodz:
Following @tobes advices I refactored the module a bit, moved xrandr output parsing to `_xrandr()` method in order to keep `get_status()` cleaner.

py3status/modules/xrandr_toggle.py
```diff
@@ -88,7 +88,7 @@ def _xrandr(self):
                         monitors.append(source)
             return monitors
         except:
-            return self._error('error while parsing xrandr', None)
+            return self._error('error while parsing xrandr')
```
mrt-prodz: This is still not the proper way to do this, I'm returning an error response instead of the monitors list.
Sorry for the mixup, I will fix this.

tobes: This is why you should remove the try/except

mrt-prodz: I was thinking of actually setting the return value to None, and in `get_status()` I would first check if monitors is set to None to notice the user of a xrandr parsing error.

Wouldn't that make it easier to catch an error from there?

py3status/modules/xrandr_toggle.py
tobes: It's a shame we will have 3 xrandr modules all doing `xrandr -q` it would be nice if they could share the output.  But that's not an issue to look at here.

Users can set `cache_timeout` to a value that works for them.

py3status/modules/xrandr_toggle.py
```diff
@@ -0,0 +1,170 @@
...
+                            source['state'] = 'off'
+                        else:
+                            source['connected'] = 1
+                            if search(regex, details[2]):
+                                source['state'] = 'auto'
+                            else:
+                                source['state'] = 'off'
+                        monitors.append(source)
+            return monitors
+        except:
```
tobes: The thing with this is we don't know the issue.  Maybe I don't have `xrandr` installed or it had some issue.  I think the `xrandr` output is fairly standardised so I'd just let errors be thrown, that way we can fix any parsing issues.  At the moment we don't get very good feedback from `py3status` of the error.  I've got some patches to improve on this but I want #226 to get merged first.

py3status/modules/xrandr_toggle.py
mrt-prodz: I was actually thinking about that when I noticed the issue with multiple xrandr calls, if there was a way to share or cache the output between modules it would limit these calls. I'm still curious if this issue is also affecting other environment. Will test this right now actually.

py3status/modules/xrandr_toggle.py
tobes: > Wouldn't that make it easier to catch an error from there?

My view is we should not catch it anywhere.  It should just break.  The parsing is fairly fundamental, and you can make it robust enough to not have any errors.

maximbaz:
Hey @mrt-prodz, may I ask a naïve question, how would you compare your new module to the existing `xrandr` module? 

From what I read in the description, sometimes you want to output everything on one display only (TV), and sometimes on two other monitors. I use `xrandr` module for a very similar purpose, it allows to select any combination using a left click and apply it using a right click. 

I simply wonder what benefits I would have if I switch to your module instead?

mrt-prodz:
Hi @maximbaz , xrandr is a complete solution to organize your multi-monitor setup while this module is just a quick way to enable/disable a fixed monitor in your setup.

I also wanted to know the xrandr module better before filling some issues I experienced with it:

- xrandr calls each 10 seconds makes my workspaces refresh hanging for a couple ms
- must be my lack of experience with it but when I DVI-I-1+DVI-I-2+HDMI-2 I have a clone (not sure why) of the same workspaces on my dual monitors and the TV is using the workspace I set in the config
- when I switch my dual monitors workspaces to the third display (TV), i3 stops working and become unresponsive, no more keybindings and no way to recover access apparently. I'm still not sure what is going on but I couldn't even reboot/logout using keybindings. Maybe the patch @tobes made could fix this, I didn't try it yet.
- I managed to lock myself 3 times while trying out different setup, granted this is me being silly while trying things out but I just wanted to activate my third monitor and thought it was getting too complicated for a single specific task.

Since I really just needed a way to quickly activate/deactivate with a single click a monitor at a fixed position in my setup I thought it would be a faster solution.

maximbaz:
I certainly understand your desire to get a fast solution, if you however have time to make a long-term investment, I would definitely recommend improving the existing module instead of creating a new one.

Why? I personally haven't seen issues 2 and 3, but if you experience them, chances are, someone does as well. People will be grateful if you fix those :) Also, if we have two modules instead of one, you have twice as many chances that other people will keep improving the module that _you_ actually use.

Issue 1 is easily to fix I think, just set a larger value to the `cache_timeout` parameter.

As for the issue 4 (overcomplication), I am also affected by this :) Just like you I have a limited number of output configurations I use (two in my case). I don't like scrolling through all configurations that `xrandr` generated, I only need two to toggle between them. 

Wouldn't it be nice if there was a parameter for `xrandr` module specifying a subset of output configurations to use (for example: TV | Mon1+Mon2), so that `xrandr` module scrolls _only_ between them? If you have two configurations, toggling between them would be as easy as Left+Right mouse clicks on the status bar. If you have 3 configurations, that would be either Left+Right, or Left+Left+Right clicks, I'm sure you will get used to this after a week and will be doing this automatically.

Think about it :smile:

mrt-prodz:
> I certainly understand your desire to get a fast solution, if you however have time to make a long-term investment, I would definitely recommend improving the existing module instead of creating a new one.

This is a very good and valid point. There is also another issue @tobes mentioned when someone wants to use (for some reason) all of them, each module will make a request with xrandr separately. And on setup such as mine this means a lot of freezing for a couple ms if they all call xrandr on interval.

The xrandr freeze is really an issue for me, this is unrelated to py3status modules because I can reproduce the issue in bash directly on my desktop and laptop. @ultrabug told me to try `--current` to avoid polling for new hardware and it actually fixed the issue, but then this means I cannot find out if a new hardware has been plugged or not. That was the reason xrandr_toggle refreshes the selected screen state only on click.

>Wouldn't it be nice if there was a parameter for xrandr module specifying a subset of output configurations to use (for example: TV | Mon1+Mon2), so that xrandr module scrolls only between them?

That would be really handy indeed, instead of toggling all possible detected configurations and would also avoid having another xrandr module since the original could fulfill this task.

maximbaz:
How about these suggestions then?

1. `xrandr` module already contains a functionality to force refresh the available devices by using "middle click" on the status bar. Let's add a parameter `autodetect_outputs` equaling True by default that is constantly polling new hardware (the current behavior today), and you would set it to False to avoid the issue you have (you will use "middle click" to refresh). When set to False, all the calls should use `--current` flag, just like you described, to avoid the issue. I assume you actually don't need to refresh the output list too often, is that correct?
2. Let's add another parameter `output_configurations` where we would list all configurations we need (None by default), and `xrandr` will use this list (when it is provided) to scroll between, instead of composing all possible configurations.

What do you think about these @mrt-prodz? Would they be able to cover everything you are trying to achieve? If "yes" and "yes", do you think you could have a look in helping implementing those? :smile:

mrt-prodz:
> I assume you actually don't need to refresh the output list too often, is that correct?

This is correct in my case it's not neceassary to refresh outputs at all actually, only when interacting with them. And the flag + middle click would solve my issues with xrandr.

> Let's add another parameter output_configurations..

I really like this idea.

> If "yes" and "yes", do you think you could have a look in helping implementing those?

You are absolutely correct, I wouldn't mind selecting one of the pre-defined configuration and right click to enable/disable them. And it actually solves all my 'not-so-serious' issues with xrandr module.

maximbaz:
Would be nice to hear others' feedback on these ideas, before we go and implement them, perhaps @tobes, @ultrabug?

@mrt-prodz do you feel like doing that all together, or do you want to split? I could have a look at making pre-defined configurations, not today but hopefully later this week.

mrt-prodz:
I talked about it with @ultrabug on IRC and he liked the idea as well, this would definitely render xrandr_toggle unnecessary (which is a good thing) and would provide other options suchs as switching the screen positions.

I would also like to hear @tobes opinion about how this could be implemented or if there is some alternative options we could set as well.

If you want to implement this yourself I have no issue at all, I really liked your ideas about improving the xrandr module. If you eventually can't find time I would be happy to have a look at it.

tobes:
I'm good with the suggestions :+1: 

It feels like a simple solution that meets a real user need.

maximbaz:
Cool, guys :+1: 

@mrt-prodz, let'd do it in parallel then to speed up, I'll take the `output_configurations` parameter since I know very well what to expect from this feature, please take the `autodetect_outputs` part, you have already a good setup with 3 screens and you will be able to easily validate if it works in all your scenarios :+1:

mrt-prodz:
That makes sense since not everybody seem to have the lagging issue with `xrandr` apparently.

tobes:
> when I switch my dual monitors workspaces to the third display (TV), i3 stops working and become unresponsive, no more keybindings and no way to recover access apparently. 

@mrt-prodz it'd be nice to get this resolved too.

If you don't mind crashing everything could you try this.

1) cause the issue (i3 locks up)

2) press `Ctrl` + `Alt` + `F3` to get to a terminal

3) send a `SIGCONT` to `xrandr` (I just do it in `htop`)  Initially it has a state of `S` the signal sends it to `Z`

4) `Ctrl` + `Alt` + `F?` back to `i3` if it worked then all is happy again. (`F?` as tends to be setup specific but one of `F1` - `F8` has always worked for me)

If that works then I know the fix, but it needs a bit of reworking of the code but I can knock a patch up.

tobes:
As there has been no action on this for just shy of a year.  I'm going to close this.

PR #243, opened 2016-04-04
Proposing new module Twitch streaming

HoratioZEDU:
Checks if twitch streamer is online or offline.

### DIFF
--- a/modules/twitch_streaming
+++ b/modules/twitch_streaming
@@ -2,9 +2,16 @@
 """
 Checks if a streamer is online. [Gross Gore or grossie_gore as example placeholder]
 
+Checks if a streamer is online using the requests library to check to see if the dictionary key 'stream' in the json response is set to None or not.
+
 Configuration parameters:
     - cache_timeout : how often we refresh this module in seconds
     - stream_name : name of streamer(twitch.tv/<streamer_name>)
+    
+Format of status string placeholders
+     - {stream_name}  :  name of the streamer (twitch.tv/<stream_name>)
+     - {status} : whether or not the streamer is online/offline
+     
 @author Alex Caswell horatioesf@virginmedia.com
 @license BSD
 """
@@ -15,29 +22,26 @@ import requests
 
 
 class Py3status:
-    """
-    The Py3status class name is mandatory.
-    Below you list all the available configuration parameters and their
-    default value for your module which can be overwritten by users
-    directly from their i3status config.
-    This examples features only one parameter which is 'cache_timeout'
-    and is set to 10 seconds (0 would mean no cache).
-    """
     # available configuration parameters
     cache_timeout = 10
     stream_name = "grossie_gore"
+    # can be customized in i3status.conf
+    format = "{stream_name} {status}"
+    format_offline = "{stream_name} {status}"
 
     def is_streaming(self, i3s_output_list, i3s_config):
         r = requests.get('https://api.twitch.tv/kraken/streams/' + self.stream_name)
         if r.json()['stream'] == None: 
-            answer=" is offline."
+            answer="is offline."
             colour=i3s_config['color_bad']
+            full_text = self.format_offline.format(stream_name=self.stream_name, status=answer)
         else: 
-            answer=" is live!"
+            answer="is live!"
             colour=i3s_config['color_good']
+            full_text = self.format.format(stream_name=self.stream_name, status=answer)
         response = {
             'cached_until': time() + self.cache_timeout,
-            'full_text': self.stream_name + answer,
+            'full_text': full_text,
             'color': colour
         }
         return response
--- a/modules/twitch_streaming
+++ b/modules/twitch_streaming
@@ -24,7 +24,7 @@ class Py3status:
     # available configuration parameters
     # can be customized in i3status.conf
     cache_timeout = 10
-    stream_name = "asdfdfgag"
+    stream_name = "grossie_gore"
     format = "{stream_name} is live!"
     format_offline = "{stream_name} is offline."
     format_invalid = "{stream_name} does not exist!"
--- a/modules/twitch_streaming
+++ b/modules/twitch_streaming
@@ -24,7 +24,7 @@ class Py3status:
     # available configuration parameters
     # can be customized in i3status.conf
     cache_timeout = 10
-    stream_name = "grossie_gore"
+    stream_name = "asdfdfgag"
     format = "{stream_name} is live!"
     format_offline = "{stream_name} is offline."
     format_invalid = "{stream_name} does not exist!"
@@ -38,6 +38,7 @@ class Py3status:
 
     def is_streaming(self, i3s_output_list, i3s_config):
         r = requests.get('https://api.twitch.tv/kraken/streams/' + self.stream_name)
+        print(r.json())
         if not self._display_name:
             self._get_display_name()
         if 'error' in r.json():
@@ -46,9 +47,12 @@ class Py3status:
         elif r.json().get('stream') == None:
             colour=i3s_config['color_bad']
             full_text = self.format_offline.format(stream_name=self._display_name)
-        else:
+        elif r.json().get('stream') != None:
             colour=i3s_config['color_good']
             full_text = self.format.format(stream_name=self._display_name)
+        else:
+            colour=i3s_config['color_bad']
+            full_text = "An unknown error has occurred."
         response = {
             'cached_until': time() + self.cache_timeout,
             'full_text': full_text,
--- a/modules/twitch_streaming
+++ b/modules/twitch_streaming
@@ -28,14 +28,18 @@ class Py3status:
     format = "{stream_name} is live!"
     format_offline = "{stream_name} is offline."
 
+    def get_display_name(self, stream_name):
+        display_name_request = requests.get('https://api.twitch.tv/kraken/users/' + self.stream_name)
+        return display_name_request.json().get('display_name')
+
     def is_streaming(self, i3s_output_list, i3s_config):
         r = requests.get('https://api.twitch.tv/kraken/streams/' + self.stream_name)
         if r.json().get('stream') == None:
             colour=i3s_config['color_bad']
-            full_text = self.format_offline.format(stream_name=self.stream_name)
+            full_text = self.format_offline.format(stream_name=self.get_display_name(self.stream_name))
         else:
             colour=i3s_config['color_good']
-            full_text = self.format.format(stream_name=self.stream_name)
+            full_text = self.format.format(stream_name=self.get_display_name(self.stream_name))
         response = {
             'cached_until': time() + self.cache_timeout,
             'full_text': full_text,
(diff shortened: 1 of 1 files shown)


tobes:
Nice simple module

It would be good if the user could customize this a little eg to use an icon instead of online if they so chose.

Adding a placeholder of `{stream_name}` and `format` and `format_offline` parameters would allow that.

then I could do 
```
twitch_streaming {
    format = "⬤ {stream_name}"
    format_offline = "⭕ {stream_name}"
}
```

If you could also fix the parameter in the docstring that would be great.

The example module has been updated today and now includes the correct formatting of the docstring and example placeholder
https://github.com/ultrabug/py3status/blob/master/doc/example_module.py

HoratioZEDU:
Alright, I'm fairly new to github so I hope I've done all this correctly... I added the functionality you suggested and everything seems to be in order. I put it as a new commit to the branch I already made, so I'm hoping I was meant to.

tobes:
Yeah just updating the branch updates this pull request.

```
Configuration parameters:
    - cache_timeout : how often we refresh this module in seconds
    - stream_name : name of streamer(twitch.tv/<streamer_name>)
```
should be
```
Configuration parameters:
    stream_name: name of streamer(twitch.tv/<streamer_name>)
```

similar with the placeholders.

modules/twitch_streaming
```diff
@@ -0,0 +1,58 @@
+
+"""
+Checks if a streamer is online. [Gross Gore or grossie_gore as example placeholder]
+
+Checks if a streamer is online using the requests library to check to see if the dictionary key 'stream' in the json response is set to None or not.
+
+Configuration parameters:
+    - cache_timeout : how often we refresh this module in seconds
+    - stream_name : name of streamer(twitch.tv/<streamer_name>)
+
```
tobes:
```
Configuration parameters:
    - cache_timeout : how often we refresh this module in seconds
    - stream_name : name of streamer(twitch.tv/<streamer_name>)
```
should be
```
Configuration parameters:
    stream_name: name of streamer(twitch.tv/<streamer_name>)
```

similar with the placeholders.

modules/twitch_streaming
```diff
@@ -0,0 +1,58 @@
...
+import requests
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 10
+    stream_name = "grossie_gore"
+    # can be customized in i3status.conf
+    format = "{stream_name} {status}"
+    format_offline = "{stream_name} {status}"
```
tobes: I'm not sure that adding the `{status}` placeholder adds much here.  If there were more placeholders it'd make sense.

Others may have opinions.

modules/twitch_streaming
```diff
@@ -0,0 +1,58 @@
...
+# import your useful libs here
+from time import time
+import requests
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 10
+    stream_name = "grossie_gore"
+    # can be customized in i3status.conf
```
tobes: this comment should be up 2 lines

tobes:
This is looking better now.  There are some niggly issues that need fixing.  line lengths etc

Looking briefly at the API it seems like you could get some interesting information about a channel to display https://github.com/justintv/Twitch-API/blob/master/v3_resources/channels.md#get-channelschannel  eg  display_name.  They might be fun to get too, but if there isn't a need then no reason to bother.  Also you'd need an extra call if the stream is offline.

modules/twitch_streaming
```diff
@@ -0,0 +1,55 @@
+
+"""
+Checks if a streamer is online. [Gross Gore or grossie_gore as example placeholder]
+
+Checks if a streamer is online using the requests library to check to see if the dictionary key 'stream' in the json response is set to None or not.
```
tobes: lines shouldn't be over 79 chars long.

modules/twitch_streaming
```diff
@@ -0,0 +1,55 @@
...
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds
+    stream_name: name of streamer(twitch.tv/<stream_name>)
+    
+Format of status string placeholders
+    {stream_name}:  name of the streamer (twitch.tv/<stream_name>)
```
tobes: `:` isn't needed

modules/twitch_streaming
```diff
@@ -0,0 +1,55 @@
+
+"""
+Checks if a streamer is online. [Gross Gore or grossie_gore as example placeholder]
```
tobes: I think this could read better. and kill the stuff in square brackets.

Checks if a Twitch.tv streamer is online.

HoratioZEDU:
I'll look into getting the display name from the api as well as perhaps functionality to use the display name rather than the stream name by default? I'm thinking it might be nicer to see "Gross Gore is offline." rather than "grossie_gore is offline".

modules/twitch_streaming
```diff
@@ -0,0 +1,55 @@
...
+    # available configuration parameters
+    # can be customized in i3status.conf
+    cache_timeout = 10
+    stream_name = "grossie_gore"
+    format = "{stream_name} is live!"
+    format_offline = "{stream_name} is offline."
+
+    def is_streaming(self, i3s_output_list, i3s_config):
+        r = requests.get('https://api.twitch.tv/kraken/streams/' + self.stream_name)
+        if r.json()['stream'] == None:
```
tobes: if the request fails (offline etc) will this fail? 

`r.json().get('stream')` is safer

 Can `r.json()` ever be `None`?

modules/twitch_streaming
```diff
@@ -0,0 +1,56 @@
+
+"""
+Checks if a streamer is online. [Gross Gore/grossie_gore as placeholder]
+
+Checks if a streamer is online using the requests library to check to see
+if the dictionary key 'stream' in the json response is set to None or not.
```
tobes: This is for the user.  They probably don't care how you did do this.  Maybe something like `queries the twitch kraken api to see if a chanel is streaming your favorite games.....`

Sorry I just skipped the line before as it was too long to read

modules/twitch_streaming
```diff
@@ -0,0 +1,59 @@
...
+
+class Py3status:
+    # available configuration parameters
+    # can be customized in i3status.conf
+    cache_timeout = 10
+    stream_name = "grossie_gore"
+    format = "{stream_name} is live!"
+    format_offline = "{stream_name} is offline."
+
+    def get_display_name(self, stream_name):
```
tobes: You need to change this function to `_get_display_name` otherwise py3status will try and run it.

Also this result is unlikely to change so you probably don't want to keep checking every 10 seconds.

modules/twitch_streaming
pydsigner: It could be if there was an error perhaps.

tobes:
Maybe I wasn't very clear.

With `_get_display_name()` The result will be the same unless the user updates their display name which is not going to happen much (unless people do update this often).

If they are online or not you want to check regularly.  The default value should be something sensible.

So if you get the display name say just once.  It gets slightly more complicated but you could do something like.

```
def __init__(self):
    # we'll use this to keep the displayed name
   self._display_name = None

def _get_display_name(self, stream_name):
    display_name_request = requests.get('https://api.twitch.tv/kraken/users/' + self.stream_name)
    # store the result
    self._display_name = display_name_request.json().get('display_name')

def is_streaming(self, i3s_output_list, i3s_config):
    # get the display name if we don't know it
    if not self._display_name:
        self._get_display_name(self)

    ...

The comments are just by way of explination

```

tobes:
Another though is if the `stream_name` the user gives does not exist.  The API returns an error.  It would be good to check this and if so tell the user that it's invalid.

HoratioZEDU:
Ah, right, I'm dumb, sorry. Fixing now.

HoratioZEDU:
I'm trying to think of something slightly nicer to see than "{stream_name} does not exist!", if you have any ideas they'd be welcome, hah. I set it as a parameter just in case the user wanted to play with it.

modules/twitch_streaming
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self):
+        self._display_name = None
+
+    def _get_display_name(self):
+        display_name_request = requests.get('https://api.twitch.tv/kraken/users/' + self.stream_name)
+        self._display_name = display_name_request.json().get('display_name')
+
+    def is_streaming(self, i3s_output_list, i3s_config):
+        r = requests.get('https://api.twitch.tv/kraken/streams/' + self.stream_name)
+        print(r.json())
```
tobes: You'll want to remove this print at some point.

modules/twitch_streaming
```diff
@@ -0,0 +1,68 @@
...
+            self._get_display_name()
+        if 'error' in r.json():
+            colour=i3s_config['color_bad']
+            full_text = self.format_invalid.format(stream_name=self.stream_name)
+        elif r.json().get('stream') == None:
+            colour=i3s_config['color_bad']
+            full_text = self.format_offline.format(stream_name=self._display_name)
+        else:
+            colour=i3s_config['color_good']
+            full_text = self.format.format(stream_name=self._display_name)
```
tobes: Here you are deciding that if nothing is 'wrong' then everything must be good.  I'd prefer that we check if things are good and if not then assume things are bad.

I haven't checked but I imagine that if we are not on a network that the module will currently show as being online.

Does that make sense?

tobes:
Also could you want to edit the title of this pull request from `Proposing new module` to `Proposing new module Twitch streaming`

modules/twitch_streaming
```diff
@@ -38,6 +38,7 @@ class Py3status:
 
     def is_streaming(self, i3s_output_list, i3s_config):
         r = requests.get('https://api.twitch.tv/kraken/streams/' + self.stream_name)
+        print(r.json())
```
tobes: this has returned ;p

`git add -p` is really useful for making sure you only add the bits you need.  It takes a little time to get used to but is great.

likewise `stream_name = "asdfdfgag"` may not be intended.

HoratioZEDU:
Oops >.> I have no idea how i managed that one, I'll fix it now

ultrabug:
Looks good to me, except that I will have to pass it through `yapf` for reformatting of some lines.

Also, you removed the `color_degraded` from the standalone test options (at the end of the file) and you shouldn't.

If nobody have something to add, I'll merge soon.

modules/twitch_streaming
```diff
@@ -0,0 +1,71 @@
...
+
+Checks if a streamer is online using the Twitch Kraken API to see
+if a channel is currently streaming or not.
+
+Configuration parameters
+    cache_timeout: how often we refresh this module in seconds
+    stream_name: name of streamer(twitch.tv/<stream_name>)
+
+Format of status string placeholders
+    {stream_name}:  name of the streamer (twitch.tv/<stream_name>)
```
tobes: being picky `{stream_name}` can be the stream name or the display name so I'd just remove the '(twitch.tv/<stream_name>)' bit here

ultrabug:
Thanks a lot for your patience and contribution @HoratioZEDU !

PR #244, opened 2016-04-04
updates to example module

tobes:
PR #243 is a new module and appears to have been based off the example module in docs.  This is expected behaviour, but the example uses old style of docstring.

I have updated the docstring to the 'new' style.

I also added a format param and a placeholder as this feels like a good place to include them.  They feel like a py3status standard.


ultrabug:
Thanks, had this in mind too :(

tobes:
Great minds...

PR #245, opened 2016-04-04
Nice module errors

tobes:
Currently when a module raises an error we are just informed of the method and a vague error.

This PR improves on this situation.  We have more information on the error:

* filename
* instance
* line number
* exception type

This makes finding/fixing errors in modules much simpler.

The code is a little nasty with the `finally` but this is to prevent circular references https://docs.python.org/2/library/sys.html#sys.exc_info

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -289,9 +289,9 @@ class Py3statusWrapper():
         We also make sure to log anything to keep trace of it.
         """
         if not self.config['dbus_notify']:
-            msg = 'py3status: {}. '.format(msg)
+            msg = 'py3status: {}.'.format(msg)
         if level != 'info':
-            msg += 'please try to fix this and reload i3wm (Mod+Shift+R)'
+            msg += ' Please try to fix this and reload i3wm (Mod+Shift+R)'
         try:
             log_level = LOG_LEVELS.get(level, LOG_ERR)
             syslog(log_level, msg)
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -290,6 +290,8 @@ class Py3statusWrapper():
         """
         if not self.config['dbus_notify']:
             msg = 'py3status: {}.'.format(msg)
+        else:
+            msg = '{}.'.format(msg)
         if level != 'info':
             msg += ' Please try to fix this and reload i3wm (Mod+Shift+R)'
         try:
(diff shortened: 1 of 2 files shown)


ultrabug:
Indeed way better to tackle problems !

One small thing : it seems we miss a space before 'please fix this...' on core `notify_user` function. Maybe a stop and a 'Please ...' would be nicer too.

tobes:
I've made the suggested change.  There was a space which I've moved but I'm not sure if you wanted an extra one.  Let me know

py3status/core.py
```diff
@@ -289,9 +289,9 @@ def notify_user(self, msg, level='error'):
         We also make sure to log anything to keep trace of it.
         """
         if not self.config['dbus_notify']:
-            msg = 'py3status: {}. '.format(msg)
+            msg = 'py3status: {}.'.format(msg)
         if level != 'info':
-            msg += 'please try to fix this and reload i3wm (Mod+Shift+R)'
+            msg += ' Please try to fix this and reload i3wm (Mod+Shift+R)'
```
ultrabug: Add a dot so it's in line with the nagbar version please :
`msg += '. Please try to fix this and reload i3wm (Mod+Shift+R)'`

py3status/module.py
```diff
@@ -324,8 +325,24 @@ def run(self):
...
-                    msg = 'user method {} failed ({})'.format(meth, err)
+                    try:
+                        # Try to get nice feedback info
+                        # we need to make sure to delete tb even if things go
+                        # wrong.
+                        exc_type, exc_obj, tb = sys.exc_info()
+                        stack = extract_tb(tb)
+                        filename = os.path.basename(stack[-1][0])
+                        line_no = stack[-1][1]
+                        msg = '{}, instance, {}, user method {}, failed ({}) line {}'.format(
```
ultrabug: There's an extra comma after `instance` imho should be:
`msg = '{}, instance {}, user method {}, failed ({}) line {}'.format(`

ultrabug:
2 comments made, thanks !

py3status/core.py
tobes: OK, I was still on the nag bar so couldn't work out what you meant.  Fixed slightly differently.

tobes: OK, I was still on the nag bar so couldn't work out what you meant.  Fixed slightly differently.

tobes:
Fixes done

tobes:
That dot issue in the messages really needs to be fixed better by having messages add their own full stop.  I'll do that in another pull request though.  Github is like mud this morning.

ultrabug:
Thanks mate !

PR #247, opened 2016-04-04
nicer module updates fixes #241

### LINKED ISSUE #241: "killall -s USR1 py3status" does nothing
I am trying to include the send signal into my script that lowers and raises the volume. I am using the module volume_status.py. The change in volume works but sending the signal does not refresh the statusbar.

example  command:
`pactl set-sink-volume $(pactl list short sinks|grep RUNNING|awk '{print $1}') -5%`
`killall -s USR1 py3status`

i3config:
`order += "volume_status"`
`volume_status {`
`	   format = "Vol.: {percentage}%"`
`	   format_muted = "Vol: mute"`
`	   device = "pulse"`
`}`

The bar only refreshes after the default 10sec. It is like py3status is completely ignoring the signal. Unfortunately, I don't know how to trace this. Could you please help me out?

tobes:
Renames `Module.clear_cache()` to `Module.force_update()`

This method now as well as clearing the cache sets a Timer to update the module.

Change py3status to only use this to force modules to update eg on click events.

As well as being cleaner it should fix #241 @neutronst4r could you see if this fixes your issue.  In my testing it works as expected.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -353,7 +353,7 @@ class Py3statusWrapper():
         For every module, reset the 'cached_until' of all its methods.
         """
         for module in self.modules.values():
-            module.clear_cache()
+            module.force_update()
 
     def terminate(self, signum, frame):
         """
@@ -379,8 +379,7 @@ class Py3statusWrapper():
         for group in groups_to_update:
             group_module = self.output_modules.get(group)
             if group_module:
-                group_module['module'].clear_cache()
-                group_module['module'].run()
+                group_module['module'].force_update()
 
     def create_output_modules(self):
         """
--- a/py3status/events.py
+++ b/py3status/events.py
@@ -111,10 +111,7 @@ class Events(Thread):
         if module is not None:
             if self.config['debug']:
                 syslog(LOG_INFO, 'refresh module {}'.format(module_name))
-            for obj in module.methods.values():
-                obj['cached_until'] = time()
-            # get module to update
-            module.run()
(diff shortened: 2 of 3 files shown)


ultrabug:
tested OK here, I like it better like that indeed

neutronst4r:
It works, but modules that use averages, such as cpu_termperature, net_rate and cpu_usage have sudden spikes. I think the forced update throws off the modules' calculations. Correcting this would require changes in all of those modules. I think the cleaner way would be to introduce signals for relevant modules and if a specific signal is sent only one modules is updated.

tobes:
@neutronst4r  what was it like in 2.9 are we much different?

neutronst4r:
I don't know, I only noticed it recently.

tobes:
@neutronst4r 

`git checkout 2.9` will switch you to that version.  If you try things there . Then you can get a comparison to this branch.  If it is an existing issue then I'm not worried but if it's new then I care.

Could you at least explain exactly which modules and configs have the issue so I can try to replicate it.

Thanks

ultrabug:
My wild guess is that it has the same behaviour of *spikes* since the beginning because the USR1 signal always refreshed all modules.

tobes:
@ultrabug I agree, I think it is an old issue that has just become noticed now

ultrabug:
Closing, thanks @tobes .

@neutronst4r, if you can test and experience this is related to the new code, please open an issue

PR #248, opened 2016-04-04
write_tmp_i3status_config() rewrite

tobes:
This PR makes some changes to the generation of the `i3status.conf` created to send to `i3status`.

Changes:

* Instead of using the user provided `general` section we just create our own from a template `I3STATUS_CONF_GENERAL`.  (less likely to break i3status)

* The interval is from the users config if set, else `DEFAULT_I3STATUS_INTERVAL` is used.

* The config file is then created using `I3Status.config['i3s_modules']`.  Before `config['order']` and `config['.group_extras']` where used.  
 
`I3Status.set_responses()` uses `config['i3s_modules']` to work out the order of the i3status modules.  In the current setup this is fine because 

```config['order'] + config['.group_extras'] == config['i3s_modules']``` 

However this is almost by chance.  This corrects this issue.

The main reason for this is that I have a new config parser that I want to submit and with that, the above assumption fails causing issues.  Including this change really complicates that pull request. 
So I've separated it out into this PR.

I've tested this across various configs for the last week or so without problems.

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -19,6 +19,15 @@ TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
 TZTIME_FORMAT = '%Y-%m-%d %H:%M:%S %Z'
 TIME_MODULES = ['time', 'tztime']
 
+DEFAULT_I3STATUS_INTERVAL = 15
+I3STATUS_CONF_GENERAL = """
+    general {
+        interval = %s
+        colors = false
+        output_format = "i3bar"
+    }
+"""
+
 
 class Tz(tzinfo):
     """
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ultrabug** [comment]
I need to take a closer look, but are we positive that we still respect the users' i3status general settings ? You *sound* like you don't.

Also, plz rebase

**Comment by tobes** [comment]
>  we still respect the users' i3status general settings?

I thought we shouldn't care except if they set an update time.  That way we can use that section for our globals too.  Something like
```
general {
    color = "#FF00FF"
}
```
makes i3status croak.

Anyhow removing that cleans up the patch and so is no bad thing.

ultrabug:
Clean one, thanks :+1:

PR #249, opened 2016-04-04
Add `output_combinations` parameter to xrandr module

maximbaz:
Implementing a way to make a user-defined list of output combinations to use, as discussed in #242.
Personally... it is *so much better* now! :smile: 

@mrt-prodz, @tobes, @ultrabug please have a look.

### DIFF
--- a/py3status/modules/xrandr.py
+++ b/py3status/modules/xrandr.py
@@ -14,6 +14,7 @@ For convenience, this module also proposes some added features:
         screen is available (handy for laptops)
     - Automatically apply this screen combination on start: no need for xorg!
     - Automatically move workspaces to screens when they are available
+    - Define your own subset of output combinations to use
 
 Configuration parameters:
     cache_timeout: how often to (re)detect the outputs
@@ -26,6 +27,19 @@ Configuration parameters:
         when the module starts (saves you from having to configure xorg)
     format_clone: string used to display a 'clone' combination
     format_extend: string used to display a 'extend' combination
+    output_combinations: string used to define your own subset of output
+        combinations to use, instead of generating every possible combination
+        automatically. Provide the values in the format that this module uses,
+        splitting the combinations using '|' character.
+        The combinations will be rotated in the exact order as you listed them.
+        When an output layout is not available anymore, the configurations
+        are automatically filtered out.
+        Example:
+            Assuming the default values for `format_clone` and `format_extend`
+            are used, and assuming you have two screens 'eDP1' and 'DP1', the
+            following setup will reduce the number of output combinations
+            from four (every possible one) down to two:
+            output_combinations = "eDP1|eDP1+DP1"
 
 Dynamic configuration parameters:
     - <OUTPUT>_pos: apply the given position to the OUTPUT
@@ -70,6 +84,7 @@ class Py3status:
     force_on_start = None
     format_clone = '='
     format_extend = '+'
+    output_combinations = None
 
     def __init__(self):
         """
@@ -138,21 +153,31 @@ class Py3status:
         Generate all connected outputs combinations and
         set the max display width while iterating.
         """
-        available_combinations = set()
+        available = set()
         combinations_map = {}
 
+        whitelist = None
+        if self.output_combinations:
+            whitelist = self.output_combinations.split('|')
+
         self.max_width = 0
-        for output in range(len(self.layout['connected']) + 1):
-            for comb in combinations(self.layout['connected'], output):
-                if comb:
-                    for mode in ['clone', 'extend']:
-                        string = self._get_string_and_set_width(comb, mode)
-                        if len(comb) == 1:
-                            combinations_map[string] = (comb, None)
-                        else:
-                            combinations_map[string] = (comb, mode)
-                        available_combinations.add(string)
-        self.available_combinations = deque(available_combinations)
+        for output in range(len(self.layout['connected'])):
+            for comb in combinations(self.layout['connected'], output + 1):
+                for mode in ['clone', 'extend']:
+                    string = self._get_string_and_set_width(comb, mode)
+                    if whitelist and string not in whitelist:
+                        continue
+                    if len(comb) == 1:
+                        combinations_map[string] = (comb, None)
+                    else:
+                        combinations_map[string] = (comb, mode)
+                    available.add(string)
+
+        # Preserve the order in which user defined the output combinations
+        if whitelist:
+            available = reversed([comb for comb in whitelist if comb in available])
+
+        self.available_combinations = deque(available)
         self.combinations_map = combinations_map
 
     def _get_string_and_set_width(self, combination, mode):


py3status/modules/xrandr.py
```diff
@@ -141,17 +156,28 @@ def _set_available_combinations(self):
...
-                            combinations_map[string] = (comb, mode)
-                        available_combinations.add(string)
+                if not comb:
+                    continue
+                for mode in ['clone', 'extend']:
+                    string = self._get_string_and_set_width(comb, mode)
+                    if whitelist is not None and string not in whitelist:
+                        continue
+                    combinations_map[string] = (comb, None if len(comb) == 1 else mode)
+                    available_combinations.add(string)
```
tobes: Personally I find this harder to follow than the existing code.

```
if len(comb) == 1:
     combinations_map[string] = (comb, None)
else:
     combinations_map[string] = (comb, mode)
```

might be longer than

```
combinations_map[string] = (comb, None if len(comb) == 1 else mode)
```
but it is much easier to maintain and is really clear what it's doing. 

There is also code churn which makes reviews harder and so increases the chance of introducing bugs.  Don't get me wrong I'm a fan of re-factoring stuff but the aim should be to make it clearer to read.

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,28 @@ def _set_available_combinations(self):
...
+
+        # Preserve the order in which user defined the output combinations
+        if whitelist is not None:
+            available = set(available_combinations)
```
tobes: isn't `available_combinations` already a set?

py3status/modules/xrandr.py
maximbaz: The reason I did was a cheap way to get rid of the warning "C901 the code is too complex" - as I didn't think the function is complex enough to split it in several, and getting rid of one extra `if` did it :smile:

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,28 @@ def _set_available_combinations(self):
...
             for comb in combinations(self.layout['connected'], output):
-                if comb:
-                    for mode in ['clone', 'extend']:
-                        string = self._get_string_and_set_width(comb, mode)
-                        if len(comb) == 1:
-                            combinations_map[string] = (comb, None)
-                        else:
-                            combinations_map[string] = (comb, mode)
-                        available_combinations.add(string)
+                if not comb:
```
tobes: Do we ever get a nothing here?  If so it feels like there is a problem somewhere else in the code.  I can understand why you are keeping it but I'm just curious.

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,28 @@ def _set_available_combinations(self):
...
+                    continue
+                for mode in ['clone', 'extend']:
+                    string = self._get_string_and_set_width(comb, mode)
+                    if whitelist is not None and string not in whitelist:
+                        continue
+                    combinations_map[string] = (comb, None if len(comb) == 1 else mode)
+                    available_combinations.add(string)
+
+        # Preserve the order in which user defined the output combinations
```
tobes: nice comment :+1:

py3status/modules/xrandr.py
maximbaz: my bad, missed that one, thanks!

tobes: `available` looks like a line length choice too ;p

py3status/modules/xrandr.py
maximbaz: Ha, nice! We do come inside, but apparently only because `range` call starts with zero a couple of lines above, fixed :stuck_out_tongue_winking_eye:

py3status/modules/xrandr.py
maximbaz: Because you discovered a nice way to remove one `if` above, I'm returning this one for you :laughing:

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,29 @@ def _set_available_combinations(self):
...
-                    for mode in ['clone', 'extend']:
-                        string = self._get_string_and_set_width(comb, mode)
-                        if len(comb) == 1:
-                            combinations_map[string] = (comb, None)
-                        else:
-                            combinations_map[string] = (comb, mode)
-                        available_combinations.add(string)
+                for mode in ['clone', 'extend']:
+                    string = self._get_string_and_set_width(comb, mode)
+                    if whitelist is not None and string not in whitelist:
```
tobes: `if whitelist and string not in whitelist:`

might be nicer here.

Looking much nicer though now

maximbaz: you've got it :+1:

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,29 @@ def _set_available_combinations(self):
...
+                    if whitelist and string not in whitelist:
+                        continue
+                    if len(comb) == 1:
+                        combinations_map[string] = (comb, None)
+                    else:
+                        combinations_map[string] = (comb, mode)
+                    available_combinations.add(string)
+
+        # Preserve the order in which user defined the output combinations
+        if whitelist is not None:
```
pydsigner: Could be just `if whitelist:`

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,29 @@ def _set_available_combinations(self):
         available_combinations = set()
         combinations_map = {}
 
+        if self.output_combinations is None:
+            whitelist = None
+        else:
+            whitelist = self.output_combinations.split('|')
```
pydsigner: How about:
```py
whitelist = None
if self.output_combinations:
    whitelist = self.output_combinations.split('|')
```

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,29 @@ def _set_available_combinations(self):
...
+                    if len(comb) == 1:
+                        combinations_map[string] = (comb, None)
+                    else:
+                        combinations_map[string] = (comb, mode)
+                    available_combinations.add(string)
+
+        # Preserve the order in which user defined the output combinations
+        if whitelist is not None:
+            available_combinations = [comb for comb in whitelist
+                                      if comb in available_combinations][::-1]
```
tobes: I don't really like `[::-1]` as it's not very clear and explicit is good.  Also explaining why we do this might be good
```
available_combinations = [comb for comb in whitelist
    if comb in available_combinations][::-1]
```
so

```
available_combinations = [comb for comb in whitelist
    if comb in available_combinations]
# why are we reversing this?
available_combinations.reverse()
```

maximbaz: Why? I think list slicing is a very pythonic thingy, it becomes quite clear as soon as you get used to it - I actually miss this simplicity in many other languages :)

We inverse because the existing code that scrolls through available combinations on mouse click is doing that in inverse order, and I didn't want to change that part of a code not to affect existing users that are already accustomed to this.

I did put a general comment that this part of code does the correct order, but not explained in details as it might get complicated.

tobes: Slicing is very pythonic but I'm not sure if using it to reverse is

> Simple is better than complex.

>There should be one-- and preferably only one --obvious way to do it.

maximbaz: I'll save myself from arguing, in my mind this is not something important and I'm fine either way :smile:
Same applies for both comments by @pydsigner.

Guys, please mark everything you don't like and I will fix it in one go, presumably on Thursday.

py3status/modules/xrandr.py
```diff
@@ -141,17 +156,29 @@ def _set_available_combinations(self):
...
 
+        if self.output_combinations is None:
+            whitelist = None
+        else:
+            whitelist = self.output_combinations.split('|')
+
         self.max_width = 0
-        for output in range(len(self.layout['connected']) + 1):
+        for output in range(1, len(self.layout['connected']) + 1):
             for comb in combinations(self.layout['connected'], output):
```
tobes: you could remove the `1`s from the line above (165) and then use `output + 1` here

py3status/modules/xrandr.py
pydsigner: @tobes `[::-1]` is pretty standard. Better `reversed(...)` though than `...; x.reverse()`.

py3status/modules/xrandr.py
pydsigner: I think it's more explicit to have the `+ 1` in the range.

tobes: I did mean that like this
```
 for output in range(len(self.layout['connected'])):
             for comb in combinations(self.layout['connected'], output + 1):
```

maybe wasn't very clear

tobes:
To be extreme here and also going against everything I've said about code churn, I'd really like `available_combinations` be renamed to `available`.  It would make the whole function more readable, and given that it is a local variable in function `_set_available_combinations()` the name would still have clear meaning.

maximbaz:
Wow guys you made quite a few suggestions, thanks! 

I've fixed all of them as per your latest suggestions, I hope that makes most of you glad :stuck_out_tongue: 

I suggest we take all future suggestions in the separate PR, this review is already hanging for quite some time.

py3status/modules/xrandr.py
```diff
@@ -26,6 +27,19 @@
         when the module starts (saves you from having to configure xorg)
     format_clone: string used to display a 'clone' combination
     format_extend: string used to display a 'extend' combination
+    output_combinations: string used to define your own subset of output
+        combinations to use, instead of generating every possible combinations
```
tobes: > every possible combinations

Should be 'combination' not 'combinations'

tobes:
I still think a comment about why we reverse would be nice, however I can live without it

Looks good to me, apart from the grammatical issue in the docstring

ultrabug:
Love it, thanks guys for this great collaboration !

PR #251, opened 2016-04-06
nicer group default format

tobes:
I've decided the default group format is stupid and this would be an improvement


ultrabug:
Yep, I agree

PR #252, opened 2016-04-06
show more info on errors

tobes:
Following on from the improved module errors.  This PR adds better information to exceptions that occur inside py3status.

Again we show `filename` and `line number` which really help with debugging issues.

```
self.config = {}
```
has been added just in case the exception happens before it has been set in `setup()`


ultrabug:
clean & cool, thx

PR #254, opened 2016-04-08
Volume buttons for volume_status

tobes:
This PR adds volume control buttons to `volume_status`.

These buttons are disabled by default and must be enabled in the `i3status.conf` as many users already will have their own `on_click` defined.

I also cleaned up the docstring and removed the odd `transformed` thing.


ultrabug:
cool one @tobes ! just a quick remark : please order variables alphabetically then I merge

tobes:
Sorry I thought I'd done that, should be fixed now

ultrabug:
Thanks !

PR #255, opened 2016-04-09
report_exception() find first issue in py3status land

tobes:
I've had an issue where the reported exception was deep inside a python lib (subprocess).  Whilst this was a useful start, ultimately it was not what I wanted.  Ideally we only care about where the problem hits our code (inside py3status or in one of our modules 'py3status land').

This PR does some simple filtering of the stack to find where in py3status land the exception first hit.


ultrabug:
:eyes: that's some kind of code ! hats off

PR #256, opened 2016-04-09
fedora_updates fix occassional error

tobes:
I was getting the odd error notification generally when restarting py3status.

This fixes the issue.

I also changed the cache timeout of the initial call as the update may as well happen immediately.


ultrabug:
Thanks mate !

Issue #250, opened 2016-04-06
class Meta - a cleaner future?


tobes:
I was having a think the last few days about py3status modules.

currently a module has update methods like this

```
 def update(self, i3s_output_list, i3s_config):
```
but I've never seen any that use `output_list`, and `i3s_config` only seems to be used to get color states.
```
 def update(self):
```
would be much nicer, but has two problems:

1) All the modules would need updating or would break including users custom modules

2) we now can't do `color_degraded` etc

But we have the current opportunity to solve both these issues and move to a cleaner calling pattern.
We can do this by using the `class Meta:`  Currently employed by the group module. This is what I would like to do.

If a Py3status module contains `class Meta:` it is taken to signify that it works with the new 'standard'.  As such it would have update calls of just `update(self)` without the parameters.  The `on_click` would likewise be `on_click(self, event):`  So we have the simplified calling but only for self-declaring modules.  All existing modules would continue to function as before.

But what about the `i3s_config` that might be useful to a module?  Well, firstly the module can still access that config.

```
    class Meta:
        include_py3_module = True
```

gives access the modules `Module` and we could easily add a `helper_get_i3_config()` method.  So that would solve that issue, although I'm not sure that the config is that useful. Apart from the `color_bad` type stuff.

With the colors many modules do 
```
 response['color'] = i3s_config['color_bad']
```
some allow custom degraded colors to be defined in their config (which is nicer)
```
response['color'] = self.color_bad or i3s_config['color_bad']
```

But I want to go further, what if my module just did
```
 response['color'] = 'bad'
```
and we then automagically change this to a config specific `color_bad` if it is supplied, else the general `color_bad`.  If an actual color is returned then we honor eg
```
 response['color'] = '#FF00FF'
```

I'd also like to change the `cache_until` so that setting it to `-1` would mean cache forever (well until an update is forced eg middle click), but again just for new style modules.

So to convert a module you would update the method arguments etc and then just add

```
class Meta:
    pass
```

Once we have this in place we can make more changes easily in the future.  eg.

```
class Meta:
    use_groovy_new_events = True
```

So this is my suggestion.  I'm interested to know what people think of it.  I could easily knock up a patch to demonstrate this but I'll wait until either there is some interest in this idea or it has been shot down in flames.

cornerman:
>  
```response['color'] = 'bad'```

I really like your idea, but I am not in favor of encoding logic in a string. I think, such a magic conversion is difficult to read and it becomes harder to detect errors.

[...]

PR #259, opened 2016-04-10
Change to new calling and py3 object for modules

tobes:
**This is outdated see comment below**

This PR follows on from discussions in issue #250  I'd like to thank @cornerman for the `self.py3` contribution.

What this PR does:

Any module that implements `class Meta` is treated specially.

* Calls to its `on_click` and update methods are called simply as `update()` and `on_click(event)` rather than the old style `update(i3s_output_list, i3s_config, event)` and `on_click(i3s_output_list, i3s_config, event)`.

* A special helper object is injected into the module as `self.py3`.  This helper module includes: 

  -  helper functions:

    - `notify_user()` send message to user.

    - `get_module_info()` get info about other modules (`group` module  needs to do this).

    - `trigger_event()` trigger an event for a module.

    - `update()` force update of self or another module.

  - constant:

    - `CACHE_FOREVER` if returned as `cache_until` causes the module to not be scheduled for an update (events would still trigger an update)

So basically it allows modules to be written in a cleaner way and have easy access to extra functionality.

here is `fish.py` an example module
```
# -*- coding: utf-8 -*-

class Py3status:

    class Meta:
        pass

    def fish(self):
        return {
            'full_text': '>(((º>',
        }

    def on_click(self, event):
        self.py3.notify_user('くコ:彡')
```

Additionally the `group` module is updated

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -11,12 +11,73 @@ from time import time
 from py3status.profiling import profile
 
 
+PY3_CACHE_FOREVER = -1
+
+
+class Py3ModuleHelper:
+    """
+    Helper object that gets injected as self.py3 into Py3status
+    modules that have not got that attribute set already.
+
+    This allows functionality like:
+        User notifications
+        Forcing module to update (even other modules)
+        Triggering events for modules
+    """
+
+    CACHE_FOREVER = PY3_CACHE_FOREVER
+
+    def __init__(self, module):
+        self._module = module
+
+    def update(self, module_name=None):
+        """
+        Update a module.  If module_name is supplied the module of that
+        name is updated.  Otherwise the module calling is updated.
+        """
+        if not module_name:
+            return self._module.force_update()
+        else:
+            module = self.get_module_info(self, module_name).get(module_name)
+            if module:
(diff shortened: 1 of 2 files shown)


tobes:
This is an updated version of the patch based on feedback from @ultrabug 

`class Meta` has been removed.

All modules can now define `update(self)` or `update(self, i3s_output_list, i3s_config)` they will be called appropriately.  same for `on_click()` and `kill()`.

There can even be a mix of old and new calling methods.

`self.py3` is injected into any module that does not have it defined as an attribute.

Example updated:

```
# -*- coding: utf-8 -*-

class Py3status:

    def fish(self):
        return {
            'full_text': '>(((º>',
        }

    def on_click(self, event):
        self.py3.notify_user('くコ:彡')
```

pydsigner:
This will make exposing logging to modules so much easier.

ultrabug:
What can I say, it's very elegant, congratulations :palm_tree:

tobes:
A big thanks to @ultrabug and @cornerman on this.  You both really helped make this with your suggestions. :+1:

PR #262, opened 2016-04-11
Notification message fixes

tobes:
There were some inconsistencies between messages to the user.  Having two routes `i3-nagbar` and `notify-send` made some of this more noticeable.

This cleans up the messages.

I'm not sure if it is still too complicated.  eg `notify_user()` wants a sentence `This is my error.` whereas `report_exception()` wants a fragment `Error happened`.  Maybe we should just add/remove any full stops so that the user does not need to care so much.

I'd be happy to add that, but it might limit us in some circumstances.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -334,7 +334,7 @@ class Module(Thread):
                         syslog(LOG_INFO,
                                'method {} returned {} '.format(meth, result))
                 except Exception:
-                    msg = 'Instance `{}`, user method `{}` Failed'
+                    msg = 'Instance `{}`, user method `{}` failed'
                     msg = msg.format(self.module_full_name, meth)
                     notify = not self.nagged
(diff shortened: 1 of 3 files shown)


py3status/module.py
```diff
@@ -334,7 +334,7 @@ def run(self):
                         syslog(LOG_INFO,
                                'method {} returned {} '.format(meth, result))
                 except Exception:
-                    msg = 'Instance `{}`, user method `{}` Failed.'
+                    msg = 'Instance `{}`, user method `{}` Failed'
```
ultrabug: maybe `failed` without the capital F would be more appropriate here

ultrabug:
I think you're right in doing this yes

py3status/module.py
tobes: updated

PR #263, opened 2016-04-11
Core to do all i3bar output communication

tobes:
Currently `i3status.py` reads the input from `i3status` and then either outputs it to `i3bar` or passes it to `core.py`.

This commit allows `core.py` to do all the out bound communication.

The basic output to i3bar is a json stream starting with version info dict and then an array of arrays.  Before we needed to worry about if we needed a comma between the items etc and would look at the i3status output to determine this.

Now we just send our 'version dict' followed by opening the array and sending an empty initial array.  After that i3bar just expects `,[......]` for each line which is simple to provide.

In turn this simplifies `i3status.py` and we can also remove the `jsonify` helper

One additional advantage (and actually the reason I started this) is that we no longer output the first json sent by `i3status` which makes the start-up a slightly cleaner user experience.

`events.py` is also 'cleaned' up to remove it's need for `jsonify()`

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -490,6 +490,9 @@ class Py3statusWrapper():
         interval = self.config['interval']
         last_sec = 0
 
+        # start our output
+        print_line('{"version": 1, "click_events": true}[[]')
+
         # main loop
         while True:
             sec = int(time())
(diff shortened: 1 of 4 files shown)


tobes:
This conflicts with #266 I'll update if that gets merged.

ultrabug:
smart & clean, thank you

PR #264, opened 2016-04-11
report on_click errors to user for py3status modules

tobes:
on_click errors were not being shown to the user.  This corrects the issue.

Also better error info is provided


ultrabug:
Perfect sense, thanks !

Issue #261, opened 2016-04-10
py3status dies if no i3status modules in order


tobes:
See issue #258 for details

py3status either needs to run with the `-s` option or have an i3status module

PR #265, opened 2016-04-11
Allow py3status to run with no defined i3status modules

tobes:
```
general {
        output_format = "i3bar"
        colors = true
        interval = 5
}

order += "static_string"

static_string {
    format = "hello world"
}
```

This simple config breaks py3status as no i3status modules are defined.

This PR fixes issue #261


ultrabug:
Good catch ! thanks

Issue #179, opened 2016-02-16
pomodoro module is not updating timer when i3bar is not visible


pedro-nonfree:
Hi,

Pomodoro's module timer is stopped when i3bar is not shown in display, situations where I found this:
- using a tty (ctrl+alt+f1)
- with [mode hide](http://i3wm.org/docs/userguide.html#_display_mode)
- in [full screen](http://i3wm.org/docs/userguide.html#_default_keybindings)

I don't know if this could affect other modules that require to be updated.

For me this is a big issue that I don't know how to fix.

ultrabug:
Hello @pedro-nonfree,

Well this behaviour is expected when running with **mode hide** only imho (I can't reproduce when doing full screen or switching to tty).

This is not related to py3status but is instead a feature of i3bar.
- When the bar gets hidden, the i3bar code sends a SIGSTOP to the whole process (py3status included) so everything is freezed.
- When the bar is unhidden, a SIGCONT is sent and the process continues.

This is here : http://code.stapelberg.de/git/i3/commit/?id=a83e7699e75cad515019aa1ad5f7d59ee236a918

The only workaround I could think of regarding pomodoro would be to calculate the remaining seconds based on a timestamp instead of a simple decrementation of an interger. But this couldn't beat the fact that you'd get no notification since the whole process is freezed by i3bar.

Hope this helps mate.

pedro-nonfree:
I'm using self compiled version:
`i3 version 4.11-31-gd3f42c2 (2015-11-16, branch "master") © 2009 Michael Stapelberg and contributors`

in Debian 8.3 Stable Jessie 

I don't know why you cannot reproduce it.

Timestamp is a nice idea!

Anon1234:
@pedro-nonfree another workaround could be recompiling i3 with reverted a83e7699e7 commit.

ultrabug:
Problem with timestamp @pedro-nonfree is that you still would never get your pomodoro's notifications ! This defeats the whole point of it don't you think ?

pedro-nonfree:
@Anon1234 Oh no, back to 2010 ? :cry: 

@ultrabug Use case is not always being in this "hide" situations. So you receive a late notification, better than nothing.
Uh, and I don't know why you cannot reproduce this for full screen (Mod + F) or going to tty.

Thinking in a better design, what about starting an independent background process for the timer and notification, and share that info with py3status to print it in the bar?

This is in general useful for other modules such as battery notifications, perhaps for new possible modules?

[...]

Issue #253, opened 2016-04-07
Asynchronous modules stop processing events on SIGSTOP


cornerman:
When i3bar is hidden, the statusline generator gets a SIGSTOP signal. Thus, the asynchronous modules cannot process new events, which currently leads to https://github.com/i3/i3/issues/2280 in i3 - ultimately, this can freeze i3. Depending on how this is solved, it might be desirable to have the event processing in a different process, so it is not stopped. What do you think?

This is also mentioned in https://github.com/ultrabug/py3status/issues/179 for a different use case.

tobes:
Interesting stuff.  Do we have an easy way to reproduce this?

cornerman:
Yeah, it is rather easy to reproduce.

First, activate an async module in your i3status.conf:
```order += "window_title_async"```

Then, hide your bar, so py3status gets stopped:
```$ i3-msg bar mode hide```

Trigger the event a few times, i.e., switch window focus a few times (until i3 freezes).

tobes:
Hmm, this is an nice problem.  So really the modules need to run even when py3ststus is stopped.  But modules like `group.py` hook quite deeply into py3status.

Maybe we need to have py3status as a thin layer that just spawns the actual py3status and reads/writes data to/from i3bar etc.  Since we can't trap SIGSTOP and have to live.

cornerman:
Yeah, at least the asynchronous modules should not be stopped. A thin py3status main process, which forks of the actual worker and prints data sounds like a good idea. Another possibility would be to centralize the event processing into another process. But then this process would need to buffer incoming events when the receiving modules are stopped - not sure whether this would be a viable solution.

tobes:
Is this due to a recent change in i3, or just a long standing issue that hasn't been noticed before?

I think this will need some thinking through to get a nice fix, as every there are lots of possible options.

cornerman:
Afaik, this is a long standing issue; the IPC interface was designed to have well-behaving clients. I agree, we should definitely take the time to think about a good solution and also see, how i3 solves it.

tobes:
So my thought is

`py3status` - spawns new process (py3) write to i3bar and gets i3bar event messages

`py3` - reads config, starts modules including i3status ones, creates i3bar ready output.

We then use shared memory to communicate.

__py3__ 

writes
- `output` - unicode output in i3bar format
- `updated` - boolean set to true when output has been updated

reads
- `event_queue` - array of event strings

__py3status__

reads
[...]

PR #266, opened 2016-04-12
Keep alive when i3bar hidden

tobes:
By default i3bar sends a -SIGSTOP to a process sending it data.  This causes
some modules interacting with i3 ipc to be suspended, which lead to locking
issues inside i3.  This patch switches to using SIGUSR2 as the suspend
signal.

The async modules eg window_title_async were affected.

see https://github.com/ultrabug/py3status/issues/253

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta, tzinfo
 from subprocess import Popen
 from subprocess import PIPE
 from syslog import syslog, LOG_INFO
+from signal import SIGUSR2
 from tempfile import NamedTemporaryFile
 from threading import Thread
 from time import time
@@ -540,6 +541,8 @@ class I3status(Thread):
                                 if 'version' in line:
                                     header = loads(line)
                                     header.update({'click_events': True})
+                                    # set custom stop signal
+                                    header['stop_signal'] = SIGUSR2
                                     line = dumps(header)
                                 print_line(line)
                             else:
@@ -548,8 +551,13 @@ class I3status(Thread):
                                     self.last_prefix = prefix
                                     self.set_responses(json_list)
                         else:
-                            err = self.poller_err.readline()
                             code = i3status_pipe.poll()
+                            # If we are 'suspended' then we recieve this
+                            # signal from i3status and can safely ignore
+                            # it.
+                            if code == -SIGUSR2:
+                                continue
+                            err = self.poller_err.readline()
                             if code is not None:
                                 msg = 'i3status died'
                                 if err:
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -523,15 +523,13 @@ class I3status(Thread):
                        'i3status spawned using config file {}'.format(
                            tmpfile.name))
 
-                def preexec():
-                    # Ignore the SIGUSR2 signal for this subprocess
-                    signal(SIGUSR2, SIG_IGN)
-
                 i3status_pipe = Popen(
                     ['i3status', '-c', tmpfile.name],
                     stdout=PIPE,
                     stderr=PIPE,
-                    preexec_fn=preexec)
+                    # Ignore the SIGUSR2 signal for this subprocess
+                    preexec_fn=lambda:  signal(SIGUSR2, SIG_IGN)
+                )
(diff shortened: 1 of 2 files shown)


tobes:
@cornerman could you just confirm that this is working for you.  It should all be good but worth double checking.

tobes:
Note:  this patch just fixes the immediate issue of i3 crashing.

Further work is needed.

* We should put i3status to sleep when we are suspended.

* We should stop py3status modules too (any that need to stay alive should be using a thread).

cornerman:
It works for me!

cornerman:
Wait, this does not work as expected. Sorry, but I overlooked that i3status actually crashes, so we cannot just continue when we receive USR2 from i3status. There is something else going on...

cornerman:
By default, signals are dispatched to all subprocesses and i3status exits when it receives a SIGUSR2. So, in order to avoid sending this signal to i3status, we could do this [when we open the subprocess](https://github.com/tobes/py3status/blob/SIGUSR2/py3status/i3status.py#L520):

```
def preexec():
    # Ignore the SIGUSR2 signal for this subprocess
    signal.signal(signal.SIGUSR2, signal.SIG_IGN)

i3status_pipe = Popen(
                ['i3status', '-c', tmpfile.name],
                stdout=PIPE,
                stderr=PIPE,
                preexec_fn=preexec)
```

We could probably also translate it to SIGSTOP there.

tobes:
Hmm this is frustrating.  Yesterday I could reproduce this bug today I can't.

@cornerman is there a minimal `i3status.conf` that will hit this for you on master?

```
general {
        output_format = "i3bar"
        colors = true
        interval = 5
}


order += "static_string"
order += "load"
order += "window_title_async"


static_string {
    format = "hello world"
}

window_title_async {
    always_show = True
}
```
feels like it should be enough but I can't trigger stuff

cornerman:
> is there a minimal i3status.conf that will hit this for you on master?

Seems that it happens with any config for me. What happens if you do:

```
killall -s USR2 i3status
```

For me, all i3status processes exit and py3status busy waits and as we now ignore that i3status exited.

cornerman:
This will also fix https://github.com/ultrabug/py3status/issues/179

tobes:
OK, I can trigger it again via `i3-msg bar mode hide` and `i3-msg bar mode dock`

Should have an updated patch soon. Thanks

tobes:
@cornerman could you test the latest version which includes your fix and also should stop i3status (which should wake up too)

cornerman:
Nice, it works!

tobes:
Nice bit of teamwork there :+1: 

@ultrabug this patch should be ready for consideration now

py3status/i3status.py
```diff
@@ -548,8 +565,13 @@ def run(self):
                                     self.last_prefix = prefix
                                     self.set_responses(json_list)
                         else:
-                            err = self.poller_err.readline()
                             code = i3status_pipe.poll()
+                            # If we are 'suspended' then we recieve this
+                            # signal from i3status and can safely ignore
+                            # it.
+                            if code == -SIGUSR2:
```
ultrabug: sorry if it should be obvious, but why negate the SIGUSR2 value here ?

ultrabug:
You guys baffle me :eyes: quick question and I'll merge. Thanks

py3status/i3status.py
```diff
@@ -516,14 +523,22 @@ def run(self):
                        'i3status spawned using config file {}'.format(
                            tmpfile.name))
 
+                def preexec():
+                    # Ignore the SIGUSR2 signal for this subprocess
+                    signal(SIGUSR2, SIG_IGN)
+
                 i3status_pipe = Popen(
```
ultrabug: why not self.i3status_pipe here directly ?

py3status/i3status.py
tobes: That was an odd thing but is no longer needed as we do not send the SIGUSR2 to i3status now.  Maybe an i3status bug?

py3status/i3status.py
tobes: I've moved this into the Popen def is that what you meant?

tobes:
I've updated this should be better now

ultrabug:
Thank you, yes.

PR #267, opened 2016-04-12
cache async modules until event was received

cornerman:
Use the new py3 style for async modules to cache their responses forever and force an update as soon as an event was received.


ultrabug:
Awesome :+1:

PR #268, opened 2016-04-12
Py3status sleep

tobes:
This extends SIGUSR2 to sleep py3status modules.  We should really sleep when we have been told to.

When sleeping we stop the update timer.  On waking we reset the timer for the requested cache_until if in the future, else do an update immediately.

```
# pause py3status
killall -USR2 py3status

# restart py3status
killall -CONT py3status
```

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -236,9 +236,9 @@ class Py3statusWrapper():
         # set the Event lock
         self.lock.set()
 
-        # SIGUSR2 will be recieced from i3bar indicating that all output should
+        # SIGUSR2 will be received from i3bar indicating that all output should
         # stop and we should consider py3status suspended.  It is however
-        # important that any processes using i3 ipc should continue to recieve
+        # important that any processes using i3 ipc should continue to receive
         # those events otherwise it can lead to a stall in i3.
         signal(SIGUSR2, self.i3bar_stop)
         # SIGCONT indicates output should be resumed.
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -486,9 +486,23 @@ class Py3statusWrapper():
         self.i3bar_running = False
         # i3status should be stopped
         self.i3status_thread.suspend_i3status()
+        self.sleep_modules()
 
     def i3bar_start(self, signum, frame):
         self.i3bar_running = True
+        self.wake_modules()
+
+    def sleep_modules(self):
+        # Put all py3modules to sleep so they stop updating
+        for module in self.output_modules.values():
+            if module['type'] == 'py3status':
+                module['module'].sleep()
+
+    def wake_modules(self):
+        # Wake up all py3modules.
+        for module in self.output_modules.values():
+            if module['type'] == 'py3status':
+                module['module'].wake()
 
     @profile
     def run(self):
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -523,15 +523,13 @@ class I3status(Thread):
                        'i3status spawned using config file {}'.format(
                            tmpfile.name))
 
-                def preexec():
-                    # Ignore the SIGUSR2 signal for this subprocess
-                    signal(SIGUSR2, SIG_IGN)
-
                 i3status_pipe = Popen(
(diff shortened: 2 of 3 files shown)


py3status/core.py
```diff
@@ -235,6 +236,14 @@ def setup(self):
         # set the Event lock
         self.lock.set()
 
+        # SIGUSR2 will be recieced from i3bar indicating that all output should
```
pydsigner: `recieced` -> `received`

py3status/core.py
```diff
@@ -235,6 +236,14 @@ def setup(self):
...
+        # stop and we should consider py3status suspended.  It is however
+        # important that any processes using i3 ipc should continue to recieve
```
pydsigner: `recieve` -> `receive`

pydsigner:
How well does this change handle notifications which could still be desired while the status bar is hidden?

tobes:
> How well does this change handle notifications which could still be desired while the status bar is hidden?

Good question.  How it works is like this.

When we ask a module to sleep we cancel any timer it has set to run in the future set `self.sleeping = False` which prevents it from setting a new timer.  If it happens to be updating at the time we let it finish.

It can still be forced to update, if its `force_update()` method is run.  So for example with `window_title_async` it can force an update from it's thread (assuming @cornerman has his PR accepted) so the module will have the correct info but it just won't display it till the bar reappears.

When we wake the module it checks it's `self.cache_until` and updates if it was due else sets the timer back up as before.

So to `stay alive` a module would need to create it's own thread like `window_title_async` that would keep awake.  It can trigger updates if it wishes and they will be processed.  So we don't really sleep we just stop calling methods on things automatically.

So modules cannot rely on being called regularly.  Any modules that are broken were anyway as the whole py3status process was stopped previously.

So to answer your question.  If the module is doing a notification by playing a sound calling `notify_user()` etc then that would work but it needs to be done from a thread.

Now I think that having something like a `self.py3.alarm_at(callback)` general helper would be great for say  `pomodoro` so every module wouldn't need to start it's own thread.

I'd also like a `self.py3.play_sound()` helper too :)

pydsigner:
Sounds good.

py3status/i3status.py
```diff
@@ -548,8 +565,13 @@ def run(self):
                                     self.last_prefix = prefix
                                     self.set_responses(json_list)
                         else:
-                            err = self.poller_err.readline()
                             code = i3status_pipe.poll()
+                            # If we are 'suspended' then we recieve this
+                            # signal from i3status and can safely ignore
+                            # it.
+                            if code == -SIGUSR2:
+                                continue
```
cornerman: Oh, this is not needed anymore!

tobes: I was thinking this too. I'll remove it

ultrabug:
As we promised the pomodoro module was fixed, I'd like to kindly request that it is taken care of in this PR please ?

tobes:
> pomodoro module fix

Ok

tobes:
>    pomodoro module fix

Actually looking at it it is not as bad as I thought.  So it sort of should be working (as in counts down but doesn't do the alert)

To really fix it I'd like to add 

*  `py3.alarm_at(callback)` so modules don't need to do any threading themselves for something as simple as this.  Could be useful for many things.

* `py3.play_sound(file)` again so a general helper for all modules (using `aplay` or `paplay` whichever is available).

That all feels it should be a separate PR.

tobes:
.... or maybe that's just over complicating things, I'll ponder

tobes:
I've had a ponder and a look at the code.

The reality is that pomodoro has got a bit messy over time (it also seems to have very fast seconds). I think it needs a clean up especially to work with a Timer.  Trying to hack it in looks harder than just doing a refresh of stuff and simplifying the logic.

Doing this will make make a mess of this PR.  So I'll do it as a separate one.  If you want to wait for that to pull this that's fine.

I'm going to just do the pomodoro though and not mess with py3 I think otherwise it'll get icky.

ultrabug:
Okay @tobes I get what you're saying and you're right thanks !

PR #269, opened 2016-04-12
Allows non-strings to be notified on dbus

tobes:
When using dbus for non `info` non-string messages would fail eg `notify_user([1, 2, 3])`

This fixes the issue.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -302,7 +302,8 @@ class Py3statusWrapper():
         if not dbus:
             msg = 'py3status: {}'.format(msg)
         if level != 'info':
-            msg += ' Please try to fix this and reload i3wm (Mod+Shift+R)'
+            fix_msg = '{} Please try to fix this and reload i3wm (Mod+Shift+R)'
+            msg = fix_msg.format(msg)
         try:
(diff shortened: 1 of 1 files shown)


py3status/core.py
```diff
@@ -302,7 +302,8 @@ def notify_user(self, msg, level='error'):
         if not dbus:
             msg = 'py3status: {}'.format(msg)
         if level != 'info':
-            msg += ' Please try to fix this and reload i3wm (Mod+Shift+R)'
+            fix_msg = '{} Please try to fix this and reload i3wm (Mod+Shift+R)'
```
ultrabug: `msg = '{} Please try to fix this and reload i3wm (Mod+Shift+R)'.format(msg)` and avoid a variable creation ?

tobes: I had that before but it was a very long line. Can it stay like this?

tobes: A compromise? (shorter message)
`msg = '{} Please fix and reload i3wm (Mod+Shift+R)'.format(msg) `

ultrabug:
Okay, it's not such a big deal

Issue #174, opened 2016-01-29
Wrong module guess for click event with multiple volume blocks


cornerman:
I have defined two volume blocks for two different pulse devices. When defining click events, clicking on either block runs the onClick action of the first module.

The configuration looks like this:
```
volume master {
    device = "pulse:0"
    on_click 1 = "exec pavucontrol"
nus"
}

volume slave {
    device = "pulse:1"
}
```

Log when running py3status with this configuration and clicking on the first block:
```
Jan 29 18:13:13 genius /py3status[24905]: received event {'y': 7, 'instance': 'pulse:0.Master.0', 'x': 1061, 'name': 'volume', 'button': 1}
Jan 29 18:13:13 genius /py3status[24905]: trying to dispatch event to module "volume master"
```

Clicking on the second block does not dispatch the event to the second module ```volume slave```, but again to the first one ```volume master```:
```
Jan 29 18:13:37 genius /py3status[24905]: received event {'y': 12, 'instance': 'pulse:1.Master.0', 'x': 1108, 'name': 'volume', 'button': 1}
Jan 29 18:13:37 genius /py3status[24905]: trying to dispatch event to module "volume master"
```

The instance in the reply does not correspond to the descriptor ```master/slave```, but is a concatenation of the ```device```, ```mixer``` and ```mixer_id```. [The guessing method for the module](https://github.com/ultrabug/py3status/blob/master/py3status/__init__.py#L806) takes this into account but relies on the presence of all three configuration parameters. If any of the three parameter is not set, the guessing will not work correctly. Instead, the check should just ignore unset parameters and get the closest match for device, mixer, and then mixer_id.

As a **workaround**, define all three configuration parameters explicitly:
```
volume slave {
    device = "pulse:1"
    mixer = "master"
    mixer_id = "0"
}
```

py3status version 2.8

ultrabug:
As discussed on IRC, this is not directly a py3status bug. It is related to the fact that i3status does not send the same instance name from its config to the event message... I'll find a workaround !

ultrabug:
fixed by #270 thanks to @tobes

PR #270, opened 2016-04-13
Correct i3status module names

tobes:
Whilst working on something different I was hit by a variation of issue #174 (i3status returning different name/instance than it was given).

This PR solves the issue in a simple way.  We ignore what i3status claims and just add the correct name/instance that are defined in `i3status.conf`.  The upshot of this is that the events are handled correctly too.

PS. Sorry I keep adding so many PRs


ultrabug:
:eyeglasses: I see dead code

@cornerman mind testing wrt issue #174 and report that this fixes the issue please ?

cornerman:
Great, this fixes the issue!

@tobes thanks for working on this :)

ultrabug:
neat ! cheers

PR #273, opened 2016-04-13
Move Py3 to own file

tobes:
This PR moves the `Py3ModuleHelper` class into it's own file.  It also renames the class to `Py3`.

Reasoning:

* Seeing as it will be mainly known through the use of `self.py3` the class name is more intuitive.
* It is likely that the functionality of the class will expand.  This will help keep that cleaner and not confuse `module.py`.
* Other parts of py3status can import the class (or more importantly the defined constants).

Internally `class Py3` is unchanged.  A small spelling correction was made to `Module` docstring

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -8,71 +8,14 @@ from collections import OrderedDict
 from syslog import syslog, LOG_INFO, LOG_WARNING
 from time import time
 
+import py3status.py3 as py3
 from py3status.profiling import profile
-
-
-PY3_CACHE_FOREVER = -1
-
-
-class Py3ModuleHelper:
-    """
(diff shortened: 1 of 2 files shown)


py3status/module.py
```diff
@@ -8,71 +8,14 @@
 from syslog import syslog, LOG_INFO, LOG_WARNING
 from time import time
 
+import py3status.py3 as py3
```
ultrabug: explicit maybe ? 

`from py3status.py3 import Py3, PY3_CACHE_FOREVER` ?

tobes:
Import now explicit.

ultrabug:
:+1:

PR #274, opened 2016-04-13
Revert "Workaround, removed PlaybackStatus query"

AnwariasEu:
This reverts commit 0f41e58099c7fc0da47f290c19e684ef923a9904.
Spotify fixed dbus playback status.


ultrabug:
Mind you share from which version this has been fixed ? I think we should add this information on the docstring so users having an old version won't open issues about this !

ultrabug:
And thanks a lot for reporting back mate !

PR #275, opened 2016-04-14
Composites fun

tobes:
This PR introduces composites, this was inspired by  #176 by @cornerman 

For full functionality i3bar 4.12 is needed but it will work (though degraded) for previous versions.

__What does this offer?__

* py3status treats a composite as it would any other normal module output (so can be added to groups etc).

* a modules response can be split into parts that together make a composite.

* each part of a composite can define it's own color.

* each part of a composite can catch it's own click events.

* the whole module can still catch events for any component part.

* `group` module now no longer conflicts with the click events of its contents

* `group` module can now be opened/closed to hide show its content.

__Demo__

The power of composites can be seen by adding the following to your `i3status.conf`.  It introduces the `timer` module.  You can change the time via mouse wheel by hours/minutes/seconds clicking starts/pauses and resets the timer.  The timer is in an updated group that can be opened/closed as well as changing the displayed content (timer or static_string)

```
order += "group demo"

group demo {
    timer {}
    static_string hello { format = 'Hello' }
}
```

for old i3bar versions that do not support `separator` and `separator_block_width` the separator will be shown but otherwise it will function as expected.

The changes to core py3status are not that big, but the group module did need quite a few changes.

__Changes__

* Instead of a response containing `full_text` it can contain `composite` which is a list of responses.

* Events now include `index` which is either the part count of a composite or the index name supplied in the response

### DIFF
--- a/py3status/events.py
+++ b/py3status/events.py
@@ -203,6 +203,13 @@ class Events(Thread):
                         dispatched = True
                         break
 
+        if not dispatched:
+            # find container that holds the module and call its onclick
+            module_groups = self.i3s_config['.module_groups']
+            containers = module_groups.get(module_name)
+            for container in containers:
+                self.process_event(container, event)
+
         # fall back to i3bar_click_events.py module if present
         if not dispatched:
             module = self.i3bar_click_events_module()
--- a/py3status/events.py
+++ b/py3status/events.py
@@ -313,6 +313,19 @@ class Events(Thread):
                     instance = event.get('instance', '')
                     name = event.get('name', '')
 
+                    # composites have an index which is passed to i3bar with
+                    # the instance.  We need to separate this out here and
+                    # clean up the event.  If index
+                    # is an integer type then cast it as such.
+                    if ' ' in instance:
+                        instance, index = instance.split(' ', 1)
+                        try:
+                            index = int(index)
+                        except ValueError:
+                            pass
+                        event['index'] = index
+                        event['instance'] = instance
+
                     # i3status module name guess
                     instance, name = self.i3status_mod_guess(instance, name)
                     if self.config['debug']:
@@ -322,7 +335,7 @@ class Events(Thread):
                                 '{} {}'.format(name, instance).strip()))
 
                     # guess the module config name
-                    module_name = '{} {}'.format(name, instance).strip()
+                    module_name = '{} {}'.format(name, instance.split(' ')[0]).strip()
                     # do the work
                     self.process_event(module_name, event)
 
--- a/py3status/events.py
+++ b/py3status/events.py
@@ -203,12 +203,11 @@ class Events(Thread):
                         dispatched = True
                         break
 
-        if not dispatched:
-            # find container that holds the module and call its onclick
-            module_groups = self.i3s_config['.module_groups']
-            containers = module_groups.get(module_name)
-            for container in containers:
-                self.process_event(container, event)
+        # find container that holds the module and call its onclick
+        module_groups = self.i3s_config['.module_groups']
+        containers = module_groups.get(module_name)
+        for container in containers:
+            self.process_event(container, event)
 
         # fall back to i3bar_click_events.py module if present
         if not dispatched:
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -172,6 +172,9 @@ class Module(Thread):
         composite = response['composite']
         if not isinstance(composite, list):
             raise Exception('Expecting composite')
+        # if list is empty nothing to do
+        if not len(composite):
+            return
         if 'full_text' in response:
             raise Exception('Conflict "full_text" and "composite" in response')
         # set universal options on last component
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -124,7 +124,11 @@ class Module(Thread):
     def get_latest(self):
         output = []
         for method in self.methods.values():
-            output.append(method['last_output'])
+            data = method['last_output']
+            if isinstance(data, list):
+                output.extend(data)
+            else:
+                output.append(data)
         return output
 
     def set_module_options(self, module):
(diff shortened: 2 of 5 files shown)


pydsigner:
This looks awesome.

pydsigner:
FYI @ultrabug, the lines changes from v2.0 to v2.9 were +6300-1100. From v2.9 to current master is +5000-3000. We might want to start thinking about wrapping up for a release (v3.0?) so that people can get their hands on this without going through git.

ultrabug:
Yes definitely I'm also willing to get a first 3.0 release ASAP, I'll open an issue to track it

tobes:
I've updated the groups module to simplify it a little and to fix format parsing.

ultrabug:
ooooook I finally reviewed the *huge* code and tested this out. I must admit that I'm troubled by the fact that click events only work on the {button} now and this was kind of unexpected so I'm sure it will also trouble regular users who simply want the simple module grouping.

2016-04-19-093220_429x173_scrot

Also, I have IndexError errors when using group without the {button} placeholder.

tobes:
I'll look at the unicode issue.

The thing with the click events and the button is that I have a volume control in a group which uses the same clicks to change as the group uses.  I think this is not an uncommon usage.  having the 'button' stops this being a problem.

> I'm sure it will also trouble regular users who simply want the simple module grouping.

I feel whilst it is only in master not released then we are free to change things.  However, I could change things so that we could have it work in both ways but we'd need to make it somehow clear to the user how it works.  I'll have a think but fix the other issues in the meantime.

tobes:
Unicode issue fixed, I think it's your font, but we should be aware that unicode support is not great for all characters so I've switched to ascii to be safe.

Index issue is also fixed.

Plus merge conflict.

tobes:
So the button issue.

currently if the group has any format content other than `{output}` eg `format = "moo {output}"` then that responds to click events but the `{output}` does not capture them.

I think this behaviour makes sense as the {output} is another module and so it should do it's clicks.  Now if that does not respond to the clicks then maybe it makes sense to pass them to the group.  But if I have 
```
group {
   module_no_clicks {}
   module_catches_clicks {}
}
```
then it would be inconsistent when we are using it.

Anyhow I have added a patch that allows uncaught events to be caught and passed to the container (group) that allows the old behaviour (but sticks on a module that captures events).  Maybe we should do some 'clever' stuff at setup eg if all contained modules have no click events and no button then pass on the events.  So anyway I think we are back in a position when all is possible via simple changes we just need to nail down the behaviour we are expecting in the different situations.

This needs to make sense to actual users.

ultrabug:
I just want to apologize for lagging behind. My free time is very limited due to my working on preparing my future flat so we can move there with our newborn asap.

This PR is next on my queue, please bear with me.

tobes:
Congratulations and take all the time you need.  Hope you are getting enough sleep!

ultrabug:
> This needs to make sense to actual users.

The initial behaviour that groups implemented is that they are click agnostic. If you add a on_click configuration to a group of modules you **stack** it to any possible click handler defined on the module and I really think it's an understandable behaviour as it sounds straightforward (to me at least).

No magic or behavioural change is made, it just **all adds up**.

tobes:
> The initial behaviour that groups implemented is that they are click agnostic. 

This is true.  However at the point of their implementation they did not have much of an alternative.  Either they allowed all clicks to propagate or none.

With the composite idea we can now determine within a module where the click has occured so can be much more nuanced about how we do things.  I would also strongly argue that whilst the group module has not been in a release, that changes to how they function is completely allowed.

>  you stack it to any possible click handler defined on the module

To me that doesn't make sense.  I would also say that the group module is the first container of other modules and as such there is no current 'default' behaviour here.

also

```
        # fall back to i3bar_click_events.py module if present
        if not dispatched:
            module = self.i3bar_click_events_module()
```
in events.py suggests that this is not true (incidentally that code should probably die soon)

Given a choice I'd also have on actioned click event stop propogation of the event but that is not part of this discussion.

Anyhow the thing is if a group contains a module that uses the same buttons as the group does then they clash.  in my mind that causes problems.  The 'button' allows for the group to have a handle to change its content and additionally hide/show which I think is a useful feature.

From my point of view I see a future where we make more use of containers (hence the config parser branch) and for those again the button is a useful concept and we have a chance now to introduce it given the current "let's not break things for he user" which I generally support.

I can easily change things so we can have both the "catch events and pass them on" and the "button" approach available to the user either an option (or actually seeing what click events exist, but that might be too magic)

Have you tried having a group that contains a module which uses the same click events?  It is frustrating and feels totally broken.

```
group tz {
    cycle = 0
format = "{output}"

volume_status {
    button_up = 4
    button_down = 5
    button_mute = 2
}

    tztime ny {
        format = "NY %H:%M:%S"
        timezone = "America/New_York"
    }

    tztime du {
        format = "DU %H:%M:%S"
        timezone = "Asia/Dubai"
    }
}
```

tobes:
With the above config try it in both master and this branch.

We can always have two modules one the way you like and another the way I do (even if out of tree) but it would be nice to keep it unified in some way if we can.

pydsigner:
@tobes The solution should be though to set `button_forward` on the group in the config rather than implementing mass amounts of more complexity. We do need a good way to propagate clicks consistently, I will not disagree on that, as the current on_click behavior gives no indication of which buttons are actually bound to events.

tobes:
> The solution should be though to set button_forward on the group in the config rather than implementing mass amounts of more complexity. 

I'd be happy to do something like that.  I'm more than happy to allow @ultrabug to get the behaviour he wants if I get mine too ;p. 

I not sure of the name. I'd prefer something like `catch_all_clicks` because technically the contained module receives the click before the group gets it.  Also I like `button_xxx` for assigning functions to mouse clicks (like group does) in the same way that `format_xxx` is used for output formats. 

> We do need a good way to propagate clicks consistently, I will not disagree on that, as the current on_click behavior gives no indication of which buttons are actually bound to events.

This is true - personally I'd like modules to return a True/False from `on_click()`, if they have actioned an event.  Existing modules would return None so we could do stuff there.  But that discussion is for another day perhaps.

A module can tell if the click events are set in the config eg `on_click 1 = '...'` although there is not an easy route but a helper in `py3` would be trivial to add.

tobes:
> I'd prefer something like `catch_all_clicks`

Actually something like `click_mode` with options `all` and `container`, that way it could be expanded in future if there is a need.

tobes:
I've changed the group module to have a `click_mode` option.  It accepts `all` or `button`.

It defaults to `all` with format of `"{output}"`.  
If set to `button` the default format is `"{output} {button}"`.

The group can only be toggled open and closed only via clicking on the `{button}` as that felt right but I can change the behaviour if people want.

So this keeps the group module as it was, but allows the don't use the clicks on content option if wanted.

ultrabug:
Finally !

Thanks so much for your patience. I'll make a quick commit to fix some conventions.

Also, the more I think of it the more I'm inclined to just honour per module on_click and not on the main group module.

This would be OK:

```
group tz {
	cycle = 10
	format = "{output}"

	tztime la {
		format = "LA %H:%M"
		on_click 1 = "exec firefox-bin https://weather.com/weather/tenday/l/Los+Angeles+CA+USCA0638:1:US"
	}

	tztime ny {
		format = "NY %H:%M"
		timezone = "America/New_York"
	}

	tztime du {
		format = "DU %H:%M"
		timezone = "Asia/Dubai"
	}
}
```


This would not:

```
group tz {
	cycle = 10
	format = "{output}"
	on_click 1 = "exec my_random_command.sh"

	tztime la {
		format = "LA %H:%M"
		on_click 1 = "exec firefox-bin https://weather.com/weather/tenday/l/Los+Angeles+CA+USCA0638:1:US"
	}

	tztime ny {
		format = "NY %H:%M"
		timezone = "America/New_York"
	}

	tztime du {
		format = "DU %H:%M"
		timezone = "Asia/Dubai"
	}
}
```

PR #276, opened 2016-04-15
[3.1] Sleeping Dogs: reduce the amount of updating needed especially for containers

tobes:
This PR is aimed at reducing the amount of updates required.

* py3status modules now only trigger an update if their content has actually changed.

* containers (currently only groups) can now register a function that allows py3status to be aware of the active content so updates are only triggered for content that is being displayed.  If they have not registered a function then all sub-module updates trigger an update.

This leads to a reduction in overall workload.

__example:__
```
group {
    time {}
    volume_status {}
}
```

When the group shows `time` we are updated every second. When we show `volume_status` we only update when the volume has changed etc.


ultrabug:
Plz rebase if it's still a point to address ?

tobes:
Yeah this is still good

ultrabug:
Very cool addition, thanks !

PR #278, opened 2016-04-15
Do not send format_time to i3status time modules - fix #177

### LINKED ISSUE #177: Support for format_time in tztime module
The tztime module from i3status introduced another ```format_time``` option to enable pango markup: https://i3wm.org/i3status/manpage.html#_tztime

As py3status handles the tztime module, this new option needs to be taken into account when parsing the time's format. Currently, when using ```format_time```, the time stops updating and adds the ```%Y,%m,%d``` formatstrings to the time (see [here](https://github.com/ultrabug/py3status/blob/master/py3status/__init__.py#L413)).

tobes:
We should not pass format_time to i3status time modules as we handle this in py3status

fixes #177

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -497,8 +497,11 @@ class I3status(Thread):
                     # Set known fixed format for time and tztime so we can work
                     # out the timezone
                     if section_name.split()[
-                            0] in TIME_MODULES and key == 'format':
-                        value = TZTIME_FORMAT
+                            0] in TIME_MODULES:
+                        if key == 'format':
+                            value = TZTIME_FORMAT
+                        if key == 'format_time':
+                            continue
                     if isinstance(value, bool):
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks !

ping @cornerman for testing plz

cornerman:
This only fixes the value error, but does not give the proper result. Using the example from the i3status manpage:
```
tztime time {
    format = "<span foreground='#ffffff'>time:</span> %time"
    format_time = "%H:%M %Z"
}
```

The result is:
```
"full_text": "2016-04-15 16:32:27 CEST"
```

But it should be:
```
"full_text":"<span foreground='#ffffff'>time:</span> 16:32 CEST
```

cornerman:
No, made a mistake in my config. Works!

ultrabug:
good news, thanks !!

PR #279, opened 2016-04-15
[3.1] Color everywhere

tobes:
Now that #270 has been merged this branch is now ready.

What does this do?

We can now use colors all over our config

```
general {
    color: '#FF00FF'
}

time {
    color: '#FFFF00'
}
```

All modules seem to have learned about color. 

It also adds `py3.COLOR_GOOD` etc.  These can be used by modules and will be replaced by the correct color (either the globally defined or one defined in the modules own config section)

`modules/dpms.py` has been updated by way of a demo.

The big change is in `modules.py`
```
-                if not config.startswith('.'):
+                if not config.startswith('.') and hasattr(self.module_class, config):
```

This means that config params only get passed to modules if they have them defined.  This prevents things like `color`, `color_bad` being passed to modules.  It is possible this will break a badly written module, but I think that fixing the module would be more important. 

Other work needed.

We should probably remove color options from modules like `static_string`, `group` etc that implement it internally.


egeskow:
Some modules using bunch of colors like:
Example:                                                                                              
    color_one = "#FFFFFF"                                                                              
    color_two = "#FFFFFF"                                                                             
    color_six = "#FFFFFF"

`vnstat` have own way:
`coloring = { 800: "#dddd00", 900: "#dd0000", } `

`keyboard_layout` able to give a string:
`"us=#FCE94F, fr=#729FCF"`
There is opened pull request #337 to pomodoro module with same idea

Should we create a general interface `colors` with random color names to all modules?
`colors = "good=#FFFFFF, bad=#FFFFFF, degraded=#FFFFFF", color=#FFFFFF, some=#FFFFFF`

py3status/py3.py
```diff
@@ -26,11 +26,13 @@ class Py3:
     """
 
     CACHE_FOREVER = PY3_CACHE_FOREVER
+
     LOG_ERROR = PY3_LOG_ERROR
     LOG_INFO = PY3_LOG_INFO
     LOG_WARNING = PY3_LOG_WARNING
 
     def __init__(self, module=None, i3s_config=None):
+        self._colors = {}
```
ultrabug: bad kitties don't like alphabetical ordering

py3status/py3.py
```diff
@@ -42,6 +44,25 @@ def __init__(self, module=None, i3s_config=None):
...
+        """
+        if not name.startswith('COLOR_'):
+            raise AttributeError
+        name = name.lower()
+        if name not in self._colors:
+            if self._module:
+                color_fn = self._module._py3_wrapper.get_config_attribute
+                color = color_fn(self._module.module_full_name, name, 'color')
+            else:
+                color = self._i3s_config.get(name)
```
ultrabug: means that color can be 'None' ?

tobes: yes but that is only if we do stand alone testing so I'm ok with that as they could add to test config if they wanted

ultrabug: fine, thanks

py3status/i3status.py
```diff
@@ -248,6 +250,11 @@ def write_tmp_i3status_config(self, tmpfile):
             section = self.config[section_name]
             self.write_in_tmpfile('%s {\n' % section_name, tmpfile)
             for key, value in section.items():
+                # don't include color values
```
ultrabug: ... except in the general section

tobes: looking at the i3status docs the modules themselves do not accept color_bad etc

or is this just about a poor comment?

tobes:
I was thinking.  Seeing as we are passing 'dynamic' config params to modules eg `DP1_pos` maybe we should just pass the `color_*` ones as well, just to make it consistent.  It will also keep any existing modules that use `color_*` config params working.

I blocked these originally so we didn't inject unwanted stuff into modules but maybe that was just over thinking things.

tobes:
Ok,  I changed stuff so all format params (not starting with `.`) are passed to the modules like before.  That makes the most sense.

I think we should also add a "Color options:" section to modules docstrings to document available colors.  I did this with dpms as an example.

Other things we could do.

* Fix for i3status modules color_good etc, currently i3status modules just use color_good etc from general section.  I could add some code so we can also define it in the config section for a module.

* Add some documentation for users / module devs

tobes:
Ok, I think this is now ready to roll as far as I'm concerned.

I've made some changes to how the color gets selected.  This means that you can now do `color=None` and it will end up as the default color defined in i3bar section of your i3 config.  i3bar is quite happy getting `'color': null`.

The next stage will be to convert the existing modules that use colors.  Some are easy, but others like wifi use `down_color` which we should change to `color_down` or `color_off` or `color_offline` or `color_disabled` I think the fewer 'standard' ones we have the better. I think we should think about separately from here.

I got carried away and updated keyboard_layout.  I've tried to keep it so it will not change for the user so we keep the old defaults.

doc/README.md
```diff
@@ -440,6 +476,72 @@ click_info {
...
+
+`self.py3` in our module has color constants that we can access, these allow the user to set colors easily in their config.
+
+__Note: py3 colors constants require py3status 3.1 or higher__
+
+
+```
+# -*- coding: utf-8 -*-
+"""
+Example module that uses colors
```
ultrabug: please terminate the sentence with a dot.

doc/README.md
```diff
@@ -88,6 +90,40 @@ imap {
...
+
+time {
+    color = 'FF00FF'
+    format = "%H:%M"
+}
+
+battery_level {
+    color_good = '#00AA00'
+    color_bad = '#AA0000'
+    color_good = '#AAAA00'
```
ultrabug: color_good is defined twice ?

py3status/constants.py
```diff
@@ -25,6 +25,14 @@
 
 I3S_SINGLE_NAMES = ['cpu_usage', 'ddate', 'ipv6', 'load', 'time']
 
+I3S_ALLOWED_COLORS = ['color_bad', 'color_good', 'color_degraded']
+
+# i3status modules that allow colors to be passed.
+# general section also allows colors so included.
```
ultrabug: "so is included" ?

py3status/core.py
```diff
@@ -27,6 +27,16 @@
 
 DBUS_LEVELS = {'error': 'critical', 'warning': 'normal', 'info': 'low', }
 
+CONFIG_SPECIAL_SECTIONS = [
+    'order',
+    'general',
```
ultrabug: bad kitty !

py3status/i3status.py
```diff
@@ -12,7 +12,10 @@
 
 from py3status.profiling import profile
 from py3status.events import IOPoller
-from py3status.constants import TIME_MODULES, TZTIME_FORMAT, TIME_FORMAT
+from py3status.constants import (
+        TIME_MODULES, TZTIME_FORMAT, TIME_FORMAT,
+        I3S_ALLOWED_COLORS, I3S_COLOR_MODULES
```
ultrabug: bad kitty !

py3status/modules/keyboard_layout.py
```diff
@@ -44,17 +46,33 @@ def __init__(self):
             self._command = self._xkblayout
         except:
             self._command = self._xset
+        self.colors_dict = {}
+        # old default values for backwards compatability
+        self.defaults = {
+            'us': '#729FCF',
+            'fr': '#268BD2',
```
ultrabug: baaaaad kitty !

ultrabug:
Merges everywhere !! :palm_tree:

PR #281, opened 2016-04-15
[3.1] Do not die even if i3status has.

tobes:
This is just a nice bug fix when i3status chokes on say a bad config eg
```
general {
    color = "#FF00FF"
}
```

This can wait for 3.1


ultrabug:
Simple and useful, the way I like it :)

Issue #280, opened 2016-04-15
3.0 release


ultrabug:
So much as been done since 2.9 it's insane and impressive !

It's time to allow the other users to benefit from it so I'd like to call in a freeze of features and issue killing spree instead.

Let's list here the issues we want to be resolved for 3.0 and agree that the rest will come after ?

tobes:
From my point of view.

1) I still have one outstanding PR that I would really like to get into version 3.0 but it relies on #248 so therefore is not queued currently.  That PR would allow us to have much richer config files.
I know that you want to close the door on 3.0 but I feel that this PR would actually hugely benefit the 3.0 release and without it I feel it would be crippled.  Basically it offers the following

* containers - like groups but anyone can create them (and #275 creates the need for a 'frame' container)

* parameters can be `str`, `int`, `bool`, `None`, `list`, `dict`, `tuple`

* really good errors eg 'CONFIG ERROR: Value expected saw \`}\` at line 102 position 7'

* bad config files do not kill py3status (it has an emergency mode - complete with error info)

* stable - I've been regularly using it for the last month.

It is fairly self contained.

2) Documentation:  I've made some headway but some of it is dependent on existing PRs eg color -everywhere. (this could wait but it would be nice to have)

3) Pomodoro - wants fixing

4) Weather-yahoo - fixing or killing

5) Oh I've also got a pull request that allows testing of new style modules that use `self.py3`

Anyhow I think we should try to bed things down for a couple of weeks, but I think calling it 3.0 yet would be premature.  Especially in view of the composite patch where I'd like to do a secondary `button` patch on that.

tobes:
One other important thing in my mind is around `Py3` currently modules can use this to get 'legally' right inside of py3status.  I'd like to close the door on that, which would be a very minor PR and I'd consider it a bug fix.

pydsigner:
@tobes I don't think that's a real issue. Considering that a malicious module could just `import os; os.system("rm -rf ~/")`, weird py3status hackery seems like a minimal issue, and closing this down could potentially prevent some useful user-written modules.

pydsigner:
@tobes Other than that, better config parsing is something that I would personally be willing to hold the 3.0 release for.

tobes:
You see I believe this.

`self.py3.get_module_info()['module']` gets me the module and then I can crawl around and should be supported.  If I do `self.py3._module` it gives me the same but I know that what I do is not supported in any way.

That's all I mean just stopping user expectation.

[...]

PR #282, opened 2016-04-15
[3.1] Config parser

tobes:
Following on from the 3.0 issue #280.  This is the config parser.

It is a basic top down parser.

We break the config into a list of tokens and travel through them to build
up a dict that corresponds to the config.

The parser allows:

* containers
```
    container {
        included_module 1 {}
        included_module 2 {}
    }
```
* nesting of containers
```
    container 1 {
        container 2 {
            included_module 1 {}
        }
        included_module 2 {}
    }
```
* many types including lists, dict and tuple values
```
    my_module {
        my_str = 'hello'
        my_int = 23
        my_bool = true
        my_list = [1, 2, 3]
        my_dict = {'x': 1, 'y': 2}
        my_tuple = (1, 'something')
        my_complex = {
            'list' : [1, 2, 3],
            'dict' : {'x': 1, 'y': 2}
        }
    }
```

We can cope with horrible stuff all on one line which I do not condone
```
group tz { cycle = 0 open = false volume_status { button_up = 4 button_down = 5 button_mute = 0 } time {} tztime la { format = "LA %H:%M:%S" timezone = "America/Los_Angeles" } tztime ny { format = "NY %H:%M:%S" timezone = "America/New_York" } tztime du { format = "DU %H:%M:%S" timezone = "Asia/Dubai" } }
```

If an error is detected we:

* Inform the user about the error as best we can

```
general {{
        colors = false
}
CONFIG ERROR: Name expected saw `{` at line 1 position 9
```

Any module can declare itself as a container via our old friend class Meta
```
    class Meta:
        container = True
```

We try to cope with the config as well as we can.  If things are too bad we use a special error config so we can start and show the error in the status bar.

The process is two stage

1) parse the config into a general dict

2) parse that dict into a py3status config 

Hopefully having the two split helps.

To see the output as py3status gets it run `python parse_config.py <path-to-config>` in no config we try `~/.i3/i3status.conf`

I am aware of a few issues where the old parser accepted things eg
```
general {
        colors = false
}}
CONFIG ERROR: Name expected saw `}` at line 3 position 2
```
the old parser would be fine here.  As we get useful error that tell us the line/char and some basic issue, this should not be a major issue for users.

~~`i3status.py` changes include the  PR write_tmp_i3status_config() rewrite #248 which makes things look more complicated than they are.~~

Anyhow it's here and I suspect it will need time to be assessed.  I will try to add a few more comments/docstrings especially around the processing for py3status `process_config()`

Please try to break it.


ultrabug:
Holly cow, this is a tough one to read :) noted, amazing thanks !

PR #283, opened 2016-04-18
pomodoro update

tobes:
Here is the pomodoro update.  We now continue alert the user when the time is up even if the status bar is hidden.

As far as I can tell the module works identically to how it used to.  I have not tested the sound as I've not got pygame etc installed but I've not touched that part of the code.


ultrabug:
LGTM, tested with hide mode, :+1:

PR #287, opened 2016-04-24
Proposed vpn_status module

OfficerHalf:
The main reason I started using py3status was that I use multiple VPNs while at home (one for school, one for work) and it would help to know which is enabled, rather than i3status's unhelpful yes or no output. So, I wrote a small module that replaces the default vpn pidfile watch that i3status uses.

The only improvement is to display the active vpn name, which is retrieved with python-NetworkManager (which is just a wrapper for dbus-python). Otherwise, it's a drop-in replacement; there's even an option to act just like i3status and check if the given pidfile exists.

I tried my best to meet the docstring format. Let me know if I missed something. My only real concern is that I don't know how it acts without NetworkManager or python-NetworkManager installed. It's not a large or complicated module by any means, but I thought I'd share.

### DIFF
--- a/py3status/modules/vpn_status.py
+++ b/py3status/modules/vpn_status.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+"""
+Drop-in replacement for i3status run_watch VPN module.
+
+Expands on the i3status module by displaying the name of the connected vpn
+using pydbus. Asynchronously updates on dbus signals unless check_pid is True.
+
+Configuration parameters:
+    cache_timeout: How often to refresh in seconds when check_pid is True.
+        (default 10)
+    check_pid: If True, act just like the default i3status module.
+        (default False)
+    format: Format of the output.
+        (default 'VPN: {name}')
+    pidfile: Same as i3status.conf pidfile, checked when check_pid is True.
+        (default '/sys/class/net/vpn0/dev_id')
+
+Format string parameters:
+    {name} The name and/or status of the VPN.
+
+Requires:
+    pydbus: Which further requires PyGi. Check your distribution's repositories.
+
+@author Nathan Smith <nathan AT praisetopia.org>
+"""
+
+from pydbus import SystemBus
+from gi.repository import GObject
+from threading import Thread
+from os import path
+from time import time, sleep
+
+
+class Py3status:
+    # Available Configuration Parameters
+    cache_timeout = 10
+    check_pid = False
+    format = "VPN: {name}"
+    pidfile = '/sys/class/net/vpn0/dev_id'
+
+    def __init__(self):
+        self.thread_started = False
+        self.active = []
+
+    def _start_handler_thread(self):
+        """Called once to start the event handler thread."""
+        # Create handler thread
+        t = Thread(target=self._start_loop)
+        t.daemon = True
+
+        # Start handler thread
+        t.start()
+        self.thread_started = True
+
+    def _start_loop(self):
+        """Starts main event handler loop, run in handler thread t."""
+        # Create our main loop, get our bus, and add the signal handler
+        loop = GObject.MainLoop()
+        bus = SystemBus()
+        manager = bus.get(".NetworkManager")
+        manager.onPropertiesChanged = self._vpn_signal_handler
+
+        # Loop forever
+        loop.run()
+
+    def _vpn_signal_handler(self, args):
+        """Called on NetworkManager PropertiesChanged signal"""
+        # Args is a dictionary of changed properties
+        # We only care about changes in ActiveConnections
+        active = "ActiveConnections"
+        # Compare current ActiveConnections to last seen ActiveConnections
+        if active in args.keys() and sorted(self.active) != sorted(args[active]):
+            self.active = args[active]
+            self.py3.update()
+
+    def _get_vpn_status(self):
+        """Returns None if no VPN active, Id if active."""
+        # Sleep for a bit to let any changes in state finish
+        sleep(0.3)
+        # Check if any active connections are a VPN
+        bus = SystemBus()
+        for name in self.active:
+            conn = bus.get(".NetworkManager", name)
+            if conn.Vpn:
+                return conn.Id
+        # No active VPN
+        return None
+
+    def _check_pid(self):
+        """Returns True if pidfile exists, False otherwise."""
+        return path.isfile(self.pidfile)
+
+    # Method run by py3status
+    def return_status(self, i3s_outputs, i3s_config):
+        """Returns response dict"""
+        # Start signal handler thread if it should be running
+        if not self.check_pid and not self.thread_started:
+            self._start_handler_thread()
+
+        # Set 'no', color_bad as default output. Replaced if VPN active.
+        name = "no"
+        color = i3s_config["color_bad"]
+
+        # If we are acting like the default i3status module
+        if self.check_pid:
+            if self._check_pid():
+                name = "yes"
+                color = i3s_config["color_good"]
+
+        # Otherwise, find the VPN name, if it is active
+        else:
+            vpn = self._get_vpn_status()
+            if vpn:
+                name = vpn
+                color = i3s_config["color_good"]
+
+        # Format and create the response dict
+        full_text = self.format.format(name=name)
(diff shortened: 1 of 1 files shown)


tobes:
> I tried my best to meet the docstring format. 

You did good.  If you could put the config params in alphabetical order it'd be even better.

>  I don't know how it acts without NetworkManager or python-NetworkManager installed. 

Currently it'll break on the `import NetworkManager` line, you could wrap that in a try and fall back to the pidfile check.



Have you considered having it as an async style module?  dbus should allow us to receive events when the vpn starts something sort of like

```
import dbus

def __init__(self):
    system_bus = dbus.SystemBus()  # session bus?
    system_bus.add_signal_receiver(_vpn_connection_handler,
        dbus_interface="org.freedesktop.NetworkManager.VPN.Connection",
        signal_name="PropertiesChanged")

def _vpn_connection_handler(self):
    self.py3.update()  # this will force the module to update

def return_status(self, i3s_outputs, i3s_config):
    .... stuff ....
    response = {
        'full_text': full_text,
        'color': color,
    }
    return response
```

This is only possible using the current master not on 2.9 but would mean that we'd be notified straight away without constantly checking. 

That wouldn't work if no network manager.  but you could do a fallback to pidfile checking and the timeout.

It might be possible to add other clients like wicd if it ever adds vpn support (planned but no expected date)

OfficerHalf:
Hadn't considered it, but that would probably be a more efficient solution. I'll look into it.

OfficerHalf:
Okay, I got it working... Kind of. Take a look at my [vpn_status_async branch](https://github.com/OfficerHalf/py3status/blob/vpn_status_async/py3status/modules/vpn_status.py). The module runs, and the _vpn_signal_handler method is called as it should be (right now it's just printing the name of the newly connected VPN). However to get that working, I had to add to the main call at the bottom to start the loop so that it would actually listen for the signal.

That main loop is needed, as far as I can tell (I've never touched dbus before today, so I might be missing something), but I'm not sure where or how I should call it and still have the module work properly with py3status as a whole, since loop.run() doesn't return.

tobes:
> Okay, I got it working... Kind of. 

cool :)

>  Would it need its own thread?

yes almost certainly.  It might be good to look at the window_title_async module as an example
https://github.com/ultrabug/py3status/blob/master/py3status/modules/window_title_async

tobes:
It makes it hard if you completely edit your comments ;p

tobes:
> I added a lame try/except block around it that just throws the exception away and it seems to work. 

Exceptions are great it's nice to hit them and eventually work out why they happen and then just catch the minimum needed (when you understand the cause)

OfficerHalf:
Heh, yeah sorry about that. I was speaking before thinking, and eventually my brain caught up, so I edited it. I'll track down that exception and see why NetworkManager is breaking.

py3status/modules/vpn_status.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _start_thread(self):
+        self.loop = GLib.MainLoop()
+        self.bus = dbus.SystemBus(self.loop)
+        self.bus.add_signal_receiver(self._vpn_signal_handler,
+                                     dbus_interface="org.freedesktop.NetworkManager.VPN.Connection",
+                                     signal_name="PropertiesChanged")
+        # Start handler thread
+        self.t.start()
+        self.thread_started = True
```
tobes: To me it feels like you are trying to define the thread and start it in the same function.  If you have the 'thread' define the loop and start it separate from the `self.t.start()` stuff it might help.

py3status/modules/vpn_status.py
```diff
@@ -0,0 +1,140 @@
...
+    {name} The name and/or status of the VPN.
+
+Requires:
+    python-NetworkManager
+
+@author Nathan Smith <nathan AT praisetopia.org>
+"""
+
+try:
+    import NetworkManager
```
tobes: do we need `NetworkManager`?  Can we get the info just from dbus?

 ` dbus-monitor --system`  start/stop vpn

OfficerHalf: No, it's not really needed, mostly just easier on the eyes. Here, I wrote:
```
for conn in NetworkManager.NetworkManager.ActiveConnections:
    # do stuff
```
The same code with pure dbus is something like:

```
bus = dbus.SystemBus()
m_proxy = bus.get_object("org.freedesktop.NetworkManager", "/org/freedesktop/NetworkManager")
manager = dbus.Interface(m_proxy, "org.freedesktop.NetworkManager")
mgr_props = dbus.Interface(m_proxy, "org.freedesktop.DBus.Properties")

s_proxy = bus.get_object("org.freedesktop.NetworkManager", "/org/freedesktop/NetworkManager/Settings")
settings = dbus.Interface(s_proxy, "org.freedesktop.NetworkManager.Settings")

active = mgr_props.Get("org.freedesktop.NetworkManager", "ActiveConnections")
for a in active:
    a_proxy = bus.get_object("org.freedesktop.NetworkManager", a)
    a_props = dbus.Interface(a_proxy, "org.freedesktop.DBus.Properties")
    # do stuff
```

py3status/modules/vpn_status.py
OfficerHalf: I see what you mean. Can do.

OfficerHalf:
Removed the python-NetworkManager dependency. There is a syncing issue where calling `active = properties.Get("org.freedesktop.NetworkManager", "ActiveConnections")` fails when disconnecting a VPN. That call ends up looking for a connection that no longer exists and throws a DBusException. I bypassed this by calling `sleep(0.3)` to let the change propagate and then get all active connections - might not be the best solution, but it seems to work. I'm not sure how to go about fixing it properly.

tobes:
Firstly, well done it's looking good.

Secondly, looking at https://wiki.python.org/moin/DbusExamples it seems that python-dbus is viewed as a legacy api.  pydbus looks like the preferred option and would be the simplest change but also relies on gi.repository which I don't like (not on pypi etc).  txdbus doesn't have that dependency but looks a little horrible.

Other than that it is minor tweaks like maybe defining `"org.freedesktop.NetworkManager"` etc as constants

 `sleep(0.3)` since in `_vpn_signal_handler()` we know we have disconnected `state == 4` do we even need to check the connections?  hmm, would that work if I have multiple vpns? Maybe sleeping is the best approach?

OfficerHalf:
I checked out pydbus, and the code would look a little nicer using that package. I can leave it as is and if I'm still here if/when PyGi is on pypi, update it. Or I can see about switching it now (or once I finish up with finals and graduate, anyway). I'm indifferent.

I've been watching and haven't had any issues with it the last few days of normal use, no excess battery drain or anything like that.

tobes:
> Or I can see about switching it now

I think that would be good.

All the best with your finals

OfficerHalf:
I need another pair of eyes on this. pydbus makes things much easier, and in fact, I managed to eliminate

OfficerHalf:
Updated to use pydbus.

py3status/modules/vpn_status.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _vpn_signal_handler(self, args):
+        """Called on NetworkManager PropertiesChanged signal"""
+        # Args is a dictionary of changed properties
+        # We only care about changes in ActiveConnections
+        active = "ActiveConnections"
+        # Compare current ActiveConnections to last seen ActiveConnections
+        if active in args.keys() and sorted(self.active) != sorted(args[active]):
+            self.py3.update()
+            self.active = args[active]
```
tobes: set `self.active` before calling `update()` since you want active set when you do the update

tobes:
Looking good.  That is so much more readable now :)

pydsigner:
Looks good, you might want to squash the commits though.

ultrabug:
Impressive contribution @OfficerHalf and reviews as usual @tobes and @pydsigner thank you !

@OfficerHalf, are you ok to take care of squashing the commits plz ?

ultrabug:
Thanks a bunch @OfficerHalf for your contribution and patience ! and ofc @pydsigner and @tobes for this great review.

PR #291, opened 2016-04-26
flake8 line length max 79

tobes:
This PR fixes line lengths to the PEP8 recommended max length of 79.

I have tried to do it in a way that minimizes the changes and maintains readability (something that yapf often fails at imho)

There are a few places where some refactoring was needed ( `xrandr_rotate.py`) but again I've tried to avoid this where possible.

### DIFF
--- a/py3status/docstrings.py
+++ b/py3status/docstrings.py
@@ -241,7 +241,8 @@ def update_docstrings():
                 out.append(row)
                 if not replaced:
                     out = out + [
-                        ''.join(_to_docstring(modules_dict[mod])).strip() + '\n'
+                        ''.join(_to_docstring(modules_dict[mod])).strip() +
+                        '\n'
                     ]
                     replaced = True
                 if lines:
@@ -285,7 +286,8 @@ def check_docstrings(show_diff=False, config=None):
                 create_readme(readme).split('\n'), create_readme(
                     modules_readme).split('\n'))))
         else:
-            print_stderr('\nUse `py3status docstring check diff` to view diff.')
+            print_stderr()
+            print_stderr('Use `py3status docstring check diff` to view diff.')
 
 
 def update_readme_for_modules(modules):
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -340,7 +340,9 @@ class I3status(Thread):
                     if group_name:
                         # update the items in the group
                         config[group_name]['items'].append(section_name)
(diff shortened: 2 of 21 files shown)


ultrabug:
I have mixed feeling about this, I'd like a third party to take part of the vote : @pydsigner ?

pydsigner:
@ultrabug 160 character lines are definitely too long but I'd rather not throw down a 79 line restriction.

tobes:
@pydsigner thanks for your response. What limit would you prefer?

@ultrabug why do you have mixed feelings?

Rather than just saying PEP8 my take would be this.  Restrictions are not always bad.  When I first started writing python I'd have had the same view, all this indentation making my lines head off to the right wtf!  But now I'd say that the whole 79 char limit has actually made me a much better coder. It makes me create much tighter code. It makes debugging stuff easier.  My code is more encapsulated.  So regardless of all the small screen (reading code on my phone) or in emails etc, which is a valid issue for me, overall it is about better code, not just to read, but mainly about helping others learn through the restrictions.

I'm an old fuck, and things change, and in many ways I don't care, but I'd not want others to not have the opportunities to improve that I've had.

Anyhow, I see two issues here.

1) are the code changes themselves acceptable?

2) can we knock the line length limit down to something more reasonable? 

A lot of this ends up in religious wars, vi/emacs or should the kernel be written in C++.  I don't claim to have any of the answers, but I'd like to think I have a small amount of insight. I'd be interested in what you think on line length and why.

pydsigner:
For my personal coding, my *goal* is 80, but I'll allow a line to go to 100 if it makes things more legible. So it isn't that I don't like 79 character lines, it's that I don't want our QC to be broken by lines that *are* better a little longer. Whatever you put in flake8 is going to be an unbudgeable standard and that standard should be less strict.

tobes:
>  lines that are better a little longer

If we really need a long line then we can always use `# noqa`.

Anyhow 100 would be a good start

ultrabug:
I must say that at work we enforce a strong PEP8 / 80 chars policy which I'm fine with but I agree that for an OSS such as ours here, a bit of flexibility should prevail.

100 is it, okay and thanks !

tobes:
> PEP8 / 80 chars 

PEP8 is 79

> 100 is it, okay and thanks !

99?  I've amended `tox.ini` if you really want 100 I'll update it again.

ultrabug:
99 is good aye, cheers !

PR #292, opened 2016-04-30
Added Insync status monitor module

Cypher1:


py3status/modules/insync.py
```diff
@@ -0,0 +1,75 @@
...
+import subprocess
+import sys
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 1
+    format = ''
+    _format_default = '{status}'
+    _line_separator = "\\n" if sys.version_info > (3, 0) else "\n"
```
tobes: This is never used.

py3status/modules/insync.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if format == '':
+        format = _format_default
```
tobes: `format =  '{status}'` is all you need `_format_default` is just creating noise

py3status/modules/insync.py
```diff
@@ -0,0 +1,75 @@
...
+        results = self.format.format(status=str(status))
+
+        response = {
+            'cached_until': time() + self.cache_timeout,
+            'full_text': results,
+            'color' : color
+        }
+        return response
+
+    def _check_insync_status(self, config):
```
tobes: Having this function just makes this module harder to follow without adding any real benefit.  You'd be better just doing everything in `check_insync()`

py3status/modules/insync.py
```diff
@@ -0,0 +1,75 @@
...
+Gets the current insync status
+
+Configuration parameters:
+    format: Display format to use *(default
+        '{status}'
+        )*
+
+Format status string parameters:
+    {status} Status of Insync
+    {files} Number of pending syncing files
```
tobes: `{files}` is not available

py3status/modules/insync.py
Cypher1: @tobes Do you mean insync is never used? I know multiple people who use it.

py3status/modules/insync.py
Cypher1: Thanks, good pick up

py3status/modules/insync.py
tobes: I mean `_line_separator`

Cypher1: Removed in 19526160e6c032f365ccda1b8308231137414233

ultrabug:
Thanks a lot @Cypher1 !

@tobes are we okay for a merge in your opinion ?

py3status/modules/insync.py
```diff
@@ -0,0 +1,60 @@
...
+        status = str(subprocess.check_output(["/usr/bin/insync", "get_status"]))
+        if len(status) > 5:
+            status = status[2:-3]
+        color = i3s_config.get('color_degraded', '')
+        if status == "OFFLINE":
+            color = i3s_config.get('color_bad', '')
+        if status == "SHARE":
+            color = i3s_config.get('color_good', '')
+            status = "INSYNC"
+        results = self.format.format(status=str(status))
```
tobes: status is already a `str` and we don't need to cast it anyhow we can just use.

`results = self.format.format(status=status)`

py3status/modules/insync.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+
+"""
+Gets the current insync status
+
+Configuration parameters:
+    format: Display format to use *(default
+        '{status}'
+        )*
```
tobes:
```
    format: Display format to use (default '{status}')
```
is  better

py3status/modules/insync.py
```diff
@@ -0,0 +1,60 @@
...
+import subprocess
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 1
+    format = '{status}'
+
+    def check_insync(self, i3s_output_list, i3s_config):
+        status = str(subprocess.check_output(["/usr/bin/insync", "get_status"]))
```
tobes: `/usr/bin/insync` is this always located here?  would not just `insync` work and be more portable?

py3status/modules/insync.py
```diff
@@ -0,0 +1,60 @@
...
+Configuration parameters:
+    format: Display format to use *(default
+        '{status}'
+        )*
+
+Format status string parameters:
+    {status} Status of Insync
+
+@author Joshua Pratt <jp10010101010000@gmail.com>
+Thanks to @author Iain Tatch <iain.tatch@gmail.com> for the script that this is based on
```
tobes: This should be either a) removed or b) reworded and moved into the main body of the docstring.

tobes:
> are we okay for a merge 

I've added a few comments for some changes that would be nice.

I'd also like to see some more info in the docstring about what insync is etc but generally this is good to me.

ultrabug:
Thank you all for this productive collaboration and @Cypher1 for this new module and your patience !

Issue #293, opened 2016-05-09
Event propagation


tobes:
@ultrabug following the merge of #275 you added

Also, the more I think of it the more I'm inclined to just honour per module on_click and not on the main group module.

This would be OK:
```
group tz {
    cycle = 10
    format = "{output}"

    tztime la {
        format = "LA %H:%M"
        timezone = "America/Los_Angeles"
        on_click 1 = "exec firefox-bin https://weather.com/weather/tenday/l/Los+Angeles+CA+USCA0638:1:US"
    }

    tztime ny {
        format = "NY %H:%M"
        timezone = "America/New_York"
    }

    tztime du {
        format = "DU %H:%M"
        timezone = "Asia/Dubai"
    }
}
```
This would not:
```
group tz {
    cycle = 10
    format = "{output}"
    on_click 1 = "exec my_random_command.sh"

    tztime la {
        format = "LA %H:%M"
        timezone = "America/Los_Angeles"
        on_click 1 = "exec firefox-bin https://weather.com/weather/tenday/l/Los+Angeles+CA+USCA0638:1:US"
    }

    tztime ny {
        format = "NY %H:%M"
        timezone = "America/New_York"
    }

    tztime du {
        format = "DU %H:%M"
        timezone = "Asia/Dubai"
    }
}
```

I have read this as if a module captures a click event using `on_click X` then we would not pass the event on to the container.  However if we did not use the click then we would pass on the event.  Is this correct?

I think this is a good idea and it would be easy to implement.

Personally I would like to go further.  We have the following situations.

* A module has an `on_click X` in the config that is used and its own `on_click()` would we pass the event to the `on_click()` on in this situation?

* A module in a container has an `on_click()` currently we don't know if this is consumed.  I would like to change the `on_click` so that it can return a `True`/`False` to signify if it has used the event and then we could pass unused events to the container.  Current modules would return `None` and so we could see if the the module 'understands' the new way of treating clicks.

ultrabug:
To maybe be more clear : the only case where i3status.conf on_click is valid is on a module, not a container.

> A module has an on_click X in the config that is used and its own on_click() would we pass the event to the on_click() on in this situation?

Yes, i3status.conf on_click adds up to any module on_click() function. Both will be executed.

[...]

PR #294, opened 2016-05-09
limit on_click config option for containers

tobes:
This patch stops containers from using `on_click X` config events re #293 

I think this does what we want.


ultrabug:
Cheers, perfect !

PR #295, opened 2016-05-09
py3 improvements

tobes:
I updated the module -> module_info


ultrabug:
easier indeed, thanks !

Issue #204, opened 2016-03-06
Load modules from XDG_CONFIG_HOME by default


tobanw:
Currently, `py3status` modules are [loaded by default](https://github.com/ultrabug/py3status/blob/master/py3status/core.py#L51) from `~/.i3/py3status/` only. I suggest adding `~/.config/i3status/py3status/` (and possibly also `~/.config/i3/py3status/`) to the `include_paths` list. I'm currently specifying the path manually with the `-i` flag.

The `i3status.conf` is already [loaded from the XDG config dirs](https://github.com/ultrabug/py3status/blob/master/py3status/core.py#L65), so this improves consistency. Also, using `~/.config/` is common: for example, the [Arch wiki presumes](https://wiki.archlinux.org/index.php/i3#Configuration) it for i3 config files.

P.S. It seems to me that [line 67-68](https://github.com/ultrabug/py3status/blob/master/py3status/core.py#L67) has an error, since [XDG_CONFIG_HOME expands to](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support#User_directories) `~/.config/`. In my case, this variable isn't set, so the `os.environ.get` call returns `home_path`. But if it were set, the resulting path would be `~/.config/.config/i3status/config`.

ultrabug:
I'm not forgetting your issue @tobanw and I think you got a good point.

It's just that for now there are some major PR open that need attention, please bear with me.

ultrabug:
fixed by @tobes on #297 thanks !

PR #297, opened 2016-05-10
fix XDG_CONFIG_HOME #204

tobes:
This fixes the XDG_CONFIG_HOME issue.

I'm not sure what other paths if any we want to add from that issue #204 

It looks like we cover all the i3status default locations https://i3wm.org/i3status/manpage.html#_options

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -55,11 +55,19 @@ class Py3statusWrapper():
         # defaults
         config = {
             'cache_timeout': 60,
-            'include_paths': ['{}/.i3/py3status/'.format(home_path)],
             'interval': 1,
             'minimum_interval': 0.1,  # minimum module update interval
             'dbus_notify': False,
         }
+
+        # include path to search for user modules
+        config['include_paths'] = [
+            '{}/.i3/py3status/'.format(home_path),
+            '{}/i3status/py3status'.format(os.environ.get(
+                'XDG_CONFIG_HOME', '{}/.config'.format(home_path))),
+            '{}/i3/py3status'.format(os.environ.get(
+                'XDG_CONFIG_HOME', '{}/.config'.format(home_path))),
+        ]
 
         # package version
         try:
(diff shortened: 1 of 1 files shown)


ultrabug:
I think @tobanw was referring to the inclusion folder for user modules, not the i3status.conf location detection itself (tho you're correctly fixing this too in this PR)

I guess he has a point there and we could extend these path to the i3 user modules inclusion directory detection ?

tobes:
I can add extra module search dirs to this PR.

tobes:
Currently it looks like we only allow one user directory `-i` overrides it.  This change would mean that we would now look in multiple directories.  So we need to worry about an order of dirs to look in (easy enough to change)

Are we still going to replace the user modules directory/directories if `-i` is used? 

For me it makes most sense to look in 
User supplied via `-i` > `~/.i3/py3status` > `~/.config/i3status/py3status/` >  `~/.config/i3/py3status/`

using the XDG_CONFIG_HOME if we have it set

although `~/.config/i3status/py3status/` feels wrong to me

tobanw:
Yup that's right, it's this line: https://github.com/ultrabug/py3status/blob/master/py3status/core.py#L58

pydsigner:
@tobes I think we should probably have an additional flag to tell py3status not to look in the standard directories.

tobes:
I've updated the user module directories that are now checked in the following order:

`~/.i3/py3status` > `~/.config/i3status/py3status/` > `~/.config/i3/py3status/`

Assuming `XDG_CONFIG_HOME`  is not defined or `~/.config`

I removed the sorted as we should look in the directories in a user set order if provided by `-i` as that is what the user would expect.  Also this means that we look in our predefined directories in a set order.

We no longer overwrite user modules once they have been found. 

if `-i` is supplied we only look in those directories as that was the current behaviour, but I'd prefer to just add those to the front of the list, but that is a behaviour change.

ultrabug:
I think this is good as is, `-i` users just point where to look themselves and the others use autodetection.

Thanks !

PR #298, opened 2016-05-11
move -b runtime option alphabetically and add to README.rst

tobes:
This should be in the main readme

* add -b, --dbus-notify option to README.rst

* move option so they are alphabetical for `py3status -h`


ultrabug:
good catch, thanks !

PR #299, opened 2016-05-11
Updates to timer module allowing sound playback via py3

tobes:
The timer module was written as a composite demo, this PR:

* Plays a sound when the countdown completes if configured by user.

* Adds `time` config parameter

* Uses a Timer so i3bar can be hidden without affecting the operation. 

* Adds time module to the README

* Sound playback is via Py3 so available to other modules (uses `paplay` or `play` if available)

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -1,5 +1,9 @@
+import os
 import re
+import shlex
 from time import time
+from subprocess import Popen, call
+
 
 PY3_CACHE_FOREVER = -1
 
@@ -20,6 +24,7 @@ class Py3:
     def __init__(self, module):
         self._module = module
         self._output_modules = module._py3_wrapper.output_modules
+        self._audio = None
 
     def _get_module_info(self, module_name):
         """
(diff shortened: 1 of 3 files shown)


py3status/py3.py
```diff
@@ -20,6 +24,7 @@ class Py3:
     def __init__(self, module):
         self._module = module
         self._output_modules = module._py3_wrapper.output_modules
+        self._audio = None
```
ultrabug: alphabet mon amour !

py3status/py3.py
```diff
@@ -93,3 +98,31 @@ def replace_fn(match):
...
+
+    def play_sound(self, sound_file):
+        """
+        Plays sound_file if possible.
+        """
+        self.stop_sound()
+        cmd = self.check_commands(['paplay', 'play'])
+        if cmd:
+            c = shlex.split('{} {}'.format(cmd, sound_file))
+            self._audio = Popen(c, shell=True)
```
ultrabug: don't you need to catch stderr and stdout pipes here in case the command outputs something ?

py3status/py3.py
```diff
@@ -93,3 +98,31 @@ def replace_fn(match):
...
+
+    def stop_sound(self):
+        """
+        Stops any currently playing sounds for this module.
+        """
+        if self._audio:
+            self._audio.kill()
```
ultrabug: what when it's None ? does it always have a kill() method really ?

tobes: It is initialized as `None` otherwise it is the result of `Popen()` which does have `kill()`  I should probably set it to `None` following the `kill()` call.

py3status/py3.py
tobes: For these specific commands I think they actually exit 0 even if the file doesn't exist etc but I'll fix this anyhow in case we add other sound players later.

tobes:
I've updated things a little.

With the Popen issue adding stdout/stderr didn't really add anything as currently it seems to work fine for me regardless of output etc.  for some reason the shlex.split() was causing issues which I wish I understood better.

py3status/py3.py
```diff
@@ -93,3 +98,33 @@ def replace_fn(match):
...
+    def play_sound(self, sound_file):
+        """
+        Plays sound_file if possible.
+        """
+        self.stop_sound()
+        cmd = self.check_commands(['paplay', 'play'])
+        if cmd:
+            sound_file = os.path.expanduser(sound_file)
+            c = shlex.split('{} {}'.format(cmd, sound_file))
+            self._audio = Popen(c)
```
ultrabug: unless I'm mistaken, this still doesn't catch any output the command may throw. It is my understanding that you **need** to catch them using PIPES.

Example : https://github.com/ultrabug/py3status/blob/master/py3status/modules/xrandr.py#L111

Am I wrong ?

tobes: I've tested with bad files etc and all seems to work fine.  I'm not really interested in the output so no need to trap it.  Catching the errors might be worth it but I'm not sure it adds much.

ultrabug:
Thanks for the extended testing !

PR #300, opened 2016-05-11
TRIVIAL: remove repeated line in module.py

tobes:
You can't see it in the diff but this line is repeated just before the diff starts

```
        self.module_full_name = module      <---- HERE
        self.module_inst = ''.join(module.split(' ')[1:])
        self.module_name = module.split(' ')[0]
        self.new_update = False
        self.module_full_name = module      <---- HERE
```


ultrabug:
:+1:

PR #301, opened 2016-05-15
modules: sysdata: substract cached mem from the used mem

pmamonov:
The `used memory` number reported by `free -m` is misleading, since it includes the memory used for disk cache. An OS kernel tries to allocate all spare memory for disk cache and readily gives it back to processes. The difference `(used_mem - cached_mem)` seems to be a more convinient measure of RAM usage.

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -81,6 +81,14 @@ class GetData:
         total_mem = int(mem_data[mem_index + 1]) / 1024.
         used_mem = int(mem_data[mem_index + 2]) / 1024.
 
+        # Substract cached memory from the used memory
+        try:
+            cached_index = mem_data.index('cached')
+            cached_mem = int(mem_data[mem_index + 1 + cached_index]) / 1024.
+            used_mem -= cached_mem
+        except:
+            pass
+
         # Caculate percentage
         used_mem_percent = int(used_mem / (total_mem / 100))
 


tobes:
> The used memory number reported by free -m is misleading

Are you sure?   `man free` for me shows

```
..the displayed columns are:

       total  Total installed memory (MemTotal and SwapTotal in /proc/meminfo)

       used   Used memory (calculated as total - free - buffers - cache)
```

so `used` should be correct.  Maybe we have different versions

```
$ free -V
free from procps-ng 3.3.10
```

pydsigner:
@tobes Same for me, I have the same version though.

pmamonov:
I run stable Debian "jessie"  which has an older version of `free` [https://packages.debian.org/jessie/procps]:
```
$ free -V
free from procps-ng 3.3.9
```

Here is it's output:
```
$ free -m
             total       used       free     shared    buffers     cached
Mem:          4045       3761        284        145        220       1991
-/+ buffers/cache:       1548       2497
Swap:         7632         74       7557
```

Clearly `used mem` number includes cached and buffers memory.
The number of interest (used - buffers - cache) is printed on the next line (`-/+ buffers/cache:`).
However, I guess there are some other versions/implementations of `free` which lacks that line.
Probably the better solution is to read /proc/meminfo rather than struggle with the zoo of `free-s`

tobes:
It seems that the output changed in December 2014 although it is not actually mentioned in the changelog.  So thanks for reporting.

> Probably the better solution is to read /proc/meminfo

I think I'd agree.  

We could also get some other info like swap whilst we are at it (but that could be done later)

ultrabug:
@pmamonov you closed because this is not an issue as-is and you don't seek anything more from it ?

tobes:
I think more that this solution isn't the one we are looking for.

pmamonov:
@ultrabug I've closed the merge request, because there is a better solution to this problem, i.e. parse `/proc/meminfo` contents instead of `free` output.

ultrabug:
Ok mates, that's fine with me, thanks !

PR #304, opened 2016-05-18
Twitch updates

tobes:
Updates:

* move to correct location and rename (something odd happened here)

* remove `grossie_gore` default as channel is banned

* add error if `stream_name` missing

* docstring cleanups

* minor code cleanups


ultrabug:
good catches, thanks !

PR #305, opened 2016-05-18
yandex updates

tobes:
docstring plus removed unneeded `str` cast


ultrabug:
so quick :speedboat: !

PR #307, opened 2016-05-18
modules: sysdata: get memory usage from /proc/meminfo

pmamonov:
The output of the `free` command varies across different
versions/implementations. So parse the contents of
/proc/meminfo instead.

Signed-off-by: Peter Mamonov <pmamonov@gmail.com>

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -71,21 +71,29 @@ class GetData:
 
     def memory(self):
         """
-        Execute 'free -m' command, grab the memory capacity and used size
+        Parse /proc/meminfo, grab the memory capacity and used size
         then return; Memory size 'total_mem', Used_mem, and percentage
         of used memory.
         """
-        # Run 'free -m' command and make a list from output.
-        mem_data = self.execCMD('free', '-m').split()
-        mem_index = mem_data.index('Mem:')
-        total_mem = int(mem_data[mem_index + 1]) / 1024.
-        used_mem = int(mem_data[mem_index + 2]) / 1024.
 
-        # Caculate percentage
-        used_mem_percent = int(used_mem / (total_mem / 100))
+        memi = {}
+        with open('/proc/meminfo', 'r') as fd:
+            for s in fd:
+                tok = s.split()
+                memi[tok[0]] = float(tok[1]) / (1 << 20)
 
-        # Results are in kilobyte.
-        return total_mem, used_mem, used_mem_percent
+        try:
+            total_mem = memi["MemTotal:"]
+            used_mem = (total_mem -
+                        memi["MemFree:"] -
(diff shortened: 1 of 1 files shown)


py3status/modules/sysdata.py
```diff
@@ -71,20 +71,30 @@ def cpu(self):
...
-        mem_data = self.execCMD('free', '-m').split()
-        mem_index = mem_data.index('Mem:')
-        total_mem = int(mem_data[mem_index + 1]) / 1024.
-        used_mem = int(mem_data[mem_index + 2]) / 1024.
+
+        memi = {}
+        with open('/proc/meminfo', 'r') as fd:
+            for s in fd:
+                tok = s.split()
+                memi[tok[0]] = float(tok[1]) / (1 << 20)
```
pydsigner: I think it makes sense to leave the values as kB instead of changing the behavior of the module.

pmamonov: I did not change the units of the return values, I've just fixed the comment.

tobes: Haha =D

py3status/modules/sysdata.py
```diff
@@ -71,20 +71,30 @@ def cpu(self):
...
+
+        try:
+            total_mem = memi["MemTotal:"]
+            used_mem = (total_mem -
+                        memi["MemFree:"] -
+                        memi["Buffers:"] -
+                        memi["Cached:"])
+        except:
+            total_mem, used_mem = -1, -1
```
tobes: If we don't know the values then setting them to `?` rather than `-1` is more usual for modules.

py3status/modules/sysdata.py
```diff
@@ -71,20 +71,30 @@ def cpu(self):
...
+                tok = s.split()
+                memi[tok[0]] = float(tok[1]) / (1 << 20)
+
+        try:
+            total_mem = memi["MemTotal:"]
+            used_mem = (total_mem -
+                        memi["MemFree:"] -
+                        memi["Buffers:"] -
+                        memi["Cached:"])
+        except:
```
tobes: Raw exception are nasty it looks like `KeyError` would be the issue if there is one.

py3status/modules/sysdata.py
pmamonov: `sysData()` expects `memory()` to return numeric values.

 probably `nan` is more appropriate than `-1`?

py3status/modules/sysdata.py
pmamonov: I've moved percentage calculation into `try` block, so more than one exception may rise here. Is it really necessary to list all possible exceptions here?

ultrabug:
Thanks @pmamonov and reviewers :+1:

Issue #302, opened 2016-05-18
Crash on screenlock


pigmonkey:
Using the latest git version, py3status seems to be unable to recover from a screen lock. The bar runs fine, but if I lock my screen with i3lock and then unlock, the bar is replaced by the i3 error `Error: status_command process exited unexpectedly (exit 1)`. Oddly I can still see the py3status process running via `ps`.

This does not happen in 2.9.

pydsigner:
I've had this exact same issue on master for a while now.

ultrabug:
I lock my screen omgwhat times a day and didn't encounter this problem... BUT I get the same problem on session login !

I will investigate this further, this is a blocker on the release clearly.

Thanks

tobes:
:( I can't reproduce this via i3lock or by starting a session.  My though is it could be related to #253 as signaling is probably involved here.

could someone see if the following commits are good/bad
```
# commit before signal changes
 git checkout ad8091d8b621d36
```

```
# signal changes
 git checkout 09ec0aa355239a
```

for what it's worth:
I'm on fedora 23, i3 v4.12, i3lock v2.7

ultrabug:
For my crash, I think it's due to the xrandr module command returning badly on login :
```
May 18 17:51:14 yazd __init__.py: command "xrandr --output eDP1 --auto --pos 0x0 --rotate normal --output DP1 --auto --right-of eDP1 --rotate normal --output DP2 --off --output HDMI1 --off --output HDMI2 --off --output VIRTUAL1 --off" exit code -12
May 18 17:51:14 yazd __init__.py: Runtime error (TypeError) module.py line 114.
May 18 17:51:14 yazd __init__.py: Runtime error (TypeError) module.py line 114. Please try to fix this and reload i3wm (Mod+Shift+R)
May 18 17:51:14 yazd __init__.py: lock cleared, exiting
```

My bar is then not updated and a few seconds later I get the status_command error displayed.

tobes:
Thanks I think I'll be able to trigger it now. Shouldn't be too hard to fix.

tobes:
@pigmonkey @pydsigner can you see if this is now fixed for you in master

pigmonkey:
Negative. Same error after screenlock.

pydsigner:
Not working here either.

pydsigner:
Interestingly, turning `--debug` on fixed the issue and turning it off brought it back.

pydsigner:
[...]

PR #308, opened 2016-05-19
Do not set timer if no cache_time

tobes:
With #302 the problem appears to be when we wake a module that does not have its cache_time set 

This PR checks for a cache_time value.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -105,12 +105,13 @@ class Module(Thread):
 
     def wake(self):
         self.sleeping = False
-        cache_time = self.cache_time
+        if self.cache_time is None:
+            return
         # new style modules can signal they want to cache forever
-        if cache_time == PY3_CACHE_FOREVER:
+        if self.cache_time == PY3_CACHE_FOREVER:
             return
         # restart
-        delay = max(cache_time - time(), 0)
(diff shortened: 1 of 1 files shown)


py3status/module.py
```diff
@@ -106,6 +106,8 @@ def sleep(self):
     def wake(self):
         self.sleeping = False
         cache_time = self.cache_time
```
ultrabug: stupid question, but why assign this to a new variable when we could use self.cache_time wherever needed in this function ?

tobes: habit I suppose (lookup costs - not relevant here).  do you want it changed

tobes:
As an aside the following modules do not have `cached_until` they probably should learn about it.

* dpms

* ns_checker

* glpi

py3status/module.py
ultrabug: I guess we should plz

ultrabug:
Also, I confirm this fixes my xrandr login crash problem (but I guess I need to fix smthing in the module too)

ultrabug:
Thanks !

tobes:
I think the issue here doesn't mean the xrandr module is broken

ultrabug:
no, but I can say it is broken with -12 error codes I don't understand :(

PR #312, opened 2016-05-19
Py3 testing framework

tobes:
Here is the new module_test to allow for Py3.

I have put it in it's own file for clarity.  I also dropped the MockPy3 from my first attempt as it seems better if the actual Py3 is used.

I've updated all the modules using `py3`.  I also updated `fedora_updates.py` as a demo that it will work for any module.

`Py3` now also has a `i3s_config()` to get the i3s_config.

The module tester can be given a simple config which I've used in a few places eg `static_string`

I also corrected a python2 unicode issue in `scratchpad_async.py`

`vpn_status.py` is a slight pain as the keyboard interrupt doesn't work as it gets caught in the gtk loop rather than the module_test (still for now it is good enough and I've tried too hard to sort it for one day)

A few random other changes as I was in the modules eg `self.py3.time_in()`

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -286,9 +286,7 @@ class Module(Thread):
 
             # Add the py3 module helper if modules self.py3 is not defined
             if not hasattr(self.module_class, 'py3'):
-                i3s_config = self.i3status_thread.config['general']
-                py3 = Py3(module=self, i3s_config=i3s_config)
-                setattr(self.module_class, 'py3', py3)
+                setattr(self.module_class, 'py3', Py3(self))
 
             # get the available methods for execution
             for method in sorted(dir(class_inst)):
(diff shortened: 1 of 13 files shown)


tobes:
I'm not sure whether we should just stick with the `sleep(1)` or actually use the `cached_until` but again for now it'll do as is.

ultrabug:
I'm not so sure about removing the default parameters from the modules methods such as `def check_updates(self, i3s_output_list, i3s_config): ` even if py3 does provide some way to get the i3s_config.

Maybe it would be best suited in another PR, I'm not keen on grouping changes like that tbh and I would rather gladly go for the module_test first !

tobes:
> I'm not so sure about removing the default parameters from the modules methods 

Fine, can I keep them removed where they are not used or do you want them replaced in all cases?

ultrabug:
> Fine, can I keep them removed where they are not used or do you want them replaced in all cases?

replaced for now plz, if we want to remove them, we'll have to update docs et al

tobes:
Updated,

in the vpn module I corrected the `i3s_output_list` param which is unused.  I'll remove it if you wish.

ultrabug:
Thanks a lot that's a great feature to have :+1:

PR #313, opened 2016-05-20
Restart i3status should it die

tobes:
Back to issue #302

It looks as though i3status dying kills py3status due to the main thread being killed when the i3status_thread stops.

This patch just restarts i3status if it is killed for whatever reason.  We only do this is i3status has actual worked correctly initially so faulty configs etc would still cause it to stop.

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -10,7 +10,7 @@ from syslog import syslog, LOG_INFO
 from signal import SIGUSR2, SIGSTOP, SIG_IGN, signal
 from tempfile import NamedTemporaryFile
 from threading import Thread
-from time import time
+from time import time, sleep
 
 from py3status.profiling import profile
 from py3status.events import IOPoller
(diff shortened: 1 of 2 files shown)


py3status/i3status.py
```diff
@@ -522,6 +522,14 @@ def suspend_i3status(self):
 
     @profile
     def run(self):
+        # if the i3status process dies we want to restart it.
+        while self.lock.is_set():
+            self.spawn_i3status()
+            # check if we never worked properly and if so quit now
+            if not self.ready:
+                break
+
```
ultrabug: we may want to throttle this a bit or add a maximal respawn count in cause this ends badly over and over

tobes: a sleep()?

ultrabug: yeah, and maybe a `for _ in range(10):` instead of the while (with a test on loc_is_set() to break ?

ultrabug:
looks sane, thanks !

PR #315, opened 2016-05-21
logging: log to file, Py3.log() etc

tobes:
This is a first draft for some logging changes.  Hopefully this will make it easier to find and fix issue.

* `py3status --log-file /path/to/logfile` option to use that instead of syslog

* `Py3.log(message, level)` helper function, message can be str, list etc for debugging

* `Py3statusWrapper.report_exception()` adds extra traceback info if using custom logfile.

In the future I'd like to add some `LOG_WARNING` constants for py3status internal use but I'd like to do that after the config_parser branch is merged/scrapped. `Py3` supplies and uses constants.

It might also be nice if debug extra logging used `DEBUG` rather than `INFO` in logging

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -3,17 +3,17 @@ from __future__ import print_function
 import argparse
 import os
 import sys
-from collections import deque
+import time
 
+from collections import deque
 from json import dumps
 from signal import signal
 from signal import SIGTERM, SIGUSR1, SIGUSR2, SIGCONT
 from subprocess import Popen
 from subprocess import call
 from threading import Event
-from time import sleep, time
 from syslog import syslog, LOG_ERR, LOG_INFO, LOG_WARNING
-from traceback import extract_tb
+from traceback import extract_tb, format_tb
 
 import py3status.docstrings as docstrings
 from py3status.events import Events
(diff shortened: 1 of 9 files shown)


py3status/core.py
```diff
@@ -124,6 +124,13 @@ def get_config(self):
                             dest="include_paths",
                             help="""include user-written modules from those
                             directories (default ~/.i3/py3status)""")
+        parser.add_argument('-l',
+                            '--log-file',
+                            action="store",
+                            dest="log_file",
+                            type=str,
+                            default=None,
+                            help="path to i3status log file")
```
ultrabug: py3status you mean

py3status/core.py
```diff
@@ -419,6 +424,19 @@ def notify_update(self, update):
...
+        log this information to syslog or user provided logfile.
+        """
+        if not self.config['log_file']:
+            # If level was given as a str then convert to actual level
+            level = LOG_LEVELS.get(level, level)
+            syslog(level, msg)
+        else:
+            with open(self.config['log_file'], 'a') as f:
+                log_time = time.strftime("%Y-%m-%d %H:%M:%S")
+                f.write('{}\t{}\t{}\n'.format(log_time, level.upper(), msg))
```
ultrabug: tabs on log files ? I'd strongly suggest avoiding that please, simple space is enough

ultrabug:
I'm really wondering why we don't use the py3.log in the core and other internal modules tbh but that's a cool enough version for a start, thanks a lot !

tobes:
> I'm really wondering why we don't use the py3.log in the core and other internal modules 

Do you mean with constants?  Just to keep the PR clean but yes that would be nice

ultrabug:
aye, let's keep this in mind for another day ;)

PR #317, opened 2016-05-22
Weather_yahoo docs update

tobes:
I noticed that the doc for this was outdated following your changes.  I also added an example config


ultrabug:
Thanks !

PR #318, opened 2016-05-23
Weather_yahoo: Prevent connectionError when network down

tobes:
Weather_yahoo fails when connection is not available.  eg when my laptop wakes.

This PR just hides the output and tries again in 30 seconds.


ultrabug:
fair enough, better than an useless notification, thx !

PR #320, opened 2016-05-25
events: cleanup and bug fix

tobes:
This PR does 3 things:

* Fixes issue with i3status modules not being updated on middle click.

* Simplifies `events.py`

* Adds `Py3.prevent_refresh()` that allows modules to prevent their `on_click` events causing a refresh.

The prevent_refresh functionality is for modules that do not need to update after the click event and have slow or resource intensive updates.  I can remove this functionality if not wanted but I think it is useful.

### DIFF
--- a/py3status/events.py
+++ b/py3status/events.py
@@ -61,29 +61,9 @@ class Events(Thread):
         self.lock = py3_wrapper.lock
         self.modules = py3_wrapper.modules
         self.on_click = self.i3s_config['on_click']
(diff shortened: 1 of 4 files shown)


tobes:
Removing the `i3bar_click_events` stuff would be nice too.  Not sure if you want to do that yet.

ultrabug:
:skull: dead code :dancer: !

Yes, we should drop the `i3bar_click_events` support, you're right.

PR #323, opened 2016-05-29
Added module to display number of unread emails

Daerandin:
I created a module which displays unread emails. Default settings works with gmail, and different imap servers can be defined in the i3config.


Daerandin:
It seems I overlooked the imap module when I looked through available modules.... *facepalm*

Please delete this request

ultrabug:
Thanks for contributing anyway @Daerandin :) hope this will give you the will to do another module someday hehe

PR #327, opened 2016-06-10
docs: fix examples class name

tobes:
Noticed that the examples got broken in a doc cleanup


ultrabug:
good catch thanks, my bad sorry :(

PR #328, opened 2016-06-13
pomodoro module: make clock more configurable

egeskow:
Added config parameter format_separator. By default it is '-' like 00-00.
Now you can set it to something like 00:00, 00x00.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -235,7 +235,7 @@ This demonstrates how to produce a simple custom module.
 """
 
 
-class py3status:
+class Py3status:
 
     def hello_world(self):
         return {
@@ -307,7 +307,7 @@ Configuration parameters:
 """
 
 
-class py3status:
+class Py3status:
 
     format = 'Hello World!'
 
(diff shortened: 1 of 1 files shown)


py3status/modules/pomodoro.py
```diff
@@ -236,9 +238,11 @@ def pomodoro(self, i3s_output_list, i3s_config):
             mins, seconds = divmod(rest, 60)
 
             if hours:
-                vals['mmss'] = u'%d-%02d-%02d' % (hours, mins, seconds)
+                vals['mmss'] = (u'%d' + self.format_separator + u'%02d'
+                    + self.format_separator + u'%02d') % (hours, mins, seconds)
             else:
-                vals['mmss'] = u'%d-%02d' % (mins, seconds)
+                vals['mmss'] = (u'%d' + self.format_separator
+                    + u'%02d') % (mins, seconds)
```
tobes: It might be nicer to fully use % formatting

eg 
```
vals['mmss'] = u'%d%s%02d' % (mins, self.format_separator, seconds)
```

PR #329, opened 2016-06-13
pomodoro module: make clock more configurable

egeskow:
Added config parameter format_separator. By default it is '-' like 00-00.
Now you can set it to something like 00:00, 00x00.


egeskow:
Thanks tobes for nice hint! 
And created separate branch for commiting

ultrabug:
Sorry @egeskow but you need to rebase your MR, you included a doc file that's been updated since

PR #333, opened 2016-06-15
Added Google Play Music Desktop Player module.

Spirotot:
See https://github.com/GPMDP


py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,56 @@
...
+class Py3status:
+    """
+    """
+    # available configuration parameters
+
+    cache_timeout = 5 
+
+    def gpmdp(self, i3s_output_list, i3s_config):
+        command = 'gpmdp-remote info'
+        result = '♫ ' + check_output(shlex.split(command)).decode('ascii','ignore').strip()
```
tobes: This will cause issues with python2 use  `u'♫ '`

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,56 @@
...
+        split = result.split()
+
+        if split[1] == 'Paused:' and len(split) == 3:
```
tobes: I'd change this line to
```
if len(split) == 3 and split[1] == 'Paused:':
```
so that we know split[1] will exist

ultrabug:
Thanks @Spirotot ! Please take the comments above into account, cheers !

Spirotot:
Sure, no problem! Should I make the recommended changes & resubmit a pull request? (Or perhaps I will find I don't actually need to completely re-submit -- I'm a bit of a GitHub noob. :) )

ultrabug:
just commit in your fork (or better, amend your commit) and it will update this PR automagically !

Spirotot:
I think I've got it updated -- let me know what you think. Thanks!

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""
+Display currently playing song from Google Play Music Desktop Player.
```
tobes: There should be a blank line here

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,56 @@
...
+Configuration parameters:
+    - cache_timeout : how often we refresh this module in seconds (5s default)
+
+Requires
```
tobes: `Requires:`

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,57 @@
...
+class Py3status:
+    """
+    """
+    # available configuration parameters
+
+    cache_timeout = 5
+
+    def gpmdp(self, i3s_output_list, i3s_config):
+        command = 'gpmdp-remote info'
+        result = u'♫ ' + check_output(shlex.split(command)).decode('ascii', 'ignore').strip()
```
tobes: Here it might be nice to allow the user to specify a format  eg `♫ {title}` like other modules do.
eg https://github.com/ultrabug/py3status/blob/master/py3status/modules/external_script.py

Spirotot:
Added custom format string functionality.

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,64 @@
...
+class Py3status:
+    """
+    """
+    # available configuration parameters
+
+    cache_timeout = 5
+    format = '{info}'
+
+    def gpmdp(self, i3s_output_list, i3s_config):
+        result = u'♫ ' + self.format
```
tobes: Would it not be better just to set the format to `format = u'♫ {info}'` then the user has more control of the output.

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,64 @@
...
+
+        if check_output(shlex.split('gpmdp-remote status')).decode('ascii', 'ignore').strip() == 'Paused':
+            result = ''
+        else:
+            cmds = ['info', 'title', 'artist', 'album', 'status', 'current',
+                    'status', 'time_total', 'time_current', 'album_art']
+
+            for cmd in cmds:
+                if str('{' + cmd + '}') in result:
+                    result = result.replace(str('{' + cmd + '}'),check_output(shlex.split('gpmdp-remote %s' % (cmd))).decode('ascii', 'ignore').strip())
```
tobes: I don't like the use of replace here.  Is there a reason you are not making a dict and then using format as is common for other modules.

Also I'm not sure about the str casts does this work with both python2 and 3?

py3status/modules/gpmdp.py
Spirotot: Sure, changed it.

py3status/modules/gpmdp.py
Spirotot: This method minimizes the number of times that `gpmdp-remote` needs to be run, as well as provides a simple 1:1 mapping between format string parameters and the output of gpmdp-remote. It's efficient (in LOC, but it's also not bad in terms of runtime), and easy to use/understand (don't know if  I would say the same for the common `format.format()` convention I keep seeing -- at least name the _variable_/_member_ `format` something like `format_str` to differentiate it from the _method_ `format`? That being said, I'm _not_ saying that I think this GPMDP method is suitable for every module...). 

The alternative dictionary method would be to parse out the artist, album, track, current time, and total track length every time the module is called. That requires 4-5 runs of gpmdp-remote, as well as a bunch of extra code. With this method, if someone wants just the default output ('info'), gpmdp-remote only gets run once. 'artist' and 'track'? Two runs.

`str()` casts work in Python2 and Python3. 

Let me know what you think.

tobes: > str() casts work in Python2 and Python3. 

but they do different things
```
Python 2.7.11 (default, Jun 20 2016, 14:45:23) 
[GCC 5.3.1 20160406 (Red Hat 5.3.1-6)] on linux2
>>> x = u'♫'
>>> str(x)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u266b' in position 0: ordinal not in range(128)

-------------------------------------------------
Python 3.4.3 (default, Jun 24 2016, 16:32:25) 
[GCC 5.3.1 20160406 (Red Hat 5.3.1-6)] on linux
>>> x = u'♫'
>>> str(x)
'♫'
```

tobes: > The alternative dictionary method would be to parse out the artist, album, track, current time, and total track length every time the module is called. 

I was suggesting something like this

```
data = {}
for cmd in cmds:
    if '{%s}' % in result:
        data[cmd] = check_output(['gpmdp-remote', cmd]).decode('utf-8').strip()

full_text = self.format.format(**data)
```

Spirotot: Nice. In this module they are only dealing with the characters in the strings in the `cmds` list, and of course a '{' and a '}'. No Unicode characters.

Either way, the `str()` casts are unnecessary, so I've removed them.

Spirotot: Regarding your dict suggestion: a double-splat? I don't know, I mean, I'll use that code if you want me to, but it's up to you. Not the way I'd write it (obviously)... this is Python. It's supposed to be easy and make sense. We could use C if we want the code to run faster, be less easy to write, and have questionable sensibility.

There's no way that using `.format()` makes a worthwhile speed difference over `.replace()` in a string of this size... This is 2016. I'm assuming that `.format()` is actually faster than multiple `.replace()` calls, although I'm not familiar with it's internals. But if this optimization is a dealbreaker for someone's machine's capabilities/performance, then I think they've got bigger problems, like running building & running Google Play Music Desktop Player...

Please just tell me what you want (with a succinct explanation of _why_, but only if you're feeling generous); neither of our solutions seems _particularly_ bad, and I don't have time to debate every line of code. I think we'd both be happier and have more time to do more worthwhile things with our lives if we didn't. If you want me to include some code that I think is completely stupid and nonsensical (this seems like an unlikely scenario), I'll let you know, and/or just not make the change/commit myself out of principal.

tobes: The simplest reason is that every other module uses `format(...)`.

It is also IMHO more readable and pythonic.

Spirotot: Just pushed changes, using `self.format.format()`. Let me know if there's anything else you'd like to see changed.

Spirotot: Also fixed all flake8 errors on my local machine -- still waiting for Travis to complete, though.

tobes:
> Just pushed changes, using self.format.format(). Let me know if there's anything else you'd like to see changed. 

* get rid of shlex dependency `check_output(shlex.split('gpmdp-remote status'))` -> `check_output(['gpmdp-remote', 'status'])` 

* docstring details (placeholders) and formatting, placeholders

* switch to new testing eg https://github.com/ultrabug/py3status/pull/352/commits/85b2d1f49f551e682753281df683d646b45065ec

* remove hashbang

* remove the `run_cmd()` function.

tobes:
> remove the run_cmd() function.

or at least call it `_run_cmd()` else you break py3status

py3status/modules/gpmdp.py
```diff
@@ -0,0 +1,68 @@
...
+    @staticmethod
+    def run_cmd(cmd):
+        return check_output(shlex.split('gpmdp-remote %s' % (cmd))).decode('utf-8').strip()
+
+    def gpmdp(self, i3s_output_list, i3s_config):
+        if self.run_cmd('status') == 'Paused':
+            result = ''
+        else:
+            cmds = ['info', 'title', 'artist', 'album', 'status', 'current',
+                    'status', 'time_total', 'time_current', 'album_art']
```
tobes: you have 'status' twice

Spirotot:
* [X] get rid of shlex dependency
* [ ] docstring details/formatting -- could you point me to a reference? I'm assuming you're referring to the details at the top of the file... I've copied formatting from another module, but maybe it was out of date? Are you referring to `modules/README.md`?
* [X] Switch to new testing -- I think? 
* [X] Remove hashbang.
* [X] Renamed `run_cmd` to `_run_cmd` to follow convention. Still a static method.

py3status/modules/gpmdp.py
Spirotot: Oops, fixed.

On Wed, Jul 6, 2016 at 8:27 AM tobes <notifications@github.com> wrote:

> In py3status/modules/gpmdp.py
> <https://github.com/ultrabug/py3status/pull/333#discussion_r69720495>:
>
> > +    """
> > +    # available configuration parameters
> > +
> > +    cache_timeout = 5
> > +    format = u'♫ {info}'
> > +
> > +    @staticmethod
> > +    def run_cmd(cmd):
> > +        return check_output(shlex.split('gpmdp-remote %s' % (cmd))).decode('utf-8').strip()
> > +
> > +    def gpmdp(self, i3s_output_list, i3s_config):
> > +        if self.run_cmd('status') == 'Paused':
> > +            result = ''
> > +        else:
> > +            cmds = ['info', 'title', 'artist', 'album', 'status', 'current',
> > +                    'status', 'time_total', 'time_current', 'album_art']
>
> you have 'status' twice
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/ultrabug/py3status/pull/333/files/387ce7ecec4248febb5a10454909dd6eb250afc8#r69720495>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAtBebHDCvR0ISOgvI3--H-26Fap5dJdks5qS58_gaJpZM4I17CO>
> .
>
-- 

-Aaron

tobes:
> docstring details/formatting -- could you point me to a reference?

This needs documenting and not all modules have yet been fixed.

something based on
```
"""
Display currently playing song from Google Play Music Desktop Player.

Configuration parameters:
    cache_timeout: how often we refresh this module in seconds (default 5)
    format: blah blah (default '♫ {info}')

    {info} blah blah blah

Requires:
    gpmdp: maybe some info url etc
    gpmdp-remote:

@author Aaron Fields https://twitter.com/spirotot
@license BSD
"""
```

Spirotot:
Cool, I modified the documentation a bit!

tobes:
Thanks for the updates.

You still have your old docstring which should be removed, and the flake8 test fails need addressing.

It is also a convention to list and describe all the status string placeholders.

Spirotot:
Addressed, sorry for the flake8 errors, I need to automate that (or figure
out how to get PEP8 integrated in to VIM...).

On Thu, Jul 7, 2016 at 6:14 AM tobes <notifications@github.com> wrote:

> Thanks for the updates.
>
> You still have your old docstring which should be removed, and the flake8
> test fails need addressing.
>
> It is also a convention to list and describe all the status string
> placeholders.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/ultrabug/py3status/pull/333#issuecomment-231038896>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAtBeRCrrjxM_9fx-UzaX7Cnu2w3QX0kks5qTNGHgaJpZM4I17CO>
> .
>
-- 

-Aaron

ultrabug:
Impressive work, how close are we to the merge in your opinion guys ?

tobes:
I think we're good.  The docstring still has some minor issues but I'm happy to fix them post merge.

ultrabug:
Thanks a lot @Spirotot ! and ofc @tobes for the nice review

Issue #332, opened 2016-06-14
Separator in Pomodoro Module


ben-e-whitney:
The `pomodoro` module uses a hyphen as a separator between the minutes and seconds left in a timer. The documentation contradicts itself, saying in one place that a hyphen is used and in another that a colon is used. I think a colon makes more sense here. Should I submit a PR changing the default behavior, or maybe instead add a parameter allowing the user to choose the separator?

tobes:
https://github.com/ultrabug/py3status/pull/330 adds a `format_separator` and is just merged.

The documentation could probably do with a tweak

pydsigner:
I think that the default should change here, if anyone actually prefers the old behavior, which I think unlikely, they are now able to get it back. It's even possible that the original hyphen was a typo IMO.

tobes:
> I think that the default should change here

Looks like the default was a `-` a `:` might be nicer but the user can set this now.  I'm not too fussed.

egeskow:
I believe it should be ":" by default. 
There is [ISO 8061](https://en.wikipedia.org/wiki/ISO_8601) standard to time formats
Which [says](https://www.w3.org/TR/NOTE-datetime): 

```
   Complete date plus hours and minutes:
      YYYY-MM-DDThh:mmTZD (eg 1997-07-16T19:20+01:00)
```
We can see "-" separator between year-month-date and ":" between others.

Also there is already [xkcd comics](https://xkcd.com/1179/)

ultrabug:
Fixed thanks to @egeskow !

PR #336, opened 2016-06-15
pomodoro module: Set default time separator to ":"

egeskow:
as defined by ISO8601
following issue #332


ultrabug:
Thanks you're right !

PR #337, opened 2016-06-15
pomodoro module: Pomodoro state colors configurable.

egeskow:
You can set own color for each state by config string:
    colors = "on=#FFFFFF, off=#FF0000, break=#FCE94F".

Format like in keyboard_layout module

### DIFF
--- a/py3status/modules/pomodoro.py
+++ b/py3status/modules/pomodoro.py
@@ -6,6 +6,8 @@ Configuration parameters:
     display_bar: display time in bars when True, otherwise in seconds
     format: define custom display format. See placeholders below
     format_separator: separator between minutes:seconds
+    colors: a comma separated string of color values for each state,
+        eg: "on=#FFFFFF, off=#FF0000, break=#FCE94F".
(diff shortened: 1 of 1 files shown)


py3status/modules/pomodoro.py
```diff
@@ -266,13 +270,21 @@ def pomodoro(self, i3s_output_list, i3s_config):
             response['urgent'] = True
             self._alert = False
 
+        if self.colors:
+            try:
+                self._colors = dict((k.strip(), v.strip()) for k, v in (
+                    color.split('=') for color in self.colors.split(',')))
+            except:
```
pydsigner: Don't use bare exceptions please.

egeskow:
Fixed this to catch only one error type.
User notification added. 

Should module just crash with config errors?

tobes:
I'm not a fan of the comma delimitated string.  If #282 gets accepted then we have proper dicts etc which would be better.  Alternatively #279 effectively solves this issue.

egeskow:
agreed with tobes

PR #339, opened 2016-06-18
Stop notify_user rate limiting notifications from modules

tobes:
Now that `core.notify_user()` does rate limiting (suppresses duplicates) `py3.notify_user()` does not show output when called if a message is a duplicate.

This is unhelpful for debugging and also it is reasonable for a module to display the same information to the user multiple times.  eg. Pomodoro time is up

This commit adds a rate_limit variable to allow notifications to now be rate limited.  It is set automatically in `py3.notify_user()` at 5 seconds.

It does not strictly limit the rate of duplicate messages but does prevent more than one being shown for every time period, this greatly simplifies things.

I use the hash function to just reduce the weight of our duplicate set.

`rate_limit` could be made available to users via `py3.notify_user()` in the future, or I can add now if wanted.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -465,7 +465,7 @@ __trigger_event(module_name, event)__
 
 Trigger an event on a named module.
 
-__notify_user(msg, level='info')__
+__notify_user(msg, level='info', rate_limit=5)__
 
 Send a notification to the user.
 `level` must be `info`, `error` or `warning`.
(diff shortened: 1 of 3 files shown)


pydsigner:
I'd be for adding it to `py3.notify_user()` now.

tobes:
Updated

* add `rate_limit` param to `py3.notify_user()`

* each module is now rate limited independently

* allow none string messages to be displayed on dbus eg int, bool (helpful when debugging)

ultrabug:
Cool addition and nice hashing trick, thanks !

PR #340, opened 2016-06-18
github module: fix button press issue due to notification count not being reset

tobes:
When the number of notification was zero our notification count was not reset.  This fixes the issue so we now go to the correct url.


ultrabug:
Thanks !

PR #341, opened 2016-06-21
add backlight plugin

tjaartvdwalt:
this is a simple plugin that shows your screen backlight level. By default it uses the `light` command to achieve this.


py3status/modules/backlight.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""
+Display the current screen backlight level.
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds (10s default)
+    format: prefix text for the backlight status
+    command: the command used to retrieve the bakglight level.
+             By default we use `light` https://github.com/haikarainen/light
+
```
tobes:
```
Format status string parameters:
    {level} brightness percentage

Requires:
    light: cli utility
```

py3status/modules/backlight.py
```diff
@@ -0,0 +1,45 @@
...
+import subprocess
+from time import time
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    format = '☼: {level}'
```
tobes: `format = u'☼: {level}'` will fix for python2

tobes:
Looks good.

Is there a reason you used `light` rather than `cat /sys/class/backlight/acpi_video0/brightness`?  Working out the percent might take a bit more work but it'd reduce the dependencies.

tjaartvdwalt:
@tobes: thanks for the feedback, I made the changes you suggested

py3status/modules/backlight.py
```diff
@@ -0,0 +1,48 @@
...
+    cache_timeout = 10
+    device = "acpi_video0"
+    format = u'☼: {level}'
+
+    def backlight(self,  i3s_output_list, i3s_config):
+        level = int(subprocess.check_output(
+            ["cat", "/sys/class/backlight/%s/brightness" % self.device]))
+        max = int(subprocess.check_output(
+            ["cat", "/sys/class/backlight/%s/max_brightness" % self.device]))
+        full_text = self.format.format(level=round((level/max) * 100))
```
ultrabug: in python2 this division will return 0.0, you should add this in the imports :
```
from __future__ import division
```

ultrabug: Also, imho displaying a float for brightness sounds overkill to me, maybe you could adapt your division like this:
```
full_text = self.format.format(level=(level * 100 // max))
```

ultrabug: Finally, I am picky but I'd advise not to use the `max` variable name since it's a reserved name for the max function in python

py3status/modules/backlight.py
```diff
@@ -0,0 +1,48 @@
...
+        response = {
+            'cached_until': time() + self.cache_timeout,
+            'full_text': full_text,
+            'color': '#FFFFFF'
```
ultrabug: Please do not force users' coloring output, #279 will bring a fix for this later.

py3status/modules/backlight.py
```diff
@@ -0,0 +1,52 @@
...
+Display the current screen backlight level.
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds (default: 10s)
+    device: The backlight device (default: "acpi_video0")
+            If you are unsure try: `ls /sys/class/backlight`
+
+Format status string parameters:
+    {level} brightness
+    {color} text color
```
tobes: This should be in the config parameter section

tjaartvdwalt:
@ultrabug thanks for some good comments. I have made the changes.

py3status/modules/backlight.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+"""
+Display the current screen backlight level.
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds (default: 10s)
+    device: The backlight device (default: "acpi_video0")
+            If you are unsure try: `ls /sys/class/backlight`
+    color:  The text color (default: "#FFFFFF")
```
tobes: `(default: "#FFFFFF")` etc you don't need the `:`

py3status/modules/backlight.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    device = "acpi_video0"
+    format = u'☼: {level}%'
+    color = '#FFFFFF'
```
tobes: ultrabug is keen on these being in alphabetical order (same with in the docstring)

With the color, I'd be inclined to have the default as `None` and only set it in the responsive if the user specifies one.
As it is #279 will move the color setting out of the modules anyway so I'd not worry too much.

ultrabug:
Ok @tjaartvdwalt thanks.

I'm still not a fan of using a subprocess to simply read the content of a file and I really think that the `device` option could be guessed (and thus should default to None).

Reading a file, I imagine something like:
```
for brightness_line in open(brightness_file_path, 'rb'):
    brightness = int(brightness_line)
    for brightness_max_line in open(brightness_max_file_path, 'rb'):
        brightness_max = int(brightness_max_line)
        level = (brightness * 100 // brightness_max)
```

And for guessing the device:
```
if self.device is None:
    for (path, devices, files) in os.walk('/sys/class/backlight/'):
        for device in devices:
            if 'brightness' in os.listdir(device) and 'brightness_max' in os.listdir(device):
                device_path = path + device
                break
```

I guess you see what I have in mind.

That being said, I know I'm being picky so I see two options before us:
1. You are okay to implement the changes
2. I merge your module as-is and possibly do it myself

tjaartvdwalt:
@ultrabug, yes I made the mistake of not thinking in terms of "reading a file", but rather "executing a command". Your solution is much more elegant.

I have some work pressure at them moment, but I will see if I can find some time tonight to make the changes you suggested.

ultrabug:
Thanks @tjaartvdwalt, you might make it for v3.0 due tomorrow ;)

ultrabug:
friendly bump @tjaartvdwalt

tjaartvdwalt:
@ultrabug, thanks for the push :)

py3status/modules/backlight.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+"""
+Display the current screen backlight level.
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds (default: 10s)
+    color:  The text color (default: "#FFFFFF")
+    device: The backlight device
+            If not specified the plugin will try to detect it automatically
```
tobes: format should be described here

py3status/modules/backlight.py
```diff
@@ -0,0 +1,65 @@
...
+               'max_brightness' in os.listdir(path + device):
+                return path + device
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    color = '#FFFFFF'
```
tobes: `device = None`

py3status/modules/backlight.py
```diff
@@ -0,0 +1,65 @@
...
+    format = u'☼: {level}%'
+
+    def backlight(self,  i3s_output_list, i3s_config):
+        if(not hasattr(self, 'device')):
+            device_path = get_device_path()
+        else:
+            device_path = "/sys/class/backlight/%s" % self.device
```
tobes: Can we not just remember the device_path nrather than having to calculate it every time?

gferon:
Any news on this? I'm using this code locally and it works wonders.

tobes:
> Any news on this? 

Currently this is awaiting @tjaartvdwalt responding to request for some minor tweaks

py3status/modules/backlight.py
tjaartvdwalt: not 100% sure what you mean by this comment?

tjaartvdwalt:
sorry for taking so long on this. Hopefully we can get it into the next release.

py3status/modules/backlight.py
tobes: You use `self.device` but it is not defined.  All config options should have a default even if it is None.

Actually having looked at things it might be good if you find a device if none is defined.  On my system I have two `acpi_video0` and `intel_backlight` it would work with either they are both the same device.  I've not tried with multiple monitors etc

py3status/modules/backlight.py
```diff
@@ -0,0 +1,68 @@
...
+                self.device_path = "/sys/class/backlight/%s" % self.device
+
+        for brightness_line in open("%s/brightness" % self.device_path, 'rb'):
+            brightness = int(brightness_line)
+        for brightness_max_line in open("%s/max_brightness" % self.device_path, 'rb'):
+            brightness_max = int(brightness_max_line)
+
+        full_text = self.format.format(level=(brightness * 100 // brightness_max))
+        response = {
+            'cached_until': time() + self.cache_timeout,
```
tobes: We now expect all modules to use `py3.time_in()`

So this line would now be `'cached_until': self.py3.time_in(self.cache_timeout),`

py3status/modules/backlight.py
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return response
+
+
+if __name__ == "__main__":
+    from time import sleep
+    x = Py3status()
+    while True:
+        print(x.backlight([], None))
+        sleep(1)
```
tobes: We also now do standalone testing differently

```
if __name__ == "__main__":
    """
    Run module in test mode.
    """
    module_test(Py3status)
```

py3status/modules/backlight.py
tjaartvdwalt: ok I think I understand now... I added `device=None`, and then removed the `hasattr` part of the test

py3status/modules/backlight.py
```diff
@@ -36,12 +36,14 @@ class Py3status:
     # available configuration parameters
     cache_timeout = 10
     color = '#FFFFFF'
+    device = None
     device_path = None
```
tobes: Only config parameters should be defined here.

the ideal thing would be to add a post_config_hook method (this gets called once the module has had its config set but before its output method is first called.
```
def post_config_hook(self):
    if not self.device:
        self.device_path = get_device_path()
    else:
        self.device_path = "/sys/class/backlight/%s" % self.device
```

py3status/modules/backlight.py
```diff
@@ -36,12 +36,14 @@ class Py3status:
...
     format = u'☼: {level}%'
 
     def backlight(self,  i3s_output_list, i3s_config):
+        print(self.device_path)
```
tobes: this isn't needed

py3status/modules/backlight.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+"""
+Display the current screen backlight level.
+
+Configuration parameters:
+    cache_timeout: How often we refresh this module in seconds (default: 10s)
```
tobes: we don't need the `s` or the `:` in the default so `(default 10)`

`(default None)` and `(default '☼: {level}%')` for device and format would be good too

py3status/modules/backlight.py
```diff
@@ -0,0 +1,68 @@
...
+        for brightness_line in open("%s/brightness" % self.device_path, 'rb'):
+            brightness = int(brightness_line)
+        for brightness_max_line in open("%s/max_brightness" % self.device_path, 'rb'):
+            brightness_max = int(brightness_max_line)
+
+        full_text = self.format.format(level=(brightness * 100 // brightness_max))
+        response = {
+            'cached_until': self.py3.time_in(self.cache_timeout),
+            'full_text': full_text,
+            'color': self.color
```
tobes: We now do color setting for all modules auto-magically, so it is no longer needed here or in the docstring/config options

py3status/modules/backlight.py
tjaartvdwalt: hi I am not exactly sure how to setup the post_config_hook. Here is a small sample app I tried:

```python
class Py3status:
    def post_config_hook(self):
        return 'post config'

    def backlight(self,  i3s_output_list, i3s_config):
        return 'response'

if __name__ == "__main__":
    """
    Run module in test mode.
    """
    module_test(Py3status)
```

but this just executes the methods (in alphabetical order). I get the following output:
```
post config
response
...
```

is there something I am missing here?

tobes: yes a bug in the test module https://github.com/ultrabug/py3status/pull/474

ultrabug:
Thanks @tjaartvdwalt ! I'm merging now to get it in 3.1 ;)

tjaartvdwalt:
great!
thanks @tobes for a very rigorous review process :)
hopefully my next plugin will have less issues.

PR #342, opened 2016-06-22
notify_user: suppress restart message if module

tobes:
This seemed to be the simplest way to do this without adding an extra param.  I've renamed the param to show it's new usage better.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -326,7 +326,7 @@ class Py3statusWrapper():
             # load and spawn i3status.conf configured modules threads
             self.load_modules(self.py3_modules, user_modules)
 
-    def notify_user(self, msg, level='error', rate_limit=None, key=''):
+    def notify_user(self, msg, level='error', rate_limit=None, module_name=''):
         """
         Display notification to user via i3-nagbar or send-notify
         We also make sure to log anything to keep trace of it.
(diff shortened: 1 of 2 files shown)


ultrabug:
Great, thanks ! :fireworks:

py3status/core.py
```diff
@@ -326,7 +326,7 @@ def setup(self):
             # load and spawn i3status.conf configured modules threads
             self.load_modules(self.py3_modules, user_modules)
 
-    def notify_user(self, msg, level='error', rate_limit=None, key=''):
+    def notify_user(self, msg, level='error', rate_limit=None, module_name=''):
```
ultrabug: Good point. For the format line 359 maybe @tobes ?

tobes: The reason I did this was just because it's a string in my mind.  If we had a module called None (unlikely) then it would clash.

But I'd keep it as switching to None would cause the line length to go over 79 chars.

It's not important really so if people want to change that's fine,  I have no strong view on it.

pydsigner: It wouldn't actually end up clashing because `None` doesn't compare equal to the string `'None'`. I'd be in favor of changing just because `None` is more Pythonistic.

py3status/core.py
```diff
@@ -356,7 +356,7 @@ def notify_user(self, msg, level='error', rate_limit=None, key=''):
                 pass
         # We use a hash to see if the message is being repeated.  This is crude
```
pydsigner: Any issues with this hasher can actually be avoided by taking the hash of a tuple of these three items rather than making them into a string.

PR #343, opened 2016-06-23
Fix version for development

tobes:
If I don't re-run setup.py the version is not updated (assuming `python setup.py develop`)

```
$ git log -n 1 --oneline 
d4cc870 version 3.1_rc from here
$ py3status -v
py3status version 2.9 (python 3.4.3)
```

also after updating

```
$ git checkout master
$ py3status -v
py3status version 3.1_rc (python 3.4.3)
$ git checkout 2.9
$ py3status -v
py3status version 3.1_rc (python 3.4.3)
```
Looking around the best option is to have a file containing the version and then both setup.py and core.py can get the version from that.  It's a bit messy but it means that `py3status -v` will give a true result.


tobes:
An alternative would be to have `__version__ = 'xxxx'` in a python file and read that.  The logical one would be `__init__` but importing that by `core.py` would be a pain and having it in `core.py` feels wrong.

Ideas welcome

PR #344, opened 2016-06-23
Version fix alternative

tobes:
This is an alternative I prefer it except for where the `__version__= '...'` line is.


tobes:
I've changed this to move `__version__` into `__init__.py`

PR #345, opened 2016-06-24
fix version setup and detection for install and develop phases

ultrabug:
I'd like to propose this for #343 and #344, what do you think ?


tobes:
I'm happy with this approach.

tobes:
ugly but keeps flake8 happy

```
def get_version():
    '''
    Parse version.py file to find the version looking for line like
    version = 'X.Y'
    '''
    core_path = os.path.join(os.path.dirname(__file__), 'py3status', 'version.py')
    with open(core_path) as f:
        for line in f:
            if line.startswith('version'):
                return eval(line.split('=')[1].strip())
```

ultrabug:
hmm I'd like to find a better way to avoid this :(

ultrabug:
Better not flake8 on setup.py tbh

PR #347, opened 2016-06-26
re-add setup.py to flake checks

tobes:
We can still use flake8 on setup.py


ultrabug:
wow nice trick, thanks !

PR #349, opened 2016-06-27
Exchange rate module

tobes:
Now that the exchange rate of my country is in free fall.

I thought an exchange rate module might be a fun way to watch the car crash.

### DIFF
--- a/py3status/modules/exchange_rate.py
+++ b/py3status/modules/exchange_rate.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-View foreign exchange rates.
+Display foreign exchange rates.
 
 The exchange rate data comes from Yahoo Finance.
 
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks for this *sad times* module, did you omit the standalone testing part on purpose ?

tobes:
> did you omit the standalone testing part on purpose ?

yes because it needs https://github.com/ultrabug/py3status/pull/312

So if you pull this I'll add the test there, or here if that is pulled first.

I need to add a timeout and a few things to this PR still, I'll ping you when I've updated

tobes:
Ok I've added the timeout/bad connection bits

ultrabug:
Let's work on #312 first then.

py3status/modules/exchange_rate.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""
+View foreign exchange rates.
```
ultrabug: Display foreign exchange rates.

(more in line with other modules)

tobes:
Test now added (and minor doc change you wanted)

ultrabug:
:+1: thank you !

PR #350, opened 2016-06-28
Add hostname/dns support for the module online_status

ritze:
Now you can check the online status via ping.

### DIFF
--- a/py3status/modules/online_status.py
+++ b/py3status/modules/online_status.py
@@ -43,7 +43,7 @@ class Py3status:
         else:
             fnull = open(os.devnull, 'w')
             return subprocess.call(['ping', '-c', '1', self.url], stdout=fnull,
-                               stderr=subprocess.STDOUT) == 0
+                                   stderr=subprocess.STDOUT) == 0
 
     def online_status(self, i3s_output_list, i3s_config):
         response = {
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks @ritze ! wow that's a chain of PRs ;)

I'd rather you use python [subprocess call](https://docs.python.org/2/library/subprocess.html#subprocess.call) please ?

py3status/modules/online_status.py
```diff
@@ -25,18 +27,24 @@ class Py3status:
...
+    format_online = u'●'
     timeout = 2
     url = 'http://www.google.com'
 
     def _connection_present(self):
-        try:
-            urlopen(self.url, timeout=self.timeout)
-        except:
-            return False
+        if re.search('://', self.url):
```
tobes: `if '://' in self.url:` would save having to use the re module

ritze:
Is this the right way to push the stdout and stderr to /dev/null?

py3status/modules/online_status.py
```diff
@@ -33,18 +33,17 @@ class Py3status:
...
             else:
                 return True
         else:
-            if os.system("ping -c 1 " + self.url + ">> /dev/null 2>&1") == 0:
-                return True
-            else:
-                return False
+            fnull = open(os.devnull, 'w')
+            return subprocess.call(['ping', '-c', '1', self.url], stdout=fnull,
+                               stderr=subprocess.STDOUT) == 0
```
tobes: use `stderr=fnull` too

ultrabug:
Thanks mate, good addition !

PR #351, opened 2016-06-28
Added MPRIS module

ritze:
Just check if some player with mpris support is running and show the song information. You can also prioritize your players.


ritze:
Hmm, there are some wrong commits of my file_status module. These should be not there...

ritze:
I opened a new issue, see #354

PR #352, opened 2016-06-28
Add new module file_status

ritze:
A simple module which shows you if a file or dir in your filesystem is available or not.

### DIFF
--- a/py3status/modules/file_status.py
+++ b/py3status/modules/file_status.py
@@ -33,8 +33,7 @@ class Py3status:
         }
 
         fnull = open(os.devnull, 'w')
-        if subprocess.call(["ls", self.path], stdout=fnull,
-                                              stderr=subprocess.STDOUT) == 0:
+        if subprocess.call(["ls", self.path], stdout=fnull, stderr=fnull) == 0:
             response['full_text'] = self.format_available
             response['color'] = i3s_config['color_good']
         else:
--- a/py3status/modules/file_status.py
+++ b/py3status/modules/file_status.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""
+Display if a file or dir exists.
+
+Configuration parameters:
+    cache_timeout: how often to run the check
+    color_available: color of format_available
+    color_unavailable: color of format_unavailable
+    format_available: what to display when available
+    format_unavailable: what to display when unavailable
+    path: check if file or dir exists
+
+@author obb, Moritz Lüdecke
+"""
+
+from time import time
+import os
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    color_available = None
+    color_unavailable = None
+    format_available = u'●'
+    format_unavailable = u'■'
+    path = '/dev/sdb1'
+
+    def file_status(self, i3s_output_list, i3s_config):
+        response = {
+            'cached_until': time() + self.cache_timeout
+        }
+
+        if os.system("ls " + self.path + " >> /dev/null 2>&1") == 0:
+            response['full_text'] = self.format_available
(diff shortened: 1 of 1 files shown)


py3status/modules/file_status.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    color_available = None
+    color_unavailable = None
```
tobes: @ultrabug I worry about having these config params.  Although the names may be appropriate for this module, there is something nice about sharing names eg color_bad as it keeps the configs consistent.

Maybe you have a nice insight on this

ritze:
Now the module uses only the global params color_good and color_bad. Additionally the module use subprocess instead of os.system.

py3status/modules/file_status.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+"""
+Display if a file or dir exists.
+
+Configuration parameters:
+    cache_timeout: how often to run the check
+    color_available: color of format_available
+    color_unavailable: color of format_unavailable
```
tobes: they are still here :)

ritze:
Maybe I'm a little bit too fast with handling three branches at the same time. :-)

py3status/modules/file_status.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+"""
+Display if a file or dir exists.
+
+Configuration parameters:
```
ultrabug: same as the other process module, please state the defaults (and path shouldn't have one imo)

py3status/modules/file_status.py
```diff
@@ -0,0 +1,55 @@
...
+        if subprocess.call(["ls", self.path], stdout=fnull, stderr=fnull) == 0:
+            response['full_text'] = self.format_available
+            response['color'] = i3s_config['color_good']
+        else:
+            response['full_text'] = self.format_unavailable
+            response['color'] = i3s_config['color_bad']
+
+        return response
+
+if __name__ == "__main__":
```
ultrabug: same comment as your process module, see github module for inspiration plz

py3status/modules/file_status.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+import subprocess
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    format_available = u'●'
```
ultrabug: totally arbitrary comment, do we feel a bullet point is a good default to state if a file is present or not? I'm not sure tbh

ritze:
Instead a bullet point we can use a [white](http://unicode-table.com/de/2690) ⚐ and [black flag](http://unicode-table.com/de/2691/) ⚑. But these symbols are not really better than the bullet point.

Or we just use text: "target exists" and "no target"

tobes:
There have been issues with unicode chars not always being available to all users.

Personally I'd like all core modules to use ascii and allow users to then specify their own formats as preferred.

Some existing modules do however use unicode in their defaults.  Also currently python2 installs cannot use unicode in the config although there is an outstanding pull request #282 that would fix this.

What I'd really like to see is my 'themes' dream get accepted but I'm waiting for my current PRs to be accepted/rejected first.  This would have a set of themes that users can quickly enable that would alter the look/feel of py3status with minimal effort for one or all modules.

```
general {
    theme = 'SuperCool'
}
```

ultrabug:
Same comment as the process_status and the bullet points, maybe the file path or basename would have been a better default tbh as I tend to agree with @tobes wrt unicode stuff.

Thanks a lot @ritze !

PR #353, opened 2016-06-28
Add new module process_status

ritze:
Just show you if a process is currently run or not.

### DIFF
--- a/py3status/modules/process_status.py
+++ b/py3status/modules/process_status.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Display if a file or dir exists.
+Display if a process is running.
 
 Configuration parameters:
     cache_timeout: how often to run the check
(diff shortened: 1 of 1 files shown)


py3status/modules/process_status.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""
+Display if a file or dir exists.
```
ultrabug: I guess this is not the right description :)

py3status/modules/process_status.py
```diff
@@ -0,0 +1,56 @@
...
+
+Configuration parameters:
```
ultrabug: please also state the defaults (and I think that process should have no default)

py3status/modules/process_status.py
```diff
@@ -0,0 +1,56 @@
...
+                           stdout=fnull, stderr=fnull) == 0:
+            response['full_text'] = self.format_running
+            response['color'] = i3s_config['color_good']
+        else:
+            response['full_text'] = self.format_not_running
+            response['color'] = i3s_config['color_bad']
+
+        return response
+
+if __name__ == "__main__":
```
ultrabug: If you feel like it, please switch to the new 3.1 module testing framework (see https://github.com/ultrabug/py3status/blob/master/py3status/modules/github.py#L229)

ultrabug:
Thanks a lot for your contributions @ritze !

ultrabug:
I'm still not convinced that displaying a bullet or a square is the best way of showing if a process is running or not but I'm not sure I have better to propose apart displaying the name of the process itself on the bar.

Anyway this is not a blocker to me !

Thanks a lot @ritze !

PR #354, opened 2016-06-28
Added MPRIS module

ritze:
Just check if some player with mpris support is running and show the song information. You can also prioritize your players.

### TODO
- [x] Use dbus listeners
- [x] Use list format for `player_priority` rather than a comma separated string
- [x] Use formatting with optional placeholders, e.g. `{state} [[[{artist} - ]{title}]|[{file}]]` and maybe delete `format_stream`
- [x] Variable placement of buttons
- [x] Get informed when a player with a higher priority is still running but is less active than a player with a lower priority
- [x] Test the module with all players of the world :-)


py3status/modules/mpris.py
```diff
@@ -0,0 +1,236 @@
...
+Requires:
+    dbus
+
+@author Pierre Guilbert, Jimmy Garpehäll, sondrele, Andrwe, Moritz Lüdecke
+"""
+
+from datetime import timedelta
+from time import time, sleep
+import re
+import dbus
```
tobes: pydbus would be better to use 
https://www.freedesktop.org/wiki/Software/DBusBindings/

ultrabug:
Which module did you base yourself upon please ? I'm asking because there is the player_control module which is meant for this AFAIK, what do you think ?

Thanks

ritze:
My code is based on the spotify plugin. player_control only supports vlc and audacious.

My daily routine is listening music with mpd/cantata and watching videos with [bomi](https://bomi-player.github.io) while the music player is paused. Based on this behavior the mpris plugin always shows the current information of the most active player. E.g. if mpd paused and bomi playing I get the video title which bomi is played. When I listening music with mpd and I just check for five seconds a video in bomi the plugin don't switch unnecessary.

tobes:
Is there a reason that player_control could not learn about mpris?

ritze:
Absolutely not. :-) Should be easy to merge this two modules. I'll do it when I find some time. Maybe already this weekend.

tobes:
I should have said spotify shouldn't I :)

ritze:
The mpris module should already includes all functionality of the spotify plugin. But I didn't test it with spotify because I have no spotify account.

tobes:
Actually having thought a bit more this is what I'd be in favour of creating a super new media_player module as follows.  We could then deprecate player_control and spotify.

* use pydbus rather than dbus python lib as it has better support etc

* only use dbus for everything getting track / artist, player controls

* use composites so multiple buttons (play, next, volume, shuffle etc)

* display track/artist etc

* it could even handle multiple players at the same time though that might be a bit mental.

Spotify is a really specific name and none of the features seem spotify specific.  Player_control is limited and written specifically for each player when there is no need we could just use dbus.  It also doesn't display info about the tracks.  Merging both these features into one module. feels sensible too me.

Maybe having track display and controls separate is a good idea.

@ultrabug what would your view be?

ritze:
These points were still on my todo list for the next days. :-)

tobes:
I'm not a fan of that last commit.  I'd prefer there to be 'real' buttons on the status bar rather that left/right clicking for actions etc.

ritze:
Didn't know that this is possible.

tobes:
It's a new feature in 3.0 and currently not well documented.  It is used in the timer module.

ritze:
Handling streaming is a little bit tricky because the metadata field is in each player different:
* Cantata doesn't set the dbus metadata property
* mpd set it to "artist - title"
* and vlc set the title just to the radio station name and additionally to that it defines a new metadata tag "vlc:nowplaying" with the artist and title in it.

I don't know how I could do this better as currently.

Also I doesn't like my detection of video and music (see is_video variable). It smells like a workaround...

tobes:
> Handling streaming is a little bit tricky because the metadata field is in each player different:

It might be worth looking at how other projects deal with this eg https://github.com/enkore/i3pystatus/blob/master/i3pystatus/now_playing.py (not that helpful but you might find something better)

ritze:
With the last commit you can show only information of one specific player or multiple specific players. Additionally you can add more instances of the mpris module in your py3status config, so you can show multiple players at the same time in your bar.

You can also show other players if your favorite player isn't running. Just add a `,*` in the **player_priority** parameter. This is the "old" way.

tobes:
I think you have too many formats etc . The clementine module https://github.com/ultrabug/py3status/blob/master/py3status/modules/clementine.py has a simple format, even it not user set-able.

This should aim to be lean first.

ritze:
So which features are expendable? I think `genre`, `date` and `track` are not so important. For me `shuffle` and `loop` are also unimportant, but maybe someone wants this. Same with `time`, `length` and `time_left`.

Personally I only use `state`, `artist` and `title`. `player` can also be useful, when you run multiple players at the same time or multiple instances

tobes:
> Personally I only use state, artist and title. player can also be useful, when you run multiple players at the same time or multiple instances

That sounds a better start

ritze:
I kept `album`, `time` and `length` because this seems useful in this plugin. Also I delete `format_no_tags` and use `format_stream` instead.

ritze:
Now the user can decide if he wants real buttons or interact with the text.

ultrabug:
This is a huge one, how close to final do you think this module is mates ?

tobes:
I'd really like to spend some time simplifying/streamlining the code.  I was going to wait for @ritze to finish making his regular changes before starting on that.

ritze:
I'm done with my changes. It's your code now. :-)

tobes:
> It's your code now. :-)

No, this is still your baby ;p

py3status/modules/mpris.py
```diff
@@ -0,0 +1,378 @@
...
+                    'full_text': ' ' + self.state_next + ' ',
+                    'index': 'next',
+                })
+            elif button == 'previous':
+                response.append({
+                    'color': '#CCCCCC',
+                    'full_text': ' ' + self.state_previous + ' ',
+                    'index': 'previous',
+                })
+
```
tobes: To me this looks like repeated code.

We know the index.  We could have a crafty dict
```
whatever = {
    'play': state,
    'stop': self.state_stop,
    ...
}
```
for the bits that are dynamic and remove a few lines of code.

tobes: We could even check for capabilities of the player (via dbus) to decide if they should be enabled or disabled the control.  Well have an appropriate color.

maybe the `whatever` dict might need to be cleverer eg
```
whatever = {
    'play': {'enabled': ?, 'icon': state, ...},
    ...
}
```

`whatever` is a terrible name

py3status/modules/mpris.py
```diff
@@ -0,0 +1,378 @@
...
+        else:
+            if button == 1:
+                if index == 'play':
+                    self._player.PlayPause()
+                elif index == 'stop':
+                    self._player.Stop()
+                elif index == 'next':
+                    self._player.Next()
+                elif index == 'previous':
+                    self._player.Previous()
```
tobes: that `whatever` dict might save a few lines here too :)

or at least make adding new ones really easy

py3status/modules/mpris.py
```diff
@@ -0,0 +1,378 @@
...
+                    length = _get_time_str(_length_ms)
+            else:
+                # use stream format if no metadata is available
+                is_stream = True
+        except Exception:
+            color = i3s_config['color_bad']
+
+        if is_stream:
+            # delete the file extension
+            title = re.sub(r'\....$', '', title)
```
tobes: Your regex `r'\....$'` only catches 3 letter extensions `r'\.[^.]+*$'` might be better.  But is hiding the extension important?  I'm not sure.  Do all streams have extensions? what if there is a `.` in a 'proper' title

py3status/modules/mpris.py
```diff
@@ -0,0 +1,378 @@
...
+            if button == 'play':
+                response.append({
+                    'color': '#CCCCCC',
+                    'full_text': ' ' + state + ' ',
+                    'index': 'play',
+                })
+            elif button == 'stop':
+                response.append({
+                    'color': '#CCCCCC',
+                    'full_text': ' ' + self.state_stop + ' ',
```
tobes: `' ' + self.state_stop + ' '`

What if I don't like these spaces?  why not just add the spaces in `state_stop = ' ◾ '` if I do?

ritze: `state_play`, `state_pause` and `state_stop` are also used for the placeholder `{state}`. If you move the spaces from the source code to the format string than you can be get an ugly text e.g. for `format = "{state} {title}"` you'll get `" ◾ We Will Rock You"`.

But yes I also don't like this code. Just couldn't find a better solution here…

py3status/modules/mpris.py
ritze: Primary I use this for hiding file extensions when playing videos. Most of the video files have no tags.

tobes:
> format = "{state} {title}"

I think that is fine but we should have.

icon_stop, icon_play, icon_pause ....

then the buttons list is 'play, stop, toggle' ...

so {state} and 'button icons' have separate purposes rather than state_play also used for a button.

Make sense?

py3status/modules/mpris.py
```diff
@@ -0,0 +1,385 @@
...
+                                        self.icon_stop,
+                                        self._player.Stop),
+            'next':     _get_ctr_struct(self._player.CanGoNext,
+                                        self.icon_next,
+                                        self._player.Next),
+            'previous': _get_ctr_struct(self._player.CanGoPrevious,
+                                        self.icon_previous,
+                                        self._player.Previous)
+        }
+
```
tobes: using `_get_ctr_struct()` makes this hard to read why not just be explicit to start with.

 ``` 
'play': {'clickable': ...,
         'icon': ...,
         'action':...},
```

Also I'd just use the attributes eg `'CanGoPrevious'` and then do `getattr(self._player, attribute)` as needed rather than having to keep generating the dict all the time.

File: py3status/modules/mpris.py [hunk 7]
```diff
@@ -0,0 +1,385 @@
...
+        state = 'pause' if self._player.PlaybackStatus == 'Playing' else 'play'
+        control_states['play_pause'] = control_states[state]
+
+        return control_states
+
+    def _get_response_buttons(self):
+        response = []
+        buttons_order = self.buttons_order.split(',')
+        available = ['pause', 'play', 'play_pause', 'stop', 'next', 'previous']
```
tobes: isn't this just the keys of `_control_state` dict so doesn't need repeating.

`for button in [e for e in buttons_order if e in self._control_states]:`

py3status/modules/mpris.py
ritze: CanGoPrevious etc are methods. I call them finally when I need the value, see `self._control_states[index]['action']()` line 378. Or does this work in another way in python?

tobes: `getattr(self._player, self._control_states[index]['action'])()` works if the value is a string.

The reason is that you don't need to keep defining the dict when `self._player` is changed.

py3status/modules/mpris.py
```diff
@@ -8,7 +8,7 @@
 to define the button parameters in the i3status config.
 
 Configuration parameters:
-    button_play: mouse button to toggle between play and pause mode
+    button_play_pause: mouse button to toggle between play and pause mode
```
tobes: is `button_toggle` a better name?

ultrabug:
Have you guys reached a good level of satisfaction here ?

py3status/modules/mpris.py
```diff
@@ -0,0 +1,382 @@
...
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    button_stop = None
+    button_toggle = 1
+    button_next = 4
+    button_previous = 5
+    buttons_control = None
```
tobes: I'm not sure of this name it is a bit too close to `button_xxx` maybe something like `show_controls`?

py3status/modules/mpris.py
```diff
@@ -0,0 +1,382 @@
...
+                              state=state,
+                              album=album,
+                              artist=artist,
+                              length=length,
+                              time=time,
+                              title=title), color)
+
+    def _get_control_states(self):
+        control_states = {
+            'pause':    {'clickable': self._player.CanPause,
```
tobes: like the action I think it would be nice to make this a string too.  We would only have to check for controls we care about rather than all of them and also this would then be static and would just need to be set up once.

Stop is a funny one as there is no `CanStop` but defining this as `None` would be easy to get around this issue.

tobes:
We are getting there :) My main concern currently is the config params as we will end up 'stuck' with them.  So getting them right is the main thing.

py3status/modules/mpris.py
```diff
@@ -0,0 +1,385 @@
...
+        return control_states
+
+    def _get_response_buttons(self):
+        response = []
+        buttons_order = self.buttons_order.split(',')
+        available = self._control_states.keys()
+
+        for button in [e for e in buttons_order if e in available]:
+            control_state = self._control_states[button]
+            clickable = getattr(self._player, control_state['clickable'], True)
```
ritze: The trick with None didn't work but this workaround should works good enough.

tobes:
Ok,  I've started tested this.

My first observations

* We don't seem to be listening to dbus.  eg I use vlc to play a track. toggle play pause the module does not update except on the schedule of default update interval.  We could subscribe to the player and then instantly respond to that state change.

* The current time should really be updating every second when it is running.  We could do this

These are things that we can improve later so I'm not too bothered about them for now (the time thing can be a bit tricky but we have done this in other modules)

The controls.

*  The spacing for these at default is too much we could do this better adding something like `'min_width': 20, 'align': 'center',` to the control output feels like a nicer way to achieve the spacing.

* the color for the controls is set and might not work for users why not just use `color_good` etc?

* placement of the controls.  Why not just have a `{controls}` status string placeholder? (again this is a bit of work but we've got similar code elsewhere).

* the controls are not very clever so the play control is 'enabled' even if the player is playing, to me it should be disabled at that point.  Also next/previous are enabled even for a single track in the playlist (this might be trickier and I'm not sure if they really should be enabled or not, they restart the song) - This is something that isn't important as it can be improved on later.

Other

* the on_click function throws an error for unknown buttons being pressed adding a `else: return` in the `if index == 'text':` would be a simple fix.

@ritze what do you think?  If you need a pointer on any of the bits let me know.

tobes:
Maybe with the `{commands}` thing go even further and have things like `{command_toggle}` `{command_next}` that way we could remove `commands`.

ritze:
- I had the idea with the `{commands}` placeholder too, but I aborted it because it were a bit too heavy at this time. I'm relatively new in python. But I like this idea too.
- There are listener in dbus so we can get every changes instantly. Especially for time and state changes.
- I only tested the module with **mpDris2** (for mpd), **Cantata** and a little bit **vlc** and **bomi**. Each player do some other strange things. For example Cantata doesn't set the url in the metadata.
- I didn't know that there is another way for spacing without blanks. So your idea is much better than my way with one blank character on each side of the symbol.
- `CanPlay`, `CanStop` and so on are controlled by the player. So only the player knows how the specific behavior behind this buttons are. For example you could listening an audio stream and when you click on the play button the player "could" restart the audio stream and clean the buffer. Yes the behavior of the control buttons could be better but we should also think about it what can be go wrong.

tobes:
> I had the idea with the {commands} placeholder too, but I aborted it because it were a bit too heavy at this time. I'm relatively new in python. But I like this idea too.

If you like I could add this and make a pull request.

> CanPlay, CanStop and so on are controlled by the player. So only the player knows how the specific behavior behind this buttons are. For example you could listening an audio stream and when you click on the play button the player "could" restart the audio stream and clean the buffer. Yes the behavior of the control buttons could be better but we should also think about it what can be go wrong.

Not sure about on other players but vlc just seems to treat these as 'yes you can send the Play command' rather than 'this is a time when this actually does something'.

I was thinking that we know that if we are playing then play should be disabled.  But I think this is fine for now.

py3status/modules/mpris.py
```diff
@@ -277,40 +281,67 @@ def _get_text(self, i3s_config):
...
+
+    def _get_response_buttons(self, i3s_config):
         response = []
         buttons_order = self.buttons_order.split(',')
         available = self._control_states.keys()
 
         for button in [e for e in buttons_order if e in available]:
             control_state = self._control_states[button]
-            clickable = getattr(self._player, control_state['clickable'], True)
+            control_state['enable'] = self._get_button_state(control_state)
```
tobes: I think this is a backward step a it makes control_states have state, which feels wrong to me.  There isn't a need for this.  You are only really needing it here.  You do use it in `on_click()` but then you are using a cached version and it has no real value.  In the on_click if you care then why not just recalculate the value?

ritze:
- [x] Use native button spaces
- [x] Use native `color_good` and `color_bad` for button state
- [x] Intelligent buttons
- [ ] Use dbus listeners
- [ ] Update time every second (when it's running)
- [ ] Variable placement of buttons
- [ ] Test the module with all players of the world :-)

Did I forget some point?

I also added an optional setting for the button state color so the user can decide…

py3status/modules/mpris.py
```diff
@@ -123,17 +125,19 @@ class Py3status:
     button_next = 4
     button_previous = 5
     buttons_order = 'previous,toggle,next'
+    color_buttons_bad = None
+    color_buttons_good = None
```
tobes: maybe `color_control_active` /  `color_control_inactive` buttons is a little confusing imho

ritze: Didn't found a better expression. I'm very bad in choosing good variable names.

py3status/modules/mpris.py
```diff
@@ -0,0 +1,422 @@
...
+        is_stream = False
+        player = UNKNOWN
+        state = UNKNOWN
+        artist = UNKNOWN
+        album = UNKNOWN
+        length = UNKNOWN_TIME
+        ptime = UNKNOWN_TIME
+        title = UNKNOWN
+
+        _left_s = None
```
tobes: @ritze generally in python a leading `_` has special meaning.  It shows that the thing should be treated as private.  Python has no real private as we can just find anything in a module/class etc, but the underscore let's people know that they should not be playing with stuff, or that it is not supported.

In Py3status objects we use it for methods names to stop them being run to get a modules output too.

`_left_s` is a local variable so it is effectively private anyhow so there is no need to have the leading underscore.  If you want to keep it that's fine but reading the code it just looks a bit out of place.

py3status/modules/mpris.py
```diff
@@ -407,6 +407,8 @@ def on_click(self, i3s_output_list, i3s_config, event):
                 index = 'next'
             elif button == self.button_previous:
                 index = 'previous'
+            else:
+                return
```
ritze: I mistyped, this commit fixes a bug…

ultrabug:
Wow that ought to be the most worked on module during a PR @ritze ! Thanks so much for the time and effort you put into this.

Guys, I trust you both on this module to ping me when I shall merge it.

Cheers !

tobes:
A few more thoughts on this

* I'm still keen on having a simpler format.  I like how mpd https://github.com/ultrabug/py3status/tree/master/py3status/modules#mpd_status does it's formatting eg `{state} [[[{artist} - ]{title}]|[{file}]]`  Personally I think this would be useful for several modules.  I've got some code that does this and could be added to the `Py3` class.  @ultrabug do you think this would be useful?  I'd probably make it an option on `Py3.safe_format()`.  Anyhow it would mean we would need one less format string.

* Now that #282 is merged we can set `player_priority` to being a list in the config file rather than using a comma separated string.  It might be worth seeing if the user set a string eg `player_priority` = 'vlc'` and just converting to a list in that case.  Also maybe consider changing the name to `players`.

Anyhow @ritze let me know if you need a hand with any of the fiddly bits.

ritze:
> I like how mpd https://github.com/ultrabug/py3status/tree/master/py3status/modules#mpd_status does it's formatting eg {state} [[[{artist} - ]{title}]|[{file}]]

I had this idea some time ago too, when I separated the format string into `format` and `format_stream`. But I discarded this to keep the code simple.

I added this two features to the TODO list at the opening pull request comment.

tobes:
PR #384 adds parse_format() for `{state} [[[{artist} - ]{title}]|[{file}]]` style stuff

ritze:
I don't know why but the `_on_change` function is never called. I played a little bit with the code but couldn't find the problem yet.

The [example code](https://github.com/LEW21/pydbus/blob/master/pydbus/tests/gnome_music.py) of pydbus works fine. Maybe because they're using `GObject.MainLoop()` instead of `sleep(X)`?

tobes:
Yeah you need the loop.

tobes:
Module vpn_status implements pydbus listening https://github.com/ultrabug/py3status/blob/master/py3status/modules/vpn_status.py

You need to run the loop in a thread a thread etc see `_start_handler_thread()`

tobes:
From your todo list

> Variable placement of buttons

I was going to look at making a nice helper function for this in Py3 as it will be cleaner to make a single implementation rather than each module that wants to do this re-implementing it.

ritze:
I tried to connect with all mpris players which are available in dbus. This had the effect that I get multiple signals from one same mpris bus/player. I don't know what's went wrong so I only check at the regular interval time for new players and if the playback state has been changed.

Can I update the text by myself when some state has been changed or do I have to wait for the regular update interval?

tobes:
> Can I update the text by myself when some state has been changed or do I have to wait for the regular update interval?

If you are getting that info then `self.py3.update()` in your module will force an update.

py3status/modules/mpris.py
```diff
@@ -317,30 +318,24 @@ def _get_text(self, i3s_config):
...
-                update = time() + 1
+        try:
+            ptime_ms = self._player.Position
+            ptime = _get_time_str(ptime_ms)
+        except Exception:
+            ptime = UNKNOWN_TIME
+
+        if '{time}' in format and self._data['state'] == PLAYING:
+            # Don't get trapped in aliasing errors!
+            update = time() + 0.5
```
ritze: This doesn't work. Can it be that the granularity of py3status is 1 second?

tobes: The minimum update time is 0.1 seconds.

tobes:
@ritze coming along nicely

A few things

* I get an error on line 426 when `self.player` is `None`.  Just adding a test for that before subscribing fixes the issue. This happens starting the module with no player running.

* Could you also change lines with unicode like `icon_pause = '▮▮'` to `icon_pause = u'▮▮'` for python2 compatibility.

* `state_stop` is not defined.

* where you delete the file extension if `self._data['title']` is `None` it fails, just check for that condition.

Hopefully #384 will get merged soon and then the formatting and adding controls via the format string will be available.

tobes:
>  I patched my local code but I get always the brackets including the content in it.

can you give an example?

There is still a python2 unicode issue with songs that have unicode names but this is probably better fixed inside safe_format so works for all modules.

ritze:
At the current state of the branch I get `[▶ The Kinks - Victoria]|[▶ Victoria]` instead of `▶ The Kinks - Victoria`

I build the string with `self.py3.safe_format(self.format, placeholders)`.

tobes:
what is you `format`? for me this works fine but I had to merge that branch into your one to test things.

ritze:
> what is you format?

`format = '[{state} {artist} - {title}]|[{state} {title}]'`

I checked out the current master branch and copied the changes #384.

tobes:
Odd I cannot reproduce this

What version of python are you running? I'm on 3.4.3

Incidentally you probably want

`format = '{state} [{artist} - {title}]|[{title}]'`

as `{state}` will always be something.

ritze:
Seems to be my fault: `build_composite`couldn't found. Something went wrong on my merge....

ritze:
I found it. I installed on my system the master branch without your patch and locally the master branch with your patch. Now you can guess whats went wrong on the execution. :-/

tobes:
I found another small issue.

If there is no player and I click on mpris I get an error.  you just need to check for `self._player`

tobes:
I had a little play getting the module to tell when a player is added or removed.  This patch seems to do the trick.  You would need to recheck players that are available eg remove the player if none are, before calling the update()

```
diff --git a/py3status/modules/mpris.py b/py3status/modules/mpris.py
index d3da8f6..8352e0f 100644
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -368,7 +368,22 @@ class Py3status:
         self._loop = GObject.MainLoop()
         self._loop.run()
 
+    def _available_players_changed(self, *args):
+        if len(args) < 6:
+            return
+        if args[5][0].startswith('org.mpris.MediaPlayer2'):
+            self.py3.update()
+
     def _start_listener(self):
+        self._dbus.con.signal_subscribe(
+            None,
+            'org.freedesktop.DBus',
+            'NameOwnerChanged',
+            None,
+            None,
+            0,
+            self._available_players_changed,
+        )
         t = Thread(target=self._start_loop)
         t.daemon = True
         t.start()
```

You are using `py3.build_composite()` incorrectly `py3.safe_format()` works for how things are now.

To use `build_composite()`

```
return {
    'cached_until': ..,
    'composite': self.py3.build_composite(....),
}
```

ritze:
All features are implemented now. It's debug time … yeah…

py3status/modules/mpris.py
```diff
@@ -0,0 +1,497 @@
...
+        if self._player is None:
+            self._player = None
+            self._player_name = 'None'
+            self._player_subscription = None
+            text = self.format_none
+            color = i3s_config['color_bad']
+            composite = ({
+                'full_text': text,
+                'color': color,
+            })
```
tobes: a composite is a list so `composite = [{ ... {[` you have an error when no player.

tobes:
:)

I've updated #384 so your new format parses correctly (shows the space between {state} and the title.

tobes:
The color isn't showing for the state (playing etc).

This is because we are expecting it for the section.  I think I'll change how the composites are handled so that the 'color' you do pass gets used.  I'll have a bit more of a think if that is the best approach.

tobes:
Currently things get quite 'jumpy' when playing/pausing a song due to `state_pause` and `icon_pause` being two characters whilst the others are one character.

I had a look but couldn't find a nice single character pause symbol.

Just using `u'▮'` rather than  `u'▮▮'` has a much nicer not 'jumpy' feel.  I'd be inclined to use that as it make the module feel more 'professional' even if it's not the normal pause symbol people are used to.

In the future I want to add themes to py3status which would give us a py3status custom font so we could have nice symbols.  This of course is not a given to happen, but users can always add a font say font-awesome and use the symbols for that as things are.

tobes:
I think your format is wrong
```
format = '{previous}{toggle}{next} {state}[ [{artist} - {title}]|{title}]'
```
should be (extra space before title)  I haven't tested this but I think it's needed
```
format = '{previous}{toggle}{next} {state}[ [{artist} - {title}]| {title}]'
```

tobes:
Good work on this @ritze 

As I see it

* we are now dependent on PR #384 and #395

* minor docstring stuff (adding defaults) eg
```
    button_stop: mouse button to stop the player (default None)
```

* we should also add in the docstring a list of known working players I have only tested with 'vlc' and ask users to report other players that work.

* you have `# TODO: This should also works with "del self._player"` this doesn't make sense to me if you do `del self._player` then every reference to it eg `if self._player:` will crash and burn.

* `cached_until = time() + self.cache_timeout` and be changed to `cached_until = self.py3.CACHE_FOREVER` then we can remove `cache_timeout` config parameter.

There are a few niggly things about dealing with multiple players both of different types and multiple instances of the same player.  I think this does not really present an issue for this PR as it can be improved later.  Maybe along with the ability to actively change the player being monitored.

I am also not sure about a couple of color names eg `color_control_inactive` in PR #279 we are working towards more universal colors that can work across modules (some of the discussions have been outside of the git log issue).  But the names are not decided yet so nothing to do here as yet.

There are also a few python2 unicode issues but they need fixing in the main code not here and we work in most situations in python2.

ritze:
> `cached_until = time() + self.cache_timeout` and be changed to `cached_until = self.py3.CACHE_FOREVER` then we can remove `cache_timeout` config parameter.

With this change we don't get recognized when a player was started with a higher priority than the current one.

tobes:
> With this change we don't get recognized when a player was started with a higher priority than the current one.

We should get a dbus notification when a player is opened/closed.

Is this when I have two players open but then play the higher priority one?  If so maybe we need to subscribe to all active players (or at least the higher priority ones) so we can update when they change state eg stopped -> playing

py3status/modules/mpris.py
```diff
@@ -0,0 +1,500 @@
...
+    button_next: mouse button to play the next entry (default 4)
+    button_previous: mouse button to play the previous entry (default 5)
+    color_control_inactive: button color if button is not clickable
+    color_control_active: button color if button is clickable
+    color_paused: text color when song is paused, defaults to color_degraded
+    color_playing: text color when song is playing, defaults to color_good
+    color_stopped: text color when song is stopped, defaults to color_bad
+    format: see placeholders below
+    format_none: define output if no player is running
+    icon_pause: text for the pause button in the button control panel (default ▮)
```
tobes: `(default ▮)` -> `(default '▮')`

pydsigner:
Shouldn't the space just be after `{state}` rather than duplicated?

tobes:
> Shouldn't the space just be after `{state}` rather than duplicated?

As things are it probably would work the same but if we ever had a state but no artist and title then it does make a difference.

tobes:
@ritze If I kill the active player then I often get this error.

```
Error: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.mpris.MediaPlayer2.vlc was not provided by any .service files (2)
  File "/home/toby/dev/py3status/py3status/module.py", line 400, in run
    elif isinstance(response, tuple):
  File "/home/toby/dev/py3status/py3status/modules/mpris.py", line 461, in mpris
  File "/home/toby/dev/py3status/py3status/modules/mpris.py", line 361, in _get_response_buttons
  File "/home/toby/dev/py3status/py3status/modules/mpris.py", line 197, in _get_button_state
  File "/usr/lib/python3.4/site-packages/pydbus/proxy_property.py", line 22, in __get__
    return instance["org.freedesktop.DBus.Properties"].Get(self._iface_name, self.__name__)
  File "/usr/lib/python3.4/site-packages/pydbus/proxy_method.py", line 62, in __call__
    0, instance._bus.timeout, None).unpack()
```

ritze:
The priority functionality is also a little bit buggy. I think I can fix this next week end…

ritze:
> Is this when I have two players open but then play the higher priority one? If so maybe we need to subscribe to all active players (or at least the higher priority ones) so we can update when they change state eg stopped -> playing

That's the problem. But as I already mentioned:

> I tried to connect with all mpris players which are available in dbus. This had the effect that I get multiple signals from one same mpris bus/player. I don't know what's went wrong so I only check at the regular interval time for new players and if the playback state has been changed.

tobes:
> That's the problem. 

Is this something I should leave with you, or do you want me to take a look?

ritze:
> Is this something I should leave with you, or do you want me to take a look?

I don't know if I've time for this the next two weeks. So you can try to fix it. :-)

tobes:
@ritze could you merge upstream master into this branch when you get a chance as the PRs we were dependent on have been merged.

tobes:
@ritze Thanks for taking those PRs.  I'm thinking that this is getting ready for a final review and then I'll be happy to recommend we get it merged.

Is there much stuff you want to change before that happens?  I know you said you were quite busy so if you need some time that is not a problem.

I'm aware of a couple of minor glitches but nothing serious.  Sometimes closing the only player takes a second to go to 'no player running' but it is really hard to hit it so a bit of a pain to bug fix.  If we get it actively used that should help shake out any final issues.

ritze:
I think we lost some priority functionality with your patch. When I set mpd to the highest priority and pause mpd. Than the module is also shown the title of the mpd player even another player is running.

tobes:
> I think we lost some priority functionality with your patch. When I set mpd to the highest priority and pause mpd. Than the module is also shown the title of the mpd player even another player is running.

this can easily be changed `WORKING_STATES = ['Playing', 'Paused']` says which should be considered active.  To me if it is paused it feels active but if you consider that inactive changing to  `WORKING_STATES = ['Playing']` would do as you want.

tobes:
> I don't know what the mainstream wants…

I think we should just make an executive decision.  And I think what you said makes sense.  I'll make a quick patch to do that.

ritze:
I've a strange bug. Two players are open and one is running. When I open a third player the output seems to be frozen. No reactions if I change a state of any player. After closing the third player or restarting py3status the module is running perfectly.

tobes:
anything in the logs?  And is it easy to reproduce?

ritze:
No output from mpris module when I start py3status in debug mode. I think the module is frozen when I start a second player. There is also no reaction from the module when I change the state of the first player until I close the second player. But this behaviour is not always.

If I've time I'll try to debug it…

> anything in the logs? And is it easy to reproduce?

tobes:
> But this behaviour is not always.

I think sometimes `self._player` ends up referring to a player that is no longer on dbus.  When it gets queried then we die.  We are inside the thread so we don't get good feedback about this happening and then `self.py3.update()` ends up not being called.

So we need to make things a bit more robust and check we always access `self._player` in a way that is safe for dbus throwing an error.

Threads can be a bit of a pain.

ritze:
I think there are more than one thread per player instance. I get a tons of debug messages from the same player about the same state change if I add a print statement to `player_on_change`

tobes:
Are all instances of py3status killed?  Sometimes they hang about after a restart.

ritze:
py3status has 16 threads which is ok. If I start the mpris module in test mode (`python mpris.py`) than the threads number is increasing constantly. That's all but not good. Now I know why my terminal couldn't start a new child process yesterday. I started the mpris module in test mode five minutes ago…

ritze:
After running the new code for a while the module hangs very often. We should fix this bug before releasing the module.

As I mentioned I'm very busy at the moment. Maybe I find some time in the next week.

tobes:
Ok, I'll see if I can take a look.  We do want this working smoothly as it's a great module.

tobes:
@ritze The plan is to merge this (along with the patch I made in your repo) so that we get it into 3.1 (due out hopefully today/tomorrow) and it will then get some real life usage (which should help flush out any remaining issues)

Please let us know if you have an objection to this.

ritze:
With the last pull of the master branch I get "invalid format". I'll check this…

py3status/modules/mpris.py
```diff
@@ -0,0 +1,552 @@
...
+            update = self.py3.CACHE_FOREVER
+
+        placeholders = {
+            'player': self._data['player'],
+            'state': state_symbol,
+            'album': self._data['album'],
+            'artist': self._data['artist'],
+            'length': self._data['length'],
+            'time': ptime,
+            'title': self._data['title'] or 'No Track',
```
ritze: When we set `'artist'` and `'title'` we get an _invalid format_ error. Without this two fields the module runs like a dog behind a cat.

Is this an issue in the py3status code?

tobes:
>When we set 'artist' and 'title' we get an invalid format error. Without this two fields the module runs like a dog behind a cat.

> Is this an issue in the py3status code?

Yes It was in the formatter code, thanks for triggering it.  I just need to write some tests then I'll push the code

tobes:
fix for this is in #460

ritze:
@tobes thx, it's working now again.

There are still two issues:
* Running a player with a high priority and starting a second player with a lower priority the module do nothing until the second player is closed.
* Sometimes the module freezes.

I think we can push the module to mainstream and open a new ticket for this two issues.

ultrabug:
Thank you so much @ritze for this amazing contribution, patience and so polished result !

Tons of thank you to you also @tobes for your guidance, review and time :)

This is by far the most collaborative and worked on module, congratulations !

PR #357, opened 2016-07-01
volume_status: color_bad for muted channel

chezstov:
Muted channel now will display with "bad color" (with default thresholds)

Reason: muted state must be equal to zero-volume.


ultrabug:
Good point and fix @chezstov thanks !

PR #363, opened 2016-07-11
Ignore ? as battery time_remaining value, because it might exist on b…

ZeiP:
…oth batteries.

It seems that the row currently assumes that only the active battery has time_remaining defined, but in fact it is always defined, with value '?' if not known. This patch fixes that to only consider the battery active if the value is also not ?.

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -273,7 +273,7 @@ class Py3status:
             active_battery = None
             inactive_battery = battery_list[:]
             for battery_id in range(0, len(battery_list)):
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks @ZeiP that's a good point, can you please make sure your PR code doesn't break flake8 validation ? (see CI)

ZeiP:
@ultrabug, of course, sorry I didn't notice that. I fixed the line length issue and all checks seem to pass now.

ultrabug:
Thanks @ZeiP !

Issue #362, opened 2016-07-11
List index ot f range with module tztime


piquev:
since the v3, I have an error when i add the module tztime

"py3status: list index out of range please try to fix this and reload i3wm" 

> My i3satus.conf

```
order += "tztime local"

tztime local{
        format = "%Y-%m-%d %H:%M:%S"
}
```

thx for any help

tobes:
Hi,

Is that your entire config?  For me using 3.0/master this works as expected.  Do you still have the issue if you run with that as the full config?

piquev:
it's not my full config file,
but when i comment every thing except this i have this error ...

there is my full config file

```
# i3status configuration file.
# see "man i3status" for documentation.

# It is important that this file is edited as UTF-8.
# The following line should contain a sharp s:
# ß
# If the above line is not correctly displayed, fix your editor first!

# ß
general {
        colors = true
        interval = 5
        output_format = i3bar
        color_good = "#2267a5"
       color_degraded = "#8c7f22"
       color_bad = "#be2422"
}

order += "netdata"
order += "whatismyip"
order += "xrandr"
order += "disk /"
order += "battery 0"
order += "whoami"
order += "tztime local"

netdata {
  nic = "wlp58s0"
}

disk "/" {
       format = " Disk: %free "
}

battery 0 {
       format = "%status %percentage %remaining "
       path = "/sys/class/power_supply/BAT0/uevent"
       low_threshold = 20
}

tztime local{
        format = "%Y-%m-%d %H:%M:%S"
}

```

tobes:
I can't replicate this here.  Is there anything in the logs?

running something like ` py3status -c ~/.i3/test.conf -l /tmp/py3status.log` should hopefully give some info.

piquev:
```
2016-07-11 16:26:54 WARNING Setup error (IOError) i3status.py line 282.
2016-07-11 16:26:54 INFO Traceback
IOError: [Errno 2] No such file or directory: '/home/vpi//.i3/test.conf'
  File "/usr/lib64/python2.7/site-packages/py3status/__init__.py", line 17, in main
    py3.setup()
  File "/usr/lib64/python2.7/site-packages/py3status/core.py", line 281, in setup
    self.i3status_thread = I3status(self)
  File "/usr/lib64/python2.7/site-packages/py3status/i3status.py", line 175, in __init__
    self.config = self.i3status_config_reader(config_path)
[...]

PR #364, opened 2016-07-12
Prevent error when i3status does not return timezone info (issue 362)

tobes:
Fixes issue #362


ultrabug:
I'm confused since #362 got closed already, is it a code hardening PR ?

tobes:
Yes, but now that it has been fixed I'm not sure if we should try to give the user a better error or just hide the issue (which is what this does)

ultrabug:
I'd say leave it as-is tbh

tobes:
Cool if we get another problem with the Timezone thing then we can look at a 'proper message' for the user.

PR #365, opened 2016-07-12
Better logging/reporting of i3status errors

tobes:
This makes issues like #362 much easier to debug as we now report the line number and provide traceback in the logs.


ultrabug:
Thank you indeed

PR #366, opened 2016-07-13
cleanup arch_updates

tobes:
* Re-order config params.

* Use new testing

* Docstring minor changes

I'd really like to change the `include_aur` to be boolean (default False) It would still work for people who use 0 etc as I'd change the test from `if include_aur == 0:` to `if not include_aur:`


ultrabug:
sure @tobes then do it :)

ultrabug:
Thanks !

PR #368, opened 2016-07-19
New module: coin_balance

fmorgner:
The 'coin_balance' module fetches the users balances of different
crypto-currencies from a user-defined server. The currently supported
coins are: Bitcoin, Dogecoin, and Litecoin. 'coin_balance' requires a
server that is compatible with the Bitcoin RPC protocol and is easily
extendable to new crypto-currencies by extending the 'COIN_PORTS'
dictionary.

### DIFF
--- a/py3status/modules/coin_balance.py
+++ b/py3status/modules/coin_balance.py
@@ -1,0 +1,138 @@
+# -*- coding: utf-8 -*-
+
+"""
+Display balances of diverse crypto-currencies
+
+This module grabs your current balance of different crypto-currents from a
+wallet server. The server must conform to the bitcoin RPC specification.
+Currently Bitcoin, Dogecoin, and Litecoin are supported.
+
+Configuration parameters:
+    cache_timeout: The time between refreshs (default 30)
+    color: The color for the output (default color_good)
+    format: The format string for the output (default 'LTC: {litecoin}')
+    host: The server hostname (default 'localhost')
+    protocol: The protocol to use to connect to the server (default 'http')
+
+Format status string parameters:
+    {bitcoin} Your bitcoin balance
+    {dogecoin} Your dogecoin balance
+    {litecoin} Your litecoin balance
+
+Requires:
+    requests: python module from pypi https://pypi.python.org/pypi/requests
+
+Example:
+
+```
+# Get your bitcoin balance
+coin_balance {
+    cache_timeout = 45
+    color = "#ffcd18"
+    format = "My BTC: {bitcoin}"
+    host = "localhost"
+    protocol = "http"
+}
+```
+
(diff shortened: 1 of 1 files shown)


py3status/modules/coin_balance.py
```diff
@@ -0,0 +1,138 @@
...
+            with open(expanduser('~/.{0}/{0}.conf').format(coin)) as cfg:
+                creds = {}
+                for line in cfg.readlines():
+                    fields = line.split('=')
+                    if len(fields) == 2 and fields[0] in CREDENTIAL_FIELDS:
+                        creds[CREDENTIAL_FIELDS[fields[0]]] = fields[1].strip()
+
+                if all(key in creds for key in CREDENTIAL_FIELDS.values()):
+                    return creds
+        except FileNotFoundError:
```
tobes: This exception  does not exist in python 2.x

py3status/modules/coin_balance.py
```diff
@@ -0,0 +1,138 @@
...
+            pass
+
+            return None
```
tobes: This return is unneeded as the default action of a function is to return None

tobes:
Does the user need a way to set their credentials?

fmorgner:
@tobes Thanks for the feedback! Since I'am somewhat new to python, would you recommend just capturing IOError instead of FileNotFoundError? Regarding credentials, i will implement that ASAP.

tobes:
`OSError` catches a wider bunch but probably fills your need here.

```
from errno import ENOENT
...
except OSError as e:
    if e.errno == ENOENT:
        # file not found
```
is the equivalent of  FileNotFoundError that works in python 2/3 but is horrible

pydsigner:
IOError would be better than OSError if it works.

ultrabug:
Thanks a lot for your proposal @fmorgner !

I'll let you take @tobes and @pydsigner remarks into account, they're great reviewers and this shall be a great way to improve while making a better module :)

py3status/modules/coin_balance.py
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+        return self._credential_cache[coin]
+
+    def _get_balance(self, coin):
+        if coin not in COIN_PORTS:
+            return 'Unsupported coin'
+
+        try:
+            credentials = self._get_credentials(coin)
```
tobes: This feels like it should be before the `try`

tobes:
@ultrabug apart from the last comment I've just added the others have been done.  I can't test this but it looks good to me.  There is a raw except in `_get_balance()` that I don't like but I think here I can live with it.

ultrabug:
Thanks a lot @fmorgner !

fmorgner:
Thanks for the support from all of you!

PR #372, opened 2016-07-25
Improve parsing in safe_format()

tobes:
Reviewing @fmorgner's PR made me realise that `Py3.safe_format()` could be improved by not using our own parser.


ultrabug:
Thanks !

Issue #373, opened 2016-07-27
UnboundLocalError: local variable 'py3' referenced before assignment


aDrz:
Hello!

After a upgrade of debian, py3status seems to be slightly unhappy. I tried to reinstall it, however I still have this error when I try to launch it:

`dude@dudladly:~$ py3status 
Traceback (most recent call last):
  File "/usr/bin/py3status", line 9, in <module>
    load_entry_point('py3status==3.0', 'console_scripts', 'py3status')()
  File "/usr/lib/python2.7/dist-packages/py3status/__init__.py", line 22, in main
    py3.report_exception('Setup error')
UnboundLocalError: local variable 'py3' referenced before assignment`

Thank you in advance for any help ! Have a good day.

tobes:
Thanks for reporting this.

Unfortunately this error is not the actual cause of the problem.

do you have a locale set?

```
$  locale
```

aDrz:
mhmh i see, i see... 

`locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
LANG=fr_FR.UTF-8
LANGUAGE=
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL=
`

aDrz:
Thank you, indeed it is due to my shitty locale variables...

PR #374, opened 2016-07-27
Better handling of init errors especially locale issues

tobes:
This would help for issue #373 

Basically check for the locale independently

Check we have `py3` available when reporting errors


ultrabug:
good point, thanks

tobes:
I was thinking do we really care if the locale is unavailable?  Maybe we should just keep going

ultrabug:
Not really, I was asking myself the same question tbh

Issue #369, opened 2016-07-19
py3status does not respect commented out modules


ultrabug:
maybe this applies to group modules only, don't have time to debug but just saw it so I wanted to share so not to forget
```
#order += "group tro_hits"
#order += "group tro_lag"
#order += "group tro_bean"
#order += "group tro_spool"
```

and on the logs I see the modules listed on those groups loaded and executed (I get their notifications!)

tobes:
I haven't been able to reproduce this yet but I'll try.

tobes:
OK,

I understand the issue better now.  I thought you were saying that the group modules were appearing in the logs, rather than the contained modules appearing.

Yes this is buggy.  The real solution IMHO is to switch to the config parser in #282 as that fixes the issue as well as all the extra niceness,  I've merged master into the branch so it is up-to-date with the latest changes.  We could fix it with the current setup but that feels like just creating more work.

ultrabug:
Ok, I'll review #282 in priority then, thanks a lot @tobes !

ultrabug:
FYI tested with your config-parser branch and I get the same result mate

tobes:
odd with config-parser branch I see

```
2016-07-26 12:50:24 INFO i3status spawned using config file /tmp/py3status_4fy1ova5
2016-07-26 12:50:24 INFO loading module "volume_status" from py3status.modules.volume_status
2016-07-26 12:50:24 INFO loading module "battery_level" from py3status.modules.battery_level
2016-07-26 12:50:24 INFO loading module "xrandr_rotate" from py3status.modules.xrandr_rotate
2016-07-26 12:50:24 INFO loading module "xrandr" from py3status.modules.xrandr
2016-07-26 12:50:24 INFO loading module "dpms" from py3status.modules.dpms
2016-07-26 12:50:24 INFO loading module "fedora_updates" from py3status.modules.fedora_updates
2016-07-26 12:50:24 INFO loading module "github" from py3status.modules.github
2016-07-26 12:50:24 INFO loading module "ticker" from py3status.modules.ticker
```

with master

```
2016-07-26 12:50:46 INFO i3status spawned using config file /tmp/py3status_0c2ffd3z
2016-07-26 12:50:46 INFO loading module "volume_status" from py3status.modules.volume_status
2016-07-26 12:50:46 INFO loading module "battery_level" from py3status.modules.battery_level
[...]

PR #376, opened 2016-07-28
allow passing unicode to i3status

tobes:
This fixes the issue re i3status unicode and python2 #369


ultrabug:
works ! thanks a lot !

PR #377, opened 2016-07-28
Catch timeouts

tobes:
I've been getting notified of some occasional timeouts this PR seems to fix the issue


ultrabug:
Good idea, thanks mate

PR #378, opened 2016-07-28
Add new module check_tcp

ritze:
Checks if a ssh service is available.

Maybe this module is too trivial to pull it upstream…


py3status/modules/ssh_status.py
```diff
@@ -0,0 +1,66 @@
...
+            response['full_text'] = self.format_up
+            response['color'] = self.color_up or i3s_config['color_good']
+        else:
+            response['full_text'] = self.format_down
+            response['color'] = self.color_down or i3s_config['color_bad']
+
+        return response
+
+if __name__ == "__main__":
+    x = Py3status()
```
ultrabug: please use the new testing module :)

ultrabug:
Silly question, but wouldn't it be worth to make a standard `check_tcp` module instead ? that would be handy !!

ritze:
> but wouldn't it be worth to make a standard check_tcp module instead ? that would be handy !!

I not only check if the service on the given port answers but also check if it's an ssh service: `ssh_connectable = 'SSH' in str(output)`

But I could delete this line and the plugin should also works for ssh even it's unclear if on this port is a ssh server running.

ritze:
Now the module runs faster because we don't need timeouts for socket. :-)

tobes:
> But I could delete this line and the plugin should also works for ssh even it's unclear if on this port is a ssh server running.

`nmap` would allow you to actually check the actual service on the port but it does take a little while for remote hosts.  It is also a dependency.

ultrabug:
> I not only check if the service on the given port answers but also check if it's an ssh service: ssh_connectable = 'SSH' in str(output)
> But I could delete this line and the plugin should also works for ssh even it's unclear if on this port is a ssh server running.

OR you could make an option for this named something like `expected_string` which checks if the answer contains the given string if specified !

ultrabug:
I'll do some QA and permit myself to change the default format, thanks again @ritze !

PR #382, opened 2016-07-31
Add module frame

tobes:
The frame module.  This allow modules to be grouped.  This is useful if you wish to add more than one module to a 'slot' in a group.

```
group {
    frame {
        volume_status {}
        battery_level {}
    }

    time {}
}
```

### DIFF
--- a/py3status/modules/frame.py
+++ b/py3status/modules/frame.py
@@ -1,21 +1,64 @@
 # -*- coding: utf-8 -*-
 """
-Group two or more modules as a single one.
+Group modules and treat them as a single one.
 
 This can be useful for example when adding modules to a group and you wish two
 modules to be shown at the same time.
 
+By adding the `{button}` placeholder in the format you can enable a toggle
+button to hide or show the content.
+
+Configuration parameters:
+    button_toggle: Button used to toggle if one in format.
+        Setting to None disables (default 1)
+    format: Display format to use (default '{output}')
+    format_button_open: Format for the button when frame closed (default '-')
+    format_button_closed: Format for the button when frame open (default '+')
+    format_separator: Specify separator between contents.
+        If this is None then the default i3bar separator will be used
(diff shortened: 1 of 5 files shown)


ultrabug:
We ought to document this more than in the module itself since it's a great feature ! Would you care updating the documentation here https://github.com/ultrabug/py3status/tree/master/doc#group_modules with this feature please ?

tobes:
Sure

tobes:
I've added some documentation as requested.

I also uncovered a bug, that I fixed in this PR.  Some modules need to have the `output_modules` dict in core.py available to get content from other modules, but were running before it was built.  We now build the dict before starting the modules.

tobes:
I'm thinking of adding an open/close button to frame like the group module has as I think this is a useful space saving feature.

Would that be acceptable?  Alternatively that could be a separate module but it feels that pushing it into frame would be more sensible.   Then any module you wanted to be open/close-able you could just put in a frame.

ultrabug:
> I'm thinking of adding an open/close button to frame like the group module has as I think this is a useful space saving feature.

Sure

tobes:
Ok, this now has the button added and a way to change the separators between the modules.

Anyhow I think I'm finished on this now.

I also added `py3.is_my_event(event)` helper that is useful for containers like group, hope that is ok.

py3status/modules/frame.py
```diff
@@ -0,0 +1,141 @@
+# -*- coding: utf-8 -*-
+"""
+Group modules and treat as a single one.
```
ultrabug: "and display them as a single one" instead ?

ultrabug:
minor docstring remark and I merge :+1: thanks a lot !

tobes:
typo fixed

ultrabug:
awesome :D

PR #384, opened 2016-08-01
Py3 improvements: safe_format() and build_composite()

tobes:
Allow substituted formatings like mpd has eg `[[{artist} - ]{title}]|{file}`

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -484,7 +484,7 @@ Returns the time a given number of seconds into the future.
 Helpful for creating the `cached_until` value for the module
 output.
 
-__safe_format(format_string, data)__
+__safe_format(format_string, param_dict)__
 
 Parser for advanced formating.
 
@@ -512,7 +512,7 @@ This will show `artist - title` if artist is present,
 `title` if title but no artist,
 and `file` if file is present but not artist or title.
 
-__build_composite(format_string, data=None, composites=None)__
+__build_composite(format_string, param_dict=None, composites=None)__
 
 Build a composite output using a format string.
 
--- a/doc/README.md
+++ b/doc/README.md
@@ -484,13 +484,40 @@ Returns the time a given number of seconds into the future.
 Helpful for creating the `cached_until` value for the module
 output.
 
-__safe_format(format_string, param_dict)__
+__parse_format(format_string, data)__
+
+Parser for advanced formating.
+
+Square brackets `[]` can be used. The content of them will be removed
+from the output if there is no valid placeholder contained within.
+They can also be nested.
+
+A pipe (vertical bar) `|` can be used to divide sections the first
+valid section only will be shown in the output.
+
+A backslash `\` can be used to escape a character eg `\[` will show `[`
+in the output.
+
+`{<placeholder>}` will be converted, or removed if it is None or
+missing.  formating can also be applied to the placeholder eg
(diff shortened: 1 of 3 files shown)


ultrabug:
This is insane! I'm tempted to make it the only format parser, what do you think ?

tobes:
It would break some use configs if this is the default but would be cleaner
if all formats behaved the same. Your call I can make it the only one if
you want

This is insane! I'm tempted to make it the only format parser, what do you
think ?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ultrabug/py3status/pull/384#issuecomment-236597554>,
or mute
the thread

tobes:
I think we should just go for it and add a big note in the change log.

Also add a note in the main README

tobes:
@ultrabug shall I action the 'one true format'?

For this I'd swap all modules to use the new `py3.format()` rather than `string.format()` so we get the behaviour for all modules. It makes sense to just call it `format()` imho.

ultrabug:
> @ultrabug shall I action the 'one true format'?

Yes, I just would like you to explain a bit more about how easily it could break users' current config plz ?

tobes:
Say if I have 
```
volume_status {
    format = 'Volume [{percent}]'
}
```
then I'd loose the `[` so breakage is a little extreme as a word more the user might have their bar format change minorly.  Before `Volume [100]` After `Volume 100`

tobes:
I've decided to stick with `safe_format()` as a name so it is easy to grep.

I'd be for merging this if it ok and then doing a new cleanup branch to swap modules to using it

tobes:
Closed and reopened to get travis test to be re-run

tobes:
I've added `py3.build_composite()` it allows composites to be correctly applied onto format strings.  It uses the same logic as `safe_format` but also allows composites to be supplied eg `{'full_text': 'moo'}`

It will be useful for the mpris module etc

py3status/py3.py
```diff
@@ -237,6 +237,18 @@ def next_item(ignore_slash=False):
             if found:
                 # A placeholder exists so this section is valid
                 parts[len(parts) - 1][1] = True
+                # We need to work out which other sections depend on this one
+                # so that we can make sure they are shown.
+                # [ [ ] | [ ] ] [show [{placeholder}] | [ ] ]
+                fix_level = level
+                for i in range(len(parts) - 2, 0, -1):
+                    part_level = parts[i][2]
```
ritze: I've got an error with my mpris module:
> IndexError: list index out of range

tobes:
@ritze sorry I fixed this but forgot to `push -f` it should now be fixed

doc/README.md
```diff
@@ -486,11 +486,40 @@ output.
...
+`title` if title but no artist,
+and `file` if file is present but not artist or title.
+
+__build_composite(format_string, param_dict=None, composites=None)__
+
+Build a composite output using a format string.
+
+Takes a format_string and treats it the same way as `safe_format` but
+also takes a composites dict where each key/value is the name of the
+placeholderand either an output eg `{'full_text': 'something'}` or a
```
ultrabug: > placeholder and (also in the docstring of the function plz)

ultrabug:
closing as per #402

PR #385, opened 2016-08-02
Refactor events and module refresh

tobes:
This PR does the following

* Moves refreshing i3status into i3status.py as `I3status.refresh_i3status()`. Rather than being implemented in both in events.py and core.py.

* `Py3statusWrapper.refresh_modules()` can refresh all or named modules.  With named modules either an exact name or starting_with can be used.  `refresh_all` events now happen without needing to send a USR1 signal to py3status.  Events code changed to use this rather than re-implementing.

* `Events.dispatch_event()` Now handles the dispatching of events rather being done in `Events.run()` This allows events to come from other sources as well as i3bar.

The reasons for this PR are.

1. Cleans up some of the code and reduces repetition.

2. Allows for some cool future functionality I've been playing with to be introduced more easily.

I can split it into 3 PRs if that helps with reviewing.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -9,7 +9,7 @@ import time
 from collections import deque
 from json import dumps
 from signal import signal, SIGTERM, SIGUSR1, SIGTSTP, SIGCONT
-from subprocess import Popen, call
+from subprocess import Popen
 from threading import Event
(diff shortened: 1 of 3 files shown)


ultrabug:
Thanks, tested out ok, nice move !

PR #386, opened 2016-08-03
Add module quodlibet

faubiguy:
This adds a module to show the currently playing track in [Quod Libet](https://quodlibet.readthedocs.io/en/latest/index.html). It's mostly based on the existing deadbeef module, modified to use Quod Libet instead.

### DIFF
--- a/py3status/modules/quodlibet.py
+++ b/py3status/modules/quodlibet.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""
+Display track currently playing in Quod Libet.
+
+Configuration parameters:
+    cache_timeout: how often we refresh usage in seconds (default: 1s)
+    format: Output format (default: "{track}")
+    quodlibet_format: Quod Libet tag pattern to use for song. See
+    https://quodlibet.readthedocs.io/en/latest/guide/tags/patterns.html
+    for Quod Libet's documentation on them. (default: "<artist~title>")
+    quodlibet_directory: Directory of Quod Libet (default: "~/.quodlibet")
+
+Format of status string placeholders:
+    {track} Current song, formatted by quodlibet_format
+
+
+Requires:
+        quodlibet:
+"""
+from subprocess import check_output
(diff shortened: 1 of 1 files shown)


tobes:
Quodlibet can provide MPRIS support (optional) so the mpris module could be used here (when it is completed) as this reduces the need for a module for each individual player.

However if MPRIS support is not available then this module makes sense.

Unfortunately Quodlibet uses dbus-python rather than pydbus for dbus/MPRIS support which adds a extra dependency.

py3status/modules/quodlibet.py
```diff
@@ -0,0 +1,81 @@
...
+            'cached_until': time() + self.cache_timeout,
+            'full_text': ''
+        }
+        return response
+
+    # Return track currently playing in quodlibet
+    def get_status(self, i3s_output_list, i3s_config):
+        expanded_directory = os.path.expanduser(self.quodlibet_directory)
+        current_file = os.path.join(expanded_directory, 'current')
+        # Test if Quod Libet is running
```
tobes: Could you not just use `quodlibet --status`?

faubiguy:
`quodlibet --status` requires the output to be checked to determine if the app is running, which could break if the error message changes in a future version. It's also much slower (On my machine about 0.13 seconds vs 0.00001 for the file check).

I hadn't noticed the MPRIS support. I agree that the MPRIS module would make this one mostly redundant and should likely be used instead.

ultrabug:
Thanks for your proposal @faubiguy , since quodlibet supports MPRIS maybe you'd be interested in helping out and testing the module being worked on in #354 ?

Cheers

tobes:
@faubiguy the mpris module has now been merged into master.  Could you give it a try and see if it works with quodlibet.  If so can we close this PR?

faubiguy:
I can confirm that it does work well. You ahead and close this.

ultrabug:
Thanks again @faubiguy !

PR #388, opened 2016-08-04
fix issue with &gt; in modules docs

tobes:
This fixes the `<` being escaped to `&lt;` in the modules README.

Also proof of concept for pushing to your repo


ultrabug:
awesome :)

Issue #387, opened 2016-08-03
group module cannot be loaded


ikzk:
I am using the master branch. I simply put the example in the doc:

```
# Create a disks group that will show space on '/' and '/home'
# Change between disk modules every 30 seconds
...
order += "group disks"
...

group disks {
    cycle = 30
    format = "Disks: {output} {button}"
    click_mode = "button"

    disk "/" {
        format = "/ %avail"
    }

    disk "/home" {
        format = "/home %avail"
    }
}
```

to my configuration file (I also tried an empty configuration file). An error is reported:

```
py3status: Setup error (RuntimeError) parse_config.py line 544. Please try to fix this and reload i3wm (Mod+Shift+R)
```

Without group module or when the group module only contains a single module it works fine.

ultrabug:
Hi @ikzk, that's strange, would you mind showing the result of `py3status -v` and if possible your entire `i3status.conf` when it's failing ?

tobes:
Could you also post the error from the logs.  `py3status -c <config-path> -l <logfile-path>` Thanks

ikzk:
@ultrabug @tobes Sorry for replying late. 

The `py3status -v` shows:

```
py3status version 3.1rc0 (python 3.5.2)
```

The `i3status.conf` is

```
general {
        colors = true
        interval = 5
}

order += "group disks"

group disks {
    cycle = 30
    format = "Disks: {output} {button}"
    click_mode = "button"

    disk "/" {
        format = "/ %avail"
    }

    disk "/home" {
        format = "/home %avail"
    }
}
```

The error log is

```
2016-08-04 14:26:26 WARNING Setup error (RuntimeError) parse_config.py line 544.
2016-08-04 14:26:26 INFO Traceback
RuntimeError: OrderedDict mutated during iteration
  File "/home/alice/.pyenv/versions/anaconda3-4.0.0/lib/python3.5/site-packages/py3status/__init__.py", line 23, in main
    py3.setup()
  File "/home/alice/.pyenv/versions/anaconda3-4.0.0/lib/python3.5/site-packages/py3status/core.py", line 276, in setup
    config = process_config(config_path, self)
  File "/home/alice/.pyenv/versions/anaconda3-4.0.0/lib/python3.5/site-packages/py3status/parse_config.py", line 562, in process_config
    get_modules(config_info)
[...]

PR #389, opened 2016-08-05
warn user on empty config

tobes:
#387 includes a report of an issue with an empty config.  This prevents that error and notifies the user that no modules are present.  I'm not sure if the notification is the best.  really we are sayingthat there are none defined in `order +=...`

### DIFF
--- a/py3status/parse_config.py
+++ b/py3status/parse_config.py
@@ -595,7 +595,7 @@ def process_config(config_path, py3_wrapper=None):
     allowed_modules = module_names()
 
     # create config for modules in order
-    for name in config_info['order']:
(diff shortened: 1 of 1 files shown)


py3status/parse_config.py
```diff
@@ -626,6 +626,8 @@ def add_container_items(module_name):
             else:
                 config[name]['format'] = TZTIME_FORMAT
 
+    if not config['order']:
+        notify_user('There are no modules in your py3status config')
```
ultrabug: > Your configuration file does not list any module to be loaded with the "order" directive.

?

ultrabug:
Thanks !

PR #390, opened 2016-08-05
Bug Fix: ensure that last_output remains correct

tobes:
There was an issue after the sleeping dogs branch was merged that affected the output from a group.  A time module in the group would not be updated regularly.

This was because i3status time modules do not always create a fresh response and therefore they corrupted the module updated checks in module.py (the time would be updated and this would end up updating the last_output - so it looked like the output hadn't changed)


ultrabug:
I see, thanks !

PR #391, opened 2016-08-06
Bug fix: parse_config for python 3.5

tobes:
Force modules.items() to a list so that we are not iterating it when it might change.

This was not needed for python 3.4/2.7 but was a problem in 3.5


ultrabug:
thanks !

PR #393, opened 2016-08-07
Documentation update example modules improvements and composites

tobes:
Make module examples use `py3.format_safe()`

Also add initial documentation on composite responses


ultrabug:
:+1:

PR #394, opened 2016-08-07
Group module minor update

tobes:
Due to how modules were started the group module had to be sure that all modules were initialised before it could produce the correct output.  Since PR #276 all modules are now initialised before any are run so we no longer need this behaviour.


ultrabug:
:skull: code !

PR #395, opened 2016-08-08
Pass color to composite parts

tobes:
In PR #354 mpris module the output wants to be colored but currently there is no easy way to do this.

This PR allows composites to provide a color
```
{
    'cached_until': ...,
    'color': <color>,
    'composite': [
        {.....},
        {.....},
    ]
}
```

The color will be added to any of the parts of the composite that do not supply color themselves.
It needs documenting but since #393 adds composite documentation. I'll add it in either this PR or that one depending which gets merged first.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -205,6 +205,7 @@ class Module(Thread):
             align = self.module_options['align']
 
         # update all components
+        color = response.get('color')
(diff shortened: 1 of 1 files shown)


ultrabug:
Interesting one :+1:

PR #396, opened 2016-08-09
Bug Fix: py3 python2 unicode fixes for notify_user() and log()

tobes:
Both `py3.notify_user()` and `py3.log()` suffered issues when being sent a unicode string under python2.

This PR fixes the issue so that they are both much more forgiving about the strings they will accept.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -464,6 +464,11 @@ __update(module_name=None)__
 Update a module. If `module_name` is supplied the module of that
 name is updated. Otherwise the module calling is updated.
 
+__is_python_2()__
+
+True if the version of python being used is 2.x
+Can be helpful for fixing python 2 compatability issues
+
 __get_output(module_name)__
 
 Return the output of the named module. This will be a list.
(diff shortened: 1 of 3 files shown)


tobes:
Add `is_python2()` documentation

tobes:
@ultrabug I've added documentation as requested for the `is_python_2` I have changed it from being an attribute to a method for consistency with the other bits of the documentation

ultrabug:
Thanks a lot !

pydsigner:
Wouldn't it be cleaner as an attribute?

tobes:
> Wouldn't it be cleaner as an attribute?

Maybe, because the name `is_python_2` it feels like it should be a function to me.

If it is an attribute it should really be a `py3.CONSTANT` so if we had a nice name I'd be happy to change as currently there are no active users.  `py3.PYTHON_VERSION` might be ok.

I'm not too bothered either way but if you have strong feelings about this then we can do something.

ultrabug:
We've had the debate on IRC and quickly decided to go for the method. I still don't have a strong opinion about this.

So Daniel if this is something you really think we should have as an attribute then I agree with @tobes proposal on the constant.

PR #397, opened 2016-08-10
Anon modules: less module clashes in containers

tobes:
If we have the following i3status.conf.  

```
order += "static_string"
order += "group"

static_string {
    format = 'one'
}

group {
    static_string {
        format = 'two'
    }
    static_string {
        format = 'three'
    }
}
```
Then currently, only one static_string module is created even though three have been defined.  This PR instead treats this as three independent modules.  It does this by creating a unique name for modules defined inside a container.  This allow them to be distinguished between and so be initialised with the correct values.


ultrabug:
Congratz, this is a smart one :)

PR #398, opened 2016-08-11
Improved reporting of module import errors

tobes:
Module import errors were not very informative.

This PR adds better logging of exceptions.

It also adds `level` parameter to `report_exception()` to allow error level to be passed.


ultrabug:
reporting is believing :)

PR #399, opened 2016-08-11
Module init() method

tobes:
* Add `init()` support for modules.

* Fixes containers so that `self.items` is always available.  This makes things easier for initialising containers as we do not need to check if self.items is defined for the module it will just be an empty list.

* Adds documentation on module methods/responses (think of this as a bribe)

Due to how py3status modules are loaded their `__init__()` is called before any configuration parameters have been set.  It is often useful to run code just once when the module has been configured.

To achieve this I have often written code like this 
```
def __init__(self):
    self._initialized = False

def _init(self):
    ... run initialization code here ...
    self._initialized = True

def module_method(self):
    if not self._initialized:
        self._init()
    ... run module code here ...
```

Other modules just run code every time in their main method which only needed to be run once.

I think it would be nice if we allowed an `init()` function to be defined in a module that would be run when the config parameters had been initialised.  Similar to how `kill()` is called in a module when it is destroyed.

So a module would look like this
```
def init(self):
    ... run initialization code here ...

def module_method(self):
    ... run module code here ...
```

Adding this functionality will only effect existing modules if they used `init` as the name of one of their methods.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -28,6 +28,7 @@ class Module(Thread):
         self.cache_time = None
         self.click_events = False
         self.config = py3_wrapper.config
+        self.has_init = False
         self.has_kill = False
         self.i3status_thread = py3_wrapper.i3status_thread
         self.last_output = []
@@ -80,6 +81,15 @@ class Module(Thread):
             py_mod = getattr(py_mod, comp)
         class_inst = py_mod.Py3status()
         return class_inst
+
+    def start_module(self):
+        """
+        Start the module running.
+        """
+        # Call modules init() method if it has one
+        if self.has_init:
+            self.module_class.init()
+        self.start()
 
     def force_update(self):
         """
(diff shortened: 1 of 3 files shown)


tobes:
I'm happy with a better name. `post_config_hook` feels a bit unwieldy not sure what would be better.

`setup` 
`run_first`

don't feel like big improvements either.

ultrabug:
Just a quick heads up and excuses as I'm in the process of moving since the last few days. I should settle down and be back next week.

I see you've been very busy and active thank you !

ultrabug:
I agree with Daniel that init() is not so understandable and does not reflect the fact that it's gonna be executed after the module config setup.

I think the concept of hooks are widely used and understood and that even if it sounds long the `post_config_hook` has the advantage to stand for what it's for so I'd go for it too (+ it has the advantage of being duplicable to other parts of modules' life cycle).

tobes:
ok `post_config_hook` it is.

ultrabug:
Thank you both !

py3status/module.py
```diff
@@ -81,6 +82,15 @@ def load_from_namespace(module_name):
         class_inst = py_mod.Py3status()
         return class_inst
 
+    def start_module(self):
+        """
+        Start the module running.
+        """
+        # Call modules init() method if it has one
```
pydsigner: This comment is out of date but could be better anyways I think, it's obvious from the code which function is being called. It should say instead why we have the hook. Just my two cents.

PR #403, opened 2016-08-16
safe_format() get attributes from module

tobes:
* `safe_format()` now can get placeholders straight from the py3status module (ignores any methods)

* `\?` special escape char added for the future

* Works with module stand alone test

* docs updated to reflect changes


ultrabug:
Very elegant, thank you so much !

PR #405, opened 2016-08-18
Cleanup: update modules test code to new method

tobes:
Swap standalone test to new method for all modules


ultrabug:
Long and cool job thanks !

PR #406, opened 2016-08-19
Fixed exception if config-file is a symlink

Stautob:
Fixed the unknown encoding "binary" exception that occurred, if the config-file happened to be a symbolic link on linux.


tobes:
Looks good to me.  Thanks @Stautob

ultrabug:
Thanks a lot @Stautob !

PR #407, opened 2016-08-19
Update modules for colors patch 1

tobes:
I've updated the modules to work with the new color system.

* Switch from `is3_config['color_good']` -> `self.py3.COLOR_GOOD` etc
* Remove `i3s_config` etc from methods
* Document colors used

As it is quite a lot of changes I've split them up to be more manageable.

This PR is all the simple updates that should not be controversial.

### DIFF
--- a/py3status/modules/check_tcp.py
+++ b/py3status/modules/check_tcp.py
@@ -4,14 +4,16 @@ Display if a TCP port is available on the given host.
 
 Configuration parameters:
     cache_timeout: how often to run the check (default 10)
-    color_down: color of format_down
-    color_up: color of format_up
     format: what to display on the bar (default '{host}:{port} {state}')
     host: check if tcp port on host is up (default 'localhost')
     port: the tcp port (default 22)
 
 Format of status string placeholders:
     {state} port state ('DOWN' or 'UP')
+
+Color options:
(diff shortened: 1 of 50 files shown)


py3status/modules/check_tcp.py
```diff
@@ -24,24 +26,22 @@ class Py3status:
...
+    def check_tcp(self):
         response = {'cached_until': self.py3.time_in(self.cache_timeout)}
 
         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         result = sock.connect_ex((self.host, self.port))
 
         if result == 0:
             state = 'UP'
-            response['color'] = self.color_up or self.py3.COLOR_GOOD
+            response['color'] = self.COLOR_UP or self.py3.COLOR_GOOD
```
ultrabug: Confused here, isn't that supposed to be self.py3.COLOR_UP ?

Just like here : https://github.com/ultrabug/py3status/pull/411/files#diff-b153ef94eda7ed9e0be8d993f1daed17R332

tobes: This is why we code review ;)

Good catch, I'll fix it

ultrabug:
Thanks a lot @tobes this is a huge and reluctant job !

PR #408, opened 2016-08-19
Color update: kdeconnector

tobes:
Color update

This module was using `i3s_config['interval']` for its timeout.

This adds `cache_timeout` param like other modules


ultrabug:
:tophat: __

ultrabug:
We ought to remember to use `self.py3.time_in(self.cache_timeout)` for 3.2

PR #409, opened 2016-08-19
scratchpad-async: urgent

tobes:
scratchpad-async has a `color_urgent` parameter.

This patch removes it and sets `urgent` in the response so that i3bar will display it as urgent.  This feels like the correct approach.


ultrabug:
Thanks :)

PR #410, opened 2016-08-19
Color updates: mpd_status

tobes:
mpd_status has the following config param

> color: enable coloring output (default False)

This clashes with setting color in the general section of the config.

I have just removed this as an option. 

Maybe we should allow the user the option to prevent colors in modules as a universal option/by module as a future enhancement.


ultrabug:
:tophat: 

Totally fine imho.

PR #411, opened 2016-08-19
Color updates: battery_level

tobes:
This is it's own PR just because it is a large change to a single module.


ultrabug:
:tophat:

PR #412, opened 2016-08-19
Color updates: misc

tobes:
This updates

* bitcoin_price:

Uses `color_index` parameter - should be fine but will want fixing

* imap:
* spaceapi

Both have colors defined as `xxx_color` in config

Anyhow these will want updating in 3.2 I have added FIXME comments to them for 3.1 they will be half done.


ultrabug:
:tophat: 

Yes you're right we should not forget to mention it in the CHANGELOG, or at least 3.2 will say it

PR #413, opened 2016-08-20
Color updates: glpi

tobes:
This module used `i3s_config['colors']` to decide if colors should be enabled (undocumented).

This removes that logic.  As in #410 if we want this as a feature we should add it universally.


ultrabug:
:tophat:

PR #415, opened 2016-08-20
module imap: add base64 encoded password support

cemiarni:


tobes:
I can understand your desire to not have your password in plain text.  I think doing it at the module level is the wrong place.  We could do it directly in the config via something like

```
imap {
    user = 'moo'
    password:base64 = 'MTIzNA=='
}
```
It would then be available to encode any strings for all modules

That way we could add other encodings over time if there was a need. base64 encoding is an easy scheme to bypass, not that a really secure solution would be easy.  It does however provide a level of obfuscation that would have uses.

I've got a proof of concept patch that does this and doesn't add much complexity.  @ultrabug what are you thoughts?  Although the security is pretty poor it doesn't seem any worse than say how firefox seems to be keeping all my passwords.

ultrabug:
Thank you @cemiarni !

I'm not sure of the best way to actually have such configuration parameters strongly protected in the i3status.conf file but this is surely not by base64 encoding them imho.

Anyway, I think that @tobes' proposal for a global support of encoding/decoding is the way to g in #418 as it will benefit any module :+1:

ultrabug:
Thanks again @cemiarni, your PR has been the inspiration for a new feature of py3status !

https://github.com/ultrabug/py3status/tree/master/doc#obfuscation

PR #416, opened 2016-08-21
Fix exchange_rate module when single currency is used

yaqwsx:
When a single currency is used in the format string, this module failed on line
68 since Yahooapis does not return single currency as a one element list, but
the element (dictionary) itself.

Solution: check if the expected element is a list or not.


tobes:
@yaqwsx Thanks for this

@ultrabug looks good

ultrabug:
Thanks a lot @yaqwsx !

PR #418, opened 2016-08-22
Allow password encodings in config

tobes:
From #415 this adds universal base64 encoding via
```
module {
   value:base64 = '<encoded>'
}
```

@ultrabug if you are happy with this concept I'll update this PR with some documentation

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -112,6 +112,24 @@ imap {
 }
 ```
 
+As well as disguising the value in your configuration py3status also tries to
+prevent the information leaking, eg into log files or i3bar output.  Due to the
+workings of python it is easy for an attacker to counter these efforts.  When
+requested in an inappropriate place `***` will be used.
+
+If you just want py3status to prevent leakage you can use `:hide` to do
+this whilst keeping your configuration in plain text.
+
+```
+# Example of private configuration
+imap {
+    imap_server = 'imap.myprovider.com'
+    password:hide = 'coconut'
+    user = 'mylogin'
+}
+```
+
+
 #### <a name="configuring_color"></a>Configuring colors
 
 Since version 3.1 py3status allows greater color configuration.
--- a/doc/README.md
+++ b/doc/README.md
@@ -95,11 +95,29 @@ imap {
 
 __New in version 3.1__
 
-In the previous example configuration the users password is in plain text.  Users may
-want to make it less easy to read. Py3status allows strings to be base64
-encoded.
+Py3status allows you to hide individual configuration parameters so that they
+do not leak into log files, user notifications or to the i3bar. Additionally
+they allow you to obfuscate configuration parameters using base64 encoding.
 
-__Base64 encoding is very simple and should not be considered secure in any way.__
+To do this you need to add an obfuscation option to the configuration
+parameter. Obfuscation options are added by adding `:hide` or `:base64` to the
+name of the parameters.
+__Obfuscation is only available for string parameters.__
+
+Example:
+
+```
+# normal_parameter will be shown in log files etc as 'some value'
+# obfuscated_parameter will be shown in log files etc as '***'
+module {
+    normal_parameter: 'some value'
+    obfuscated_parameter:hide 'some value'
+}
+```
+
+In the previous example configuration the users password is in plain text.
+Users may want to make it less easy to read. Py3status allows strings to be
+base64 encoded.
 
 To use an encoded string add `:base64` to the name of the parameter.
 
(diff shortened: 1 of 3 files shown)


py3status/parse_config.py
```diff
@@ -441,6 +442,12 @@ def parse(self, dictionary=None, end_token=None):
                     dictionary[name] = value
                 # assignment of value
                 elif t_value == '=':
+                    # if we have a colon in the name of a setting then it
+                    # indicates that it has been encoded.
+                    if ':' in name:
+                        (name, scheme) = name.split(':')
+                        if scheme == 'base64':
```
ultrabug: if this ought to be extendable, I'd rather have this done in a dedicated method please

py3status/parse_config.py
```diff
@@ -100,7 +101,7 @@ class ConfigParser:
         '|(?P<literal>'
         r'("(?:[^"\\]|\\.)*")'  # double quoted string
         r"|('(?:[^'\\]|\\.)*')"  # single quoted string
-        '|([a-z_][a-z0-9_]*)'  # token
+        '|([a-z_][a-z0-9_]*(:[a-z0-9_]+)?)'  # token
```
ultrabug: I think `a-z0-9_` equivalent is `\w`

py3status/parse_config.py
```diff
@@ -4,6 +4,7 @@
 import glob
 import os
 import imp
+import base64
```
ultrabug: bad kitty

ultrabug:
I pretty like the concept of generalization here yes tho I'm still not convinced that base64 encoding anything increases the security of #415

tobes:
> I'm still not convinced that base64 encoding anything increases the security

I was off the same opinion but it is still useful.  It means I can look at you config and not just read your password by accident.  But it is completely insecure

We could do something proper secure and have py3status prompt for a pass phrase and not start affected modules till it has one.  But I'm not sure that it is worth the effort at the moment.

py3status/parse_config.py
tobes: sometimes explicit is good eg is `-` in `\w`

ultrabug: true

tobes:
I've moved the processing into it's own method.  It means we can now do some insane stuff in our configs, not that it should be supported.

```
static_string test {
    format:base64 = 'MTIzNA=='
    format += '=='
    format:base64 += 'MTIzNA=='
}
```

doc/README.md
```diff
@@ -7,6 +7,7 @@ py3status documentation
 * [Available modules](#available_modules)
 * [Ordering modules](#ordering_modules)
 * [Configuring modules](#configuring_modules)
+* [Obfuscation](#obfuscation)
```
ultrabug: **Configuration obfuscation** maybe more understandable ?

doc/README.md
```diff
@@ -83,14 +84,34 @@ imap {
...
     user = 'mylogin'
     on_click 1 = "exec thunderbird"
 }
 ```
 
+#### <a name="obfuscation"></a>Obfuscation
+
+__New in version 3.1__
+
+In the previous example configure the users password is in plain text.  Users may
```
ultrabug: ... example configuration ...

tobes:
Made those doc fixes

tobes:
So I was having a think about the whole privacy stuff and decided that it was an interesting topic.

py3status would easily leak any information that you encode eg if you use logging, notifications, put it in your format string.  So I've updated this PR to stop this leakage.  It is not an easy problem as we need to ensure that modules work, especially if they use external libs eg requests.

This implementation allows the value of 'private' config options only to be known to modules with the same name as the config option is given for.  And when used outside this `***` is used as the value.

All unicode methods should be available and work both when valid and not.

The whole thing is academic as all I need is to add a module that monkey patches this out and I can get the real values.  So this is more about keeping things semi-private and a bit of fun.

As well as `:base64` I have added `:hide` that is plain text in the config but will not leak to logs, i3bar etc

tobes:
A few more updates.

* We are more resistant to modules trying to change the Private classes

* Only allow this for py3status modules

* Only allow for strings

* Check the 'scheme' exists

ultrabug:
I'm tempted to use `:private` instead of `:hide`

Also, since :base64 et al are used to obfuscate, why don't we hide them too by default ?

ultrabug:
> I'm tempted to use :private instead of :hide

dunno, :hide seems more explicit after some more thoughts

tobes:
> Also, since :base64 et al are used to obfuscate, why don't we hide them too by default ?

We do the :hide is just so I can have my password hidden in the logs all :xxx are hidden in this way.

Do I need to make the docs clearer?

tobes:
Hopefully clearer documentation

ultrabug:
Thanks this is a neat addition :)

PR #420, opened 2016-08-22
Improved formatter

tobes:
So `_safe_format()` had a few issues where it didn't behave as i would have liked.

eg `'{empty}|nothing here'` would return `''` rather than `'nothing here'`

Additionally the code was starting to become quite hard to read, possibly buggy, and prevented expansion of new ideas eg block commands things like `[\?max_length=10 {file-name}]` `[\?if=param Hello]`

Anyhow I have written a formatter it uses a simple parser similar to the config parser.  It also makes py3.py much shorter and easier to read.

Plus it has tests `python formatter.py`

I think seeing as we are going to shift modules to safe_format() that having a solid implementation is an improvement.

### DIFF
--- a/py3status/formatter.py
+++ b/py3status/formatter.py
@@ -1,0 +1,632 @@
+# -*- coding: utf-8 -*-
+import re
+import sys
+
+try:
+    from urllib.parse import parse_qsl
+except ImportError:
+    from urlparse import parse_qsl
+
+
+class Composite:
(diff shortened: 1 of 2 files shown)


ultrabug:
That's a huge one :astonished: thanks !

Tested OK !

Note: those tests are awesome, but we shall move them to a proper py.test way later

PR #422, opened 2016-08-26
Imporvements to py3.time_in()

tobes:
this is the py3.time_in() magic

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -179,12 +179,54 @@ class Py3:
             my_info = self._get_module_info(self._module.module_full_name)
             my_info['content_function'] = content_function
 
-    def time_in(self, seconds=0):
+    def time_in(self, seconds=None, sync_to=None, offset=0):
         """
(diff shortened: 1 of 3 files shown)


py3status/py3.py
```diff
@@ -179,12 +179,54 @@ def register_content_function(self, content_function):
...
+            # If we have a sync_to then seconds can be 0
+            if sync_to and sync_to > 0:
+                seconds = 0
+            else:
+                try:
+                    # use py3status modules cache_timeout
+                    seconds = self._py3status_module.cache_timeout
+                except AttributeError:
+                    # use default cache_timeout
+                    seconds = self._module.config['cache_timeout']
```
ultrabug: here self._module represents the global py3status config ? (meaning the -t command line parameter) ?

tobes: Yes this is correct.  We could use general {interval=x} but that feels wrong to me

ultrabug: And it would indeed be wrong since it's an i3status only config.

PR #423, opened 2016-08-26
formatter updates

tobes:
So this does the following

* ~~Bug fix: we were passing `self._module` (Module) not `self._py3status_module` (Py3status) to the formatter.~~ 

* Auto select `self.format` for format_string if not passed by a module

* Change calling parameters so that we can deal with `()`, `(format, dict)`, `(dict)` etc

* auto-magically grab the `full_text` if missing from the response dict

If this is ok then I'll update the docs etc in this PR

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -439,8 +439,13 @@ class Module(Thread):
                     else:
                         # validate the response
                         if 'full_text' not in result:
-                            err = 'missing "full_text" key in response'
-                            raise KeyError(err)
+                            try:
+                                # try to set the full_text automatically
+                                full_text = self.module_class.py3.safe_format()
+                                result['full_text'] = full_text
+                            except:
+                                err = 'missing "full_text" key in response'
+                                raise KeyError(err)
                         # set universal module options in result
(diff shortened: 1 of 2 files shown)


ultrabug:
`full_text` is a specification of the i3bar protocol so I think this should be mandatory in the response dict.

tobes:
> full_text is a specification of the i3bar protocol so I think this should be mandatory in the response dict.

yes I'm just saying that if it is not provided by the module we see if we can generate it via `Py3status.py3.safe_format()`

ultrabug:
Ok but I also think that we should not guess the format tbh, this being mandatory to safe_format() looks sane to me

tobes:
> Ok but I also think that we should not guess the format tbh, this being mandatory to safe_format() looks sane to me

Well if we do that then this whole PR is void.

I think it would have been nice but maybe explicit is best

tobes:
Or did you just mean for a missing 'full_text' in the response.  If so reopen this and I'll remove that bit and fix the merge issue

PR #427, opened 2016-08-27
Use format placeholders: in docstrings

tobes:
Change `Format of status string placeholders:` to `Format placeholders:`


tobes:
@ultrabug This shouldn't break anything and will just cause merge conflicts if it goes stale

ultrabug:
Thanks, makes sense

PR #428, opened 2016-08-27
modules to use safe_format()

tobes:
Get modules to use `safe_format()`


ultrabug:
huge one for 3.2, thank you !

PR #430, opened 2016-08-27
keyboard_layout fix docstring

tobes:
Fix for modules/README.md rendering


tobes:
#434 is a far better fix this was just a temporary botch not a proper solution

PR #432, opened 2016-08-28
kdeconnect device is paired error

tobes:
When connecting to a reachable but unpaired device d-bus device['isPaired'] is actually a method so should be called as the function definition is treated as a truth.  This is quite hard to hit.

@ritze can you confirm this is good for your setup, I've done quite a bit of playing and this seems robust.

### DIFF
--- a/py3status/modules/kdeconnector.py
+++ b/py3status/modules/kdeconnector.py
@@ -21,7 +21,7 @@ Format of status string placeholders:
     {charge} the battery charge
     {name} name of the device
     {notif_size} number of notifications
-    {notif_status} shows if a notification is available or nor
+    {notif_status} shows if a notification is available or not
 
 Color options:
     color_bad: Device unknown, unavailable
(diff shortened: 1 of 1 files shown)


tobes:
@ultrabug I think we can just go with this as @ritze hasn't got back and it is a real bug

ritze:
@tobes both are available (method and property). So it makes no difference how you call `isPaired`.

> When connecting to a reachable but unpaired device d-bus device['isPaired'] is actually a method so should be called as the function definition is treated as a truth. This is quite hard to hit.

tobes:
> both are available (method and property).

@ritze this is not true.  If you use something like d-feet you can check sometimes the property does not exist the method does though

you have to completely unpair your device etc to see it

ritze:
Ok, I just disconnected the PC from my device…

> you have to completely unpair your device etc to see it

tobes:
@ritze the main thing is that it does not break things for you as without this kdeconnect gives me an error.

Incidentally, do you know a way to prevent kdeconnect from sending messages to dbus notification.  I get them (via dunst) and it would be nice if your module handled them completely (eg sent any user notifications) and could for example silence the notifications.

Nice module :)

@ultrabug I think this is good to merge

ritze:
@tobes I installed the kde packages so you can open the setting window with `kcmshell5 kcm_kdeconnect` and activate or deactivate each module like on your smartphone.

I wrote a script to show me the notification. This script is called if I click on the kdeconnect module in my py3status bar. ;-)

Maybe we can use the code to expand the module?

ultrabug:
Thanks a lot guys !

PR #433, opened 2016-08-28
py3status docstrings: Allow checking docstring for specific modules

tobes:
When using `py3status docstring check` allow modules to be specified eg  `py3status docstring check xrandr`


ultrabug:
nice, we need this kind of QA, thanks !

PR #435, opened 2016-08-28
Cpu temp

guiniol:
This PR adds a module to show the CPU temperature using lm_sensors. The zone to show is configurable.


guiniol:
I'll work on sysdata and close this and #440 and #443.

PR #436, opened 2016-08-28
Disk free space

guiniol:
This PR adds a module to show the remaining free space on a given disk. The disk is chosen by filesystem path.

### DIFF
--- a/py3status/modules/disk_free_space.py
+++ b/py3status/modules/disk_free_space.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+"""
+Display the amount of free space of given disk.
+This modules requires the psutil python module.
+
+Configuration parameters:
+    path: filesystem path to check
+        (default: "/")
+    threshold_degraded / threshold_bad: thresholds for color change
+"""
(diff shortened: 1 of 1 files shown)


pydsigner:
How about using https://docs.python.org/3.3/library/os.html#os.statvfs to avoid the library dependency?

guiniol:
I'll look into it. I do want to remove the psutil dependency where possible.

py3status/modules/disk_free_space.py
```diff
@@ -0,0 +1,55 @@
...
+    format: format string
+        (default: "{space:.1f}G")
+
+Format string parameters:
+    {space} space left (in GB)
+    {path} path checked
+"""
+
+# import your useful libs here
+import psutil
```
tobes: Is there a reason you are using psutil?  I've got a disk space module that uses `df` to get the info.

tobes: opps just saw @pydsigner's comment

guiniol:
Closing this until #476 gets merged.

PR #437, opened 2016-08-28
Remove additional separator at the end of frames

guiniol:
This PR removes the additional separator that were added at the end of frames. This is only if "separator" is set. If this was an expected behaviour, I'll add an option to remove it.


tobes:
@guiniol I think you are right about this

@ultrabug please pull this

ultrabug:
Thanks @guiniol !

guiniol:
Happy to help ^^

PR #438, opened 2016-08-28
Net iplist

guiniol:
This PR adds a module to display the list of IP of the computer. IPs can be blacklisted by prefix. By default the loopback is blacklisted.

### DIFF
--- a/py3status/modules/net_iplist.py
+++ b/py3status/modules/net_iplist.py
@@ -110,7 +110,7 @@ class Py3status:
         ip_re = re.compile(r'\s+inet (\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3})/')
         ip6_re = re.compile(r'\s+inet6 ([\da-f:]+)/')
 
-        txt = check_output(['ip', 'address', 'show']).decode('utf-8').split('\n')
+        txt = check_output(['ip', 'address', 'show']).decode('utf-8').splitlines()
         len_txt = len(txt)
 
         idx = 0
--- a/py3status/modules/net_iplist.py
+++ b/py3status/modules/net_iplist.py
@@ -67,6 +67,11 @@ class Py3status:
     ip_sep = ','
     remove_empty = True
 
+    def __init__(self):
+        self.iface_re = re.compile(r'\d+: (\w+):')
+        self.ip_re = re.compile(r'\s+inet (\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3})/')
+        self.ip6_re = re.compile(r'\s+inet6 ([\da-f:]+)/')
+
     def ip_list(self):
         response = {
             'cached_until': self.py3.time_in(seconds=self.cache_timeout),
@@ -108,17 +113,13 @@ class Py3status:
         return response
 
     def _get_data(self):
-        iface_re = re.compile(r'\d+: (\w+):')
-        ip_re = re.compile(r'\s+inet (\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3})/')
-        ip6_re = re.compile(r'\s+inet6 ([\da-f:]+)/')
-
         txt = check_output(['ip', 'address', 'show']).decode('utf-8').splitlines()
         len_txt = len(txt)
 
         idx = 0
         data = {}
         while idx < len_txt:
-            iface = iface_re.match(txt[idx])
+            iface = self.iface_re.match(txt[idx])
             idx += 1
 
             if iface is None:
@@ -126,13 +127,13 @@ class Py3status:
 
             iface = iface.groups()[0]
             iface_data = {}
-            while idx < len_txt and not iface_re.match(txt[idx]):
-                ip = ip_re.match(txt[idx])
+            while idx < len_txt and not self.iface_re.match(txt[idx]):
+                ip = self.ip_re.match(txt[idx])
                 if ip:
                     iface_data.setdefault('ip', []).append(ip.groups()[0])
                     idx += 1
                     continue
-                ip6 = ip6_re.match(txt[idx])
+                ip6 = self.ip6_re.match(txt[idx])
                 if ip6:
                     iface_data.setdefault('ip6', []).append(ip6.groups()[0])
                     idx += 1
--- a/py3status/modules/net_iplist.py
+++ b/py3status/modules/net_iplist.py
@@ -1,0 +1,92 @@
+# -*- coding: utf-8 -*-
+"""
+Display the list of current IPs. This excludes loopback IPs and displays
+"no connection" if there is no connection.
+
+Configuration parameters
+    ignore: list of IPs to ignore. Can use shell style wildcards.
+        (default: ['127.*'])
+    no_connection: string to display if there are no non-ignored IPs
+        (default: 'no connection')
+    separator: string to use between IPs.
+        (default: ' ')
+"""
+
+# import your useful libs here
+import socket
+import struct
+import fcntl
+import array
+from fnmatch import fnmatch
+
+
+class Py3status:
+    cache_timeout = 30
+    separator = ' '
+    no_connection = 'no connection'
+    ignore = ['127.*']
+
+    def __init__(self):
+        pass
+
+    def ip_list(self):
+        response = {
+            'cached_until': self.py3.time_in(seconds=self.cache_timeout),
+            'full_text': ''
+        }
+
+        ip = []
+        ifaces = self._list_ifaces()
+        for iface in ifaces:
+            addr = self._get_ip(iface)
+            add = True
+            for ignore in self.ignore:
+                if fnmatch(addr, ignore):
+                    add = False
+                    break
+            if add:
+                ip.append(addr)
+        if len(ip) == 0:
+            response['full_text'] = self.no_connection
+            response['color'] = self.py3.COLOR_BAD
+        else:
+            response['full_text'] = self.separator.join(ip)
+            response['color'] = self.py3.COLOR_GOOD
+
+        return response
+
+    def _list_ifaces(self):
+        SIOCGIFCONF = 0x8912
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        max_possible = 128  # arbitrary. raise if needed.
+        data = max_possible * 32
+        names = array.array('B', [0]) * data
+        outbytes = struct.unpack('iL', fcntl.ioctl(sockfd, SIOCGIFCONF,
+                                 struct.pack('iL', data,
+                                             names.buffer_info()[0])))[0]
+        namestr = names.tostring()
+        lst = []
+        for i in range(0, outbytes, 40):
+            name = namestr[i:i+16].split(b'\x00', 1)[0]
+            lst.append(name)
+        return lst
+
+    def _get_ip(self, iface):
+        SIOCGIFADDR = 0x8915
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifreq = struct.pack('16sH14s', iface, socket.AF_INET, b'\x00'*14)
+        try:
+            res = fcntl.ioctl(sockfd, SIOCGIFADDR, ifreq)
+        except:
+            return None
+        ip = struct.unpack('16sH2x4s8x', res)[2]
+        return socket.inet_ntoa(ip)
+
+if __name__ == "__main__":
+    """
+    Test this module by calling it directly.
+    """
+    from py3status.module_test import module_test
+    module_test(Py3status)
--- a/py3status/modules/net_iplist.py
+++ b/py3status/modules/net_iplist.py
@@ -1,16 +1,21 @@
 # -*- coding: utf-8 -*-
 """
-Display the list of current IPs. This excludes loopback IPs and displays
-"no connection" if there is no connection.
+Display the list of current IPs.
 
-Configuration parameters
+This can exclude specified IPs and displays "no connection" if there are no IPs
+to display.
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds.
+        (default 30)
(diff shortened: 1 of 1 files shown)


py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,90 @@
...
+import array
+
+
+class Py3status:
+    def __init__(self):
+        self.cache_timeout = 30
+        self.separator = ' '
+        self.no_connection = 'no connection'
+        self.ignore_prefix = ['127.']
+        pass
```
tobes: These should be defined at the class level rather than in __init__()

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,90 @@
...
+Display the list of current IPs. This excludes loopback IPs and displays
+"no connection" if there is no connection.
+
+Configuration parameters
+    separator: string to use between IPs.
+        (default: ' ')
+    no_connection: string to display if there are no non-ignored IPs
+        (default: 'no connection')
+    ignore_prefix: list of prefixes to ignore.
+        (default: ['127.'])
```
tobes: Hmm,  think I'd be inclined to do ignoring as reg expressions (althought all the `.` make it a pain)  maybe use wild cards so `ignore = ['127.*', '192.168.*.100']` type thing to be honest I've not got too many Ips on my machines to worry about this too much.

guiniol:
I'll see what I can do. Adding regexp is easy, but if we want users to type '127.*' and it should be understood as '127\..*', that's something else.

guiniol:
Added support for wildcards. The `fnmatch` module already does it.

tobes:
Ok, I have the following questions

1) What is the purpose of this module?  Is it purely about showing my local IP addresses?

2) Is there other information that I might later want eg which interface is connected to each IP address?

3) Is this only for IPv4?  what about IPv6?

4) Is there a reason that you are not using something like `ifconfig` or `ip addr show`?

5) Is there a reason for no `format` config parameter?  This is something that we are trying to get all modules to have for consistency etc.

guiniol:
> 1) What is the purpose of this module? Is it purely about showing my local IP addresses?

Yes, that's the idea.
 
> 2) Is there other information that I might later want eg which interface is connected to each IP address?

That's actually a good idea since I already have the information. I'll add that.

> 3) Is this only for IPv4? what about IPv6?

IPv4 only for now, I'll look into IPv6

> 4) Is there a reason that you are not using something like ifconfig or ip addr show?

Because I can't be sure ifconfig or ip addr will be installed. This is the only reliable way to find it that I could find. It's what this does too: https://pypi.python.org/pypi/netifaces

> 5) Is there a reason for no format config parameter? This is something that we are trying to get all modules to have for consistency etc.

I have an idea but the format is a bit weird:
1. `format` has the `{iface_list}` placeholder (I don't remember what the actual word for "placeholder" is)
2. `iface_list` is of the form `iface_format``iface_separator``iface_format``iface_separator``iface_format`...
3. `iface_format` is a sub format with the `{iface}` and `{ip_list}` placeholders
4. `ip_list is of the form `ip_format``ip_separator``ip_format``ip_separator``ip_format`...
5. `ip_format` is a sub format with the `{ip}` placeholder
6. `*_separators` are just strings

It's a bit complex, but I don't see how to do it simpler than that.

guiniol:
@tobes: new answers now that I rewrote the module

> 1) What is the purpose of this module? Is it purely about showing my local IP addresses?

Still yes.

> 2) Is there other information that I might later want eg which interface is connected to each IP address?

You can do this now (actually, it does it by default)

> 3) Is this only for IPv4? what about IPv6?

Handles both now, and it's easy to choose which to show.

> 4) Is there a reason that you are not using something like ifconfig or ip addr show?

I switch to using `ip addr`. It was the easiest way to handle multiple ipv4 addresses on one interface.

> 5) Is there a reason for no format config parameter? This is something that we are trying to get all modules to have for consistency etc.

Added. I'm not sure if there is a better solution than what I did (`format` + `format_iface`) but I don't really see how else to handle the list of interfaces/IPs in `format`.

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+            for ip in ips.get('ip6', []):
+                if self._check_blacklist(ip):
+                    connection = True
+                    txt_ip6 += ip + self.ip_sep
+            txt_ip6 = txt_ip6[:-len(self.ip_sep)]
+            iface_txt += self.py3.safe_format(self.format_iface, {'iface': iface,
+                                                                  'ip': txt_ip,
+                                                                  'ip6': txt_ip6})
+            iface_txt += self.iface_sep
+            iface_txt = iface_txt[:-len(self.iface_sep)]
```
tobes: This just removes the text you added, I think you meant to do it outside the loop.

Anyhow personally I'd put all the bits in a list and the join them as that way you don't need to worry about the extra bit on the end

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+        len_txt = len(txt)
+
+        idx = 0
+        data = {}
+        while idx < len_txt:
+            iface = iface_re.match(txt[idx])
+            idx += 1
+
+            if iface is None:
+                continue
```
tobes: why not do the blacklisting here?  better not collect the data than to have to remove it imho

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+            response['full_text'] = self.py3.safe_format(self.format,
+                                                         {'format_iface': iface_txt})
+            response['color'] = self.py3.COLOR_GOOD
+
+        return response
+
+    def _get_data(self):
+        iface_re = re.compile(r'\d+: (\w+):')
+        ip_re = re.compile(r'\s+inet (\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3})/')
+        ip6_re = re.compile(r'\s+inet6 ([\da-f:]+)/')
```
tobes: Probably not a huge performance deal but if you compiled these in `__init__()` then you would only need to do it once.

Did you consider using groups?  eg `(?P<name>....)`

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+
+class Py3status:
+    cache_timeout = 30
+    format = 'Network: {format_iface}'
+    format_iface = '{iface}: v4{{{ip}}} v6{{{ip6}}}'
+    iface_blacklist = ['lo']
+    iface_sep = ' '
+    ip_blacklist = []
+    ip_sep = ','
+    no_connection = 'no connection'
```
tobes: `format_no_connection` would be more to the py3status naming conventions

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+            response['color'] = self.py3.COLOR_GOOD
+
+        return response
+
+    def _get_data(self):
+        iface_re = re.compile(r'\d+: (\w+):')
+        ip_re = re.compile(r'\s+inet (\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3})/')
+        ip6_re = re.compile(r'\s+inet6 ([\da-f:]+)/')
+
+        txt = check_output(['ip', 'address', 'show']).decode('utf-8').split('\n')
```
tobes: `splitlines()` is nicer than `split('\n')` because it understands newlines better but on unix output not such an issue

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+        len_txt = len(txt)
+
+        idx = 0
+        data = {}
+        while idx < len_txt:
```
tobes: why not `for line in txt:`  we know interface name lines don't start with `' '` so this whole loop can probably be reduced in complexity.

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+                    idx += 1
+                    continue
+                idx += 1
+
+        to_del = set()
+        for iface, ips in data.items():
+            if ips == {}:
+                to_del.add(iface)
+        for iface in to_del:
+            del data[iface]
```
tobes: again if we didn't add this data we wouldn't have to remove it later

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+        (default 30)
+    format: format of the output.
+        (default 'Network: {format_iface}')
+    format_iface: format string for the list of IPs of each interface.
+        (default '{iface}: v4{{{ip}}} v6{{{ip6}}}')
+    iface_blacklist: list of interfaces to ignore.
+        (default ['lo'])
+    iface_sep: string to write between interfaces.
+        (default ' ')
+    ip_blacklist: list of IPs to ignore. Accepts shell-style wildcards.
```
tobes: an example would be good, either here or as an example at the end (see https://github.com/ultrabug/py3status/blob/master/doc/README.md#docstring)

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,152 @@
...
+    {iface} name of the interface.
+    {ip} list of IPv4 of the interface.
+    {ip6} list of IPv6 of the interface.
+
+Color options:
+    color_bad: no IPs to show
+    color_good: IPs to show
+
+Requires:
+    iproute2
```
tobes: i'd change this to `ip: utility found in iproute2 package`

py3status/modules/net_iplist.py
guiniol: Sure.

py3status/modules/net_iplist.py
guiniol: Ok.

py3status/modules/net_iplist.py
guiniol: > Anyhow personally I'd put all the bits in a list and the join them as that way you don't need to worry about the extra bit on the end

Good thinking.

py3status/modules/net_iplist.py
guiniol: I didn't know about `splitlines`

py3status/modules/net_iplist.py
guiniol: I'm not sure that's possible. I explicitly move the index because I don't know how many lines there are per interface. For now, the code really shows that we look for an interface, and once we have one, we parse until the next interface. I could rewrite it with a variable to remember the current interface. I'll see if it makes it cleaner/more readable.

guiniol:
@tobes Done!

For the `for line in txt:` loop, it (invisibly) assumes that we will match an iface before matching IPs. Otherwise, it'll error.

For the blacklisting, like I said, I'd rather keep it in the main loop.

I also added the option to keep interfaces without IPs, and interfaces can be blacklisted with wildcards too (useful if you want to ignore all `tap*` devices or something).

tobes:
@guiniol looking good, I think I'd be inclined to change the `ip` references to `ip4` just to make it slightly clearer.

I find this useful to reduce the number of commits, just update my last one
```
git add -p  # add any changes
git commit --amend   # you can change the commit message here too
git push -f  # need to force the updates
```

I'll do a proper review tomorrow when I have some time.

guiniol:
> @guiniol looking good, I think I'd be inclined to change the ip references to ip4 just to make it slightly clearer.

Thanks ^^
Actually, `ip addr` shows `inet` and `inet6` which is why I went with `ip` and `ip6`.

> I find this useful to reduce the number of commits, just update my last one
> 
> git add -p  # add any changes
> git commit --amend   # you can change the commit message here too
> git push -f  # need to force the updates

I actually like lots of small commits, but I do use `--amend` when I have something to change. If you're referencing c52c6ac, It must be merged in ca58e69

> I'll do a proper review tomorrow when I have some time.

Sure. Take your time.

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,151 @@
...
+
+import re
+from subprocess import check_output
+from fnmatch import fnmatch
+
+
+class Py3status:
+    cache_timeout = 30
+    format = 'Network: {format_iface}'
+    format_iface = '{iface}: v4{{{ip}}} v6{{{ip6}}}'
```
tobes: how about `'{iface}:[ {ip}][ {ip6}]'`

One of my interfaces has no IPv6 and it looks funny being blank.  Also I find the `{` an unnecessary distraction.

tobes:
> Actually, ip addr shows inet and inet6 which is why I went with ip and ip6.

I think that is because IPv6 was not really in use when ip was first written

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,151 @@
+# -*- coding: utf-8 -*-
+"""
+Display the list of current IPs.
+
+This can exclude specified IPs and displays "no connection" if there are no IPs
+to display.
```
guiniol: Done.

py3status/modules/net_iplist.py
```diff
@@ -0,0 +1,151 @@
...
+    format_no_ip = 'no connection'
+    iface_blacklist = ['lo']
+    iface_sep = ' '
+    ip_blacklist = []
+    ip_sep = ','
+    remove_empty = True
+
+    def __init__(self):
+        self.iface_re = re.compile(r'\d+: (?P<iface>\w+):')
+        self.ip_re = re.compile(r'\s+inet (?P<ip>\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3})/')
```
guiniol: Don't you mean `r'\s+inet (?P<ip>[0-9\.]+)/'` :þ

guiniol:
> I think that is because IPv6 was not really in use when ip was first written

I was on the fence for that. Done.

ultrabug:
It's about time I start merging @guiniol contributions, can we have an updated labeling on ready for merge ones please ?

How do you feel about them @guiniol ?

guiniol:
I think this one is done. @tobes?

ultrabug:
Looks very nice, thanks !

PR #439, opened 2016-08-28
Net rate add colors

guiniol:
This PR adds colors to the net_rate module, but doesn't change the default behaviour. The color can be set depending on the upload/download or total transfer rate.

### DIFF
--- a/py3status/modules/net_rate.py
+++ b/py3status/modules/net_rate.py
@@ -11,6 +11,8 @@ Configuration parameters:
     interfaces: comma separated list of interfaces to track
     interfaces_blacklist: comma separated list of interfaces to ignore
     precision: amount of numbers after dot
+    color: colorise output according to the value of 'up|down|total'
+    threshold_bad / threshold_degraded: thresholds for color change
 
 Format of status string placeholders:
     {down} download rate
@@ -50,6 +52,9 @@ class Py3status:
     interfaces = ''
     interfaces_blacklist = 'lo'
     precision = 1
+    color = False
+    threshold_bad = 1
+    threshold_degraded = 1024
 
     def __init__(self, *args, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


py3status/modules/net_rate.py
```diff
@@ -59,6 +48,19 @@ def __init__(self, *args, **kwargs):
...
+        self.format = "{interface}: {total}"
+        self.format_no_connection = ''
+        self.hide_if_zero = False
+        self.interfaces = ''
+        self.interfaces_blacklist = 'lo'
+        self.precision = 1
+        self.color = False
+        self.threshold_bad = 1
+        self.threshold_degraded = 1024
+
```
tobes: The defaults should stay where they were as it makes things easier to read and they are class defaults rather than instance ones.

py3status/modules/net_rate.py
```diff
@@ -59,6 +48,19 @@ def __init__(self, *args, **kwargs):
...
+        self.all_interfaces = True
+        self.cache_timeout = 2
+        self.devfile = '/proc/net/dev'
+        self.format = "{interface}: {total}"
+        self.format_no_connection = ''
+        self.hide_if_zero = False
+        self.interfaces = ''
+        self.interfaces_blacklist = 'lo'
+        self.precision = 1
+        self.color = False
```
tobes: Color is sort of a reserved config parameter as it is used to color module output.

I'd be inclined to just make things colored or change the option.  I think this is a topic of debate.  Personally I'm inclined to use the `colors` setting in the general section of the i3status.config to decide if to use colors or not.  If we go for that we can just strip modules color settings anyhow at the low level.

py3status/modules/net_rate.py
```diff
@@ -11,6 +11,8 @@
     interfaces: comma separated list of interfaces to track
     interfaces_blacklist: comma separated list of interfaces to ignore
     precision: amount of numbers after dot
+    color: colorise output according to the value of 'up|down|total'
+    threshold_bad / threshold_degraded: thresholds for color change
```
tobes: These should be in alphabetical order and threshold_bad / threshold_degraded should be documented individually.

py3status/modules/net_rate.py
tobes: It also makes your changes harder to read

py3status/modules/net_rate.py
tobes: A threashold_xxx of None could also indicate that we don't want to use colors

tobes:
@guiniol thanks this seems a good enhancement of the module

guiniol:
I went with `threshold_value` instead of `color`. It's explained in the docs.

guiniol:
Closing this until #471 gets merged.

PR #440, opened 2016-08-28
Sys cpu

guiniol:
This PR adds a module to display the CPU percent.

### DIFF
--- a/py3status/modules/sys_cpu.py
+++ b/py3status/modules/sys_cpu.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""
+Display the percent CPU used.
+This modules requires the psutil python module.
+
+Configuration parameters:
+    threshold_degraded / threshold_bad: thresholds for color change
+"""
+
+# import your useful libs here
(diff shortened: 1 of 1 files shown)


py3status/modules/sys_cpu.py
```diff
@@ -0,0 +1,48 @@
...
+import psutil
+
+
+class Py3status:
+    def __init__(self):
+        self.threshold_degraded = 50
+        self.threshold_bad = 80
+        self.cache_timeout = 1
+        self.format = "☢{cpu:3d}%"
+        pass
```
tobes: this pass probably isn't needed

tobes:
@guiniol is this adding something that sysdata is currently lacking or is there a reason that you are unhappy with that module?

Improving the existing modules is always a preference over adding new ones.

guiniol:
If I remember correctly, at the time I wrote this module, py3status would execute every method not starting with a `_` and this meant that I couldn't have cpu percent without having ram percent too. I'll take a look and see if I can get my current config with sysdata.

PR #441, opened 2016-08-28
Sys io

guiniol:
This PR add a module to show the current disk IO accross all disks


tobes:
I've not looked in detail but could we not read `/proc/diskstats` also this feels it should be part of a disk_info module than being independent.  Again this is the small modules vs uber modules debate.

guiniol:
I could merge the `disk_free_space`  #436 and this one together.

guiniol:
Closing this. I'll merge it into #436 once #471 is merged.

PR #429, opened 2016-08-27
modules to use time_in()

tobes:
Get modules to use `time_in()`

PR #442, opened 2016-08-28
Sys load

guiniol:
This PR adds a module to show the system load.

### DIFF
--- a/py3status/modules/sys_load.py
+++ b/py3status/modules/sys_load.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""
+Display the load average.
+
+Configuration parameters:
+    threshold_degraded / threshold_bad: thresholds for color change
+"""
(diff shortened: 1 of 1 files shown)


tobes:
I'm sort of inclined to add this functionality into sysdata.  @ultrabug I don't know how you feel about the big uber module vs small specialist ones?  To me they all feel as if they should be wrapped into one.

As I was sort of saying in another PR from this bunch that I think sysdata should become more component based so that we can do some thing like `format = '{cpu_temp} {load_5min} {ram_precent}'` and have individual coloration.

This is also part of one of my world domination strategies involving format lists but that is awaiting #429 before it appears

guiniol:
I can do that. I just pushed an improved `sysdata`. Adding this should be trivial.

guiniol:
Opened #454 to add this to sysdata.

PR #443, opened 2016-08-28
Sys ram

guiniol:
This PR adds a module to display the RAM usage

### DIFF
--- a/py3status/modules/sys_ram.py
+++ b/py3status/modules/sys_ram.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""
+Display the percent RAM used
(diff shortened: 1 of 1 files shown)


py3status/modules/sys_ram.py
```diff
@@ -0,0 +1,49 @@
...
+
+# import your useful libs here
+import psutil
+
+
+class Py3status:
+    def __init__(self):
+        self.threshold_degraded = 40
+        self.threshold_bad = 75
+        self.cache_timeout = 1
```
tobes: Do we really need to check every second as the default?

tobes:
@guiniol does this add something over what is available in sysdata?

PR #444, opened 2016-08-28
Volume status pa

guiniol:
This PR adds a volume status module using pulse audio (pamixer) instead of alsa


tobes:
@guiniol thanks for this.

Could we not adapt the volume_status module to also know about pusle audio?  That way we keep the number of modules limited but with greater functionality.

guiniol:
I'll look into it

guiniol:
Opening new PR with pulseaudio added to volume_status

PR #445, opened 2016-08-28
Frame module has a python2 issue

tobes:
python2 list has no copy() use [:] instead

Also copy the item that we are going to change.


ultrabug:
Thanks !

PR #447, opened 2016-08-29
add cmus playback status module

tobanw:
A simple `cmus` module that takes advantage of `cmus`'s ability to execute a shell script at every status change. This module reads a text file generated by cmus and swaps in status icons and colorizes the output. See the docstring for details.

I can update the modules README if you like, just let me know.

### DIFF
--- a/py3status/modules/cmus.py
+++ b/py3status/modules/cmus.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+
+"""
+Display the cmus playing status and current track.
+
+Cmus can execute a shell script at every status change by setting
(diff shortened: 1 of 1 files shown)


tobes:
@tobanw Thanks for the contribution.

Looking at the source of cmus it looks like it supports the mpris d-bus interface.  The mpris module PR #354 is intended to work with all d-bus enabled media players.  Could you see if that works with cmus and would meet your needs?

tobanw:
Good catch, I see `mpris` support was just added to `cmus` a couple months ago (should be [released soon](https://github.com/cmus/cmus/issues/506)). I'll go ahead and close this PR, and I'll wait for the `mpris` module to be released.

PR #449, opened 2016-08-29
Sysdata

guiniol:
This PR adds the possibility to choose which values are used for colors.
It also adds the possibility of choosing a zone for the cpu temperature. However, it changes the output of `cpu_temp`: it doesn't contain the trailing '°C' anymore (because it can't be compared to set the color otherwise)

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -10,6 +10,12 @@ Configuration parameters:
         (default 75)
     med_threshold: percent to consider CPU or RAM usage as 'medium load'
         (default 40)
+    padding: length of space padding to use on the left
+        (default: 0)
+    precision: precision of values
+        (default: 2)
+    zone: thermal zone to use. If None try to guess CPU temperature
+        (default: None)
 
 Format placeholders:
     {cpu_temp} cpu temperature
@@ -22,6 +28,11 @@ Color options:
     color_bad: Above high_threshold
     color_degraded: Above med_threshold
     color_good: Below or equal to med_threshold
+
+Color conditionals:
+    cpu: change color based on the value of cpu_usage
+    mem: change color based on the value of mem_used_percent
+    temp: change color based on the value of cpu_temp
 
 NOTE: If using the `{cpu_temp}` option, the `sensors` command should
 be available, provided by the `lm-sensors` or `lm_sensors` package.
@@ -92,7 +103,7 @@ class GetData:
         # Results are in gigabytes
         return total_mem, used_mem, used_mem_p
 
-    def cpuTemp(self):
+    def cpuTemp(self, zone):
         """
         Tries to determine CPU temperature using the 'sensors' command.
         Searches for the CPU temperature by looking for a value prefixed
@@ -100,16 +111,21 @@ class GetData:
         out temperatures of all codes if more than one.
         """
 
-        sensors = subprocess.check_output(
-            'sensors',
-            shell=True,
-            universal_newlines=True,
-        )
-        m = re.search("(Core 0|CPU Temp).+\+(.+).+\(.+", sensors)
-        if m:
-            cpu_temp = m.groups()[1].strip()
+        if zone:
+            cpu_temp = float(subprocess.check_output(['sensors', zone])
+                             .split()[7].decode("utf-8")[1:-2])
+
         else:
-            cpu_temp = 'Unknown'
+            sensors = subprocess.check_output(
+                'sensors',
+                shell=True,
+                universal_newlines=True,
+            )
+            m = re.search("(Core 0|CPU Temp).+\+(.+).+\(.+", sensors)
+            if m:
+                cpu_temp = float(m.groups()[1].strip()[:-2])
+            else:
+                cpu_temp = 0
 
         return cpu_temp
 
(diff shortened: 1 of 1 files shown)


guiniol:
I'll add load average and swap.

tobes:
@guiniol I think you should split this out into separate PRs as you are adding several different features zones,  threshold changes, format changes etc.

> it changes the output of cpu_temp: it doesn't contain the trailing '°C' 

This seems an unwanted change and I do not understand why it is necessary.

The format changes you make will also break the output for users of the module who have their own formats defined in the i3status.config

guiniol:
> @guiniol I think you should split this out into separate PRs as you are adding several different features zones, threshold changes, format changes etc.

Alright. I'll work on one thing at a time.

> > it changes the output of cpu_temp: it doesn't contain the trailing '°C'

> This seems an unwanted change and I do not understand why it is necessary.

This is necessary so that the color can be changed according to the value of `cpu_temp`. I also think it makes the value more flexible in case people don't want to show the unit.


> The format changes you make will also break the output for users of the module who have their own formats defined in the i3status.config

I tried to keep the defaults the same. Except for `cpu_temp` which part did I break?

tobes:
> This is necessary so that the color can be changed according to the value of cpu_temp. I also think it makes the value more flexible in case people don't want to show the unit.

`{cpu_temp}` output should stay the same but you can have it different internally

guiniol:
I hesitated and figured the added flexibility was worth it. I'll make the changes.

guiniol:
Done.

guiniol:
Currently, there's only the "flexibility" part of the changes, and that will change again. See discussion in #454

guiniol:
@tobes: maybe we should continue our talk about this PR here ^^

guiniol:
I feel like I should close all my other PRs and wait for this formatter-colors to be in. Most of my PRs are about colors

guiniol:
Since #471 has been merged, I added the new version with all the bells and whistles

py3status/modules/sysdata.py
```diff
@@ -10,19 +10,38 @@
         (default 75)
     med_threshold: percent to consider CPU or RAM usage as 'medium load'
         (default 40)
+    padding: length of space padding to use on the left
+        (default 0)
+    precision: precision of values
+        (default 2)
```
tobes: I think we shouldn't have these.  Users can do all this in the format string eg `format = '{mem_used:05.2f}'`

py3status/modules/sysdata.py
```diff
@@ -92,27 +123,55 @@ def memory(self):
...
-            cpu_temp = m.groups()[1].strip()
+        if zone:
+            cpu_temp = float(subprocess.check_output(['sensors', zone])
+                             .split()[7].decode("utf-8")[1:-2])
+
         else:
-            cpu_temp = 'Unknown'
+            sensors = subprocess.check_output(
+                'sensors',
+                shell=True,
```
tobes: `shell=True` is not needed here, .  It should be avoided anyhow as it has the potential to be abused when combined with user input even if that is not an issue here.

Consider using the new `py3.command_output()` helper function (and not just here).

py3status/modules/sysdata.py
```diff
@@ -92,27 +123,55 @@ def memory(self):
...
+        """
+        Parse /proc/meminfo, grab the memory capacity and used size
+        then return; Memory size 'total_mem', Used_mem, and percentage
+        of used memory.
+        """
+
+        memi = {}
+        with open('/proc/meminfo', 'r') as fd:
+            for s in fd:
+                tok = s.split()
```
tobes: `splitlines()` is more robust than `split()`

py3status/modules/sysdata.py
```diff
@@ -123,59 +182,89 @@ class Py3status:
...
 
         # if specified as a formatting option, also get the CPU temperature
         if '{cpu_temp}' in self.format:
-            cpu_temp = self.data.cpuTemp()
-        else:
-            cpu_temp = ''
+            cpu_temp = self.data.cpuTemp(self.zone)
+            self.values['cpu_temp'] = (value_format + '°C').format(cpu_temp)
+            if cpu_temp > self.high_threshold:
+                self.py3.COLOR_TEMP = self.py3.COLOR_BAD
```
tobes: `self.py3.COLOR_TEMP` is a constant you should not be trying to redefine it

py3status/modules/sysdata.py
```diff
@@ -10,19 +10,38 @@
         (default 75)
     med_threshold: percent to consider CPU or RAM usage as 'medium load'
         (default 40)
+    padding: length of space padding to use on the left
+        (default 0)
+    precision: precision of values
+        (default 2)
+    zone: thermal zone to use. If None try to guess CPU temperature
+        (default None)
```
tobes: Is there much call for this?  If so maybe we can be more clever so we can have output from multiple 'zones'.

py3status/modules/sysdata.py
```diff
@@ -10,19 +10,38 @@
...
     color_bad: Above high_threshold
     color_degraded: Above med_threshold
     color_good: Below or equal to med_threshold
 
+Color conditionals:
+    cpu: change color based on the value of cpu_usage
+    load: change color based on the value of load1
+    mem: change color based on the value of mem_used_percent
+    temp: change color based on the value of cpu_temp
+    swap: change color based on the value of swap_used_percent
```
tobes: Hmmm,  I think we can do things in a more transparent way for the user eg `{mem_used}` would display in the correct color automatically.  But we do have the issue about how we provide the flexibility to the user.

I think this needs some thought.

tobes:
@guiniol #532 thresholds branch has been merged this should help you with this.

```
sysdata {
    format = "CPU [\?color=cpu {cpu_usage}%] ..."
    thresholds = {
       'cpu': [(0, 'bad), (40, 'degraded'), (70, 'bad')],
        ...
    }
}
```

guiniol:
Thanks. I'll adapt the code for that and resubmit a PR if necessary, and then I'll submit a PR each for swaps, loads, etc...

Issue #451, opened 2016-08-30
Thresholds


tobes:
Several modules have threshold and I think we should look at trying to standardize how these are treated. 

__names__

Sometime they are things like `critical`, `high_threshold`, `threshold_high`.  I think we should look at unifying them as  `threshold_xxx`

Alternatively we could fully switch to the `thresholds = {..}` approach see later

Sometime they are things like `threshold_high` other times `threshold_bad` I prefer the more descriptive names  `threshold_high`

__showing thresholds__

Currently there is no system for this that I can see. Options could be

* set `threshold_xxx = None` to disable (this would mean changing all modules)

*  Use `colors` parameter that is applied in core that strips any added colors from a modules output.  This could be added to any module or on the general section. it would work as it does for i3status. This would be my preference.

__custom thresholds__

vnstat allows fully customized thresholds which is an interesting idea as it give the user full flexibility
`coloring = { 800: "#dddd00", 900: "#dd0000", }` 

I think this is an interesting idea that we should look into more (maybe allowing named colors) eg
`thresholds = { 800: "good", 900: "degraded", 1000: '#FF00FF' }` 

__multiple thresholds__

sysdata has thresholds that it uses for more than one use cpu/memory and realistically is missing ones for cpu temp. options are

* multiple params - `threshold_cpu_high`

* using a dict `thresholds = {'cpu': {'high': 95, 'low': 50 }, 'memory': {...}}`

* split dicts `thresholds_cpu = {'high': 95, 'low': 50 }` or even `thresholds_cpu = {'high': (95, 'bad'), 'low': (50, '#FFFF00') }`

Currently there is a problem about which value to care about PR #449, #439 try to solve this by adding `threshold_value` parameter to say which is to be used.  To me this feels the wrong solution as we should just use the components functionality to allow cpu/memory/cpu temp to be colored independently. 

__helpers__

It might well be that adding some helpers into py3 would help keep consistency between modules and potentially allow us to deal with the above issues in more than one way eg

```
module {
    thresholds_cpu = {'high': (95, 'bad'), 'low': (50, 'good') }
}
```
would be the same as
```
module {
    thresholds_cpu_high = 95
    thresholds_cpu_low = 50
    color_thresholds_cpu_high = 'bad'
    color_thresholds_cpu_low = 'good'
}
```

Anyway the main thing is getting a consistent approach that we can use across all relevant modules.  We can start small and increase the features/flexibility as we go.

tobes:
[...]

Issue #401, opened 2016-08-13
How can I get mem_used in MB?


Rapiz1:
How can I get mem_used in MB?

tobes:
Unfortunately the sysdata module currently only gives memory sizes in GB.

 The code would be easy enough to change https://github.com/ultrabug/py3status/blob/master/py3status/modules/sysdata.py#L72 is where the conversion to GB happens (1 << 10) would give it in MB.

ritze:
Maybe this can be a little bit more intuitively? When you have only 256 MB than MB is shown as unit. If you have 2 GB, than GB is shown as unit and so on…

pferate:
I submitted a PR (#520) that will report mem_used in MB if total_mem is less than 1GB.  Otherwise, it gets reported in GB.

It would be trivial to add an option to force it to a certain unit.

PR #454, opened 2016-08-30
Add load average to sysdata module

guiniol:
This PR adds `{load1}`, `{load5}` and `{load15}` to sysdata.

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -12,6 +12,9 @@ Configuration parameters:
 Format of status string placeholders:
     {cpu_temp} cpu temperature
     {cpu_usage} cpu usage percentage
+    {load1} one minute load average
+    {load5} five minute load average
+    {load15} fifteen minute load average
     {mem_total} total memory
     {mem_used} used memory
     {mem_used_percent} used memory percentage
@@ -31,6 +34,7 @@ from __future__ import division
 
 import re
 import subprocess
+import os
 
 
 class GetData:
@@ -111,6 +115,8 @@ class GetData:
 
         return cpu_temp
 
+    def load(self):
+        return os.getloadavg()
 
 class Py3status:
     """
@@ -150,6 +156,12 @@ class Py3status:
             (self.values['mem_total'],
              self.values['mem_used'],
              self.values['mem_used_percent']) = self.data.memory()
+
+        # get load
+        if '{load' in self.format:
+            (self.values['load1'],
+             self.values['load5'],
+             self.values['load15']) = self.data.load()
 
         response = {
             'cached_until': self.py3.time_in(seconds=self.cache_timeout),
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -6,6 +6,8 @@ Configuration parameters:
     format: output format string
     high_threshold: percent to consider CPU or RAM usage as 'high load'
     med_threshold: percent to consider CPU or RAM usage as 'medium load'
+    threshold_values: list of values to use for setting the color
+        (default: ['cpu_usage', 'mem_used_percent'])
 
 Format of status string placeholders:
     {cpu_temp} cpu temperature
@@ -29,7 +31,6 @@ from __future__ import division
 
 import re
 import subprocess
-from time import time
 
 
 class GetData:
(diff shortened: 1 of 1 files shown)


guiniol:
I built this PR on top of #449

tobes:
@guiniol you have the zone changes in this

guiniol:
sorry. I think I need to reorganise the PR. All three PRs need to be on top of 0ccd248. So, one PR for 0ccd248, and then one for each feature.

tobes:
0ccd248 is the PR That is the least likely to get merged as it involves threshold changes and that it something I'm keen on getting standardized in py3status see #451 

Also really we want to allow all thresholds to work with this module so would show correct color for `{mem_used}` `{cpu_usage}` etc so that should be done at the same time

guiniol:
Alright. I should have started there.
Do you have an example of components?
Since I have started working on this, I might as well do the job in the right way.

If I understand correctly, components allow for coloring part of the output of one module. In that case, each we should have one component for each type of value, and have the respective thresholds. It does raise a new question: how do we handle `mem_used` vs `mem_used_percent`? We could choose one value which would give the color for this component, eg. `mem_used_percent`.
If that sounds right, I'll do it tonight.

guiniol:
Let's continue the discussion in #449. I'll add the functionnality once we agree on the base ^^

tobes:
> Do you have an example of components?

Documentation here
https://github.com/ultrabug/py3status/blob/master/doc/README.md#composites

modules group, frame, timer and rainbow use it

> how do we handle mem_used vs mem_used_percent? 

they would colour the same way

guiniol:
> Documentation here
https://github.com/ultrabug/py3status/blob/master/doc/README.md#composites
> modules group, frame, timer and rainbow use it

Thanks

> > how do we handle mem_used vs mem_used_percent?

> they would colour the same way

My point exactly. So, I'm for using `mem_used_percent` to decide on the color.

guiniol:
I'm still a bit unclear on some things.
For example, let's say we keep the default
```python
format = "CPU: {cpu_usage:.2f}%, Mem: {mem_used:.2f}/{mem_total:.2f} GB ({mem_used_percent:.2f}%)"
```
This should be split in two composites:
`"CPU: {cpu_usage:.2f}%"` and `"Mem: {mem_used:.2f}/{mem_total:.2f} GB ({mem_used_percent:.2f}%)"`. What about the `, ` in between? Which color should it be?
I don't think we can know without the user telling us where to split. Which means that the user should instead give:
```python
format= ["CPU: {cpu_usage:.2f}%", "Mem: {mem_used:.2f}/{mem_total:.2f} GB ({mem_used_percent:.2f}%)"]
```
This really changes the expectation for format. Additionally, the current behaviour is completely different: The *WHOLE* format changes color depending on the *MAX* between cpu and mem.

Unless I'm missing something obvious, we have two choices:
1. We change the expectations for format and use composites. I think this is more elegant but I don't see a way to replicate the "max between mem and cpu" part.
2. We do something like I did and allow for a list of values to check to change the color. This used to be useful because we couldn't color part of the output independently. I think it is obsolete now that we can.

What do you think?

tobes:
> {cpu_usage:.2f}

The default format uses `{cpu_usage}` etc formatting is done by the module __we should not change this__.

The format should be left as it is

What i would do is use py3.build_composite() to do the work

so we do something like

```
composites  = {
    'cpu_usage': {'full_text': <usage>, 'color': <color>},
    ...
}

composite = self.py3.build_composite(self.format, param_dict={...}, composites=composites)

return {
    'composite': composite,
    'caced_until': ...,
}
```

that way all our text is in one color and the placeholders that use thresholds have the colors

tobes:
This is actually quite tricky to get right for the end user

guiniol:
What would the configuration in i3status.conf look like?

Also, I don't agree that the module should do the formatting. Ideally, `cpu_usage` should be the raw value, that way, if someone wants two digits precision, they can have it, but if I want 0 digit precision, I can also have it.

tobes:
>  i3status.conf 

```
order += sysdata
```

all the defaults should remain as they are and more importantly if I have settings in my i3config they must be respected.

> Also, I don't agree that the module should do the formatting. Ideally, cpu_usage should be the raw value, that way, if someone wants two digits precision, they can have it, but if I want 0 digit precision, I can also have it.

but if I am using them now they should still work for me just like they always did  now if you want to add {cpu_usage_full} then that would allow what you want but not ruin it for the user.

tobes:
Add new placeholders separately to doing the composite to keep things simple

guiniol:
I totally agree that the default shouldn't change if possible. However, with the composites, I don't see a way of replicating the color changing based on the max of cpu and mem.
For example, let's say the user wants to display:
```
"CPU: {cpu_usage}%, Mem: {mem_used_percent}%"
```
and have the text change on the max of `cpu_usage` and `mem_used_percent`. How would they specify this?

guiniol:
> Add new placeholders separately to doing the composite to keep things simple

Something like `max_cpu_ram`? But that would only make sense for the colors, and then I have to add one placeholder per combination. If I add swap and load that's 5 values, and 120 combinations.
So, I'm guessing I don't understand what you mean. Could you give me an example?

tobes:
Option 1

Text is not coloured

Option 2

We add some magic to the formatter but that is quite problematic and would need changes to how we do composites. 

So we have something like `[color=cpu CPU: {cpu usage}%]...`

Option 3

Something else

guiniol:
hum, additonal options:
3. we forget about composites. People can use frames to compose what they want. Each module can then have only one color. We can decide on a name for the list of values which are used to compute the color. Each value could have a different threshold, and only the highest threshold of all values applies its color. Or we just use max on the list of values if that's too complex.
4. Use composite. Users have to specify the list of elements in the composite and for each element, they can choose which value is used for the color.


I like your option 2 but that seems to make things more complex on the formatter and composite sides.

tobes:
Your option 3 is already available and is the simplest so that is a plus

option 4 is a bit complicated from the user perspective.

option 2 is not too hard but it would mean that safe_format could create a list rather than a string. We could work around that.  My main concern is that I'd want any format commands eg `[\?color=xxx ...]` to work in all modules.

Anyway we will have some time to decide as this will be for 3.2 which is likely a few months away (3.1 is due for release in the next day or so)
 
Regardless sysdata is in need of some love and has issues eg #401

guiniol:
> option 2 is not too hard but it would mean that safe_format could create a list rather than a string. We could work around that. My main concern is that I'd want any format commands eg `[\?color=xxx ...]` to work in all modules.

Wouldn't that simply mean that every module must use `safe_format`?

I'll take a look at #401 (this weekend probably).

tobes:
> Wouldn't that simply mean that every module must use safe_format?

This is part of the grand plan for 3.2 see #428 

The actually main changes are in module.py as currently response must either have `{'full_text': <string>}` or `{'composite': <list>}`  if we do the `[\?color=... ]` thing then we would get `{'full_text': <list>}`.  The code changes are pretty trivial but the documentation changes etc are quite large and it makes things more complex for the casual module writer.

guiniol:
Ok. Is there an ETA on deciding whether to allow for `{'full_text': <list>}`? Or should I just go with option 3 until it is?

tobes:
> Or should I just go with option 3 until it is?

If you want this now then yes.  I think decisions like `{'full_text': <list>}` might take a while as from my point of view naming conventions and the handling of deprecated config options is the current big priority.

guiniol:
So my current PR #449 would be mostly fine as is? Maybe change the name of `threshold_values`?

tobes:
> Maybe change the name of threshold_values?

I am opposed to the threshold_values thing as it feels a half solution.

Currently the module uses the 'correct' threshold if only cpu or mem is being displayed.  If both are being displayed it uses the most 'critical'  This is far from perfect but is better than nothing.

Really this needs a full solution.

Also that PR also messes up the the format for existing users.

guiniol:
> Also that PR also messes up the the format for existing users.
Yep, on my TODO list.

> Currently the module uses the 'correct' threshold if only cpu or mem is being displayed. If both are being displayed it uses the most 'critical' This is far from perfect but is better than nothing.

`threshold_values` does exactly this but is more generic because you can add `cpu_temp`

tobes:
@guiniol I've had a play at the option 2 color block stuff.  It also fixes an issue with max_length so I think I will try to push for it in 3.2  there is a rough cut at https://github.com/tobes/py3status/tree/formatter-colors if you wanted to have a play.  It needs a cleanup plus documentation.

Having thought some more I've decided that this is a really cool feature.

So the format could look something like
`format = '[\?color=cpu CPU: {cpu_usage}] [\?color=memory RAM: {mem_percent}%]'`

Anyhow this will still be some way off

guiniol:
That's cool.
So,  `format = '[\?color=cpu CPU: {cpu_usage}] [\?color=memory RAM: {mem_percent}%]'` means color `cpu_usage` according to `threshold_cpu_*` and the value of `cpu_usage`. Does that mean `[\?color=cpu` should be `[\?color=cpu_usage`?
I'll try playing with it.

tobes:
the color like `cpu` is either from the module or the config eg would be `self.py3.COLOR_CPU`

```
Py3status:

    method(self):
        self.color_cpu = '#FF00FF'
        return {'full_text'; self.py3.safe_format(self.format)}
```

guiniol:
How do you integrate the thresholds then? `color_cpu_high` and `threshold_cpu_high`?

tobes:
> How do you integrate the thresholds then? 

If it was me I'd just do (pseudo code)

```
if cpu_usage > threshold_high:
   color_cpu = COLOR_BAD
else:
   color_cpu = COLOR_GOOD
```

IMHO separate thresholds (`threshold_cpu_high` rather than `threshold_high` ) for cpu/mem would be a whole new PR and that needs thought again see #451 but basically work in the same way.

Personally I think `color_cpu_high` is unneeded and adds too much crazy

guiniol:
I finally got around to adding formatter-colors support to the (almost) basic sysdata. I only added the option to specify the thermal zone as well as the padding and precision of values. It's in #449. What do you think?

PR #455, opened 2016-08-30
Add format_charging and format_full to battery_level

guiniol:
This allows for changing the displayed string based on the status of the battery.

Also switched to `self.py3.safe_format` and `self.py3.time_in`

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -16,6 +16,10 @@ Configuration parameters:
         default is "⚡"
     format: string that formats the output. See placeholders below.
         default is "{icon}"
+    format_charging: string that formats the output when charging. Defaults to
+        the value of format. See placeholders below.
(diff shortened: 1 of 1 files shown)


tobes:
hmmm, we can do this via the new formatting

```
battery_level {
    format '[if=charging <charging format>]|[<general format>]'
}
```

This behaviour is new with the formatter updates in 3.1  there is no way to see if we are full though

Maybe many formats are easier for users to use.  I think this is an issue that needs some thought.

The aim for the 3.2 release is to try to make modules more organised/uniform

guiniol:
I could add a `full` variable same as charging and then it would work.

guiniol:
> The aim for the 3.2 release is to try to make modules more organised/uniform

That's a very good thing.

guiniol:
I'll close this until #471 gets merged.

PR #456, opened 2016-08-30
Tests

tobes:
This adds pytest support.  Currently the only testing is the formatter.

 Interestingly this uncovered a bug in pypy


ultrabug:
Thank you so much, I'm so unhappy that py3status isn't tested properly that this is a huge step in the right direction !

PR #458, opened 2016-08-30
Support Rotation

coxley:
This change adds rotation support to the `xrandr` plugin.

### DIFF
--- a/py3status/modules/xrandr.py
+++ b/py3status/modules/xrandr.py
@@ -51,6 +51,8 @@ Dynamic configuration parameters:
     - <OUTPUT>_workspaces: comma separated list of workspaces to move to
         the given OUTPUT when it is activated
         Example: DP1_workspaces = "1,2,3"
+    - <OUTPUT>_rotate: rotate the output as told
+        Example: DP1_rotate = "left"
 
 Color options:
     color_bad: Displayed layout unavailable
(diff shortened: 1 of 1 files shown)


py3status/modules/xrandr.py
```diff
@@ -248,15 +250,16 @@ def _apply(self, force=False):
             #
             if output in combination:
                 pos = getattr(self, '{}_pos'.format(output), '0x0')
+                rotation = getattr(self, '{}_rotate'.format(output), 'normal')
```
tobes: it is probably worth checking that the value or rotation is legal eg in  ['normal', 'inverted', 'left', 'right']

ultrabug:
Hi @coxley , thank you for your contribution :)

I wanted to understand the motivation in adding this in the xrandr module rather than using the **xrandr_rotate** module please ?

ultrabug:
Ok timeout @coxley with a lot of thank you. I'll implement the value validation then.

Cheers

coxley:
@ultrabug: Crap, so sorry I let this fall-through. Didn't mean to be a drive-by contributor.

Yeah I forget exactly why I didn't use xrandr_rotate -- maybe it does the same thing but for some reason I thought it didn't meet my need. I see you merged it, but I'll try to play with it if I get some time.

ultrabug:
@coxley np at all, thanks for contributing :)

PR #464, opened 2016-09-02
Process status

ykhodorkovskiy:
allow for --full search of a procress. Original users will see no difference, but you can now pgrep for things like "ssh 10.*"

### DIFF
--- a/py3status/modules/process_status.py
+++ b/py3status/modules/process_status.py
@@ -7,6 +7,7 @@ Configuration parameters:
     format_not_running: what to display when process is not running
         (default '■')
     format_running: what to display when process running (default '●')
+    full: if True, match against the full command line and not just the process name
     process: the process name to check if it is running (default None)
 
 Color options:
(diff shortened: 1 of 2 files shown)


tobes:
@ykhodorkovskiy You seem to have got some commits in with this that shouldn't be here.  Could you try to either clean up this PR to just your commit e09aaee or create a fresh PR containing just that.

> Original users will see no difference

This is not my experience
```
$ pgrep -l py3status
3116 py3status
$ pgrep -fl py3status
3116 py3status
3127 i3status
```

ykhodorkovskiy:
@fmorgner Alright, I'll add a new field.

py3status/modules/process_status.py
```diff
@@ -30,10 +31,16 @@ class Py3status:
     format_running = u'●'
     format_not_running = u'■'
     process = None
+    full = None
 
     def _get_text(self):
         fnull = open(os.devnull, 'w')
-        if subprocess.call(["pgrep", self.process],
+        PGREP = ["pgrep", self.process]
```
tobes: ALL_CAPS names should be constants

py3status/modules/process_status.py
```diff
@@ -30,10 +31,16 @@ class Py3status:
     format_running = u'●'
     format_not_running = u'■'
     process = None
+    full = None
```
tobes: False might be a nicer default

py3status/modules/process_status.py
```diff
@@ -7,6 +7,7 @@
     format_running: what to display when process running
     format_not_running: what to display when process is not running
     process: the process name to check if it is running
+    full: use the full command line to match the process
```
tobes: > use the full command line to match the process

`if True match against the full command line not just the process name`

feels clearer to me

ultrabug:
Thank you @ykhodorkovskiy, would you be so kind to rebase from master please so this can get in ?

Also, please order the variables alphabetically (`full` should be before `process` everywhere)

ykhodorkovskiy:
@ultrabug Fixed as requested. Thanks

ultrabug:
Thanks again @ykhodorkovskiy !

PR #468, opened 2016-09-02
fix log exception when message contains special characters, by @Lujeni

ultrabug:
reported on IRC

@Lujeni was affected also and found the problem and this patch.

I'm doing the PR since he's busy and I'd like 3.1 out.


tobes:
looks good to me

PR #471, opened 2016-09-02
Formatter colors

tobes:
OK, so this is one of my crazy PRs

__History:__

With the sysdata module we display information on memory and cpu usage.  We have thresholds so that we can show cpu usage as good/bad etc and the same with memory usage.  However we cannot show the color for both if they are different.

This PR solves this problem by introducing color blocks so we can then specify
```
sysdata {
    format = "[\?color=cpu CPU: {cpu_usage}%], [\?color=memory Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]"
}
```

The module would need some tweaking so that we set the modules `self.color_cpu` and `self.color_memory` to  be the wanted values

We first try to get the color from the module, then we try to get from the modules config. So `[\?color=bad ERROR]` would use `color_bad` from the config if not overriden by the module setting a value.

Additionally the user can also be explicit eg `Hello [\?color=#FF00FF World!]`

__Issues:__

* To work modules need to use `py3.safe_format()` 

* Because we do this by creating composites on the fly `py3.safe_format()` can return a string (as before) or a composite (list) eg `[{'full_text': 'Hello '}, {'full_text': 'World!', color: '#FF00FF'}]` from the previous example.  Therefore modules.py needs to now accept a list as `response['full_text']` as well as a string.  To me this seems fine and maybe we should do this in favour of the `response['composite']`

__Also:__

Makes the `max_length` block command work with composites.

We now merge composites together to reduce the number where we can do this without causing problems. eg `[{'full_text': 'foo'}, {'full_text': 'bar'}]` becomes `[{'full_text': 'foobar'}]`
 
Thanks to the formatter tests we can see that we have not broken the existing functionality.


py3status/formatter.py
```diff
@@ -11,15 +11,44 @@
...
+        commands = dict(parse_qsl(commands_str, keep_blank_values=True))
+
+        self._if = commands.get('if', self._if)
+        self.color = commands.get('color', self.color)
+
+        max_length = commands.get('max_length')
+        if max_length is not None:
+            self.max_length = int(max_length)
+
+        self.show = 'show' in commands or self.show
```
ultrabug: I'm never a fan of pseudo booleans, here self.show can be either True or None

ultrabug:
Awesome :+1: thanks a lot

PR #473, opened 2016-09-04
Fix for occassional timezone bug

tobes:
On rare occasions I've had this error.  I'm not sure quite what causes it but this patch will stop the error.

```
  File "/home/toby/dev/py3status/py3status/i3status.py", line 349, in spawn_i3status
    self.set_responses(json_list)
  File "/home/toby/dev/py3status/py3status/i3status.py", line 227, in set_responses
    if self.i3modules[conf_name].update_from_item(item):
  File "/home/toby/dev/py3status/py3status/i3status.py", line 105, in update_from_item
    self.set_time_zone()
  File "/home/toby/dev/py3status/py3status/i3status.py", line 143, in set_time_zone
    i3s_time_tz = parts[2]
```

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -140,7 +140,11 @@ class I3statusModule:
         # get datetime and time zone info
         parts = i3s_time.split()
         i3s_datetime = ' '.join(parts[:2])
-        i3s_time_tz = parts[2]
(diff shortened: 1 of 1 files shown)


ultrabug:
okay that's strange

PR #474, opened 2016-09-06
Fix module_test for post_config_hook staticmethods etc

tobes:
There were some bugs in `module_test.py` for standalone running of modules.

* `@staticmethod` etc were not excluded from module methods to test.

* `post_config_hook()` was not excluded from module methods to test.

* `post_config_hook()` was not executed before module was tested.

* `self.items` was not set for containers.

This PR fixes these issues as well as adding some comments for clarity 
`method_name` -> `attribute_name` as that it more descriptive.


ultrabug:
:fire: !

PR #475, opened 2016-09-07
Allow else in a \?if=.. block

tobes:
if we have the format `[\?if=condition Hello|Goodbye]`

If condition is True then we get `'Hello'`

Currently if condition is False we get `''` with this patch we get `'Goodbye'` this feels the correct behaviour and it also makes it easier to write format strings.


ultrabug:
Okay thanks :) hopefully I'll get to this CHANGELOG soon

PR #476, opened 2016-09-08
Run all post config hooks before starting any modules

tobes:
Because the `post_config_hook()` was run when a module started, it was possible for content of groups to be updated and trigger an update of the group before the group had its `post_config_hook()` run.

This PR ensures that all modules have their `post_config_hook()` run before any modules are started.

This also updates group module to use post_config_hook()


ultrabug:
Aye this is sane indeed, thank you !

PR #477, opened 2016-09-08
Move sleep to start of main loop in core.py

tobes:
By moving the sleep from the end of the main loop to the beginning we allow more modules to have provided output when we first display.  This makes the initial display when we Mod+Shift+R etc much smoother.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -700,6 +700,11 @@ class Py3statusWrapper():
 
         # main loop
         while True:
+            # sleep a bit to avoid killing the CPU
+            # by doing this at the begining rather than the end
+            # of the loop we ensure a smoother first render of the i3bar
+            time.sleep(0.1)
+
             while not self.i3bar_running:
                 time.sleep(0.1)
 
@@ -742,9 +747,6 @@ class Py3statusWrapper():
                 out = ','.join([x for x in output if x])
                 # dump the line to stdout
                 print_line(',[{}]'.format(out))
-
-            # sleep a bit before doing this again to avoid killing the CPU
-            time.sleep(0.1)
 
     def handle_cli_command(self, config):
         """Handle a command from the CLI.


ultrabug:
I must say I've considered this and I thought it was best to output as fast as possible than having a small but still palpable blank time on the bar. Then you see your other modules appearing as they output which also looked more interactive to me.

I'd like some other users' impressions about this since we have a sort of stale feelings here :)

ultrabug:
ping @guiniol @pydsigner @pferate @maximbaz 

I'd like your user experience and advising on this matter please so we can decide to merge or close this PR ! Thanks

guiniol:
I don't have a very strong opinion on this one, but I would slide with @ultrabug on this one. It's nice to see that it's working, and then modules appear. However, I think it can also be disturbing at first that some things appear and not others, so maybe from the point of view of a new user, it would be better to wait until all modules have a output and then display.

maximbaz:
I tried this branch, and _personally_ I like it more than the current behavior in master. 

But at the same time I think the difference is so negligible that I wouldn't spend too much time arguing for one of the options :)

How about we have a simple poll? :joystick: 

* Those who are in favor of merging this pull request, please upvote :+1: the first comment in this thread. 
* Those who are against merging this pull request, please downvote :-1: it. 

The poll closes on Monday 31.10.2016, so that everyone has weekends ahead to see this notification, and based on the counters @tobes either merges or abandons this pull request.

pydsigner:
I'll pull it down and check here at some point.

ultrabug:
Ok let's vote :)

let's bring more possible votes in, friendly ping to @btall @ritze @rixx @Horgix @guilbep @Lujeni 

thanks !

ritze:
Maybe each module can deliver a static init string which is shown the first time on startup when the module is uninitialized. Then we can load the module at the end of the while loop.

rixx:
That actually sounds like a nice idea, I think. +1

ultrabug:
> Maybe each module can deliver a static init string which is shown the first time on startup when the module is uninitialized. Then we can load the module at the end of the while loop.

I had the idea of having a standardized "loading icon" similar are those we usually see on websites, I'm sure the effect would be awesome :)

pydsigner:
The way I'd implement this is just to show the entry — so it'd be 
something like `clementine | battery_level | disk /home | static_text`

ultrabug:
Thanks everyone for your reports and votes ! merged then :)

Issue #469, opened 2016-09-02
modules on a container reporting as 'urgent' should switch the group display to them


ultrabug:
so their urgent output is seen on the bar without having to scroll down to them

tobes:
This makes sense but might need some playing with to work smoothly from the users point of view.  Maybe we switch to the urgent item?

ultrabug:
That's what I meant in my poor english ;)

PR #478, opened 2016-09-08
Groups switch to urgent items

tobes:
This implements #469

`Py3.register_content_function()` has now been replaced with a more general `Py3.register_function()` and it is now documented.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -446,7 +446,7 @@ class Py3statusWrapper():
         """
         raise KeyboardInterrupt()
 
-    def notify_update(self, update):
+    def notify_update(self, update, urgent=False):
         """
         Name or list of names of modules that have updated.
         """
@@ -464,6 +464,10 @@ class Py3statusWrapper():
         for container in containers_to_update:
             container_module = self.output_modules.get(container)
             if container_module:
+                # If the container registered a urgent_function then call it
+                # if this update is urgent.
+                if urgent and container_module.get('urgent_function'):
+                    container_module['urgent_function'](update)
                 # If a container has registered a content_function we use that
                 # to see if the container needs to be updated.
                 # We only need to update containers if their active content has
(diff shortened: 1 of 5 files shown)


py3status/module.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, module, user_modules, py3_wrapper):
         self.prevent_refresh = False
         self.sleeping = False
         self.timer = None
+        self.urgent = False
```
ultrabug: I don't understand the use of a class parameter for this ?

py3status/module.py
```diff
@@ -145,8 +146,16 @@ def set_updated(self):
                 output.append(data)
         # if changed store and force display update.
         if output != self.last_output:
+            # has the modules output become urgent?
```
ultrabug: tbh, I don't get the meaning of the comment :(

py3status/modules/group.py
```diff
@@ -131,6 +131,15 @@ def _content_function(self):
         '''
         return set([self.items[self.active]])
 
+    def _urgent_function(self, module_list):
+        '''
+        This returns a set containing the actively shown module.  This is so we
```
ultrabug: actually this method doesn't return anything right ?

py3status/core.py
```diff
@@ -464,6 +464,10 @@ def notify_update(self, update):
         for container in containers_to_update:
             container_module = self.output_modules.get(container)
             if container_module:
+                # If the contain registered a urgent_function then call it
```
ultrabug: "container"

py3status/module.py
tobes: The reason for the class parameter is that we want to know when a module first becomes urgent so we can switch the group to show it.  The user may then change the group to show something else.  If we didn't have this then the group would keep showing the urgent module each time it was updated.

py3status/module.py
tobes: I don't know how to explain it better.  Maybe the last comment will have helped

py3status/modules/group.py
tobes: evidence of copy/paste I'll fix

PR #479, opened 2016-09-08
Make events and i3status threads daemons

tobes:
If I run py3status at the command line and then press `^C` I usually have to wait several seconds for the process to exit.  By making the event and i3status threads daemons the process ends instantly.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -296,6 +296,7 @@ class Py3statusWrapper():
             i3s_mode = 'mocked'
         else:
             i3s_mode = 'started'
+            self.i3status_thread.daemon = True
             self.i3status_thread.start()
             while not self.i3status_thread.ready:
                 if not self.i3status_thread.is_alive():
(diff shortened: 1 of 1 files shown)


ultrabug:
From the command line perspective I understand your feeling about this *time to properly quit* period but I think it is necessary to properly make sure that all threads py3status has spawned are properly killed before exiting.

That being said, I ran some tests and it looks like we're not clean on killing our children even in master. Just try to CTRL+R multiple times in a row and you should see :

```
➜  py3status git:(master) pgrep -l i3status
16965 i3status
17068 i3status
17190 i3status
17314 i3status
17565 i3status
```

That's bad :-1: , I'd rather we fix this first and then see if we remain clean with daemon threads.

ultrabug:
Shall we open an issue for this and link it to this PR ?

tobes:
```
➜  py3status git:(master) pgrep -l i3status
16965 i3status
17068 i3status
17190 i3status
17314 i3status
17565 i3status
```

I've got a solution to this but it'll be in a new PR

PR #485, opened 2016-09-19
Fix hide_seconds when a user has two batteries

gferon:


ultrabug:
Thanks a lot @gferon ! Since I don't have two batteries I'll trust you tested it but the fix LGTM :+1:

PR #486, opened 2016-09-22
kdeconnector: API fix

ritze:
`isPaired` is deprecated and module doesn't work anymore with the new kdeconnector version.

### DIFF
--- a/py3status/modules/kdeconnector.py
+++ b/py3status/modules/kdeconnector.py
@@ -127,7 +127,7 @@ class Py3status:
             device = {
                 'name': self._dev.name,
                 'isReachable': self._dev.isReachable,
-                'isPaired': self._dev.isPaired,
+                'isTrusted': self._dev.isTrusted,
             }
         except Exception:
             return None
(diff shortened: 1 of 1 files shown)


tobes:
@ritze you should make it so that it will work with both the old and new versions.

ritze:
@tobes as an Archlinux user I forget the "legacy" distributions. I'll push a new version of the patch.

py3status/modules/kdeconnector.py
```diff
@@ -119,6 +119,17 @@ def _get_device_id(self, bus):
 
         return None
 
+    def _get_isTrusted(self):
+        try:
+            # New method which replaced 'isPaired' in version 1.0
+            return self._dev.isTrusted
+        except Exception:
```
tobes: This should probably be `AttributeError` rather than `Exception` just to be more specific.

py3status/modules/kdeconnector.py
```diff
@@ -119,6 +119,17 @@ def _get_device_id(self, bus):
...
+            try:
+                # Deprecated
```
tobes: maybe add the version of kdeconnect when this happened?

py3status/modules/kdeconnector.py
```diff
@@ -119,6 +119,17 @@ def _get_device_id(self, bus):
...
+                return self._dev.isPaired
+            except Exception:
+                return None
```
tobes: when you do `device['isTrusted']()` later this will fail.

maybe do the function call inside this function and just return True/False

ultrabug:
Thank you guys !

PR #487, opened 2016-09-24
Fix safe format refactoring regressions

maximbaz:
`self.format` is a string that needs to be passed into `safe_format()`, not `self.format.format`, which is a function.


ultrabug:
Thank you @maximbaz !

Issue #482, opened 2016-09-10
Module imports (dependencies)


tobes:
For testing purposes it would be good if modules could be loaded even if they do not have their dependencies installed.  Doing this could also allow us to provide better information to the user on why the module cannot be run etc.

eg. `aws_bill` uses the `boto` library

We could do something like (my preference)
```
try:
    import boto
except ImportError:
    boto = None
```

or
```
def post_config_hook(self):
    import boto
```

I would also like to alter the modules method output to allow a response like
```
{'error': 'Dependency', 'requires': [{'name': 'boto', 'url': 'https://pypi.python.org/pypi/boto'}]}
```

ultrabug:
I like the idea of course this would be better than the current "crash".

I'm not sure about the import in the post_config_hook but maybe we could enforce the check in this very function

```python
try:
    import boto
except ImportError:
    boto = None

def post_config_hook(self):
    if boto is None:
        raise bla
```

tobes:
> ```
def post_config_hook(self):
    if boto is None:
        raise bla
```

I'm not convinced by raising an exception here.  I was looking at this and to me `self.py3.disable()` or some similarly named method could be used to prevent the module running/displaying output.

pydsigner:
@tobes I think an exception with a good message is the best way to go, but it could be something like `raise self.py3.MissingDependencyException('The epic foobar module requires pyfoobar to be installed!')`.

tobes:
@pydsigner ok this now makes sense to me.  It's better for py3status to decide what to do about the module, rather than each module deciding.  It's a nice and simple approach.

tobes:
I'm going to close this as I can do the testing I wanted via ast which actually makes more sense

PR #488, opened 2016-09-26
Tests: module docstrings

tobes:
This PR adds testing of docstrings.

We check that 

* docstring contains all the config options the module provides.

* correct default values are given

* config parameters listed in alphabetical order

Some modules have odd defaults eg `save_path` in `screenshot` module these are excluded from the tests.

Some modules cannot be imported due to missing dependencies eg `aws_bill` these are currently excluded but if we agree the proposal in #482 then we could also test these modules.

As well as ensuring more accurate docstrings this PR will help ensure new modules are correctly documented.

Also included are some fixes for existing modules to pass the tests

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -4,9 +4,14 @@ Display the current screen backlight level.
 
 Configuration parameters:
     cache_timeout: How often we refresh this module in seconds (default 10)
-    device:        The backlight device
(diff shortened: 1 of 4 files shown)


ultrabug:
This is the end of my *bad kitty* comments ! :smirk_cat:

tobes:
There will always be need for bad kitty comments

PR #490, opened 2016-10-03
battery_level.py recommended changes

inclementweather:
Module battery_level.py lines 343 and 350 raise an AttributeError when the Py3status class is instantiated and the battery is below either threshold. 
The self.last_known_status variable isn't declared yet so the threshold conditionals fail. Otherwise if the battery starts above the threshold and drops below it, the error is avoided.
Declaring last_known_status = '' at the top of Py3status fixes the problem.

A subprocess.call() to notify-send on lines 159 and 364 of battery_level.py raises a FileNotFoundError (python3.5) when libnotify is not installed. Neither libnotify nor dunst is listed in the module requirements.
The module should probably use the py3status notify_user() method so people can utilize the -b flag.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -488,13 +488,16 @@ class Py3statusWrapper():
             level = LOG_LEVELS.get(level, level)
             syslog(level, u'{}'.format(msg))
         else:
-            with open(self.config['log_file'], 'a') as f:
+            # Binary mode so fs encoding setting is not an issue
+            with open(self.config['log_file'], 'ab') as f:
                 log_time = time.strftime("%Y-%m-%d %H:%M:%S")
                 out = u'{} {} {}\n'.format(log_time, level.upper(), msg)
                 try:
-                    f.write(out)
-                except UnicodeEncodeError:
+                    # Encode unicode strings to bytes
                     f.write(out.encode('utf-8'))
+                except (AttributeError, UnicodeDecodeError):
+                    # Write any byte strings straight to log
+                    f.write(out)
 
     def report_exception(self, msg, notify_user=True, level='error'):
         """
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -115,10 +115,12 @@ class Py3status:
     threshold_bad = 10
     threshold_degraded = 30
     threshold_full = 100
-    last_known_status = 'good'
     # obsolete configuration parameters
     mode = None
     show_percent_with_blocks = None
+    # internal status variable
+    last_known_status = ''
+
 
     def battery_level(self):
 
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -118,8 +118,6 @@ class Py3status:
     # obsolete configuration parameters
     mode = None
     show_percent_with_blocks = None
-    # internal status variable
-    last_known_status = ''
 
 
     def battery_level(self):
@@ -336,6 +334,8 @@ class Py3status:
             self.percent_charged >= self.threshold_full else self.full_text
 
     def _set_bar_color(self):
+        # Prevent AttributeError when starting below threshold.
+        last_status = getattr(self, 'last_known_status', '')
         if self.charging:
             self.response['color'] = self.py3.COLOR_CHARGING or "#FCE94F"
             battery_status = 'charging'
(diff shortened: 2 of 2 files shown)


inclementweather:
Ok!  So docstring errors with the way I declared last_known_status at first, I didn't exactly understand the build failure message, but after realizing my error I modified it to use getattr() at the top of the method to check and set a default value.  Just some minor formatting issues after that. I tested everything on my laptop with and without the -p flag, everything seems to be working.

tobes:
@inclementweather thanks for this fix.  Also thanks for the feedback regarding the new docstring tests.  I'll try to get the documentation around them added/improved.

inclementweather:
@tobes My pleasure. This was a good exercise for me. 

Also, I had thought about this more and realized I could declare the last_known_status variable in a class __init__ instead of creating a new variable in the method.  Would that be the more pythonic solution?

tobes:
> Would that be the more pythonic solution? 

Yes it probably would be, but there are always lots of different ways to achieve the same thing :)

You can also update existing commits in git via something like the following
```
git add -p
git commit --amend
git push -f
```

inclementweather:
Ok. I'll make those changes and update this pull request.  Thanks for the tip on amending my commits.

inclementweather:
Modified and tested.  Amended the last CL.

tobes:
This is now a much cleaner looking patch :+1: 

I'm not happy that github has not noticed the code has changed since my last review.  This sort of makes the feature worthless in my opinion.

inclementweather:
I thought that was kind of weird too.

py3status/core.py
```diff
@@ -488,13 +488,16 @@ def log(self, msg, level='info'):
...
             syslog(level, u'{}'.format(msg))
         else:
-            with open(self.config['log_file'], 'a') as f:
+            # Binary mode so fs encoding setting is not an issue
+            with open(self.config['log_file'], 'ab') as f:
                 log_time = time.strftime("%Y-%m-%d %H:%M:%S")
                 out = u'{} {} {}\n'.format(log_time, level.upper(), msg)
                 try:
+                    # Encode unicode strings to bytes
+                    f.write(out.encode('utf-8')
```
tobes: you've failed to close this, you need to add `)`

tobes:
@inclementweather is it possible for you to revert that last commit in this PR and do it as a separate PR.  It makes it easier to review changes that way and allows some PRs to be merged and others not

You'll need to create a new branch for that PR.

In the future he easiest way is keep you master branch the same as upstream master and then do something like
```
git checkout master  # make sure we are on the master branch
git checkout -b my-new-branch  # create a new branch
# make changes
git add -p # add changes
git commit -m'my commit message'
git push # push your new branch you'll have to set an upstream but git will prompt you how
```
Your master currently contains these extra commits so that won't work for you here.

To fix your master will a pain as really you want to reset it to commit c2fcb866e2 and then make branches from there (or at least branch from there and then add the commits via cherry picking etc).  If you are happy with doing that it'd be great but if it is too complicated (not sure if you've used git much) I can fix it up as I think that both of these solutions are good (excluding the minor bugfix).

It maybe that @ultrabug is happy to take this as is, but in future separate branches is the way to go.

py3status/core.py
inclementweather: Thanks. Looking at how to revert now.

ultrabug:
Thanks a lot @inclementweather for your reports, debugging and cool work on fixing stuff. It is not everyday that I see first contributions being on the py3status core :+1: 

@tobes I'll be okay to merge this as-is for this time, I'm sure @inclementweather will think about separate PRs for his next contributions ! :)

Also friendly thanks @maximbaz for the review.

@tobes this PR looks ready for merge to me, am I wrong ?

ultrabug:
If you participate in hacktoberfest, I've flagged this PR ;)

inclementweather:
Sorry! I didn't realize my newest commit would go to the pull request too.

tobes:
@ultrabug 

> @tobes this PR looks ready for merge to me, am I wrong ?

I think for merging straight after 3.2

@inclementweather 
> Sorry! I didn't realize my newest commit would go to the pull request too. 

It can take a little while to work out how things work, but it gets easy quickly

inclementweather:
@tobes Thanks. I appreciate the git help.  

I'll wait until the changes are merged then reset.

tobes:
> I'll wait until the changes are merged then reset.

Once they are merged then your master will be happy again because it will have the same commits as master here.  To get the latest stuff from master you'll do something like this
```
git checkout master  #make sure we are on master branch
git pull upstream master  # pull latest upstream
```
the odd `git fetch upstream --tags` once in a while is also good to get any upstream tags (version releases are tagged)

inclementweather:
> git checkout master  #make sure we are on master branch
> git pull upstream master  # pull latest upstream

@tobes would it revert my commits if I did that now?

tobes:
> @tobes would it revert my commits if I did that now? 

No it would just get the latest commits in master here.  The only time it would be an issue is if there was a merge conflict but you'll be fine here.  You will find that there will be commits after your ones if you do this `git log` will show the commits in order.

Generally keep your master branch as a copy of upstream and do any changes in fresh branches.

Git is pretty good at sorting things out so as long as there are not changes in the same bit of code you are generally fine.  merge conflicts can be a little confusing to start with but are fine once you've got used to them.

ultrabug:
Thanks a lot @inclementweather et al for this PR and review :+1:

PR #491, opened 2016-10-04
Import dependencies

tobes:
Following from #482 here is an implementation.  Currently it only logs that the dependencies are not met, and disables the module.

If this seems good to people I'll add some documentation etc.

As far as notifying the user goes, I'm quite keen to start doing this via the i3bar with a clever module but that is for another day. We need a nice way to avoid spamming users too much.


ultrabug:
> As far as notifying the user goes, I'm quite keen to start doing this via the i3bar with a clever module but that is for another day. We need a nice way to avoid spamming users too much.

That's a great point yes

ultrabug:
Shouldn't we try to catch ImportError exceptions instead and act upon it instead of having to handle everything ourselves on the modules ?

tobes:
> Shouldn't we try to catch ImportError exceptions instead and act upon it instead of having to handle everything ourselves on the modules ?

If we do that then we can't do things like check the doc strings.  Or at least it makes it harder as we need to use the ast.

tobes:
I've now got my docstring checks working using ast so we can do this differently.  I'll close this and make a new PR once I clean that up.

I like the exception thrown in `post_config_hook()` idea but that can be a PR of it's own

PR #494, opened 2016-10-08
Add pulseaudio support to volume_status

guiniol:
Adds a new variable `backend` which can be either "alsa" or "pulseaudio". Adding other backend is only a matter of creating a wrapper class which extends the abstract `AudioBackend` class. If the backend specified doesn't exist, an exception is raised in the `post_config_hook`. This stops py3status.

In addition, I added a new variable to optionally color the `format_muted` string depending on the volume. I use it to know if it's safe to unmute.

### DIFF
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -72,30 +72,23 @@ import re
 
 from subprocess import check_output, call, DEVNULL
 
-from abc import ABCMeta, abstractmethod
-from six import with_metaclass
 
-
-class AudioBackend(with_metaclass(ABCMeta)):
+class AudioBackend():
     def __init__(self, device, channel):
         self.device = device
         self.channel = channel
 
-    @abstractmethod
     def get_volume(self):
-        pass
+        raise NotImplemented
 
-    @abstractmethod
     def volume_up(self, delta):
-        pass
+        raise NotImplemented
 
-    @abstractmethod
     def volume_down(self, delta):
-        pass
+        raise NotImplemented
 
-    @abstractmethod
     def toggle_mute(self):
-        pass
+        raise NotImplemented
 
 
 class AlsaBackend(AudioBackend):
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -69,7 +69,6 @@ NOTE:
 """
 
 import re
-import shlex
 
 from subprocess import check_output, call, DEVNULL
 
@@ -100,6 +99,10 @@ class AudioBackend(with_metaclass(ABCMeta)):
 
 
 class AlsaBackend(AudioBackend):
+    def __init__(self, device, channel):
+        AudioBackend.__init__(self, device, channel)
+        self.cmd = ['amixer', '-q', '-D', self.device, 'sset', self.channel]
+
     # return the current channel volume value as a string
     def _get_percentage(self, output):
 
@@ -134,8 +137,7 @@ class AlsaBackend(AudioBackend):
     def get_volume(self):
 
         # call amixer
-        output = check_output(shlex.split('amixer -D {} sget {}'.format(
-            self.device, self.channel))).decode('utf-8')
+        output = check_output(['amixer', '-D', self.device, 'sget', self.channel]).decode('utf-8')
 
         # get the current percentage value
         perc = self._get_percentage(output)
@@ -147,18 +149,15 @@ class AlsaBackend(AudioBackend):
 
     def volume_up(self, delta):
         # volume up
-        cmd = 'amixer -q -D {} sset {} '.format(self.device, self.channel)
-        call(shlex.split('{} {}%+'.format(cmd, delta)), stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ['{}%+'.format(delta)], stdout=DEVNULL, stderr=DEVNULL)
 
     def volume_down(self, delta):
         # volume down
-        cmd = 'amixer -q -D {} sset {} '.format(self.device, self.channel)
-        call(shlex.split('{} {}%-'.format(cmd, delta)), stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ['{}%-'.format(delta)], stdout=DEVNULL, stderr=DEVNULL)
 
-    def toggle_mute(self, delta):
+    def toggle_mute(self):
         # toggle mute
-        cmd = 'amixer -q -D {} sset {} '.format(self.device, self.channel)
-        call(shlex.split('{} toggle'.format(cmd)), stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ['toggle'], stdout=DEVNULL, stderr=DEVNULL)
 
 
 class PulseaudioBackend(AudioBackend):
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -78,6 +78,7 @@ class AudioBackend():
     def __init__(self, device, channel):
         self.device = device
         self.channel = channel
+        self.devnull = open(devnull, 'wb')
 
     def get_volume(self):
         raise NotImplemented
@@ -143,18 +144,15 @@ class AlsaBackend(AudioBackend):
 
     def volume_up(self, delta):
         # volume up
-        DEVNULL = open(devnull, 'wb')
-        call(self.cmd + ['{}%+'.format(delta)], stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ['{}%+'.format(delta)], stdout=self.devnull, stderr=self.devnull)
 
     def volume_down(self, delta):
         # volume down
-        DEVNULL = open(devnull, 'wb')
-        call(self.cmd + ['{}%-'.format(delta)], stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ['{}%-'.format(delta)], stdout=self.devnull, stderr=self.devnull)
 
     def toggle_mute(self):
         # toggle mute
-        DEVNULL = open(devnull, 'wb')
-        call(self.cmd + ['toggle'], stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ['toggle'], stdout=self.devnull, stderr=self.devnull)
 
 
 class PulseaudioBackend(AudioBackend):
@@ -165,25 +163,21 @@ class PulseaudioBackend(AudioBackend):
         self.cmd = ["pamixer", "--sink", self.device]
 
     def get_volume(self):
-        DEVNULL = open(devnull, 'wb')
         perc = check_output(self.cmd + ["--get-volume"]).decode('utf-8').strip()
-        muted = (call(self.cmd + ["--get-mute"], stdout=DEVNULL, stderr=DEVNULL) == 0)
+        muted = (call(self.cmd + ["--get-mute"], stdout=self.devnull, stderr=self.devnull) == 0)
         return perc, muted
 
     def volume_up(self, delta):
         # volume up
-        DEVNULL = open(devnull, 'wb')
-        call(self.cmd + ["-i", str(delta)], stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ["-i", str(delta)], stdout=self.devnull, stderr=self.devnull)
 
     def volume_down(self, delta):
         # volume down
-        DEVNULL = open(devnull, 'wb')
-        call(self.cmd + ["-d", str(delta)], stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ["-d", str(delta)], stdout=self.devnull, stderr=self.devnull)
 
     def toggle_mute(self):
         # toggle mute
-        DEVNULL = open(devnull, 'wb')
-        call(self.cmd + ["-t"], stdout=DEVNULL, stderr=DEVNULL)
+        call(self.cmd + ["-t"], stdout=self.devnull, stderr=self.devnull)
 
 
 class Py3status:
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -78,7 +78,10 @@ class AudioBackend():
     def __init__(self, device, channel):
         self.device = device
         self.channel = channel
-        self.devnull = open(devnull, 'wb')
+
+    def run_cmd(self, cmd):
+        with open(devnull, 'wb') as dn:
+            call(cmd, stdout=dn, stderr=dn)
 
     def get_volume(self):
         raise NotImplemented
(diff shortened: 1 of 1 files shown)


py3status/modules/volume_status.py
```diff
@@ -18,6 +20,9 @@
         (default 10)
     channel: Alsamixer channel to track.
         (default 'Master')
+    color_muted: set to 1 to color 'format_muted' depending on the volume
+        percent.
+        (default 0)
```
tobes: `color_xxx` is reserved and should not be used in this way.  It is for defining colors which should be accessed as `self.py3.COLOR_XXXX` in the module.

py3status/modules/volume_status.py
```diff
@@ -53,7 +58,8 @@
 ```
 
 Requires:
-    alsa-utils: (tested with alsa-utils 1.0.29-1)
+    alsa-utils: alsa backend (tested with alsa-utils 1.0.29-1)
+    pamixer: pulseaudio backend
```
tobes: We only need one of these.

py3status/modules/volume_status.py
guiniol: Indeed. I forgot. Anyhow, per your previous comment, I'll reserve this for another PR once this one is done.

py3status/modules/volume_status.py
guiniol: One or the other, which is why I added the `XXX backend`. I'll make it more explicit.

guiniol:
>     I think it is making things over complicated, especially the whole extra class stuff. Also if we did go down that route then we could make it much simpler eg just pass self to the AudioBackend and then get any vars we need rather than passing each button etc. But anyhow I think the AudioBackend is not needed.
> 
>     does using abc add anything? To me it just adds complication for no great gain
> 
>     Why not just have things like
> 
> def _volume_up(self):
>     if self.backend == 'pulseaudio':
>         ...
>     else:
>         ...
> 
> this feels more readable to me.
> 
>     If you really want to go down the AudioBackend route then why not just
> 
> class PulseAudio:
> 
>     def get_volume(self, device, channel):
>         ...
> 
>     def volume_up(self):
>         ...
> ...
> 

Actually, I prefer it with classes and abc for several reasons.
1. I think it makes the code more readable because there's only ONE `if`, in the `post_config_hook` instead of everywhere.
2. By using ABC, we are sure that each classes defines the necessary functions, and also, the `__init__` function only needs to be declared once.

>     is there a reason you are using run() rather than call()?

No particular reason. It's what I usually use, but I'm far from an expert. I'll switch to call.

>     the more code we can share eg in the on_click() the better, also if we add another backend later then it would be easier.

I agree, and you are right. I will instead add a `volume_up`, `volume_down`, `toggle_mute` function to each backend.

>     I'd do the whole color_mute idea separately as we should probably fix it but allow any colour trhe user wants. And it is not really a pulseaudio issue

Ok. Will do.

tobes:
> Actually, I prefer it with classes and abc for several reasons.
1. I think it makes the code more readable because there's only ONE if, in the post_config_hook instead of everywhere.

you haven't really answered all my poists especially why not
```
class PulseAudio:

    def get_volume(self, device, channel):
        ...

    def volume_up(self):
        ...
...
```
that would be a compromise and is nice and simple

> 2. By using ABC, we are sure that each classes defines the necessary functions, and also, the __init__ function only needs to be declared once.

I think that the `__init__` is
a) unneeded
b) passing too much and so creating noise

I think the whole ABC thing is overkill if you are writing a module plugin system then maybe but this is a really simple module and all the code is in a single file.

guiniol:
> I think that the __init__ is
> a) unneeded
> b) passing too much and so creating noise
> 
> I think the whole ABC thing is overkill if you are writing a module plugin system then maybe but this is a really simple module and all the code is in a single file.

This is because I wrote a `on_click` function in each class instead of having `on_click` in `Py3status` calling `volume_up` etc on the backend. With this modification, `__init__` only takes `device` and `channel` as args.
I really don't see why we shouldn't use class and inheritance. This is where these concepts are useful. `AlsaBackend` and `PulseaudioBackend` are classes which do the same thing but use different tools. They have the same functions and only the specifics of the tools they use are different.

py3status/modules/volume_status.py
```diff
@@ -68,44 +71,35 @@
...
-    threshold_degraded = 50
-    volume_delta = 5
 
-    # compares current volume to the thresholds, returns a color code
-    def _perc_to_color(self, string):
-        try:
-            value = int(string)
-        except ValueError:
-            return self.py3.COLOR_BAD
+class AudioBackend(with_metaclass(ABCMeta)):
```
pydsigner: I'd like to second the request to not use `ABCMeta`. I'm completely fine with using classes and inheritance, but instead of using `@abstractmethod`, either `raise NotImplemented` or just ignore it.

py3status/modules/volume_status.py
```diff
@@ -149,6 +143,97 @@ def current_volume(self):
         # get info about channel mute status
         muted = self._get_muted(output)
 
+        return perc, muted
+
+    def volume_up(self, delta):
+        # volume up
+        cmd = 'amixer -q -D {} sset {} '.format(self.device, self.channel)
```
pydsigner: For these, wouldn't it be simpler to do a line like this rather than passing through `shlex.split()`?
```cmd = ['amixer', '-q', '-D', self.device, self.channel, '{}%-'.format(delta)]```

Comment (tobes) [inline py3status/modules/volume_status.py]: True, to be fair `volume_up()` etc should probably just return a command to run maybe even change the name to `volume_up_cmd()`

File: py3status/modules/volume_status.py (hunk 3 above)
Comment (tobes) [inline py3status/modules/volume_status.py]: I still think there is no benefit to inheritance in this instance as the class is just a holder for some methods.

File: py3status/modules/volume_status.py (hunk 4 above)
Comment (guiniol) [inline py3status/modules/volume_status.py]: That was the original code from volume_status so I didn't change it. Will do.

@tobes: I don't think they should return command because hypothetically, a new backend could not have a `toggle` command and would need to execute two commands, one to get the status and one to set the new status. This is just an example, but this is why I think the backend classes should do the interaction.

File: py3status/modules/volume_status.py (hunk 3 above)
Comment (pydsigner) [inline py3status/modules/volume_status.py]: For one thing, it's overkill; (see http://stackoverflow.com/a/19328146/1795505)
For another, you've just pulled in a dependency on `six` that wasn't there before, for no measurable gain. Better to keep the dependencies low and accessibility high unless there's a strong use case.

Comment (guiniol) [inline py3status/modules/volume_status.py]: Very well. I'm making the changes to `raise NotImplemented`.

File: py3status/modules/volume_status.py (hunk 4 above)
Comment (guiniol) [inline py3status/modules/volume_status.py]: > Although I agree this is a valid concern I'm not sure if this is likely. I'm in favour of keeping things as simple as possible if we needed to refactor for such an event then fine but we could do it then, rather than having the complications and never using it.

I don't see how it's a complication. I understand that this means there are two `call` instead of one, but it makes sense to keep it in the backend code and I don't think it makes the code less readable.

> I just want nice easy readable code that is easy to maintain. where the call is made doesn't matter too much

In that case, I'd rather have it in the backend to avoid possible refactoring at a later date.

Comment (tobes) [inline py3status/modules/volume_status.py]: fine but can we do the `call()` in one place this file now has `DEVNULL = open(devnull, 'wb')` seven times which feels overkill to me if you are going to have a base class at least use it for something

Comment (guiniol) [inline py3status/modules/volume_status.py]: Done. I open it in the `__init__` of the base class. I didn't do it at first because I try to open files as little as possible.

Comment (tobes) [inline py3status/modules/volume_status.py]: I was actually thinking more like this in the base class
```
def run_cmd(self, cmd):
    with open(devnull, 'wb') as dn:
        call(cmd , stdout=dn, stderr=dn)
```

Comment (guiniol) [inline py3status/modules/volume_status.py]: Much better. I'll put it in.

File: py3status/modules/volume_status.py [hunk 5]
```diff
@@ -135,20 +130,107 @@ def _get_muted(self, output):
...
         # get the current percentage value
         perc = self._get_percentage(output)
 
         # get info about channel mute status
         muted = self._get_muted(output)
 
+        return perc, muted
+
+    def volume_up(self, delta):
+        # volume up
```
tobes: This comment does not add much the method is already called `volume_up()`

py3status/modules/volume_status.py
```diff
@@ -66,45 +69,37 @@
...
-        text = self.py3.safe_format(format, {'percentage': percentage})
-        return text
+    def volume_up(self, delta):
+        raise NotImplemented
+
+    def volume_down(self, delta):
+        raise NotImplemented
+
+    def toggle_mute(self):
+        raise NotImplemented
```
tobes: I think these methods add little as we will get a AttributeError exception anyway.

py3status/modules/volume_status.py
```diff
@@ -135,20 +130,107 @@ def _get_muted(self, output):
...
+    device = 'default'
+    format = u'♪: {percentage}%'
+    format_muted = u'♪: muted'
+    threshold_bad = 20
+    threshold_degraded = 50
+    volume_delta = 5
+
+    def post_config_hook(self):
+        if self.backend == 'alsa':
+            self.backend = AlsaBackend(self.device, self.channel)
```
tobes: Why not just `self.backend = AlsaBackend(self)`?

Personally I think you should use a different name not overwrite `backend`

py3status/modules/volume_status.py
```diff
@@ -135,20 +130,107 @@ def _get_muted(self, output):
...
+        # toggle mute
+        self.run_cmd(self.cmd + ['toggle'])
+
+
+class PulseaudioBackend(AudioBackend):
+    def __init__(self, device, channel):
+        AudioBackend.__init__(self, device, channel)
+        if self.device == 'default':
+            self.device = "0"
+        self.cmd = ["pamixer", "--sink", self.device]
```
tobes: I don't like the whole `AudioBackend.__init__(self, device, channel)` thing here.

You could just get `AudioBackend.__init__()` to call `self.setup()` that might be nicer to read.  We should spend much more time reading code than writing it.

py3status/modules/volume_status.py
guiniol: But it shows which methods need to be implemented and the `NotImplemented` error is quite explicit.

py3status/modules/volume_status.py
guiniol: Indeed. Copy-pasta'd. I'll remove them/

py3status/modules/volume_status.py
guiniol: I like that, but at the same time, it kind of goes around the point of `__init__`.

py3status/modules/volume_status.py
tobes: True, but this is a py3status module not a framework.  To me it is just added noise.

If someone writes a new backend
a) they should look at the code
b) it should get reviewed

guiniol: Ok. It's in the latest commit.

py3status/modules/volume_status.py
tobes: > it kind of goes around the point of `__init__`

Does it?  They have different functions `__init__()` is about instance creation.  `setup()` would be a helper a bit like `post_config_hook()`  So they have different functions,  but ultimately readability is key

guiniol: It's in the latest commit anyhow.

tobes:
Seeing as this is such a re-write of the module it would be nice to also clean it up whilst we are here.

The alsa get_volume() is imho pretty nasty, it could be rewriten much more compactly eg using a single regex and all the code stay in get_volume() rather than in little functions.

The module should also be able to guess the correct backend to use eg `py3.check_commands(['amixer', 'pamixer'])`

guiniol:
> The alsa get_volume() is imho pretty nasty, it could be rewriten much more compactly eg using a single regex and all the code stay in get_volume() rather than in little functions.

I agree. I didn't do it because I'm not the original owner of the module. Will do, gladly.

> The module should also be able to guess the correct backend to use eg py3.check_commands(['amixer', 'pamixer'])

I don't think there is an easy way. For example, I have both commands installed on my system, but I want to use pamixer. I don't know if that's usual or not, but it wouldn't probably be robust. Or it should only guess if the variable is not specified. Is it possible inside a module to know if the variable is specified in the configuration?

tobes:
> Is it possible inside a module to know if the variable is specified in the configuration?

The easy way is just set the default value to None, then you can test for that and set it if the user hasn't set it.

I also think that we should change the `backend` configuration parameter to `command` or something like that.  My reasoning is that for pulseaudio we have other options eg `pactl` which other users might prefer (installed by default etc)

guiniol:
> The easy way is just set the default value to None, then you can test for that and set it if the user hasn't set it.

Fair enough. In case of both commands available, I can default to the previous default ^^

> I also think that we should change the backend configuration parameter to command or something like that. My reasoning is that for pulseaudio we have other options eg pactl which other users might prefer (installed by default etc)

Hum. Ok. But it still requires a backend class.

tobes:
> But it still requires a backend class.

Yeah it's just about future proofing.  so currently `command = alsa | pamixer`  then later we might add a `pactl` option.

The alsa get_volume() looks much nicer now Thanks :)

guiniol:
> Yeah it's just about future proofing. so currently command = alsa | pamixer then later we might add a pactl option.

Just making sure we're on the same page.

> The alsa get_volume() looks much nicer now Thanks :)

No worries ^^

py3status/modules/volume_status.py
```diff
@@ -16,9 +16,12 @@
...
         (default 10)
-    channel: Alsamixer channel to track.
+    channel: Alsamixer channel to track (ignored by pulseaudio)
         (default 'Master')
-    device: Alsamixer device to use.
+    command: Choose between "amixer" and "pamixer". If None, try to guess based
+        on available commands.
+        (default None)
+    device: Device to use.
         (default 'default')
```
tobes: I think we should change the default to `None`.

My thinking is that as it is `default` for alsa but `0` for pulseaudio that having it as `default` is misleading.  We can then just set it depending on which backend is used like pamixer already does.

The documentation can then be something like  `device: Device to use. Will use default device if none specified.`

py3status/modules/volume_status.py
```diff
@@ -66,11 +70,72 @@
 """
 
 import re
-import shlex
+from os import devnull
```
tobes: We don't really need this blank line (I'm getting really picky now)

py3status/modules/volume_status.py
guiniol: Ok (you're talking about `device`, right?). This way the default is backend dependant. I'll do the same for `channel`.

tobes:
One last thing from the start of the docstring

> Display current sound volume using amixer.

Maybe just `Display current sound volume.` would be better

guiniol:
And done!

tobes:
Cool good work @guiniol

ultrabug:
Awesome work, congratz @guiniol and thanks for the extensive review @tobes and @pydsigner

guiniol:
Glad it's in. Thanks @tobes for all the handholding, and thanks @pydsigner for siding with me at least once ^^

PR #499, opened 2016-10-11
Docstring test improvements and docstring updates

tobes:
The docstring tests are improved.

We now use ast to examine the module.  This allows us to test modules with unfulfilled dependencies.  It also allows the checking of the defined parameters in a module for alphabeticalness.  It also now catches unicode python2 issues.

Docstring/attribute order/unicode fixes for various modules are here to


ultrabug:
Impressive work :open_mouth: 

That's some serious ast mastery, thanks !

Issue #502, opened 2016-10-11
How to copy printed value when clicking on it ?


Spheerys:
I'm looking for a way to copy the printed value with a left click on it.
What is the best way to do this ?

tobes:
Currently there is no method for doing this.

A basic solution could be added without too much work but I wonder if there are other similar use cases eg sending output to other commands that would benefit from a more flexible solution.

Spheerys:
Thanks for your answer.

My use case is the following : I'm printing an OTP auth code on the bar, which is refresh each 30 seconds. 
It could be useful for me to copy this value easily with a mouse click to past it on my website

tobes:
now #504 is in master you can now do this.

```
my module {
    on_click 1 = 'exec echo $OUTPUT | xclip -i'
}
```

Spheerys:
:+1:

PR #504, opened 2016-10-12
Allow module output to be used in commands

tobes:
#502 (copying module output to the clipboard) was an interesting idea.

So I've come up with a way that we can do this.  The idea is to be flexible so users can find their own uses for this.

With this PR I can now copy the output of a module to the clipboard on a left click

```
my module {
    on_click 1 = 'exec echo $OUTPUT | xclip -i'
}
```

Where `$OUTPUT` will be replaced by the output of the module.  `$OUTPUT_PART` will give part of the output if the module is a composite.  I'm quite happy to change the names for these substitutions if people can think of better ones.

In the future we could also extend this to get variables out of the module eg the output filename from screenshot module so a little thought around the naming for that might be nice too.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -15,7 +15,7 @@ py3status documentation
 [Custom click events](#on_click)
 
 * [Special on_click commands](#on_click_commands)
-* [Example config](#on_click_example)
+* [Module data and on_click commands](#on_click_data)
 
 [Writing custom modules](#writing_custom_modules)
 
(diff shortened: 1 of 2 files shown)


ultrabug:
I find the idea and naming good. I only think that we're missing documentation here please ?

tobes:
I've added some documentation.  I ended up splitting the on_click example into the sections they relate to as it made more sense.

ultrabug:
Sorry @tobes but since you removed the `Example on_click usage on i3status.conf` section, the example displayed below the refresh and refresh_all special section is not relevant

tobes:
Sorry my mistake I just saw something starting with r

should be fixed now

ultrabug:
Thanks a lot !

PR #505, opened 2016-10-12
fedora_updates: Add option for checking security updates

jakubjedelsky:
Commit adds option `check_security` to be able to ignore info about
security fixes. It's useful info, but
 * not everyone is interesting in it (e.g. when running update everytime
   he/she sees a number > 0 :))
 * `dnf updateinfo` reads from external sources, which can be
   time and CPU consuming

The option is enabled by default because of backwards compatibility.

### DIFF
--- a/py3status/modules/fedora_updates.py
+++ b/py3status/modules/fedora_updates.py
@@ -9,6 +9,8 @@ Additionally check if any update security notices.
 Configuration parameters:
     cache_timeout: How often we refresh this module in seconds
         (default 600)
+    check_security: Check for security updates
+        (default True)
     format: Display format to use
         (default 'DNF: {updates}')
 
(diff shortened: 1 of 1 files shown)


py3status/modules/fedora_updates.py
```diff
@@ -61,15 +64,15 @@ def check_updates(self):
...
+            if self.check_security and not self._security_notice and self._updates != updates:
                 output, error = subprocess.Popen(
                     ['dnf', 'updateinfo'],
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE).communicate()
                 notices = str(output)
                 self._security_notice = len(self._reg_ex_sec.findall(notices))
                 self._updates = updates
-            if self._security_notice:
+            if self.check_security and self._security_notice:
```
tobes: We don't actually need to check `self.check_security` here as `self._security_notice` is always False, but I don't think it matters.

tobes:
@jakubjedelsky Thanks for this enhancement.

ultrabug:
Thank you @jakubjedelsky !

Issue #484, opened 2016-09-14
mpd_status (UnicodeDecodeError) i3status.py line 282, with Python 2 and unicode in track names


jurajpetrik:
I just tried installing py3status and on starting it (by i3 or through shell) I get an error saying

> py3status: Setup error (UnicodeDecodeError) i3status.py line 282. Please try to fix this and reload i3wm.

screenshot: http://i.imgur.com/ysLDRAJ.png

```
$ py3status --version
py3status version 3.0 (python 3.5.2)
```

bar config from i3 config file

```
bar {
    font pango: System San Francisco Display, FontAwesome 10
    status_command py3status
    #tray_output primary
    colors {
        background $bg-color
        separator #757575
        #                  border             background         text
        focused_workspace  $bg-color          $bg-color          $text-color
        inactive_workspace $inactive-bg-color $inactive-bg-color $inactive-text-color
        urgent_workspace   $urgent-bg-color   $urgent-bg-color   $text-color
    }
}
```

tobes:
Thanks for reporting this could you add your i3status.conf.

jurajpetrik:
I don't have any custom i3status.conf. There is only the default one in `/etc/i3status.conf`

ultrabug:
Thanks for reporting this @jurajpetrik !

Since a lot has changed on the i3status.conf parser since 3.0, would you please try by installing 3.1 and testing again first ?

Thanks !

YetAnotherNerd:
getting this error from multiple modules, guess it is related

> Instance `sysdata`, user method `sysData` failed (UnicodeDecodeError) sysdata.py line 153.
> Instance `mpd_status`, user method `current_track` failed (UnicodeDecodeError) mpd_status.py line 92.

```
% py3status --version
py3status version 3.1 (python 2.7.10)
```

tobes:
@YetAnotherNerd thanks for the report, could you share the config settings for those 2 modules.
It looks like a different issue to the original report but will need fixing up.

@ultrabug some of this should be fixed by the safe_format changes but some like mpd_status look like they will still need fixing though the fix fill likely be `''` -> `u''`

YetAnotherNerd:
@tobes 
```
mpd_status {
	hide_when_paused = true
	hide_when_stopped = true
	format = '[[[%artist% - ]%title%]|[%file%]]'
}
sysdata {
	format = "CPU: {cpu_usage}% {cpu_temp}"
}
```
seems mpd_status only fails for some tracks

[...]

PR #506, opened 2016-10-12
Formatter add attr_getter

tobes:
Issue #484 really needs mpd_status to be updated to use safe_format().  The issue with this is that the module gets information (placeholders) via a function, rather than providing a dict like most modules.  This could also potentially be useful for other modules.

This PR adds the ability to provide an attr_getter function to safe_format() that values will be got from.  This functionality is also available via build_composite().

If this PR is merged we can then update mpd_status to use safe_format().  We can add a shim to the module so that both `%...%` and `{...}` formats can still be used.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -923,9 +923,10 @@ The following functions can be registered
 >
 > Note: This function should only be used by containers.
 
-__safe_format(format_string, param_dict=None)__
+__safe_format(format_string, param_dict=None, force_composite=False,
+attr_getter=None)__
 
-Parser for advanced formating.
+Parser for advanced formatting.
 
 Unknown placeholders will be shown in the output eg `{foo}`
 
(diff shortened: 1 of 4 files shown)


ultrabug:
This one I have trouble understanding right, maybe my current state of health has something to do with this.

Meanwhile, maybe a rebase could help this PR

tobes:
The idea is that we can now pass a function to safe_format().  When the formatter is looking for a value to substitute it

looks in the param_dict for the value, then it looks in the py3status module and with this PR it will then call the `attr_getter` function if it exists with the name and use the returned value.

The reason is how mpd_status works.  The other option is to rewrite mpd_status much more invasivly.

ultrabug:
I see, it's not obvious at first glance thank you :confused:

PR #507, opened 2016-10-12
Don't go to repo when notifications == 'all'

rixx:
If you are monitoring all notifications, but there are none, a click on
the module took you to the repo page (which would probably be
unconfigured). Now it will take you to the GitHub homepage (since the
empty notification page would be kind of useless).

### DIFF
--- a/py3status/modules/github.py
+++ b/py3status/modules/github.py
@@ -16,8 +16,8 @@ Configuration parameters:
         see above.
         (default None)
     button_action: Button that when clicked opens the Github notification page
-        if notifications, else the project page for the repository . Setting to
-        `0` disables.
+        if notifications, else the project page for the repository if there is
+        one (otherwise the github home page). Setting to `0` disables.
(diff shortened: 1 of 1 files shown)


tobes:
For my usage I want to be able to visit the repo when there are no notifications, however I can see that for some users they might want to just go to the main github page when noo notifications.

I think we need a cleverer way to decide.  One option would be if no repo was set to None.  Maybe we can find a better solution.

rixx:
Huh, yeah, I didn't think about a usecase where you would receive notifications for `all`, but have `repo` set regardless (I'd have thought that you would have `notifications = "repo"` in that case?)

But I've adjusted the PR to your request regardless.

tobes:
I think it'd also be good to update the documentation to reflect this change.

rixx:
Good point, done.

ultrabug:
Thanks @rixx !

PR #508, opened 2016-10-13
Go to a space's website onclick

rixx:
I started enjoying onclick events more and more, so I thought going to a hackerspace's website (which is in the data provided via spaceapi) might be nice.

### DIFF
--- a/py3status/modules/spaceapi.py
+++ b/py3status/modules/spaceapi.py
@@ -4,6 +4,8 @@
 Display if your favorite hackerspace is open or not.
 
 Configuration parameters:
+    button_url: Button that when clicked opens the URL sent in the space's API.
+        Setting to None disables. (default 3)
     cache_timeout: Set timeout between calls in seconds (default 60)
     closed_color: color if space is closed (default None)
     closed_text: text if space is closed, strftime parameters
@@ -23,12 +25,15 @@ import codecs
 import datetime
 import json
 import urllib.request
+import shlex
+import subprocess
 
 
 class Py3status:
     """
     """
     # available configuration parameters
+    button_url = 3
     cache_timeout = 60
     closed_color = None
     closed_text = 'closed'
@@ -56,6 +61,7 @@ class Py3status:
             reader = codecs.getreader("utf-8")
             data = json.load(reader(json_file))
             json_file.close()
+            self._url = data.get('url')
 
             if(data['state']['open'] is True):
                 if self.open_color:
@@ -89,6 +95,14 @@ class Py3status:
             response['full_text'] = ''
 
         return response
+
+    def on_click(self, event):
+        button = event['button']
+        if self._url and self.button_url == button:
+            cmd = 'xdg-open {}'.format(self._url)
+            subprocess.call(shlex.split(cmd))
+            self.py3.prevent_refresh()
+
 
 if __name__ == "__main__":
     """


py3status/modules/spaceapi.py
```diff
@@ -56,6 +61,7 @@ def check(self):
             reader = codecs.getreader("utf-8")
             data = json.load(reader(json_file))
             json_file.close()
+            self.url = data['url']
```
tobes: you should probably use `data.get('url')` in case the `url` key is not in data.

py3status/modules/spaceapi.py
```diff
@@ -23,12 +25,15 @@
...
 import urllib.request
+import shlex
+import subprocess
 
 
 class Py3status:
     """
     """
     # available configuration parameters
+    button_action = 3
```
tobes: would `button_open` or some other name be more descriptive?  open doesn't feel perfect but action just feels too vague.

py3status/modules/spaceapi.py
```diff
@@ -90,6 +96,15 @@ def check(self):
 
         return response
 
+    def on_click(self, event):
+        button = event['button']
+        if self.button_action and self.button_action == button:
```
tobes: Here we could just use `if self.button_action == button:` as button can only be 1, 2, 3, 4 or 5.  I have used the longer version in the past so not a big deal.

It could also be combined with the `if self.url:`

py3status/modules/spaceapi.py
```diff
@@ -4,6 +4,8 @@
 Display if your favorite hackerspace is open or not.
 
 Configuration parameters:
+    button_action: Button that when clicked opens the URL sent in the space's API.
+        Setting to `0` disables. (default 3)
```
tobes: Although other modules use `0` I think that `None` to disable sounds better but both will work in all the uses I've seen.

CHANGELOG
```diff
@@ -6,6 +6,7 @@ version 3.3 (milestone)
 * fedora_updates module: add check_security option, by @jakubjedelsky
 * github module: open github home page if no repo is configured and you're gettin all notifications, by @rixx
 * volume_status module: add pulseaudio support, by @guiniol
+* spaceapi module: go to hackerspace website onclick, by @rixx
```
tobes: Generally we haven't added to the change log in PRs I think it may also lead to lots of unnecessary merge conflicts if we start doing this.

@ultrabug what is your take on this

py3status/modules/spaceapi.py
rixx: I think `url_button` might work. I was mirroring the `github` module here. Changed.

py3status/modules/spaceapi.py
rixx: It wouldn't be well-formed in the spaceapi standard, but you're right; better safe than sorry. Changed.

py3status/modules/spaceapi.py
rixx: Good to know. Changed.

CHANGELOG
rixx: I'll remove this then. Sorry.

py3status/modules/spaceapi.py
tobes: I'm very keen for all these type things to be `button_xxxx` to be consistent across modules.  Historically it's been a bit of a free for all.

rixx:
Pushed the requested changes. Thank you for the review! Would you look at it again? :)

py3status/modules/spaceapi.py
tobes: Never trust data is my view

tobes:
@rixx looks good to me.

rixx:
@tobes your reviews are very appreciated, thanks!

py3status/modules/spaceapi.py
tobes: > why do you redefine the url configuration parameter here?

You are right it should be `self._url` or something like that.  I was looking at the patch more than the whole file

rixx: D'oh. Yeah, totally overlooked that `url` was already taken. Changed to `_url`

rixx:
@ultrabug I fixed the attribute I was overriding, any other things you'd want me to change?

ultrabug:
No thanks a lot @rixx !

PR #510, opened 2016-10-14
modules/graphite: Allow to set the http and https proxies

btall:


tobes:
I don't use this module but is there a reason that someone might supply both a http and https proxy?

Could we not just have a `proxy` config parameter and then work out which schema http/https automatically?

btall:
> Could we not just have a proxy config parameter and then work out which schema http/https automatically?

For lot of reasons, e.g:
* For doesn't lose the flexibility offered by python-requests, _If the module `python-requests` allows it, `py3status` should also make it possible._
* Explicit is better than implicit. (PEP 20: Zen of python) :)

tobes:
>   For doesn't lose the flexibility offered by python-requests, If the module python-requests allows it, py3status should also make it possible.

The library that a module uses should not be relevant to the user and maybe we will change it at some point.  From a quick scan of requests we will only use one of the supplied proxies anyhow as graphite url will either be http or https.

>    Explicit is better than implicit. (PEP 20: Zen of python) :)

When you are coding sure, but for adding config parameters to py3status modules (fewer is better: Tau of @tobes)

As I say I've not used this module but I think that we should just supply a single proxy to the module and we should pass that to requests for both the http and https proxies there is little value in suppliying a config option that is then ignored.

Lujeni:
Imho, this kind of module (using requests) should use `kwargs` to support all extra parameters.

> The library that a module uses should not be relevant to the user

I agree, however in this case you should implement each parameters of the `requests` module.
For example, this module support SSL without the verification part (http://docs.python-requests.org/en/master/user/advanced/?highlight=SSL#ssl-cert-verification).

Use `kwargs` (code side) + `extra_params` (configuration side) could solve all problems:
- [x] Explicit is better than implicit. (PEP 20: Zen of python) :)
- [x] For doesn't lose the flexibility offered by python-requests, If the module python-requests allows it, py3status should also make it possible.
- [x] fewer is better: Tau of @tobes

ultrabug:
@Lujeni Mr. Consensus :lollipop: 

> Use kwargs (code side) + extra_params (configuration side) could solve all problems:

How would those kwargs become configuration parameters exactly ? I'm not sure I understand how you want to do it.

tobes:
As an aside,

I'm quite keen to make requests a formal dependency of py3status, this would allow us to add a `Py3.get_url()` style function that could deal with all the timeouts etc to simplify things for the modules whilst ensuring robustness.  Doing this might also allow us to move some of these settings like timeouts out of the modules configs and be more general (they could still be used in specific modules eg like colors are).

I think the SSL support is an interesting thing and could be useful but it is important to keep modules configs separate from library specifics.  Ultimately requests is just a wrapper for the likes of urllib.

py3status/modules/graphite.py
```diff
@@ -11,14 +11,11 @@
...
-    http_proxy: You can configure individual requests for the scheme HTTP,
-        see http://docs.python-requests.org/en/master/user/advanced/#proxies
-        (default: None)
-    https_proxy: You can configure individual requests for the scheme HTTPS,
-        see http://docs.python-requests.org/en/master/user/advanced/#proxies
-        (default: None)
     http_timeout: HTTP query timeout to graphite.
         (default 10)
+    proxy: You can configure the proxy with HTTP or HTTPS.
+        See: http://docs.python-requests.org/en/master/user/advanced/#proxies
```
tobes: I'd remove this link.

1) It is specific to requests and the module should be neutral about the lib it happens to use.

2) The requests documentation is not very helpful.  I had to read the code to work out what the http/https was all about.

3) a simple example would be better for users to understand how to use

ultrabug:
Thanks a lot @btall !

PR #511, opened 2016-10-15
Battery measurement using sys interface

m45t3r:
Implement measurement of battery information using `/sys/class/power_supply` provided by the kernel itself instead of acpi. More robust, however it may give a wonky time measurement and may not work in all cases, for example this one: https://bbs.archlinux.org/viewtopic.php?id=131025

### DIFF
--- a/py3status/modules/README.md
+++ b/py3status/modules/README.md
@@ -248,11 +248,18 @@ Configuration parameters:
   - `hide_when_full` hide any information when battery is fully charged (when
     the battery level is greater than or equal to 'threshold_full')
     *(default False)*
+  - `measurement_mode` either 'acpi' or 'sys'. 'sys' should be more robust and
+  does not have any extra requirements, however the time measurement may be off
+  or not work in some cases
+    *(default sys)*
   - `notification` show current battery state as notification on click
     *(default False)*
   - `notify_low_level` display notification when battery is running low (when
     the battery level is less than 'threshold_degraded')
     *(default False)*
+  - `sys_battery_path` set the path to kernel sys' battery interface. Used only
+  when `measurement_mode=sys`.
+    *(default /sys/class/power_supply/)*
   - `threshold_bad` a percentage below which the battery level should be
     considered bad
     *(default 10)*
@@ -295,7 +302,7 @@ Obsolete configuration parameters:
     *(default None)*
 
 Requires:
-  - the `acpi` command line
+  - the `acpi` command line (only if `measurement_mode='acpi'`)
 
 **author** shadowprince, AdamBSteele, maximbaz, 4iar
 
--- a/py3status/modules/README.md
+++ b/py3status/modules/README.md
@@ -248,17 +248,17 @@ Configuration parameters:
   - `hide_when_full` hide any information when battery is fully charged (when
     the battery level is greater than or equal to 'threshold_full')
     *(default False)*
-  - `measurement_mode` either 'acpi' or 'sys'. 'sys' should be more robust and
-  does not have any extra requirements, however the time measurement may not
-  work in some cases
-    *(default acpi)*
+  - `measurement_mode` either 'acpi' or 'sys', or None to autodetect. 'sys'
+  should be more robust and does not have any extra requirements, however the
+  time measurement may not work in some cases
+    *(default None)*
   - `notification` show current battery state as notification on click
     *(default False)*
   - `notify_low_level` display notification when battery is running low (when
     the battery level is less than 'threshold_degraded')
     *(default False)*
   - `sys_battery_path` set the path to kernel sys' battery interface. Used only
-  when `measurement_mode=sys`.
+  when `measurement_mode=sys`
     *(default /sys/class/power_supply/)*
   - `threshold_bad` a percentage below which the battery level should be
     considered bad
--- a/py3status/modules/README.md
+++ b/py3status/modules/README.md
@@ -249,9 +249,9 @@ Configuration parameters:
     the battery level is greater than or equal to 'threshold_full')
     *(default False)*
   - `measurement_mode` either 'acpi' or 'sys'. 'sys' should be more robust and
-  does not have any extra requirements, however the time measurement may be off
-  or not work in some cases
-    *(default sys)*
+  does not have any extra requirements, however the time measurement may not
+  work in some cases
+    *(default acpi)*
   - `notification` show current battery state as notification on click
     *(default False)*
   - `notify_low_level` display notification when battery is running low (when
@@ -304,7 +304,7 @@ Obsolete configuration parameters:
 Requires:
   - the `acpi` command line (only if `measurement_mode='acpi'`)
 
-**author** shadowprince, AdamBSteele, maximbaz, 4iar
+**author** shadowprince, AdamBSteele, maximbaz, 4iar, m45t3r
 
 **license** Eclipse Public License
 
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -83,6 +83,7 @@ Requires:
 from __future__ import division  # python2 compatibility
 from time import time
 from re import findall
+from glob import iglob
 
 import math
 import subprocess
@@ -178,38 +179,7 @@ class Py3status:
             # Backwards compatibility for '{}' option in format string
             self.format = self.format.replace('{}', '{percent}')
 
-    def _extract_battery_information_from_acpi(self, acpi_battery_lines):
-        """
-        Extract the percent charged, charging state, time remaining,
-        and capacity for a battery, given a list of two strings from acpi
-        """
-        battery = dict()
-        battery["percent_charged"] = int(findall("(?<= )(\d+)(?=%)",
-                                                 acpi_battery_lines[0])[0])
-        battery["charging"] = "Charging" in acpi_battery_lines[0]
-        battery["capacity"] = int(findall("(?<= )(\d+)(?= mAh)",
-                                          acpi_battery_lines[1])[1])
-
-        # ACPI only shows time remaining if battery is discharging or charging
-        try:
-            battery["time_remaining"] = ''.join(findall(
-                "(?<=, )(\d+:\d+:\d+)(?= remaining)|"
-                "(?<=, )(\d+:\d+:\d+)(?= until)", acpi_battery_lines[0])[0])
-        except IndexError:
-            battery["time_remaining"] = '?'
-
-        return battery
-
-    def _hms_to_seconds(self, t):
-        h, m, s = [int(i) for i in t.split(':')]
-        return 3600 * h + 60 * m + s
-
-    def _seconds_to_hms(self, secs):
-        m, s = divmod(secs, 60)
-        h, m = divmod(m, 60)
-        return "%d:%02d:%02d" % (h, m, s)
-
-    def _refresh_battery_info(self):
+    def _extract_battery_information_from_acpi(self):
         '''
         Get the battery info from acpi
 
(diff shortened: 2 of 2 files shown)


tobes:
Nice work, a few thoughts.

If I run 2 copies of the module one using acpi and the other sys I get different time remaining.  I think you are doing a different calculation than acpi is.  Is there a reason for this?  Interestingly i3status/battery gives quite different info.

You are checking all my batteries not just the one specified by battery_id

Is there a reason that you don't just read /sys/class/power_supply/BAT0/uevent to get the info?

measurement_mode this should for now default to acpi as using sys changes behaviour (my battery now claims to be doing much better) but should default to sys if no acpi command exists,

tobes:
you need to watch for this 

```
ZeroDivisionError: division by zero
  File "/home/toby/dev/py3status/py3status/module.py", line 433, in run
    response = method()
  File "/home/toby/dev/py3status/py3status/modules/battery_level.py", line 140, in battery_level
    self._refresh_battery_info()
  File "/home/toby/dev/py3status/py3status/modules/battery_level.py", line 285, in _refresh_battery_info
    battery_list = self._extract_battery_information_from_sys()
  File "/home/toby/dev/py3status/py3status/modules/battery_level.py", line 267, in _extract_battery_information_from_sys
    time_in_secs = battery["capacity"] / power_now * 3600
```

m45t3r:
> If I run 2 copies of the module one using acpi and the other sys I get different time remaining. I think you are doing a different calculation than acpi is. Is there a reason for this? Interestingly i3status/battery gives quite different info.

Solved in commit c08fc8e.

> You are checking all my batteries not just the one specified by battery_id

Same behavior as old acpi module. It is `_refresh_battery_status` module that actually select the correct battery in battery_list to display.

If you think it would be better, I can just parse the batteries that the user selected.

> Is there a reason that you don't just read /sys/class/power_supply/BAT0/uevent to get the info?

There was. I thought that parsing uevent file would be less safe than just parsing the variables itself. However parsing uevent is a one time event, so I decided to switch. Commit e159176.

> measurement_mode this should for now default to acpi as using sys changes behaviour (my battery now claims to be doing much better) but should default to sys if no acpi command exists,

Test commit c08fc8e, it should be the same now since I copied the code from acpi itself.

> you need to watch for this

Fixed in commit d5933aa. However, I am just emulating the behavior in acpi module here, we could just return that the battery is fully charged instead of the string "?". Thoughts?

tobes:
> I can set acpi to default, however I saw their code and it basically does what this code does now. 

I think this would be good.  If nothing else we can check how the sys version works and fix any bugs we encounter before flipping if we decide to do that.  For the moment I'll run both versions and see how they do.

We could try to be clever eg check if `/sys/class/power_supply` or what ever user supplied exists and if acpi program is installed to decide.  It would be easy to do in the modules `post_config_hook()` we can check for acpi via `self.py3.check_commands(['acpi'])` etc.  Ideally we can get rid of the whole mode thing.  Also not sure how this works for BSD etc. Maybe @ultrabug has a view.

Anyhow I'll do some testing of what you've done and see how it's looking.

> battery is fully charged instead of the string "?". Thoughts?

I never really liked the '?' however it won the vote.  maybe now we have better formatting we can revisit this decision but I think that might be better to do outside of this PR.

m45t3r:
> I think this would be good. If nothing else we can check how the sys version works and fix any bugs we encounter before flipping if we decide to do that. For the moment I'll run both versions and see how they do.

Commit 2c80d55.

> We could try to be clever eg check if /sys/class/power_supply or what ever user supplied exists and if acpi program is installed to decide. It would be easy to do in the modules post_config_hook() we can check for acpi via self.py3.check_commands(['acpi']) etc. Ideally we can get rid of the whole mode thing. 

I don't really like to be too clever, however I will see what is possible to do.

> Also not sure how this works for BSD etc. Maybe @ultrabug has a view.

I thought acpi client was Linux only.

m45t3r:
> We could try to be clever eg check if /sys/class/power_supply or what ever user supplied exists and if acpi program is installed to decide. It would be easy to do in the modules post_config_hook() we can check for acpi via self.py3.check_commands(['acpi']) etc.

Commit c160319

py3status/modules/battery_level.py
```diff
@@ -95,6 +104,8 @@
 FORMAT = u"{icon}"
 FORMAT_NOTIFY_CHARGING = u"Charging ({percent}%)"
 FORMAT_NOTIFY_DISCHARGING = u"{time_remaining}"
+SYS_BATTERY_PATH = u"/sys/class/power_supply/"
```
tobes: This doesn't need the `u` however it doesn't do any harm.

Also I'm not sure why these constants are defined rather than just explicitly defined in the module like most other do.

This comment can be ignored it is more observations.

py3status/modules/battery_level.py
```diff
@@ -225,6 +219,26 @@ def _refresh_battery_info(self):
         Battery 1: Unknown, 98%
         Battery 1: design capacity 1879 mAh, last full capacity 1370 mAh = 72%
         '''
+
+        def _parse_battery_info(acpi_battery_lines):
+            battery = dict()
```
tobes: personally I prefer `battery = {}` but not a big issue

py3status/modules/battery_level.py
```diff
@@ -225,6 +219,26 @@ def _refresh_battery_info(self):
...
+                                              acpi_battery_lines[1])[1])
+
+            # ACPI only shows time remaining if battery is discharging or
+            # charging
+            try:
+                battery["time_remaining"] = ''.join(findall(
+                    "(?<=, )(\d+:\d+:\d+)(?= remaining)|"
+                    "(?<=, )(\d+:\d+:\d+)(?= until)", acpi_battery_lines[0])[0])
+            except IndexError:
+                battery["time_remaining"] = '?'
```
tobes: I'd be inclined to have the `'?'` as a constant in case we ever decide to change it, as it is used in 2 places and that would make the change easier imho

py3status/modules/battery_level.py
```diff
@@ -237,8 +251,69 @@ def _refresh_battery_info(self):
...
+        Only available in kernel 2.6.24(?) and newer. Before kernel provided
+        a similar, yet incompatible interface in /proc
+        """
+        def _parse_battery_info(sys_path):
+            """
+            Extract battery information from uevent file, already convert to
+            int if necessary
+            """
+            raw_values = dict()
+            with open(sys_path + "/uevent") as f:
```
tobes: `sys_path + "/uevent"` I'd use `os.path.join(sys_path, "uevent")`

py3status/modules/battery_level.py
```diff
@@ -237,8 +251,69 @@ def _refresh_battery_info(self):
...
+                for var in f.read().split():
```
tobes: use `splitlines()` not `split()` here as my battery has a space in the serial number which causes an error.

py3status/modules/battery_level.py
```diff
@@ -83,9 +90,11 @@
 from __future__ import division  # python2 compatibility
 from time import time
 from re import findall
+from glob import iglob
 
 import math
 import subprocess
+import os
```
tobes: I'd be inclined to do `import os.path` as that is all we want,  this might be a just me sort of thing.

py3status/modules/battery_level.py
```diff
@@ -237,8 +251,69 @@ def _refresh_battery_info(self):
...
+            return raw_values
+
+        battery_list = []
+        for path in iglob(os.path.join(self.sys_battery_path, "BAT*")):
+            r = _parse_battery_info(path)
+            battery = dict()
+            battery["capacity"] = r["POWER_SUPPLY_ENERGY_FULL"]
+            battery["charging"] = "Charging" in r["POWER_SUPPLY_STATUS"]
+            battery["percent_charged"] = math.floor(
+                r["POWER_SUPPLY_ENERGY_NOW"] / battery["capacity"] * 100)
```
tobes: This returns a float eg `5.0` so we always have the `.0`  using `int()` instead of `math.floor()` would fix this

py3status/modules/README.md
```diff
@@ -295,9 +302,9 @@ Obsolete configuration parameters:
     *(default None)*
 
 Requires:
-  - the `acpi` command line
+  - the `acpi` command line (only if `measurement_mode='acpi'`)
 
-**author** shadowprince, AdamBSteele, maximbaz, 4iar
+**author** shadowprince, AdamBSteele, maximbaz, 4iar, m45t3r
```
tobes: these changes should be made in the module docstring and the README can be updated automatically by running `py3status docstring update battery_level`

the `acpi` command line __utility__ (only if `measurement_mode='acpi'`)

m45t3r: I couldn't run this command. I get "FileNotFoundError", so I just fixed the docstring. Hope someone can fix the README for me.

py3status/modules/battery_level.py
```diff
@@ -74,18 +81,21 @@
         (default None)
 
 Requires:
-    - the `acpi` command line
+    - the `acpi` the acpi command line utility (only if
+    `measurement_mode='acpi'`)
```
tobes: This docstring should look like this to be parsed correctly (I realize that initially it was incorrect)
```
    acpi: command line utility (only if
        `measurement_mode='acpi'`)
```
If you could change this then I'll recommend this for merging.  Thanks for your hard work.

m45t3r: Even in f6bf676 l got a FileNotFOundError. Should be fixed though.

tobes: don't worry about the README.md changes we can fix them later

this change is the really important line from my point of view
```
    acpi: command line utility (only if
```

> FileNotFOundError.

Do you have a traceback for that?  I think other people have had this issue and it may depend on how you have things installed

m45t3r:
```
2016-10-17 10:40:07 WARNING Setup error (FileNotFoundError) docstrings.py line 26.
2016-10-17 10:40:07 INFO Traceback
FileNotFoundError: [Errno 2] Arquivo ou diretório não encontrado: '/home/m45t3r/Source/py3status/env/lib/python3.5/site-packages/py3status-3.3rc0-py3.5.egg/py3status/modules/README.md'
  File "/home/m45t3r/Source/py3status/env/lib/python3.5/site-packages/py3status-3.3rc0-py3.5.egg/py3status/__init__.py", line 23, in main
    py3.setup()
  File "/home/m45t3r/Source/py3status/env/lib/python3.5/site-packages/py3status-3.3rc0-py3.5.egg/py3status/core.py", line 277, in setup
    self.handle_cli_command(self.config)
  File "/home/m45t3r/Source/py3status/env/lib/python3.5/site-packages/py3status-3.3rc0-py3.5.egg/py3status/core.py", line 787, in handle_cli_command
    docstrings.update_readme_for_modules(cmd[2:])
  File "/home/m45t3r/Source/py3status/env/lib/python3.5/site-packages/py3status-3.3rc0-py3.5.egg/py3status/docstrings.py", line 331, in update_readme_for_modules
    readme = parse_readme()
  File "/home/m45t3r/Source/py3status/env/lib/python3.5/site-packages/py3status-3.3rc0-py3.5.egg/py3status/docstrings.py", line 26, in parse_readme
    with open(readme_file) as f:

2016-10-17 10:40:07 ERROR py3status: Setup error (FileNotFoundError) docstrings.py line 26. Please try to fix this and reload i3wm (Mod+Shift+R)
```

Seems it is append my virtualenv isntead of looking at the relative path of the file.

tobes:
@ultrabug there are a few minor docstring cleanups but they already existed.  So I'm happy for you to merge this if you see fit.  I can always do them later.

@m45t3r thanks for this PR I think removing the acpi requirement is a good thing

ultrabug:
@m45t3r wow that's some kind of cool enhancement, thanks a lot !

Would you mind rebasing your branch please ?

m45t3r:
If I did not make any mistakes in rebasing I think this is ready. @ultrabug, @tobes?

tobes:
Looks good to me

ultrabug:
Thanks a lot @m45t3r !

Issue #509, opened 2016-10-13
battery_status.on_click() notification not working when battery is full


m45t3r:
I never get the string format set in `format_notify_charging` to work, including if my battery is charging or full charged (and the charger still plugged). When my battery is charging I get the ETA to fully charge my battery, when my battery is fully charged I get a "?", that probably comes from https://github.com/ultrabug/py3status/blob/master/py3status/modules/battery_level.py#L199.

I think the problem is in https://github.com/ultrabug/py3status/blob/master/py3status/modules/battery_level.py#L143:

```python
        if self.time_remaining:
            format = self.format_notify_discharging
        else:
            format = self.format_notify_charging
```
`self.time_remaining` is never `None` in my system, causing the bug (probably).

JFY, when my battery is full I get this from ACPI:

```
~ » acpi                                                    ruby-2.3.1p112
Battery 0: Full, 100%
```

tobes:
What is you output of `acpi -b -i` as that is the actual command called.

Also what version of py3status are you running

m45t3r:
```
~ » acpi -b -i                                                                                                                          ruby-2.3.1p112
Battery 0: Full, 100%
Battery 0: design capacity 4029 mAh, last full capacity 3556 mAh = 88%
```
```
~ » LANG=C pacman -Qi py3status                                                                                                         ruby-2.3.1p112
Name            : py3status
Version         : 3.2-1
Description     : An extensible i3status replacement/wrapper written in python
Architecture    : any
URL             : http://www.ultrabug.fr/tag/py3status/
Licenses        : custom: Simplified BSD
Groups          : None
Provides        : None
Depends On      : python  python-setuptools
Optional Deps   : i3status: for some of the functionality [installed]
Required By     : py3status-modules
Optional For    : None
Conflicts With  : None
Replaces        : None
Installed Size  : 1248.00 KiB
Packager        : Unknown Packager
Build Date      : Thu Oct 13 07:09:25 2016
Install Date    : Thu Oct 13 07:09:26 2016
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : None
```

tobes:
Ok,  I think that this is just some confusion over the settings (which there are many of and not so clear).

`format_notify_discharging` and `format_notify_charging` are used for notifications when you click on the module. But `notification` must be set to True for this.

`format` is used to display the actual output to the i3bar.

m45t3r:
Yeah, I know:

[...]

PR #512, opened 2016-10-15
on_click() logic fix

m45t3r:
Instead of checking `self.charging`, the code was checking `self.time_remaining`. Since `self.time_remaining` is always a non-empty String (even in case of error, where you would get a `?` instead of None or something else), this check always evaluated to true and `self.format_notify_charging` was never used.

Fix issue #509


ultrabug:
Good catch @m45t3r !

PR #514, opened 2016-10-16
cleanup get_config_attribute() remove unused code

tobes:
I happened upon some code that is never used, also no user used the default parameter so cleaned that up too.


ultrabug:
:skull_and_crossbones: code !

PR #516, opened 2016-10-16
Formatter bugfix display 0.0

tobes:
Testing the battery_level changes led to an error in the formatter.

`0.0` was not displayed as it is treated as False.  This PR treats `0.0` as a valid value to display.  The test looks odd due to how python treats False as a number.  A test has also been added for this behaviour.


ultrabug:
:+1:

PR #517, opened 2016-10-16
Add option to color text when mute too

guiniol:
By default, when muted the text is the color of `color_bad`.
This adds the option to have it colored depending on the volume percent in order to have an idea of the volume when unmuting.

### DIFF
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -27,6 +27,8 @@ Configuration parameters:
         (default '♪: {percentage}%')
     format_muted: Format of the output when the volume is muted.
         (default '♪: muted')
+    mute_with_colors: Set to 1 to color format_muted depending on the volume percent.
+        (default 0)
     threshold_bad: Volume below which color is set to bad.
         (default 20)
     threshold_degraded: Volume below which color is set to degraded.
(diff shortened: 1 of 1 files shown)


pydsigner:
Maybe `muted_color_hint` or `always_colored`?

guiniol:
> Maybe `muted_color_hint` or `always_colored`?

I like `always_colored`.

py3status/modules/volume_status.py
```diff
@@ -27,6 +27,8 @@
         (default '♪: {percentage}%')
     format_muted: Format of the output when the volume is muted.
         (default '♪: muted')
+    mute_with_colors: Set to 1 to color format_muted depending on the volume percent.
+        (default 0)
```
tobes:
```
my_module {
    value = True
    other_value = false
}
```
should work

tobes:
`always_colored` is ambiguous imho as color_bad is a color.

As a different approach PR #518 achieves the effect of this PR by setting `color_muted = None`

guiniol:
> As a different approach PR #518 achieves the effect of this PR by setting `color_muted = None`

#518 looks like a better solution. I'll close this PR.

PR #518, opened 2016-10-17
Allow None colors and use in volume_status color_mute

tobes:
This PR adds extra functionality to dealing with colors and demonstrates this in the volume_status module.

In a module a common usecase is

`color = self.py3.COLOR_MUTED or self.py3.COLOR_BAD`

We would expect color to be `color_bad` if `color_muted` is not in the config and this is how things currently are.  However if `color_muted = None` then I think that this should be respected and color should be `None`.

To enable this we have to know if `color_muted` is actually defined in the config or not.  This PR uses class `NoneColor` to help.  We also add `Py3.is_color()` helper function.

Module volume_status is updated:

1) volume_muted not in config (default behavior muted is color_bad)

2) `volume_muted = '#FF00FF'` the specified color is used when muted

3) `volume_muted = None` When muted the color depending on the volume will be used.

I am not sure what we should call this new config option `color_muted` or `color_mute` but we should try for a consistent tense across all modules.

This achieves the outcome of PR #517 via the use of `color_muted` rather than a new config parameter.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -778,6 +778,20 @@ __update(module_name=None)__
 Update a module. If `module_name` is supplied the module of that
 name is updated. Otherwise the module calling is updated.
 
+__is_color(color)__
+
+Tests to see if a color is defined.
+Because colors can be set to None in the config and we want this to be
+respected in an expression like.
+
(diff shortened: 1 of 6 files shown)


ultrabug:
I like the idea and implementation, thanks a lot !

PR #520, opened 2016-10-18
sysdata: Add dynamic memory unit

### LINKED ISSUE #401: How can I get mem_used in MB?

pferate:
Report memory usage in MB if total memory is less than 1GB, otherwise it is reported in GB.

Fixed bug where percent used was being rounded, which reported as "XX.00%".

Resolves #401

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -7,6 +7,8 @@ Configuration parameters:
     format: output format string
         *(default '[\?color=cpu CPU: {cpu_usage}%], '
         '[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]')*
+    mem_unit: the unit of memory to use in report, case insensitive.
+        ['dynamic', 'mb', 'mib', 'gb', 'gib'] (default 'GiB')
     padding: length of space padding to use on the left
         (default 0)
     precision: precision of values
@@ -20,6 +22,7 @@ Format placeholders:
     {cpu_temp} cpu temperature
     {cpu_usage} cpu usage percentage
     {mem_total} total memory
+    {mem_unit} unit for memory
     {mem_used} used memory
     {mem_used_percent} used memory percentage
 
@@ -74,7 +77,7 @@ class GetData:
         # return the cpu total&idle time
         return total_cpu_time, cpu_idle_time
 
-    def memory(self):
+    def memory(self, unit='GiB'):
         """
         Parse /proc/meminfo, grab the memory capacity and used size
         then return; Memory size 'total_mem', Used_mem, percentage
@@ -82,7 +85,6 @@ class GetData:
         """
 
         memi = {}
-        mem_unit = ''
         with open('/proc/meminfo', 'r') as fd:
             for s in fd:
                 tok = s.split()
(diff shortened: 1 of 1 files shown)


tobes:
@pferate again thanks for this PR.

A problem here is that if I am using this module and have my own format say `{mem_used} GB` and less than 1GB then it will be incorrect.  We try to reduce the impact of our changes on the end user.  With this module this is probably not the end of the world.  I'm not sure what @ultrabug's view will be.  We can see if the user has changed the format or is using the new placeholder etc. so there are ways of dealing with this.

Units.  whilst we are fixing this we should at least use the correct units eg MiB GiB

`{mem_unit}` should be in the Format placeholders section of the docstring

pferate:
Alright, so I'm thinking of having it be configurable through a new parameter `mem_unit`:
- dynamic: the behavior I added
- mb/mib: force values to MiB
- gb/gib: force values to GiB

We can have gb/gib be the default, so that current behavior isn't changed.

What do you think of that?

pferate:
I added an implementation for configurable memory units.

Valid values for `mem_unit`: ['dynamic', 'mb', 'mib', 'gb', 'gib']

If a user manually configures it to something else, it will raise an exception, which will in turn show `Mem: nan/nan UNKOWN (nan%)`, similar to the previous output of error `Mem: nan/nan GB (nan%)`.

I'm thinking of doing a lot more refactoring here, keeping #348 in mind.  Currently, since any exception within `GetData.memory()` gets ignored, I can't log it, since `Py3status` doesn't easily know about it.  It will be a lot cleaner soon (hopefully).

tobes:
>  Currently, since any exception within GetData.memory() gets ignored

It would probably be easier to just check/sort `mem_unit` in the modules `post_config_hook()` that way you can log things easier.

Also `GB` and `GiB` are different units 1 GB = 1000^3 Bytes, 1 GiB = 1024^3 Bytes

ultrabug:
Yep, that GB / GiB should be addressed please !

Thanks @pferate !

pferate:
I was trying to track down a way to reliable find out what unit `/proc/meminfo` reports in (my display shows `kB`).  I came across [this in the RHEL docs](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/s2-proc-meminfo.html).

> While the file shows kilobytes (kB; 1 kB equals 1000 B), it is actually kibibytes (KiB; 1 KiB equals 1024 B). This imprecision in /proc/meminfo is known, but is not corrected due to legacy concerns - programs rely on /proc/meminfo to specify size with the "kB" string.

Also [this post in the Ubuntu forums](https://ubuntuforums.org/showthread.php?t=1165076).

>I'm sure that it must be kiB, here's why:
>
> free, references /proc/meminfo, If I cat /proc/meminfo it gives me:
> MemTotal: 2797796 kB
> 
> Now, if I run free -b, I get 2864943104 bytes total.
> 
> 2864943104 / 1024 = 2797796
> 
> That is the same amount that is reported by /proc/meminfo

tobes:
Yeah `/proc/meminfo` says kB but means KiB but that is about not breaking userspace which is important for the kernel.

I have a helper that I need to cleanup around bytes > nice units.  I just need to check it against possible usecases by modules.  But you can always do your own for now.

ultrabug:
So... basically that means that we're talking about KiB and @pferate can remove the mb/gb/kb options right ?

pferate:
I thought that I pushed an update yesterday, but I guess I forgot.  I added functionality for kB/MB/GB/KiB/MiB/GiB, in their true units.

I'm okay with dropping kB/MB/GB since memory is commonly reported in KiB/MiB/GiB.

ultrabug:
Yeah just drop'em, it's more confusing than it's useful tbh

Thanks

pferate:
Rebased and dropped kB, MB, and GB.  PTAL.

ultrabug:
Thanks @pferate 

Would you please be so kind and both rebase and use the new `self.py3.format_units()` helper which would suit this PR perfectly ?

pferate:
@ultrabug, Sorry that I've been MIA.  I just started a new job, and have been really busy with that.

I rebased onto your master branch.  I'll try to take a look into the `format_units` helper later.

Travis looks like it's in a weird state right now.  The 2.7 job says that `tox` completed successfully and doesn't look like it's doing more processing, but the job timer continues to climb.  The 3.4 job hasn't kicked off yet, but the 3.5 and pypy jobs finished successfully.

ultrabug:
np at all @pferate, congratz on your new job, I hope you enjoy it :)

and thanks again for your contributions ! this one will be the last contrib making it to 3.3 :)

Issue #497, opened 2016-10-11
Module glpi has no cache_timeout


tobes:
Module will be using lots of unneeded cpu etc

pferate:
#521

PR #521, opened 2016-10-18
GLPI: Add cache_timeout

pferate:
Setting default cache_timeout to 5 minutes.

Resolves: #497

### DIFF
--- a/py3status/modules/glpi.py
+++ b/py3status/modules/glpi.py
@@ -6,6 +6,7 @@ It features thresholds to colorize the output and forces a low timeout to
 limit the impact of a server connectivity problem on your i3bar freshness.
 
 Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds (default 300)
     critical: set bad color above this threshold
     db: database to use
     host: database host to connect to
@@ -28,6 +29,7 @@ class Py3status:
     """
     """
     # available configuration parameters
+    cache_timeout = 300
     critical = 20
     db = ''
     host = ''
@@ -61,6 +63,7 @@ class Py3status:
                 )
             response['full_text'] = '%s tickets' % open_tickets
         mydb.close()
+        response['cached_until'] = self.py3.time_in(self.cache_timeout)
 
         return response
 


tobes:
Thanks for this PR.

The config parameters in both the class and the docstring should be in alphabetical order.  If you could fix that it would be great.

pferate:
No problem.  Shouldn't `test_module_doc.py` catch that?  It says that it does:

> * config parameters listed in alphabetical order

I can try to look into that to see if there is a problem in the test.  I had some other ideas for testing that I'll post in IRC for some opinions.

tobes:
> I can try to look into that to see if there is a problem in the test. 

PR #499 improves the tests so probably worth waiting till that is merged or dropped.

>  I had some other ideas for testing that I'll post in IRC for some opinions.

this sounds good we could do with more

pferate:
@tobes: PR has been updated.

tobes:
> I can try to look into that to see if there is a problem in the test. 

The reason is that this module has dependencies and so can't be imported without them installed.  the updated test used ast so doesn't need to import the module.  So currently this is blacklisted from the docstring tests.

pferate:
I see.  The current tests only check the docstrings, but not the actual modules, right?  What do you think of running the modules during the unit tests?  Obviously, those dependencies would need to be installed for testing, but we could add them to a `requirements-dev` or something similar.

Locally, I added a `test_modules.py` and modified the `module_test.module_test()` to run once, rather than continuously.  It has a lot a failures right now, but once it gets all massaged, it would be trivial to test in py2 and py3, as well as handle any regression.

That was one of the things I was going to bring up in IRC.

tobes:
> What do you think of running the modules during the unit tests? 

I think this would be good I have done some work in branch `tests-modules` some modules are just hard to test.  So I have the basics.  It has sort of stalled around dbus as I want to test modules using pydbus.

The issue is around python-gi not being a proper module.  I've been hacking on pydbus to get it to be independent of python-gi but need to do some more I've got glib working but need to get Gio and MainLoop sorted.

pferate:
> I think this would be good I have done some work in branch `tests-modules`

Thanks, I'll take a look at that branch

> some modules are just hard to test.

I agree.  It will be a process to get them all covered.

ultrabug:
Thank you so much @pferate !

PR #522, opened 2016-10-18
SpaceAPI: Make Python 2 compatible

### LINKED ISSUE #459: Module spaceapi is not python2 compatible
No module named urllib.request in python2

pferate:
Resolves #459

### DIFF
--- a/py3status/modules/spaceapi.py
+++ b/py3status/modules/spaceapi.py
@@ -21,10 +21,7 @@ Configuration parameters:
 @license WTFPL <http://www.wtfpl.net/txt/copying/>
 """
 
-import codecs
 import datetime
-import json
-import urllib.request
 import shlex
 import subprocess
 
@@ -44,9 +41,7 @@ class Py3status:
 
     def check(self):
 
-        response = {
-            'cached_until': self.py3.time_in(self.cache_timeout),
-            }
+        response = {'cached_until': self.py3.time_in(self.cache_timeout), }
 
         try:
             # if color isn't set, set basic color schema
@@ -57,10 +52,7 @@ class Py3status:
                 self.closed_color = ''
 
             # grab json file
-            json_file = urllib.request.urlopen(self.url)
-            reader = codecs.getreader("utf-8")
-            data = json.load(reader(json_file))
-            json_file.close()
+            data = self.py3.get_json_data(self.url)
             self._url = data.get('url')
 
             if(data['state']['open'] is True):
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -1,11 +1,21 @@
 from __future__ import division
 
+import json
 import sys
 import os
 import shlex
 from math import log10
 from time import time
 from subprocess import Popen, PIPE
+
+try:
+    # Python 3
+    from urllib.error import URLError
+    from urllib.request import urlopen
+except ImportError:
+    # Python 2
+    from urllib2 import URLError
+    from urllib2 import urlopen
 
 from py3status.formatter import Formatter, Composite
 
@@ -679,3 +689,16 @@ class Py3:
         setattr(self._py3status_module, color_name, color)
 
         return color
+
+    @staticmethod
+    def get_json_data(url):
+        """
+        Deserialize JSON from `url` into a Python object.
+
+        Returns None on error.
+        """
+        try:
+            data = json.loads(urlopen(url).read().decode())
+        except URLError:
+            data = None
+        return data
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -1,5 +1,6 @@
 from __future__ import division
 
+import codecs
 import json
 import sys
 import os
@@ -690,15 +691,30 @@ class Py3:
 
         return color
 
-    @staticmethod
-    def get_json_data(url):
+    def get_url(self, url):
         """
-        Deserialize JSON from `url` into a Python object.
+        Get the HTTP response for the HTTP GET request to `url`.
 
-        Returns None on error.
+        Raises `URLError` on errors.
         """
         try:
-            data = json.loads(urlopen(url).read().decode())
+            response = urlopen(url)
         except URLError:
-            data = None
+            self.log('Unable to access [{}]'.format(url),
+                     level=self.LOG_ERROR)
+            raise
+        return response
+
+    def get_json_data(self, url):
+        """
+        Deserialize JSON from `url` into a Python object.
+        """
+        response = self.get_url(url)
+        reader = codecs.getreader('utf-8')
+        try:
+            data = json.load(reader(response))
+        except (TypeError, ValueError) as e:
+            self.log('Error in json.load for [URL: {}] [{}]'.format(url, e),
+                     level=self.LOG_ERROR)
+            raise
         return data
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -741,7 +741,7 @@ class Py3:
 
         return color
 
-    def get_url(self, url):
+    def _get_url(self, url):
         """
         Get the HTTPResponse for the HTTP GET request to `url`.
 
@@ -779,7 +779,7 @@ class Py3:
             response = url
         else:
             try:
-                response = self.get_url(url)
+                response = self._get_url(url)
             except self.HTTPError:
                 # Any Exceptions extended from `Py3.Py3Exception.HTTPError`
                 return None
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -796,7 +796,12 @@ class Py3:
                 return None
         try:
             # Using py3status.py3.HTTPResponse Object
-            json_data = response.json()
+            try:
+                json_data = response.json()
+            except (TypeError, ValueError) as e:
+                self.log('Error in json.load for [URL: {}] [{}]'.format(url, e),
+                         level=self.LOG_ERROR)
+                return None
         except:
             # Using the built-in HTTPResponse Object
             reader = codecs.getreader('utf-8')
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -60,27 +60,34 @@ class HTTPResponse:
     @TODO: Implement the header attribute
     """
 
-    def __init__(self, resp_obj):
-        self._resp_obj = resp_obj
+    def __init__(self, original_response):
+        self._original_response = original_response
+        self._status = None
         self._data = None
         self._json = None
 
-    def __getattr__(self, name):
-        # If trying to access an attribute that doesn't exist in this class,
-        # try accessing that attribute from the response object.
-        return getattr(self._resp_obj, name)
-
     @property
     def status(self):
-        return self._resp_obj.getcode()
+        """
+        The HTTP status code for the response.
+        """
+        if self._status is None:
+            self._status = self._original_response.getcode()
+        return self._status
 
     @property
     def data(self):
+        """
+        A bytestring representing the content.
+        """
         if self._data is None:
-            self._data = self._resp_obj.read()
+            self._data = self._original_response.read()
         return self._data
 
     def json(self):
+        """
+        A deserialized Python object, generated from the response data.
+        """
         if self._json is None:
             self._json = json.loads(self.data.decode('utf-8'))
         return self._json
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -51,6 +51,41 @@ class NoneColor:
         return 'None'
 
 
+class HTTPResponse:
+    """
+    This class is like a simplified version of `urllib3.response.HTTPResponse`.
+
+    The HTTPResponse object provides status, data, and json attributes.
+
+    @TODO: Implement the header attribute
+    """
+
+    def __init__(self, resp_obj):
+        self._resp_obj = resp_obj
+        self._data = None
+        self._json = None
+
+    def __getattr__(self, name):
+        # If trying to access an attribute that doesn't exist in this class,
+        # try accessing that attribute from the response object.
+        return getattr(self._resp_obj, name)
+
+    @property
+    def status(self):
+        return self._resp_obj.getcode()
+
+    @property
+    def data(self):
+        if self._data is None:
+            self._data = self._resp_obj.read()
+        return self._data
+
+    def json(self):
+        if self._json is None:
+            self._json = json.loads(self.data.decode('utf-8'))
+        return self._json
+
+
 class Py3:
     """
     Helper object that gets injected as self.py3 into Py3status
@@ -708,25 +743,24 @@ class Py3:
 
     def get_url(self, url):
         """
-        Get the HTTP Response for the HTTP GET request to `url`.
+        Get the HTTPResponse for the HTTP GET request to `url`.
 
         Raises `Py3.URLError` on Connection Errors.
         Raises `Py3.ClientError` on HTTP 4xx Errors.
         Raises `Py3.ServerError` on HTTP 5xx Errors.
         """
         try:
-            response = urlopen(url)
+            response = HTTPResponse(urlopen(url))
         except URLError as e:
             self.log('Unable to access [{}]'.format(url),
                      level=self.LOG_ERROR)
             raise self.URLError(e)
-        return_code = response.getcode()
         status_message = '{url} returned a {http_status} status.'.format(
-            url=url, http_status=return_code)
-        if 400 <= return_code < 500:
+            url=url, http_status=response.status)
+        if 400 <= response.status < 500:
             self.log(status_message, level=self.LOG_ERROR)
             raise self.ClientError(status_message)
-        elif 500 <= return_code < 600:
+        elif 500 <= response.status < 600:
             self.log(status_message, level=self.LOG_WARNING)
             raise self.ServerError(status_message)
         return response
@@ -735,11 +769,12 @@ class Py3:
         """
         Deserialize JSON from `url` into a Python object.
 
-        `url` can be str or a HTTP Response object.
+        `url` can be str or a HTTPResponse object.
 
         Returns JSON object on success, or None on error.
         Does not raise any exceptions to caller.
         """
+        json_data = None
         if hasattr(url, 'getcode'):
             response = url
         else:
@@ -752,15 +787,20 @@ class Py3:
                 self.log('An uncaught exception was raised [{}]'.format(e),
                          level=self.LOG_ERROR)
                 return None
-        reader = codecs.getreader('utf-8')
         try:
-            data = json.load(reader(response))
-        except (TypeError, ValueError) as e:
-            self.log('Error in json.load for [URL: {}] [{}]'.format(url, e),
-                     level=self.LOG_ERROR)
-            return None
-        except Exception as e:
-            self.log('An uncaught exception was raised [{}]'.format(e),
-                     level=self.LOG_ERROR)
-            return None
-        return data
+            # Using py3status.py3.HTTPResponse Object
+            json_data = response.json()
+        except:
+            # Using the built-in HTTPResponse Object
+            reader = codecs.getreader('utf-8')
+            try:
+                json_data = json.load(reader(response))
+            except (TypeError, ValueError) as e:
+                self.log('Error in json.load for [URL: {}] [{}]'.format(url, e),
+                         level=self.LOG_ERROR)
+                return None
+            except Exception as e:
+                self.log('An uncaught exception was raised [{}]'.format(e),
+                         level=self.LOG_ERROR)
+                return None
+        return json_data
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -95,6 +95,21 @@ class Py3:
                 self._i3s_config = config
             self._py3status_module = module.module_class
 
+    class Py3Exception(Exception):
+        """ Base exception class """
+
+    class HTTPError(Py3Exception):
+        """ HTTP exception class """
+
+    class URLError(HTTPError):
+        """ A unified interface for URLError """
+
+    class ClientError(HTTPError):
+        """ A 4xx Client Error was returned """
+
+    class ServerError(HTTPError):
+        """ A 5xx Server Error was returned """
+
     def __getattr__(self, name):
         """
         Py3 can provide COLOR constants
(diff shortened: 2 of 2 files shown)


tobes:
@pferate Thanks,

@ultrabug this does now make requests a dependency for this module.  Personally I think that's fine.  It would be good to think about the whole dependencies thing at some point.  Requests being a good example it is used by many modules but we could easily provide the functionality using just the python standard libs via Py3 if we want to be dependency free.

ultrabug:
Thanks a lot @pferate but as much as I like requests I don't see the advantage it offers here.

@tobes I still believe in a dependency free py3status and it is an important aspect of this project's philosophy (no added config file, no deps) in my opinion.

Having the py3 helper provide a request-like method could indeed be a great solution for all those modules, maybe this challenge could be of interest for you @pferate ! :)

pferate:
I could work on that.  Would `helpers.py` be an appropriate location for it?

tobes:
> I could work on that. Would helpers.py be an appropriate location for it?

well it wants to be available to modules via `self.py3.request_url()` but you might find it easiest to create a new file for it and then just import into `py3.py` like the formatter stuff is.  That way you can keep all the code in one place.

This will be a bit of work as ideally we want it to mirror the functionality of requests if even if simplified.

Useful functionality is catching errors, timeouts, dealing with json, handeling redirects.

We can start with basic capabilities and slowly improve it and add new functionality, spaceapi is a good starting point

pferate:
Right now, in this module there is no error handling, if you can't get a response.  Currently `urlopen()`, just raises `URLError` on errors, but there is no exception handling.

I'm thinking of having this library function raise a module level exception (similar to passing along `URLError`) and documenting that it is the caller's responsibility to handle the exception rather than killing the entire process.  That way the modules can handle it however they want to, such as displaying a custom error message.

What do you think about module level exceptions?  Especially since `URLError` is in different locations in py2 and py3.

pferate:
Correction, there was some error handling here.  If any exception is raised, it displays an empty string.

pferate:
I added `py3.get_json_data()`, which can be used in other modules.  Internally, `get_json_data()` catches any `URLError` exceptions, and returns `None`, rather than passing the exception up.  Not sure how everyone feels on handling that, but it can be easily changed.

py3status/py3.py
```diff
@@ -464,3 +474,16 @@ def stop_sound(self):
         if self._audio:
             self._audio.kill()
             self._audio = None
+
+    @staticmethod
```
tobes: I'm not a fan of static methods in Py3.  My reasoning is that I might want to do some logging etc in functions without having to change things about.  It also keeps all the methods similar.

Also in this actual function I think it would be good to log the error.

py3status/py3.py
```diff
@@ -464,3 +474,16 @@ def stop_sound(self):
...
+    def get_json_data(url):
+        """
+        Deserialize JSON from `url` into a Python object.
+
+        Returns None on error.
+        """
+        try:
+            data = json.loads(urlopen(url).read().decode())
+        except URLError:
```
tobes: You are going to get many more exceptions thrown here.  json will definitely be throwing ones.

I'd do this in a much more broken down way so we can actually categorize the errors.  There are lots of different ways you can go about this.

We need to try to get this to be as robust as possible.

There is also the issue about how we handle errors.  The simple solution you have here (returning None) may well be the simplest but it doesn't give modules much to go on but maybe that is all they need.  Other options I can think of include returning a richer object like requests does so the module can get the return code etc.  I haven't thought about this in too much detail and have only looked at a couple of the modules.

tobes:
I think it might be good to make a `open_url()` function that gets a rich object back and then also have a `get_json_from_url()` function that is a wrapper and just returns json or (None/Exception).  Maybe this is overthinking it.

py3status/py3.py
pferate: Understood.  I changed it and added some logging.

py3status/py3.py
pferate: I split it up and added some other exception handling.

pferate:
@tobes:  I split up `get_json_data()` and added some logging.  PTAL.

tobes:
It's nicer but let's take a step back and just think about `get_json_data()`

We have various situations that can occur.

1. everything is good and we return the decoded json - This is straight forward.

2. we get our json but it is broken

3. getting the url fails for some reason (no connection, times out, some error 404, 403, 500 etc)

How are we going to deal with these and what is useful for the module writer and makes their life easy.

So in my mind the first question is does the module care what went wrong?

I think the module probably doesn't care it just wants to have the data or know it hasn't

Do we want to tell the user?

Well in some occasions we probably do.  For example if we have a 404 (url not found), 403 (not authorized) because that seems like they may have not set things up correctly.  But for a 500 I think that is an issue for the website etc not our user so is just like a not available and I don't think the user needs to be informed.  We can still log stuff but not bother them.

If we are telling the user is this something that something that we can do in the function or is something that the module should decide?

I'm not sure on this one but currently I can't think of a reason.

So I'm sort of inclined to just return the decoded json or None and the helper function can log it and notify about the errors.

Now using the `get_url()` I think can allow more options for the module writer if they want more control.  I also think it should actually return a Response object that we can examine eg has `json`, `return_code`, `response` as we need.

pferate:
I already have `get_url()` returning a Response object.  I updated `get_json_data()` to accept either a `str` or a Response object (not yet pushed).  With that, the module author can call `get_url()` themselves, analyze the response, then pass that to `get_json_data()` if they want to.  Or they can just call `get_json_data()` with a URL and any exceptions would be logged.

For `URLError` in `get_url()`, I'm logging an error and re-raising the exception.  For `get_json_data()`, I'm doing the same with `TypeError` and `ValueError`, the errors raised on invalid JSON input.

Maybe I can also have `get_json_data()` just return `data, response` all of the time, but set `data` to `None` on error.  If we go this route, I won't be able to re-raise any exceptions.

I'm not sure where we want to draw the line in the sand, in regards to notifying the end user.  How much should the core package do and how much should the modules be responsible for?

tobes:
> I already have get_url() returning a Response object. 

sorry missed that but I think we might want to return our own object not urllib2s.  But that's for another day.

>  For get_json_data(), I'm doing the same with TypeError and ValueError, the errors raised on invalid JSON input.

My point I don't think get_json_data() should ever raise an error back to the module.  It should either return data or None.  But I think the user should be notified for 404 or 403 but via Py3 not the module

tobes:
From irc @pydsigner wrote

> If we're considering reimplementing requests, why not just include requests?  There's a point at which we have to be careful that "light and self-contained" doesn't become a variant of Not Made Here syndrome.

This was asked of @ultrabug but I'm going to answer

I agree that not invented here is a bad thing and requests is a nice library that I use in preference to urllib.  But I think for most py3status modules both are overkill.  What I mean by that is the common usecase is to get some json data from the web.

In my opinion these modules would be happy with just a `py3.get_json_data()` helper that either returns the data or None.  If each module worries about timeouts and other errors we have that code in many modules, some will handle it well, some badly. They will do it in their own ways and notify the user differently.  I think that py3status should be uniform in how it reports (or not) this information to the user.

So really we are just wrapping functionality rather than re-inventing.  We then have the what do we use to implement this urllib or requests?  In my mind requests is an extra requirement and does not really offer much extra for our simple needs.  In this case using something already in the standard library makes sense.  If we later decide that requests is a better fit we can change all modules in one go, by just changing our wrappers internals.

ultrabug:
TBH I feel like you guys try to cover too much cases in one go.

I'd be in favor of creating ONE py3 method named `get_json_from_url()` which

- returns the JSON

- returns None if the response is not a valid JSON

- raises the urllib Exception that the module writer will have to handle for now

pferate:
I like having the idea of having a way for the modules to inspect the Response object, if they feel inclined to, but it should be optional.  I also think that while returning `None` would be sufficient, that would mean 2 different ways that modules would check for errors in the same function call: checking the return value and catching an exception.  

I think that there should be a consistent way of reporting errors back to users.  This may be a bigger design decision.

Maybe we can use custom Module or Class exceptions, and have modules catch the `py3status` base exception in the most general case, at least when calling `py3status` functions.

Also, for `requests`, `six`, and other such libraries: if we don't add them as a requirement, what are your thoughts on adding them as "vendors" instead?  `requests` currently has `urllib3` as a vendor, which in turn has `six` as a vendor.

ultrabug:
> I like having the idea of having a way for the modules to inspect the Response object

Then let's settle on `get_response_from_url()` which will have a json method right ?

> I also think that while returning None would be sufficient, that would mean 2 different ways that modules would check for errors in the same function call: checking the return value and catching an exception.
> I think that there should be a consistent way of reporting errors back to users. This may be a bigger design decision.

You're right and I think this is beyond the scope of this PR. So I see two options here, we put this PR on hold, open a dedicated issue and work on a new PR or we accept to have a preliminary work done here on this use case which will facilitate the discussion (I prefer option 2)

Maybe we could have 3 HTTP generic exceptions and have the method either return the succesful Response or raise an exception ?

- Http4xxError
- Http5xxError
- HttpConnectionError

> Also, for requests, six, and other such libraries: if we don't add them as a requirement, what are your thoughts on adding them as "vendors" instead? requests currently has urllib3 as a vendor, which in turn has six as a vendor.

Just to give you guys more context on my stubborn position about this, one of my other occupations in the OSS community is being a Gentoo Linux developer. Packaging and caretaking packages means dealing with dependencies !

- i3wm users, i3status users, tiling window manager users are lightweight thinking people mostly (I certainly am). Keeping deps to zero stays in this philosophy and has nothing to do with reinventing the wheel imho

- bundling dependencies in a project's source (vendoring) means duplicating code, lacking behind releases / bug fixes and most importantly security fixes. I see this as the worst of what a project can do and have been dealing/discussing with upstreams about this for ages.

- adding a dependency, as much as easy they can be to use, means imposing it to any packager of Linux distros. So you have to specify clearly which versions of the given dep your project relies on and not just say "requests" or "six" because one day, one of those deps can break their API or something.

- finally, as a leading conclusion to the point above, adding a dep means adding a constraint on your project, you have to keep track of the development and features of the dependencies.

So yes, I'm stubborn on this topic but as you can see I have strong opinions on the matter.

tobes:
I agree with @ultrabug on the dependency stuff.  Vendoring just seems to be an acceptance that python packaging is a mess and feels like a defeatist attitude that is bug prone and user abusive.  It makes requests no longer seem the nice project I believed it to be.  I'm also not a great fan of six

Anyhow back to this issue.

I think an `get_response_from_url()` that returns an object or throws an exception is a good thing.  I'm not sure we need so many exceptions but maybe that's fine they should all be subclassed anyhow so not a big deal.   It would have a few fancy methods to get the json, http return code etc.

However I still strongly think that py3 also wants a helper `get_json_from_url()` that will just wrap `get_response_from_url()` and only return json or None (it would silence all exceptions).  I think this is the most useful for actual module writers but  `get_response_from_url()` gives them a richer object if the want especially if they care about exceptions, otherwise all client modules need to be doing exception handling that most will ignore.

Anyhow I think this all makes for interesting discussion topics.
For this PR I'd be happy for `get_response_from_url()` to not actually be implemented fully and kept private as it will need to be really robust.

Also back to the exceptions.  They want to be available inside `Py3` maybe namespaced eg `Py3.exceptions.HttpConnectionError` currently this PR has `from py3status.py3 import URLError` and I don't think a module should ever do `import py3status.xxx`.

ultrabug:
Ok then @pferate you have all the cards in your hand. We'll trust and go with your modified implementation and we all agree that it will be a good start to iterate from.

Thanks a lot again @pferate, hope to see this happen soon :)

pferate:
I added some nested exception classes to the `Py3` class.  What do you think of the implementation?

`Py3.get_url()` is essentially a wrapper around `openurl()`.  It raises one of the 3 new exceptions on error.

`Py3.get_json_data()` returns either the JSON object, or `None`.  Any exceptions are logged, but `None` is returned.

Also, `Py3` is getting pretty big; what do you think of composing it of Mix-Ins rather than imported Classes?  That way, you wouldn't need duplicated functions (Composite) or classes assigned to attributes (Formatter); they would all be first-class methods.

py3status/py3.py
```diff
@@ -84,6 +95,18 @@ def __init__(self, module=None, i3s_config=None, py3status=None):
                 self._i3s_config = config
             self._py3status_module = module.module_class
 
+    class Exceptions(Exception):
```
tobes: Exceptions is a bit odd maybe Py3statusException?

py3status/py3.py
```diff
@@ -679,3 +702,62 @@ def threshold_get_color(self, value, name=None):
...
+        """
+        Get the HTTP Response for the HTTP GET request to `url`.
+
+        Raises `Py3.URLError` on Connection Errors.
+        Raises `Py3.HttpClientError` on HTTP 4xx Errors.
+        Raises `Py3.HttpServerError` on HTTP 5xx Errors.
+        """
+        try:
+            response = urlopen(url)
+        except URLError as e:
```
tobes: I think we need to worry about more Exceptions being thrown here.

But maybe we will improve this as we go.  I think some tests might be nice at some point.

py3status/py3.py
```diff
@@ -679,3 +702,62 @@ def threshold_get_color(self, value, name=None):
...
+        elif 500 <= return_code < 600:
+            self.log(status_message, level=self.LOG_WARNING)
+            raise self.HttpServerError(status_message)
+        return response
+
+    def get_json_data(self, url):
+        """
+        Deserialize JSON from `url` into a Python object.
+
+        `url` can be str or a HTTP Response object.
```
tobes: I'm not a fan of this overloading I think `get_url()` should return an object with a `json()` method

and `get_json_data()` should just take a str url

py3status/py3.py
```diff
@@ -679,3 +702,62 @@ def threshold_get_color(self, value, name=None):
...
+        reader = codecs.getreader('utf-8')
+        try:
+            data = json.load(reader(response))
+        except (TypeError, ValueError) as e:
+            self.log('Error in json.load for [URL: {}] [{}]'.format(url, e),
+                     level=self.LOG_ERROR)
+            return None
+        except Exception as e:
+            self.log('An uncaught exception was raised [{}]'.format(e.with_traceback()),
+                     level=self.LOG_ERROR)
```
tobes: We can actually use py3status' `report_exception()` to do this.  But for now this is ok

py3status/modules/spaceapi.py
```diff
@@ -57,10 +52,10 @@ def check(self):
...
 
             # grab json file
-            json_file = urllib.request.urlopen(self.url)
-            reader = codecs.getreader("utf-8")
-            data = json.load(reader(json_file))
-            json_file.close()
+            try:
+                data = self.py3.get_json_data(self.url)
+            except self.py3.Exceptions:
+                self.py3.notify_user('{} cannot be reached.'.format(self.url))
```
tobes: This should be updated we don't get an exception and need to handle the None case

tobes:
> Py3 is getting pretty big; 

True but I'm not too worried at the moment, and it is an easily solved problem.

> what do you think of composing it of Mix-Ins rather than imported Classes?

Personally I hate mix-ins intensely they usually seem a great way to make code unreadable.  Be explicit and all that.

tobes:
>  That way, you wouldn't need duplicated functions (Composite) or classes assigned to attributes (Formatter); they would all be first-class methods.

Actually this is exactly what I want to avoid.

py3status/py3.py
pferate: How about `Py3Exception`?  I was initially thinking of something along those lines, but I had the whole name in mind and didn't want to be repetitive (`Py3.Py3StatusException`).  I'm thinking that this may be expanded into other types of exceptions down the road.

I had something like this in mind:
```
Py3
 +-- Py3Exception
      +-- SomeError
      |    +-- SomeNestedError
      +-- AnotherError
      +-- StatusError
      +-- HTTPError
           +-- URLError
           +-- HttpClientError
           +-- HttpServerError
```

tobes: Yeah that looks good,  It's nice when they group well.

py3status/py3.py
pferate: > I'm not a fan of this overloading I think get_url() should return an object with a json() method

Are you thinking of adding a new object, mimicking or extending on Response?

> and get_json_data() should just take a str url

I just didn't want to make a second HTTP request if the author used the other function to inspect for errors, which would not be the norm.

py3status/py3.py
pferate: I still need to learn the rest of the code base.

py3status/modules/spaceapi.py
pferate: oops.  will fix this.

pferate:
Updated the PR again.  Nested custom exceptions, and handle `None` in the module.

Should we add `Py3.report_exception()`?  It looks like it may be as simple as copying the `Py3.log()` code, but calling `self._module._py3_wrapper.report_exception()` instead (and adding the `notify_user` parameter).  Not sure if I am missing something here.

tobes:
> Should we add Py3.report_exception()? It looks like it may be as simple as copying the Py3.log() code, but calling self._module._py3_wrapper.report_exception() instead (and adding the notify_user parameter). Not sure if I am missing something here.

I'd be happy to add it but only as Py3._report_exception() ie private and not available to modules.

py3status/modules/spaceapi.py
```diff
@@ -57,10 +52,9 @@ def check(self):
                 self.closed_color = ''
 
             # grab json file
-            json_file = urllib.request.urlopen(self.url)
-            reader = codecs.getreader("utf-8")
-            data = json.load(reader(json_file))
-            json_file.close()
+            data = self.py3.get_json_data(self.url)
+            if data is None:
+                self.py3.notify_user('{} cannot be reached.'.format(self.url))
```
tobes: imho I think we shouldn't notify the user like this.  eg if there is no network.  We can either show them cached info on the i3bar or a 'unavailable' message.

Before it just crashed so anything is an improvement.

py3status/py3.py
tobes: I'd have get_url() return a simple object that has a json method but for the moment maybe just have get_url() private and worry about that later.  The module and most others that this would benefit only want get_json_data() anyhow.

That way we would be pretty much ready to go with this.

py3status/modules/spaceapi.py
pferate: I'll set the error short and full text to "Unavailable" and "Site Unavailable", respectively.  At least for now.

pferate:
I added a new "simple" `HTTPResponse` object, based on `urllib3` and `requests`.  I need to do more testing before I update this PR, so I'll push it out tomorrow hopefully.

py3status/py3.py
```diff
@@ -679,3 +740,67 @@ def threshold_get_color(self, value, name=None):
...
+    def _get_url(self, url):
+        """
+        Get the HTTPResponse for the HTTP GET request to `url`.
+
+        Raises `Py3.URLError` on Connection Errors.
+        Raises `Py3.ClientError` on HTTP 4xx Errors.
+        Raises `Py3.ServerError` on HTTP 5xx Errors.
+        """
+        try:
+            response = HTTPResponse(urlopen(url))
```
tobes: Why not just pass the url to HTTPResponse and let it decide what to do?  that way it's all self contained.

py3status/py3.py
```diff
@@ -40,6 +51,41 @@ def __repr__(self):
...
+
+    def __init__(self, resp_obj):
+        self._resp_obj = resp_obj
+        self._data = None
+        self._json = None
+
+    def __getattr__(self, name):
+        # If trying to access an attribute that doesn't exist in this class,
+        # try accessing that attribute from the response object.
+        return getattr(self._resp_obj, name)
```
tobes: hmm,

I'd remove this.  the reason is that if you allow this you are effectively creating an api that you need to support forever.  By limiting the methods eg json() then it is easy to support them even it the internals change.  Also are responses from python2/3 100% identical?

py3status/py3.py
```diff
@@ -40,6 +51,41 @@ def __repr__(self):
...
+
+    @property
+    def data(self):
+        if self._data is None:
+            self._data = self._resp_obj.read()
+        return self._data
+
+    def json(self):
+        if self._json is None:
+            self._json = json.loads(self.data.decode('utf-8'))
```
tobes: should we do the decoding in `self.data`?

py3status/py3.py
```diff
@@ -679,3 +740,67 @@ def threshold_get_color(self, value, name=None):
...
+            try:
+                json_data = json.load(reader(response))
+            except (TypeError, ValueError) as e:
+                self.log('Error in json.load for [URL: {}] [{}]'.format(url, e),
+                         level=self.LOG_ERROR)
+                return None
+            except Exception as e:
+                self.log('An uncaught exception was raised [{}]'.format(e),
+                         level=self.LOG_ERROR)
+                return None
```
tobes: What's going on here? Surely this belongs in `HTTPResponse.json()` which should either return some nice json or raise some sort of InvalidJSON exception that we can catch here to return None.

py3status/modules/spaceapi.py
```diff
@@ -57,10 +52,7 @@ def check(self):
                 self.closed_color = ''
 
             # grab json file
-            json_file = urllib.request.urlopen(self.url)
-            reader = codecs.getreader("utf-8")
-            data = json.load(reader(json_file))
-            json_file.close()
+            data = self.py3.get_json_data(self.url)
```
tobes: you seem to have lost the cope with None being returned logic

py3status/py3.py
pferate: I did it partly, because I had the mindset of working with `HTTPRequest` and `HTTPResponse` objects.  When building the `HTTPRequest` object, the URL is defined there, along with other attributes.  It may do some processing prior to making the full request, which would return the `HTTPResponse` object.

Now, we aren't implementing the full feature set, but `urlopen(url)` is, in a sense, taking the place of `http.request()`, since it returns a `Response` object.

py3status/py3.py
pferate: Makes sense.  I added it initially so that all of the `I/O`-type methods would work (`read()`, `readline()`), but we don't need those exposed with the functions we have defined.

py3status/py3.py
pferate: I left `.data` as bytes to be consistent with all other libraries, since, by definition, the data is actually `bytes`.  That is actually how `urllib3` [shows how to get JSON data](http://urllib3.readthedocs.io/en/latest/user-guide.html#json-content).

tobes: Cool you are the http expert now :)

Maybe add something in the docstrings so it's obvious that data is bytes and the json is unicode

py3status/py3.py
pferate: Generally, module authors should use `get_json_data()`, which will mask all errors with `None`.  If someone is working with the `HTTPResponse` object directly, I imagine that any exceptions should not be hidden.

I had the logic of trying the `.json()` method, then if it failed for any reason, fall back to the original method.  This would catch someone passing in the output of `urlopen()` and also dealing with invalid JSON.  With removing the pass through API, this would not work with invalid JSON.

I'll clean this up a bit more.

py3status/modules/spaceapi.py
pferate: It was being implicitly checked, since the `try` block was catching every exception (which I am not a fan of).  `None` will throw an `TypeError` or `AttributeError` when accessing the attributes in the following lines.

I had other improvements, but I removed them because they would deviate from the previous behavior (an empty string for `full_text`).  I will address that when implementing the format update.

tobes: Sorry, I should have paid a bit more attention to the whole function.

Yeah I hate the whole catch any exception stuff as it tends to hide problems more than it helps.  But as you say cleaning that up is probably outside of this PR.

pferate:
I updated it again.  I was going to have `HTTPResponse.json()` raise a custom exception, but since it is not within `Py3` class, like the exceptions are, it didn't seem right to have it raise one from that tree.

Not sure how you want to handle custom exceptions for the main code and custom exceptions exposed to the modules, nested within the `Py3` class.

Is there a reason we need to have the exceptions within the `Py3` class?  Is it just so that the modules don't have to import them?

tobes:
Yeah the exceptions in py3 just makes it easy for module. They don't need to be defined there just available. I'll take a look at the changes when I get a chance

pferate:
I'm thinking of moving the `HTTPResponse` and the custom exceptions into their own modules.  Watch out for an update soon.

ultrabug:
Ok @pferate I guess it's time we move on this, I'd love to get your work merged ! How should we proceed from here ?

tobes:
@pferate I realize you've been busy with things but it'd be nice to get this PR moving.  Do you think you will have time to revisit this over the next month or so?  If not no problems but as you put a lot of work into this it would be good to see it getting used.

tobes:
This functionality was added via #778

PR #523, opened 2016-10-19
Don't call updates on modules before they have been fully initialised

tobes:
I found an issue where i3status modules are in containers and try to update the container module before it has had its post_config_hook() called.

This PR prevents this from happening

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -55,6 +55,7 @@ class Py3statusWrapper():
         self.notified_messages = set()
         self.output_modules = {}
         self.py3_modules = []
+        self.py3_modules_initialized = False
         self.queue = deque()
 
     def get_config(self):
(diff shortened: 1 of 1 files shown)


ultrabug:
Is that a British revendication that got you name it `initialise` ? :)

tobes:
Despite some evidence, sometimes I find it hard to mis-spell words ;p

py3status/core.py
```diff
@@ -454,6 +455,11 @@ def notify_update(self, update, urgent=False):
             update = [update]
         self.queue.extend(update)
 
+        # if all our py3status modules are not ready to recieve updates then we
```
ultrabug: `recieve` : tell me about it ;)

tobes:
Ok spellings should be good now

ultrabug:
Thanks Mr UK :)

PR #524, opened 2016-10-20
ns_checker: Add format and cache_timeout

pferate:
Adding format and cache_timeout to the `ns_checker` module.

@tobes: I have a few of these coming, should I put multiple modules in the same PR?  or would you rather have a PR for each module? (I'm not sure of the entire code review process here)

I ask, because I did some major refactoring, along with adding `format`, in another module (`rate_counter`) and felt that that one should be separate.

Working towards #98

### DIFF
--- a/py3status/modules/ns_checker.py
+++ b/py3status/modules/ns_checker.py
@@ -10,6 +10,7 @@ in nameservers list.
 The default resolver can be overwritten with my_resolver.nameservers parameter.
 
 Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds (default 300)
     domain: domain name to check
     lifetime: resolver lifetime
     nameservers: comma separated list of reference DNS nameservers
@@ -33,13 +34,15 @@ class Py3status:
     """
     """
     # available configuration parameters
+    cache_timeout = 300
     domain = ''
     lifetime = 0.3
     nameservers = ''
     resolvers = ''
 
     def ns_checker(self):
-        response = {'full_text': ''}
+        response = {'cached_until': self.py3.time_in(self.cache_timeout),
+                    'full_text': ''}
         counter = 0
         error = False
         nameservers = []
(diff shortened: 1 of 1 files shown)


tobes:
@pferate 

> I have a few of these coming, should I put multiple modules in the same PR? or would you rather have a PR for each module? 

Putting each module in a PR makes most sense.  Some will be quite trivial and some more controversial.  With some you might find that even splitting the changes in one module into more than one PR makes sense.

> (I'm not sure of the entire code review process here)

The best thing is to just do some PRs and get a feel for it that way.  just to let you know that things may take a little while to get merged sometimes.

> Working towards #98

Cool

py3status/modules/ns_checker.py
```diff
@@ -33,16 +40,21 @@ class Py3status:
     """
     """
     # available configuration parameters
+    cache_timeout = 300
     domain = ''
+    format = '{ns_total} NS {ns_status}'
```
tobes: You haven't added this to the docstring but to be honest it might be good to wait till #499 is merged due to docstring conflicts.

tobes:
One thing to bear in mind with these changes is that we want to not change things for existing users so they should really see the same output as they saw before, unless there is good reason to break things.  You seem to have changed the output of this module, but as it's not one I use I'm not really that sure as I've only skimmed it briefly

pferate:
The default output would be the same as before.  I just added additional placeholders that I felt would be useful.

`counter` is not needed because it equivalent to `len(nameservers)`, there is no need to add 1 to a counter for every entry in `nameservers` within the loop.

The previous behavior would use `NOK` if there was at least 1 failure, otherwise it would be `OK`.  I went with the logic to use `OK` as the default, but if any of the nameserver checks failed, then it would use `NOK`.

tobes:
> The default output would be the same as before. 

Cool as I said I only skimmed it

> I just added additional placeholders that I felt would be useful.

This is always nice

> counter is not needed because it equivalent to len(nameservers)

Yeah, kill any code you can :+1:

pferate:
I added `format` to the Configuration part of the docstring.

I also dropped the `ns_` prefix from the placeholders.

ultrabug:
Great enhancement, thanks @pferate !

PR #525, opened 2016-10-20
rate_counter: Add format and majorly refactor code

pferate:
Added full text format and money format.

Cost was majorly refactored.  It was previously calculated with way too much work, and would fail if timer ran over a month (it would ignore months and essentially reset to 0).

Working towards #98

### DIFF
--- a/py3status/modules/rate_counter.py
+++ b/py3status/modules/rate_counter.py
@@ -29,8 +29,8 @@ Money placeholders:
     {price} numeric value of money
 
 Color options:
-    color_bad: Running
-    color_good: Stopped
+    color_running: Running, default color_good
+    color_stopped: Stopped, default color_bad
 
 @author Amaury Brisou <py3status AT puzzledge.org>
 """
@@ -123,10 +123,10 @@ class Py3status:
     def counter(self):
         running_time = 0.0
         if self.running:
-            color = self.py3.COLOR_GOOD
+            color = self.py3.COLOR_RUNNING or self.py3.COLOR_GOOD
             running_time = self.current_time - self.start_time
         else:
-            color = self.py3.COLOR_BAD
+            color = self.py3.COLOR_STOPPED or self.py3.COLOR_BAD
             running_time = self.saved_time
 
         days, hours, mins, secs = self.secs_to_dhms(running_time)
--- a/py3status/modules/rate_counter.py
+++ b/py3status/modules/rate_counter.py
@@ -93,32 +93,37 @@ class Py3status:
         return days, hours, mins, secs
 
     def _start_timer(self):
-        self.start_time = self.current_time - self.saved_time
-        self.running = True
+        if not self.running:
+            self.start_time = self.current_time - self.saved_time
+            self.running = True
 
     def _stop_timer(self):
-        self.saved_time = self.current_time - self.start_time
-        self.running = False
+        if self.running:
+            self.saved_time = self.current_time - self.start_time
+            self.running = False
+
+    def _toggle_timer(self):
+        if self.running:
+            self._stop_timer()
+        else:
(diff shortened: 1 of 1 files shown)


pferate:
Sorry for only using one commit here.  I got halfway through adding the format, then I came across the mess of calculating the cost.  If it will make things easier, I can go back and split this into multiple commits.

tobes:
> If it will make things easier, I can go back and split this into multiple commits.

It'll be fine I think.  Sometimes things do just need a rewrite rather than just lots of polishing

pferate:
> Looking at the code it seems that _reset() should probably only reset when not running. It sort of does that but it feels more like luck than anything else.

I didn't change any logic in `_reset()`, but that sounds reasonable.  It has that behavior because `self.start_time` doesn't get reset; which, as you say, is luck and is not reliable.  I will fix the logic in `_reset()`.
 
> The documentation on the color options is incorrect too (My fault as I added them) It'd be great if you could fix that too in this PR.

What do you think of the names `COLOR_RUNNING` and `COLOR_STOPPED`?

> I think changing mins to be zero padded is an improvement over the original (even if it is changing the output - but it feels a bug fix), but I'd not do it for the hours.

Will do.

ultrabug:
ping @amaurybrisou, I'd like your opinion please since you wrote this module :) thanks !

py3status/modules/rate_counter.py
```diff
@@ -7,101 +7,157 @@
...
+    {mins} The remaining number of whole minutes in running timer
+    {secs} The remaining number of seconds in running timer
+    {subtotal} The subtotal cost (time * rate)
+    {tax} The tax cost, based on the subtotal cost
+    {total} The total cost (subtotal + tax)
+    {total_hours} The total number of whole hours in running timer
+    {total_mins} The total number of whole minutes in running timer
+
+Money placeholders:
+    {price} numeric value of money
```
tobes: Hmm, How we document this sort of stuff is a question.  I'm happy with this for now but maybe we need to decide as some other modules also do this 'subformatting'

py3status/modules/rate_counter.py
```diff
@@ -7,101 +7,157 @@
...
 
-    def kill(self):
-        if self.started is True:
+    @property
+    def current_time(self):
+        """Get the current time.
+
+        Using a helper property to make it easy to keep consitency.
+        """
+        return time.time()
```
tobes: I really don't like this function.  As in my mind it obscures whats going on.
I'm not sure that `self.current_time` is any more readable than `time.time()`

If the function was doing more work I could understand this but it feels it is just adding noise. @pferate maybe you have a justification as to why this should be used

Saying that I don't have such a strong opinion that I'd say we should block the PR for this.

pferate: I was thinking that having the property called `current_time` would make things look more uniform throughout the rest of the code, as well as making any future refactoring easier.

The previous implementation had `self.current_time = time.time()` and `self.start_time = time.time()` sprinkled throughout,

As well as:
`self.current_time = time.time()`
`self.diff_time = time.gmtime(self.current_time - self.start_time)`
and:
`self.saved_time = time.time() - self.start_time`

I thought that having:
`self.saved_time = self.current_time - self.start_time`
would be more consistent and readable.

I was also thinking that we could maybe transition to a different library, like `datetime`, so that we could just use the built-in `datetime.timedelta` instead of calculating it ourselves.  Having it defined in the one place would make it somewhat trivial.

ultrabug: okay, it's good as it is until we want to add more sugar to it thanks !

ultrabug:
Thanks @pferate for the nice work and @amaurybrisou for your review !

PR #526, opened 2016-10-20
GLPI: Add format

pferate:
There may be more useful data to pull from the database (opened/closed per time period, number closed by a user, etc.), but I'm not familiar with GLPI, so I'm just keeping the format as-is.

Working towards #98

### DIFF
--- a/py3status/modules/glpi.py
+++ b/py3status/modules/glpi.py
@@ -9,11 +9,15 @@ Configuration parameters:
     cache_timeout: how often we refresh this module in seconds (default 300)
     critical: set bad color above this threshold (default 20)
     db: database to use (default '')
+    format: format of the module output (default '{num_open} tickets')
     host: database host to connect to (default '')
     password: login password (default '')
     timeout: timeout for database connection (default 5)
     user: login user (default '')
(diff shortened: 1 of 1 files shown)


tobes:
Yes,  It is much easier if you only change one thing at a time as it makes the reviews much easier.

py3status/modules/glpi.py
```diff
@@ -13,6 +13,9 @@
     user: login user
     warning: set degraded color above this threshold
 
+Format placeholders:
+    {num_open} The number of open tickets
```
pferate: Either of those are fine with me.  `{open}` seems too ambiguous without context.  `{tickets_open}` would be what I would lean towards.

tobes: Let's go with `{tickets_open}`

ultrabug:
@pferate thanks for this, please check the result of the travis CI which shows that the default format should be documented in the docstring as well !

When travis will pass, I'll merge :+1:

pferate:
@ultrabug, I rebased this branch and added the missing docstring entry.  PTAL

ultrabug:
Awesome, thanks @pferate !

PR #527, opened 2016-10-21
Formatter and Composite updates

tobes:
Updates to the formatter.

* Composites in the formatter now return a `Composite` object rather than a list (if a composite is created eg when `\?color=XXX` is used in a format string).  This allows us to add composites via `param_dict`.

* A `Composite` is basically a wrapped list but allows us to know that it is a composite rather than just a list.  It also allows easier appending/creating as it understands strings, lists, dicts and Composites.  It also has some helper functions.

* Seeing as how `safe_format()` can return a `Composite` and add them via `param_dict` there is less need for `build_composite()` so it can be deprecated.

* `safe_format()` can be forced to return a composite.  This is helpful if the module will manipulate the result etc.

* `composite_join()`, `composite_update()`, `composite_create()`, `is_composite()` helper functions added to `Py3`.

* Fix various bugs in the formatter and add new tests to catch them

* Merging of Composite components improved

* `module.py` accepts Composites in responses

* Hex colors passed to formatter via `\?color=#xxxxxx` are checked, upper-cased to aid merging and now values like `#xxx` are also allowed

* As well as some general formatter fixes this allows us to start doing some more interesting things for example with a few minor changes `weather_yahoo` can allow me to choose custom colors for each weather icon (without having to add any extra config parameters)

* Most modules will just continue to work as they are with no changes, only ones that manipulate the result of `safe_format()` need to worry about these changes and they already need to care to some extent so the disruption to module writers is minimal.

I still need to do some more documentation and write some more Composite tests but the actual code shouldn't be changing so any reviews would be helpful

### DIFF
--- a/tests/test_composite.py
+++ b/tests/test_composite.py
@@ -1,0 +1,85 @@
+from py3status.composite import Composite
+
+
+# Composite initialize
+
+def test_Composite_init_1():
+    result = Composite('moo').get_content()
+    assert result == [{'full_text': 'moo'}]
+
+
+def test_Composite_init_2():
+    result = Composite({'full_text': 'moo'}).get_content()
+    assert result == [{'full_text': 'moo'}]
+
+
+def test_Composite_init_3():
+    result = Composite([{'full_text': 'moo'}]).get_content()
+    assert result == [{'full_text': 'moo'}]
+
+
+def test_Composite_init_4():
+    result = Composite(Composite('moo')).get_content()
+    assert result == [{'full_text': 'moo'}]
+
+
+# Composite append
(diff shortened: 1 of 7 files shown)


tests/test_composite.py
```diff
@@ -0,0 +1,109 @@
+from py3status.composite import Composite
+
+
+# Composite initialize
+
+def test_Composite_init_1():
+    result = Composite('moo').get_content()
+    assert(
```
tobes: Sometimes I forget

ultrabug:
Impressive, really

PR #528, opened 2016-10-23
fix graphite for tests in master

tobes:
Master breaks due to new more precise documentation testing.

This fixes things


tobes:
@ultrabug I merged this as it is a trivial fix and we don't want a broken master branch

PR #529, opened 2016-10-23
Make the net_rate module compatible with formatter colors

guiniol:


py3status/modules/net_rate.py
```diff
@@ -5,41 +5,53 @@
...
+    format: format of the module output
+        (default '{interface}: {total}')
+    format_no_connection: when there is no data transmitted from the start of the plugin
+        (default '')
+    hide_if_zero: hide indicator if rate == 0
+        (default False)
+    initial_multi: initial multiplier, if you want to get rid of first bytes, set to 1 to disable
+        (default 1024)
+    interfaces: comma separated list of interfaces to track
+        (default '')
```
tobes: reformatting makes PRs much harder to review, they should probably be their own PR but I'm not sure they add much value.

py3status/modules/net_rate.py
```diff
@@ -51,9 +63,15 @@ class Py3status:
     format = "{interface}: {total}"
     format_no_connection = ''
     hide_if_zero = False
+    initial_multi = 1024
```
tobes: I don't think these should be parameters.  really we need to move this sort of stuff into common code so that all modules can use the same paths rather than having lots of modules doing this independently.  I've got a load of such common functions that I should really make into a branch.

From the user point of view I think specifying a unit eg show me in 'Gb/s' makes sense but the details of the algorithm feel out of place.  I'm all for simplifying configs where we can whilst still allowing flexibility for the user.

py3status/modules/net_rate.py
```diff
@@ -123,16 +141,44 @@ def currentSpeed(self):
...
+                self.py3.COLOR_TOTAL = self.py3.COLOR_DEGRADED
+            else:
+                self.py3.COLOR_TOTAL = self.py3.COLOR_GOOD
+
+            if delta['up'] < self.threshold_slow:
+                self.py3.COLOR_UP = self.py3.COLOR_BAD
+            elif delta['up'] < self.threshold_fast:
+                self.py3.COLOR_UP = self.py3.COLOR_DEGRADED
+            else:
+                self.py3.COLOR_UP = self.py3.COLOR_GOOD
```
tobes: As in another PR don't change constants.  We need to find a nice way to do this.

py3status/modules/net_rate.py
guiniol: Since I'm working on this module, I figured I might as well do it.

py3status/modules/net_rate.py
guiniol: Ok. I thought that was the idea, but obviously I'm wrong.

if someone specifies `[\?color=up text]`, then the color of the text should be the color of `up`, but this varies with threshold, no? We could maybe mix colors and conditionals, but it may get ugly: `[\?critical[\?color=up_crit text]][\?degraded[\?color=up_degr text]]...`

guiniol:
Let's put #449 on hold until we figure out how we want to use formatter colors.

ultrabug:
Can we fix this PR now that thresholds are here maybe @guiniol ?

guiniol:
Actually, I still don't know how to use thresholds. I would expect `safe_format` but it doesn't seem to.
For example, here is the config I use for net_rate:
```
net_rate {
	format = "[\?color=down ⬇{down}] [\?color=up ⬆{up}]"
	cache_timeout = 1
}
```

If I understand things correctly, when I call, `safe_format`, it sets the colors, but only if those have been defined, which is why I was doing things like:
```
if delta['down'] < self.threshold_slow:
        self.py3.COLOR_DOWN = self.py3.COLOR_BAD
elif delta['down'] < self.threshold_fast:
        self.py3.COLOR_DOWN = self.py3.COLOR_DEGRADED
else:
        self.py3.COLOR_DOWN = self.py3.COLOR_GOOD
```

Should I do:
```
self.py3.COLOR_DOWN = self.py3.threshold_get_color(delta['down'])
```

But @tobes told me I shouldn't change the value of `self.py3.COLOR_DOWN`, so I'm lost.

tobes:
@guiniol does this help?

```
net_rate {
    format = "[\?color=down ⬇{down}] [\?color=up ⬆{up}]"
    # use same thresholds for everything
    thresholds = [(0, 'bad'), (20, 'degraded'), (50, 'green')]
}

net_rate {
    format = "[\?color=down ⬇{down}] [\?color=up ⬆{up}]"
    # different ones
    thresholds = {
        'up' : [(0, 'bad'), (20, 'degraded'), (50, 'green')],
        'down' : [(0, '#F0F'), (40, 'degraded'), (50, 'green')],
    }
}


Py3status:
    ....

    do_stuff(self):
        up_value, down_value = self._get_values()

        # we can ignore the return value as it sets the colors
        # so then they will magically be there
        self.py3.threshold_get_color(up_value, 'up')
        self.py3.threshold_get_color(down_value, 'down')

        # we don't need to pass the colors
        full_text = self.py3.safe_format(
            self.format, {'up': up_value, 'down': down_value}
        )        
```

guiniol:
@tobes it works! Didn't realise the magic was being done behind the scenes.

guiniol:
@tobes I have a problem with the tox check. It doesn't like tuples as default values in the docstring (`'thresholds' default not in docstring add (default ['UNKNOWN Tuple', 'UNKNOWN Tuple', 'UNKNOWN Tuple'])`). I think `test_module_doc.py` needs to be updated. I'll see if I understand how it works.

tobes:

If you can add tuples great

guiniol:

I saw that, but if we have to add every single `thresholds`, it will get old pretty fast.

tobes:
>  it will get old pretty fast.

True but it means you can get your module to pass the tests, and then we can update the doc_checker to understand tuples

guiniol:
@tobes opened #540 to fix that.

tobes:
@guiniol 

#540 is merged so the test can now pass

#535 is merged which does formatting of things in MiB etc

guiniol:
@tobes You work too fast. Anyhow, here it is, using both #540 and #535

py3status/modules/net_rate.py
```diff
@@ -74,12 +76,12 @@ def currentSpeed(self):
...
-        # == 6 characters (from MULTIPLIER_TOP + dot + self.precision)
+        # == 6 characters (from self.multiplier_top + dot + self.precision)
         if self.precision > 0:
-            self.left_align = len(str(MULTIPLIER_TOP)) + 1 + self.precision
+            self.left_align = len(str(self.multiplier_top)) + 1 + self.precision
         else:
-            self.left_align = len(str(MULTIPLIER_TOP))
-        self.value_format = "{value:%s.%sf} {unit}" % (
+            self.left_align = len(str(self.multiplier_top))
+        self.value_format = "{value:%s.%sf} {unit:>4s}" % (
```
tobes: I think that this should be made available to the user as `format_total = '{value} {unit}'` that way they can change it as the want - you can set a sensible default.

Look at the rate_counter modules `format_money` as an example

py3status/modules/net_rate.py
```diff
@@ -53,36 +63,31 @@ class Py3status:
     hide_if_zero = False
     interfaces = ''
     interfaces_blacklist = 'lo'
-    precision = 1
+    thresholds = [(0, "bad"), (1024, "degraded"), (1024*1024, "good")]
+    value_format = "{value:5.1f} {unit:>4s}"
```
tobes: can we call this `format_value` for consistency ie all format_xxx can use the formatter rules [`?color=... etc

py3status/modules/net_rate.py
```diff
@@ -5,41 +5,51 @@
...
+    devfile: location of dev file under /proc
+        (default '/proc/net/dev')
+    format: format of the module output
+        (default '{interface}: {total}')
+    format_no_connection: when there is no data transmitted from the start of the plugin
+        (default '')
+    hide_if_zero: hide indicator if rate == 0
+        (default False)
+    interfaces: comma separated list of interfaces to track
+        (default '')
```
guiniol: I can do that. By default, if it isn't a list, the module will try to split at `,` for backward compatibility

py3status/modules/net_rate.py
```diff
@@ -5,41 +5,51 @@
...
     {down} download rate
     {interface} name of interface
     {total} total rate
     {up} upload rate
 
+Value placeholders:
+    {unit} current unit
+    {value} numeric value
+
+Color conditionals:
```
guiniol: Probably, yes, now that we have the `threshold_get_color` function.

py3status/modules/net_rate.py
```diff
@@ -53,36 +63,31 @@ class Py3status:
     hide_if_zero = False
     interfaces = ''
     interfaces_blacklist = 'lo'
-    precision = 1
+    thresholds = [(0, "bad"), (1024, "degraded"), (1024*1024, "good")]
```
guiniol: The module actually switch to output the speed of the fastest interface, so there's no good way to specify the thresholds if you want something different between wifi and lan. These replicate the defaults values from before. Also, it kind of makes sense since even on wifi, if you're doing 3kbps, it's slow but if you're doing 1mpbs it's probably enough.

py3status/modules/net_rate.py
```diff
@@ -163,17 +176,12 @@ def dev_filter(x):
         except StopIteration:
             return None
 
-    def _divide_and_format(self, value):
+    def _format(self, value):
```
tobes: `_format_value()`?

py3status/modules/net_rate.py
```diff
@@ -163,17 +176,12 @@ def dev_filter(x):
...
+        Return formatted string
         """
-        for i, unit in enumerate(UNITS):
-            if value > MULTIPLIER_TOP:
-                value /= UNIT_MULTI
-            else:
-                break
-
-        return self.value_format.format(value=value, unit=unit)
+        value, unit = self.py3.format_units(value, unit="b/s", si=True)
```
guiniol: That was the default, so I kept it that way. I'll change it.

py3status/modules/net_rate.py
guiniol: Sure

py3status/modules/net_rate.py
```diff
@@ -50,39 +65,36 @@ class Py3status:
     devfile = '/proc/net/dev'
     format = "{interface}: {total}"
     format_no_connection = ''
+    format_value = "{value:5.1f} {unit:>5s}"
```
tobes: I think this should be None so we can tell if it has been user set.  If user set we use it if not then we may set using precision if set then just default.

Otherwise backward comptability is hard to handle even if this is all a pain

py3status/modules/net_rate.py
```diff
@@ -50,39 +65,36 @@ class Py3status:
...
-            self.left_align = len(str(MULTIPLIER_TOP))
-        self.value_format = "{value:%s.%sf} {unit}" % (
-            self.left_align, self.precision
-        )
+        if self.precision is not None:
+            if self.precision > 0:
+                self.left_align = 3 + 1 + self.precision
+            else:
+                self.left_align = 3
+            self.format_value = "{value:%s.%sf} {unit}" % (self.left_align, self.precision)
```
tobes: you see if you have precision set you kill my format_value here

py3status/modules/net_rate.py
```diff
@@ -5,41 +5,56 @@
...
+        (default "{value:5.1f} {unit:>5s}")
+    hide_if_zero: hide indicator if rate == 0
+        (default False)
+    interfaces: comma separated list of interfaces to track
+        (default [])
     interfaces_blacklist: comma separated list of interfaces to ignore
         (default 'lo')
-    precision: amount of numbers after dot (default 1)
+    si_units: use SI units
+        (default False)
```
tobes: This I need to think about

py3status/modules/net_rate.py
guiniol: This should be equivalent to the default if `precision` is set to `1` (which used to be the default). Of course, I need to fix this to be `{unit:>5s}` instead of `{unit}`. If the user has set `format_value`, then they shouldn't set `precision`.

py3status/modules/net_rate.py
guiniol: The user shouldn't leave `precision` if they set `format_value`. By giving precedence to `precision` we force its removal if they want to use `format_value`.

py3status/modules/net_rate.py
guiniol: Of course, but I think it has its place. I'd rather we ask the user to be explicit than try and guess what they want.

py3status/modules/net_rate.py
guiniol: Seeing how #546 is in, should I rewrite the default using the new `min_length`? (this also applies to #530 and #541)

tobes: I think we should get one of these PRs done first because once we get one done the others should all be easy.  but yes I think it'd be nicer if nothing else we don't get the 2 spaces issue for B/s which I think looks a little nasty

guiniol: Sure. I'll get something done.

tobes: I cannot see any reason why we should give precision the light of day.

> The user shouldn't leave precision if they set format_value. By giving precedence to precision we force its removal if they want to use format_value.

I think if they set format_value precision should be ignored as they are setting a new param and just have not removed an obsolete one.

guiniol:
@tobes I didn't actually reinstate `precision`. I left the `:.1f` part and used `min_length` for the padding. What do you think?

tobes:
That's fine but can you fix format_value so that if the user sets it it does not get changed

guiniol:
> can you fix format_value so that if the user sets it it does not get changed

Like I said, I don't think it's a good idea. I made it explicit in the doc that `precision` is deprecated. By giving it precedence over `format_value`, we ensure people who try to use both will see there is a problem immediately. It forces people to remove the obsolete configuration parameter, otherwise they probably wouldn't.

tobes:
>  It forces people to remove the obsolete configuration parameter, otherwise they probably wouldn't.

I think it makes it seem broken. IMHO it also makes it hard to manage the depreciation of config params as we lose the ability to see what the user has set as parameters.  I think if someone actually sets the new param we should respect that rather than making it not work.

Maybe @ultrabug has a view

guiniol:
I get that, but I think it's not a good trade off because we have more code and less robust tests. The user simply needs to remove the obsolete parameter.

tobes:
> The user simply needs to remove the obsolete parameter.

how about if they set both you do py3.notify_user('....')

guiniol:
> how about if they set both you do py3.notify_user('....')

but it still requires more code and breaks the tests all the same.

guiniol:
@tobes c0efcf3 does as you asked. Still, I would rather we revert that commit.

guiniol:
@tobes is there anything still blocking this?

guiniol:
I need to switch to the new Meta class for deprecating parameters.

ultrabug:
Yep @guiniol plz do

py3status/modules/net_rate.py
```diff
@@ -5,41 +5,52 @@
...
 
 Format placeholders:
     {down} download rate
     {interface} name of interface
     {total} total rate
     {up} upload rate
 
+Value placeholders:
+    {unit} current unit
+    {value} numeric value
```
tobes: these should be alphabetical

guiniol: I don't understand. U is before V.

tobes: I went to a special school

guiniol:
> I'd like a nicer thresholds default that took the speed of the interfaces into account, but I think that may be too painful to introduce without adding extra dependencies.

If you mean thresholds that change depending on the max speed of the interface, then yes, that's data we don't have for now. For wifi, it would change with the signal strength. And even for ethernet, it would change depending on whether it's ADSL or fiber, ... The best way would be to have user specify thresholds per interface. There could be code doing
```
cust_threshold = config.get('threshold_' + interface, self.threshold)
self.py3.threshold_get_color(delta['down'], 'down', thresholds=cust_thresholds)
```
That means modifying `threshold_get_color` to have a third optional arg which would default to `self.thresholds`.

tobes:
> If you mean thresholds that change depending on the max speed of the interface, then yes, that's data we don't have for now. For wifi, it would change with the signal strength. And even for ethernet,

yes, and I think for now let's not worry, but I'd go for max possible speed eg wifi 56kb/s nic 1Gb/s

But I think we should just get this merged first and have that as a nice dream for the future

py3status/modules/net_rate.py
```diff
@@ -5,41 +5,52 @@
...
+        exact unit will ever be used
+        (default "B/s")
 
 Format placeholders:
     {down} download rate
     {interface} name of interface
     {total} total rate
     {up} upload rate
 
+Value placeholders:
```
tobes: Change this and I think we are merge ready.  It'll be consistent with the way you have in sysdata

guiniol: @tobes done

PR #530, opened 2016-10-23
New module: Diskdata, shows both usage data and IO data

guiniol:


py3status/modules/diskdata.py
```diff
@@ -0,0 +1,226 @@
...
+    {write} writing rate
+
+Color conditionals:
+    {free} Change color based on the value of free
+    {used} Change color based on the value of used_percent
+    {read} Change color based on the value of read
+    {total} Change color based on the value of total
+    {write} Change color based on the value of write
+
+@author shadowprince
```
ultrabug: you're the author @guiniol :)

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,226 @@
...
+@license Eclipse Public License
```
ultrabug: should be BSD plz

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,226 @@
+# -*- coding: utf-8 -*-
+"""
+Display the current disk status.
```
ultrabug: Display advanced disk usage information.

Might be a better description ?

py3status/modules/diskdata.py
guiniol: the curse of copy/paste ^^

py3status/modules/diskdata.py
guiniol: See above

py3status/modules/diskdata.py
guiniol: Yes it is. Changed.

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,171 @@
...
+            self.values['write'] = self._format(write)
+            self.py3.threshold_get_color(read, 'read')
+            self.py3.threshold_get_color(total, 'total')
+            self.py3.threshold_get_color(write, 'write')
+
+        if '{free}' in self.format or '{used' in self.format:
+            free, used, used_percent = self._get_free_space(self.disk)
+
+            self.values['free'] = self.format_space.format(free)
+            self.values['used'] = self.format_space.format(used)
```
tobes: these should use safe_format

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,171 @@
...
+                    if data[1] == '0':
+                        read += int(data[5]) * self.sector_size
+                        write += int(data[9]) * self.sector_size
+        return read, write
+
+    def _format(self, value):
+        """
+        Return formatted string
+        """
+        value, unit = self.py3.format_units(value, unit="b/s", si=True)
```
tobes: As before I'm not sure si is a good idea as people expect none si eg 2^10 units even if they write `MB` we can help educate.

Also is this actually bits/second I suspect it is `B/s`

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    cache_timeout = 10
+    disk = None
+    format = "{disk}: {used_percent}% ({total})"
+    format_rate= "{value:5.1f} {unit:>4s}"
```
tobes: I'd prefer `format_rate= "{value} {unit}"` but saying that I haven't run the module yet so not sure how jumpy it is but I think we can do better on the displaying non jumpy.  But may need some more formatter  fun

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,171 @@
...
+    {free} free space on disk in GB
+    {used} used space on disk in GB
+    {used_percent} used space on disk in %
+    {read} reading rate
+    {total} total IO rate
+    {write} writing rate
+
+Rate format placeholders:
+    {unit} name of the unit
+    {value} numeric value of the rate
```
tobes: Space format has no love it should have things defined too

maybe `format_rate placeholders:`

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,171 @@
...
+    disk: disk whose stat to check. Set to 'all' to get global stats.
+        (default None)
+    format: format of the output.
+        (default "{disk}: {used_percent}% ({total})")
+    format_rate: format for the rates value
+        (default "{value:5.1f} {unit:>4s}")
+    format_space: format for the disk space values
+        (default "{value:3.1f}")
+    sector_size: size of the disk's sectors.
+        (default 512)
```
tobes: I thought this is fixed in the linux codebase and so will never change

py3status/modules/diskdata.py
```diff
@@ -0,0 +1,171 @@
+# -*- coding: utf-8 -*-
+"""
+Display advanced disk usage information
+
+Configuration parameters:
+    cache_timeout: how often we refresh this module in seconds.
+        (default 10)
+    disk: disk whose stat to check. Set to 'all' to get global stats.
+        (default None)
```
tobes: is this not about partition/ filesystem not disk

guiniol: This is about partitions or disks, not mount points.

py3status/modules/diskdata.py
guiniol: Forgotten child ^^ Fixed

guiniol:
@tobes this one doesn't have an  obsolete `precision`, so once the checks are fixed, it should be good to go.

@ultrabug I did the requested changes, but your change request is still blocking this PR. Do they need a manual review? (the automated checks are also blocking this PR until a proper fix for #545 is merged, so no hurry)

guiniol:
@ultrabug now that #545 has been merged and that the checks pass, can you re-review this? It's still blocked by the changes you requested even though I implemented them.

PR #531, opened 2016-10-24
i3status time modules fix

tobes:
There were some bugs in the i3status module around handling times.

1) Sometimes the module was not updated when it should have been.  This occurs when py3status is first started and the modules time format is the default.  The module can take several seconds to appear.  We now update as required.

2)  The logic for recreating the Tz timezones once a minute was incorrect and we were regenerating them every second.


ultrabug:
interesting old bug, good catch !

PR #532, opened 2016-10-25
Thresholds

tobes:
Following my plans set out in issue #451 here is an implementation of thresholds.

* `threshold_check()` function added to Py3.  This both returns the correct threshold color for a value but also sets the color for use in formatting.

* Threshold colors added to module as `color_threshold` accessed as `threshold` or as `color_threshold_<name>` accessed as `<name>` this makes things easy for the user whilst reducing the likelihood of accidentally overwriting variables. 

* single or multiple thresholds can be easily defined.

* thresholds can now have as many steps as the user wishes to define as opposed to just the ones of the module writer.

* Some basic documentation added.

* volume_status module updated as an example.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -9,6 +9,7 @@ py3status documentation
 * [Configuring modules](#configuring_modules)
 * [Configuration obfuscation](#obfuscation)
 * [Configuring colors](#configuring_color)
+* [Configuring thresholds](#configuring_thresholds)
 * [Grouping modules](#group_modules)
 
 [Custom click events](#on_click)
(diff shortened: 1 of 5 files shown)


ultrabug:
I find the idea and implementation very smart, thanks a lot

Two remarks:

1. the documentation does not explain clearly the boundaries of the threshold color selection, I think an example should be provided (with >= XX >= YY stuff)
2. the `threshold_chek` method name is not explicit to me, I'd rather have it named `get_color_from_threshold` or something similar

tobes:
>    the documentation does not explain clearly the boundaries of the threshold color selection, I think an example should be provided (with >= XX >= YY stuff)

I've added some more explanation, it could probably still be made clearer

>    the threshold_chek method name is not explicit to me, I'd rather have it named get_color_from_threshold or something similar

yeah I wasn't very happy with the name (naming stuff is hard) how about `threshold_get_color()`?  the reason I chose this is so if we add more threshold commands they get grouped.  Py3 is starting to get quite big so we need to start trying to keep it managable.  I'm happy to use `get_color_from_threshold()` if you prefer.

ultrabug:
`threshold_get_color()` is good thanks :)

PR #533, opened 2016-10-27
Allow multiple i3status configs and switcher module

tobes:
Ok so this is a bit of a monster.

So what's included
- i3status is further separated from core so now there is a I3statusRunner that manages the i3status process better.  We can now stop/start i3status processes including properly killing i3status so that we no longer end up with extra i3status processes.  I3status is also created in a thread that stops the i3status startup blocking the rest of our starting up. 
- py3status modules are now started in threads that allows a much faster startup when lots of modules are in use.
- The real reason for this is that we can now specify multiple config files eg `py3status -c config_1 config_2 ...` and with the help of the new config_switcher module switch between them.

I am sort of toying with the idea of automatically injecting the config_switcher module into the i3status.conf when more than one is specified but the config_switcher is not included as that sort of makes sense so users can then change configs.  Anyhow I'll think about that for a bit and see how I feel about it.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -11,14 +11,14 @@ from json import dumps
 from platform import python_version
 from signal import signal, SIGTERM, SIGUSR1, SIGTSTP, SIGCONT
 from subprocess import Popen
-from threading import Event
+from threading import Event, Thread
 from syslog import syslog, LOG_ERR, LOG_INFO, LOG_WARNING
 from traceback import extract_tb, format_tb, format_stack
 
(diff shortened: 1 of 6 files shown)


ultrabug:
since this is huge, I'm waiting for more time and focus to properly dive into it, sorry

tobes:
>  I'm waiting for more time and focus to properly dive into it, sorry

No rush

tobes:
I'm going to close this and split it and create some future PRs that will be simpler to merge

PR #535, opened 2016-10-28
Py3: helper function format_units()

tobes:
This adds a helper function to py3

The idea is to create human readable output for modules like sysdata it also means we have one true implementation and so should be less subject to errors etc.  We can force the output if needed say to MiB or we can use something like 'bits/second' and get Mibits/second etc.

The output is a numeric (so users can format in their format strings) and a unit.

I think this would be a useful helper for various modules that need this functionality.

some examples
```
format_units(1024) ->  (1.0, 'KiB')
format_units(1024 * 1024) ->  (1.0, 'MiB')
format_units(1024 * 1024, unit='KiB') ->  (1024, 'KiB')
format_units(1024, unit='b/sec') ->  (1.0, 'Kib/sec')
```

It also adds some tests

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -856,6 +856,33 @@ they have.
 Returns True if the event name and instance match that of the module
 checking.
 
+___format_units(value, unit='B', optimal=5, auto=True, si=False)__
+
+Takes a value and formats it for user output, we can choose the unit to
+use eg B, MiB, kbits/second.  This is mainly for use with bytes/bits it
+converts the value into a human readable form.  It has various
+additional options but they are really only for special cases.
+
+The function returns a tuple containing the new value (this is a number
+so that the user can still format it if required) and a unit that is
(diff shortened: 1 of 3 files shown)


doc/README.md
```diff
@@ -856,6 +856,29 @@ they have.
 Returns True if the event name and instance match that of the module
 checking.
 
+___format_units(value, unit='B', optimal=5, auto=True, si=False)__
```
ultrabug: auto is not documented, what is it for please ?

tobes: I've added a note.  It is about forcing the units to use rather than us trying to make them nice.

format_units(1024)  -> (1.0, 'KiB')
format_units(1024, auto=False) -> (1024, 'B')

ultrabug:
Awesome, thanks a lot :)

PR #536, opened 2016-10-28
volume_status: add pactl support

tobes:
I've added a new 'backend' for volume_status to use `pactl`

My main reason for this is pulseaudio allows a volume greater than 100% useful for quiet youtube videos etc.

I've added a `max_volume` parameter to control this.  @guiniol how does pamixer deal with setting volumes above 100%?  maybe you could add max_volume support for it.

### DIFF
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -18,8 +18,8 @@ Configuration parameters:
         (default 10)
     channel: channel to track. Default value is backend dependent.
         (default None)
-    command: Choose between "amixer" and "pamixer". If None, try to guess based
-        on available commands.
+    command: Choose between "amixer", "pamixer" or "pactl".
+        If None, try to guess based on available commands.
         (default None)
     device: Device to use. Defaults value is backend dependent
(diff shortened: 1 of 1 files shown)


guiniol:
I'll have to check. Pretty new to pulseaudio, I didn't know I could do that ^^

ultrabug:
Thanks @tobes , what if we're not using pulseaudio ? wouldn't 100 be a better and safer default ?

guiniol:
@tobes I checked pamixer, but while there's a way to allow going past 100%, I can't use the increase/decrease and have to set the volume explicitely (so I also need to get it). I don't think I'll add support for that on this backend. However, it might be good to rename the backend from `PulseaudioBackend` to `PamixerBackend`.

@ultrabug It has to be used by backends explicitely, so it's ignored by the alsa one. No problem there.

ultrabug:
@guiniol thanks for clarifying, okay

tobes:
> However, it might be good to rename the backend from PulseaudioBackend to PamixerBackend.

done

PR #539, opened 2016-10-29
backlight module: allow chaging brightness with mouse scroll

ultrabug:
and setting initial brightness on start !

I was frustrated that the backlight module didn't provide the functionality so I figured that it was about time I got to code again :D

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -3,6 +3,16 @@
 Display the current screen backlight level.
 
 Configuration parameters:
+    brightness_delta: Change the brightness by this step.
+        (default 5)
+    brightness_initial: Set brightness to this value on start.
+        (default None)
+    brightness_minimal: Don't go below this brightness to avoid black screen
+        (default 1)
+    button_down: Button to click to decrease brightness. Setting to 0 disables.
+        (default 5)
+    button_up: Button to click to increase brightness. Setting to 0 disables.
+        (default 4)
     cache_timeout: How often we refresh this module in seconds (default 10)
     device: The backlight device
         If not specified the plugin will detect it automatically
@@ -15,6 +25,9 @@ Configuration parameters:
 
 Format status string parameters:
     {level} brightness
+
+Requires:
+    xbacklight: need for changing brightness, not detection
 
 @author Tjaart van der Walt (github:tjaartvdwalt)
 @license BSD
@@ -37,6 +50,11 @@ class Py3status:
     """
     """
     # available configuration parameters
+    brightness_delta = 5
+    brightness_initial = None
+    brightness_minimal = 1
+    button_down = 5
+    button_up = 4
     cache_timeout = 10
     device = None
     device_path = None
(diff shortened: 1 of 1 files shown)


ultrabug:
lol failed CI for being a bad kitty... Am I getting old ? (jk) ahah

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
tobes: I'm not sure this should be here

py3status/modules/backlight.py
```diff
@@ -3,19 +3,34 @@
 Display the current screen backlight level.
 
 Configuration parameters:
+    brightness_delta: Change the brightness by this step
+        (default 5)
+    brightness_initial: Set brightness to this value on start
+        (default 100)
```
tobes: how about having this default as None, but if set then will set that value at startup?  That way we have one less parameter.

py3status/modules/backlight.py
```diff
@@ -3,19 +3,34 @@
...
+    force_on_start: Should we set initial brightness on start
+        (default False)
     format: Display brightness, see placeholders below
         (default '☼: {level}%')
 
 Format status string parameters:
     {level} brightness
 
+Requires:
+    xbacklight: only for click events, not for brightness detection
```
tobes: > click events

it is really for changing the brightness as this includes the initial setting

py3status/modules/backlight.py
```diff
@@ -37,24 +52,55 @@ class Py3status:
...
         for brightness_line in open("%s/brightness" % self.device_path, 'rb'):
             brightness = int(brightness_line)
         for brightness_max_line in open("%s/max_brightness" % self.device_path, 'rb'):
             brightness_max = int(brightness_max_line)
+        return brightness * 100 // brightness_max
 
-        full_text = self.format.format(level=(brightness * 100 // brightness_max))
+    def backlight(self,  i3s_output_list, i3s_config):
+        level = self._get_backlight_level()
+        full_text = self.format.format(level=level)
```
tobes: We should change this to use safe_format so we can have crazy formats

py3status/modules/backlight.py
```diff
@@ -37,24 +52,55 @@ class Py3status:
...
+
+    def _get_backlight_level(self):
         for brightness_line in open("%s/brightness" % self.device_path, 'rb'):
             brightness = int(brightness_line)
         for brightness_max_line in open("%s/max_brightness" % self.device_path, 'rb'):
             brightness_max = int(brightness_max_line)
+        return brightness * 100 // brightness_max
 
-        full_text = self.format.format(level=(brightness * 100 // brightness_max))
+    def backlight(self,  i3s_output_list, i3s_config):
```
tobes: maybe remove `,  i3s_output_list, i3s_config` as you're here

py3status/modules/backlight.py
ultrabug: done

py3status/modules/backlight.py
ultrabug: done

py3status/modules/backlight.py
ultrabug: indeed

ultrabug:
Thanks, addressed your cool review !

PR #540, opened 2016-10-29
Add tuples support to doc checker

guiniol:
I simplified `RE_PARAM` to
```
RE_PARAM = re.compile(
    '^  - `(?P<name>[^`]*)`.*?('
    '\*\(default (?P<value>('
    '.*'
    '))\)\*)?$'
)
```
And it seems to work for all the tests. If it is now too broad (and matches things it shouldn't), I can restrict it a bit, but this allows for nested tuples and everything. And even if it matches things it shouldn't (like unmatched quotes), that should be caught by the eval.

### DIFF
--- a/tests/test_module_doc.py
+++ b/tests/test_module_doc.py
@@ -47,11 +47,7 @@ OBSELETE_PARAM = [
 RE_PARAM = re.compile(
     '^  - `(?P<name>[^`]*)`.*?('
     '\*\(default (?P<value>('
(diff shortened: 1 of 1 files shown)


tobes:
Yeah not sure why my RE was so complex if the simpler one works anyway.  I'll run a few checks to see if it's happy and catching errors

tobes:
Ok looks good

ultrabug:
Good job, thanks !

PR #541, opened 2016-10-29
Sysdata: add support for colors/thresholds

guiniol:
This PR adds support for the new colors/thresholds utilities.

(I removed the swap and load support from this PR and open new PRs once this one is in)

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -18,9 +18,6 @@ Configuration parameters:
 Format placeholders:
     {cpu_temp} cpu temperature
     {cpu_usage} cpu usage percentage
-    {load1} load average over the last minute
-    {load5} load average over the five minutes
-    {load15} load average over the fifteen minutes
     {mem_total} total memory
     {mem_used} used memory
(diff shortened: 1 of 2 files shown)


ultrabug:
Two things

Add yourself in the @author

I simply tested the module by loading it without config from master and then yours

- master one showed red
- yours without color

is that expected?

guiniol:
> Add yourself in the @author

Should I do that for every module I make modifications to?


> is that expected?

No, but I didn't test it, so... I'll figure out what is going on.

guiniol:
> I'll figure out what is going on.

That was easy. I wasn't even trying. Fixed ^^

tobes:
I think we need to remove precision and padding parameters from this but I can make a PR for that

guiniol:
@tobes you're right. I completely forgot. I'll do it like in #529 (with less complaining) unless you already started.

guiniol:
@tobes I see you've added magic things for deprecating config parameters, so I tried using that in #553, but I can't get it to work. Once I know how to deprecate arguments using that, I'll make the PR for removing precision and padding.

PR #542, opened 2016-10-30
Check that value != 0 before taking the log

guiniol:
`log(0)` is not something we want to do, so check that value != 0 before doing any such thing.

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -238,7 +238,7 @@ class Py3:
         if can_round and optimal and value:
             # we will try to make the output value the desired size
             # we need to keep out value as a numeric type
-            places = int(log10(value))
+            places = int(log10(abs(value)))
             if places >= optimal - 2:
                 value = int(value)
             else:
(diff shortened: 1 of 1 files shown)


py3status/py3.py
```diff
@@ -235,7 +235,7 @@ def format_units(self, value, unit='B', optimal=5, auto=True, si=False):
                 value /= size
                 can_round = True
 
-        if can_round and optimal:
+        if can_round and optimal and not value == 0:
```
tobes: good catch

`if can_round and optimal and value:` feels more readable

guiniol: Well, my code caught it ^^

guiniol:
Also, use the `abs` in the log, in case the value is negative.

tobes:
Ok looks good I'll add some tests as I have a branch for that including the 0 case and better errors but it needs these fixes

ultrabug:
Thanks !

PR #545, opened 2016-10-30
Multi-line default values don't end with ')'

guiniol:
Only add `*` around the default value if it's single line (ends with a `)`)

### DIFF
--- a/py3status/docstrings.py
+++ b/py3status/docstrings.py
@@ -111,7 +111,7 @@ re_from_status = re.compile('^\s+({\S+})($|[ \t])(.*)$')
 re_from_item = re.compile('^\s+-')
 re_from_data = re.compile('^@(author|license|source)($|[ \t])')
 re_from_tag = re.compile('((`[^`]*`)|[<>&])')
-re_from_defaults = re.compile('(\(default.*\))')
+re_from_defaults = re.compile('(\(default.*\))\s*$')
 
 
 def _reformat_docstring(doc, format_fn, code_newline=''):


py3status/docstrings.py
```diff
@@ -111,7 +111,7 @@ def create_readme(data):
 re_from_item = re.compile('^\s+-')
 re_from_data = re.compile('^@(author|license|source)($|[ \t])')
 re_from_tag = re.compile('((`[^`]*`)|[<>&])')
-re_from_defaults = re.compile('(\(default.*\))')
+re_from_defaults = re.compile('(\(default.*\))$')
```
tobes: We do something a little more clever? but it quickly becomes unreadable :)

```
import re

r1 = re.compile('(\(default.*\))$')


r2 = re.compile('(\(default ('
               '[^(\'"]*'
               '|'
               '\([^(]*\)' # tuple
               '|'
               r'("(?:[^"\\]|\\.)*")'  # double quoted string
               '|'
               r"|('(?:[^'\\]|\\.)*')"  # single quoted string
               ')\))')

r3 = re.compile('(\(default ('
               '[^(\'"]*'
               '|'
               '\((' # tuple start
               '[^(\'"]*'
               '|'
               r'("(?:[^"\\]|\\.)*")'  # double quoted string
               '|'
               r"|('(?:[^'\\]|\\.)*')"  # single quoted string
               ')*\)' # tuple end
               '|'
               r'("(?:[^"\\]|\\.)*")'  # double quoted string
               '|'
               r"|('(?:[^'\\]|\\.)*')"  # single quoted string
               ')\))')


items = [
    '(default "moo")',
    '(default "moo") ',
    '(default ("moo"))',
    '(default ("m(oo"))',
    '(default ("m(oo", \'moo\'))',
    '(default ("moo",)',
    '(default ("mo)o",)',
    '(default "mo(o")',
]

for item in items:
    p1 = bool(r1.match(item))
    p2 = bool(r2.match(item))
    p3 = bool(r3.match(item))
    print(item, p1, p2, p3)
```

gives

```
(default "moo") True True True
(default "moo")  False True True
(default ("moo")) True True True
(default ("m(oo")) True False True
(default ("m(oo", 'moo')) True False True
(default ("moo",) True False False
(default ("mo)o",) True False False
(default "mo(o") True True True
```

We should get a proper parser going

guiniol: > We should get a proper parser going

Definitely. Because you're still missing the regexp with a tuple inside a tuple, and the one with a tuple inside a tuple inside a tuple, and so on.
Ideally, we could increment a counter whenever we find a `(` outside a string, decrement it whenever we find a `)` outside a string and check if the counter equals 0 a the end, but apparently python `re` doesn't support that.

tobes: Let's keep it simple do you want to just allow whitespace between the `)` and the end of the line and then we can think about a proper parser another day

guiniol: @tobes missed your last comment. It is done.

tobes:
`\s*` would be better than ` *` as it'll catch any whitespace not just space

guiniol:
Of course. Done.

tobes:
Thanks

@ultrabug this helps with the docsting tests when a tuple is defined.  It is not perfect as really we need a better parser rather than just trying more clever regexs.  But it fixes an issue that has affected @guiniol

ultrabug:
Okay, thanks guys ! until we do better

PR #544, opened 2016-10-30
Bugfix: py3.build-composite() merge error

tobes:
I did a bad merge this fixes it.

PR #547, opened 2016-10-31
fix py3 call to formatter passing an unexistent 'composites' key

ultrabug:
that did throw an 'invalid format' on the bar when using group !


tobes:
yeah this is the same as #544

PR #549, opened 2016-11-01
Universal module config parameter deprecation

tobes:
Currently we have modules that have a variety of configuration parameters that follow different standards.  It would be nice to have consistency in module parameter names eg `format_xx` for format strings, thresholds etc.  This PR allows us to change module configuations.

* Existing configuration files will continue to work for the user.

* If users set parameters they will not be overwritten, although they may be corrected (currently only {} being updated).

* Reduce the complication of code modules have dealing with legacy config options, eg battery_level has an obsolete `mode` parameter.

* Allow updated configurations to be applied to modules as though they were the user supplied ones eg colors will be available via `py3.COLOR_XXX`

* Deprecations across all modules are dealt with the same, logging etc.

* Allows defaults to be set in modules without causing complications.

* Improves docstring tests as less parameters need excluding as defaults can be set etc.

* All module deprecation information is in a single place in the module.  Although the format is slightly verbose it should be easy to understand and is flexible and so can allow for future expansion if needed.

This PR also provides updates to the following modules.  These demonstrate the types of changes that we can easily make to modules if this PR is accepted.

__battery_level__

Replaces the old way of dealing with obsolete parameters - also removes them from the documentation as the updates needed are self documented.

__file_status__

We move the module to having a single more flexible format.  Additionally we also allow `~` to be used in the path and simplify the check.

__spaceapi__

Switch from using `closed_color`, `open_color` to `color_closed`, `color_open`

__volume_status__

Replace `threshold_bad`, `threshold_degraded` with new style `thresholds`.

Documentation
============

Sometimes it is necessary to deprecate configuration parameters.  Modules
are able to specify information about deprecation so that it can be done
automatically.  Deprecation information is specified in the Meta class of a
py3status module using the deprecated attribute.  The following types of
deprecation are supported.

The deprecation types will be performed in the order here.

__rename__

The parameter has been renamed.  We will update the configuration to use the
new name.

```
class Py3status:

    class Meta:

        deprecated = {
            'rename': [
                {
                    'param': 'format_available',  # parameter name to be renamed
                    'new': 'icon_available',   # the parameter that will get the value
                    'msg': 'obsolete parameter use `icon_available`',  # message
                },
            ],
        }
```

__format_fix_unnamed_param__

Some formats used `{}` as a placeholder this needs to be updated to a named
placeholder eg `{value}`.

```
class Py3status:

    class Meta:

        deprecated = {
            'format_fix_unnamed_param': [
                {
                    'param': 'format',  # parameter to be changed
                    'placeholder': 'percent',  # the place holder to use
                    'msg': '{} should not be used in format use `{percent}`',  # message
                },
            ],
        }
```

__substitute_by_value__

This allows one configuration parameter to set the value of another.

```
class Py3status:

    class Meta:

        deprecated = {
            'substitute_by_value': [
                {
                    'param': 'mode',  # parameter to be checked for substitution
                    'value': 'ascii_bar',  # value that will trigger the substitution
                    'substitute': {
                        'param': 'format',  # parameter to be updated
                        'value': '{ascii_bar}',  # the value that will be set
                    },
                    'msg': 'obsolete parameter use `format = "{ascii_bar}"`',  #message
                },
            ],
        }
```

__function__

For more complex substitutions a function can be defined that will be called with the config as a parameter.  This function must return a dict of key value pairs of parameters to update

```
class Py3status:

    class Meta:

        # Create a function to be called
            # This function must return a dict
            return {'thresholds': [
                        (0, 'bad'),
                        (config.get('threshold_bad', 20), 'degraded'),
                        (config.get('threshold_degraded', 50), 'good'),
                    ],
            }

        deprecated = {
            'function': [
                {
                    'function': deprecate_function,  # function to be called
                },
            ],
        }
```

__remove__

The parameters will be removed.

```
class Py3status:

    class Meta:

        deprecated = {
            'remove': [
                {
                    'param': 'threshold_bad',  # name of parameter to remove
                    'msg': 'obsolete set using thresholds parameter',  #message
                },
            ],
        }
```

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -28,6 +28,7 @@ py3status documentation
 * [Py3 module helper](#py3)
 * [Composites](#composites)
 * [Module documentation](#docstring)
+* [Deprecation of configuration parameters](#deprecation)
 * [Module testing](#testing)
 
 [Contributing](#contributing)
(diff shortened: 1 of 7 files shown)


ultrabug:
Took me a while to grasp but I'm amazed by it, thank you !

I rebased so I can merge !

PR #550, opened 2016-11-02
vpn_status: remove shebang

jakubjedelsky:
because it should depends on OS setting, not to have harcoded usage
of python3 via env.

Uncovered by rpmlint while building package for Fedora.


ultrabug:
good catch thank you @jakubjedelsky

PR #552, opened 2016-11-04
Add load support to sysdata

guiniol:
This reverts commit 4d90fd69d75b83c5e15244f5bf1a043b3b0f83fb, and thus, re-adds load support to sysdata. (There'll be another one for swap).
This adds three new placeholders `load1`, `load5` and `load15` and one new threshold `load` compared to the value of `load1`.

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -358,7 +358,7 @@ class Py3status:
             self.py3.threshold_get_color(swap_used_percent, 'swap')
 
         # get load average
-        if '{load' in self.format:
+        if self.py3.format_contains(self.format, 'load*'):
             load1, load5, load15 = self.data.load()
             self.values['load1'] = load1
             self.values['load5'] = load5
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -248,6 +248,9 @@ class Py3status:
             return {
                 'cpu_usage': format_vals,
                 'cpu_temp': format_vals,
+                'load1': format_vals,
+                'load5': format_vals,
+                'load15': format_vals,
                 'mem_total': format_vals,
                 'mem_used': format_vals,
                 'mem_used_percent': format_vals,
@@ -292,6 +295,9 @@ class Py3status:
                     'placeholder_formats': {
                         'cpu_usage': ':.2f',
                         'cpu_temp': ':.2f',
+                        'load1': ':.2f',
+                        'load5': ':.2f',
+                        'load15': ':.2f',
                         'mem_total': ':.2f',
                         'mem_used': ':.2f',
                         'mem_used_percent': ':.2f',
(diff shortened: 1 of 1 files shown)


guiniol:
Needs to wait for #555 to be merged

guiniol:
Now that #555 is merged, I updated this one to use the new tools for the load support in sysdata.
After that, I've only got the swap left.

guiniol:
@tobes updated this branch to resolve conflicts (apparently, someone else already added swap support \o/). Should be good to go unless new helpers have been created since last time I hacked at it.

py3status/modules/sysdata.py
```diff
@@ -326,6 +348,14 @@ def sysData(self):
             self.values['swap_unit'] = swap_unit
             self.py3.threshold_get_color(swap_used_percent, 'swap')
 
+        # get load average
+        if self.py3.format_contains(self.format, 'load*'):
+            load1, load5, load15 = self.data.load()
+            self.values['load1'] = load1
+            self.values['load5'] = load5
+            self.values['load15'] = load15
+            self.py3.threshold_get_color(load1, 'load')
```
tobes: I think this should be `self.py3.threshold_get_color(load1 * 100, 'load')` that way we can give thresholds using whole numbers eg `thresholds = [(0, "#FF0000"), (30, "#FFFF00"), (80, "#00FF00"),]`

Other than this I think we're looking good

guiniol: It seems weird to not use the same values as the load for the load threshold. Also, I would expect people to use a threshold based on their number of core (1x cores or less : OK, 2x cores or more: NOK), no?

tobes: we sort of do this with cpu because that's a percent as well.  It's just that otherwise if I have thresholds then they trigger for everything but load.  I'll have a think

@ultrabug you have a view here?

guiniol: @tobes load is not a percent, it's a flat value.

guiniol: The problem is that the explanation above works if you have one core. If you have two cores, a load of 2 is perfectly fine, it means each of your core had just enough work to keep them busy. You get the idea. Load is the average number of active process at once over the specified amount of time. So it's really a number of processes. If we really want to see it as percentage, we have to do `load * 100 / n_cpus`. And I really don't think it'll match people expectations. If for everything else, the threshold is compared against the value and for this value only it's compared against the load percent per cpu, that's going to surprise people.

guiniol:
@tobes Sorry to bother you again, but is there anything in particular holding up this PR?
I rebased the code and checked to see if anything changed in the way of doing things but everything seemed fine (the latest release notes mentioned improvement in the way to write modules, but I think sysdata was an early adopter).

tobes:
> Sorry to bother you again

That's fine.  I think this just got held up by the thresholds issue.  I'd like it if they sort of worked out of the box with the defaults.  Anyhow I think I can live with this and having the loads is a useful feature. 

Sorry this has dragged on so long and also that it didn't get into 3.5

Thanks for this new feature

guiniol:
Thanks for merging it in. And I really think there's no issue with the thresholds because load isn't a percent ^^

PR #553, opened 2016-11-05
Use new Meta/deprecate_function in sysdata

guiniol:
Try to use the new Meta/deprecated function for deprecated config args. Tox doesn't like it.

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -29,12 +29,6 @@ Color thresholds:
     mem: change color based on the value of mem_used_percent
     temp: change color based on the value of cpu_temp
 
-Obsolete configuration parameters:
-    high_threshold: percent to consider CPU or RAM usage as 'high load'
-        (default 75)
-    med_threshold: percent to consider CPU or RAM usage as 'medium load'
-        (default 40)
-
 NOTE: If using the `{cpu_temp}` option, the `sensors` command should
 be available, provided by the `lm-sensors` or `lm_sensors` package.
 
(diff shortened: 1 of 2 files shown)


py3status/modules/sysdata.py
```diff
@@ -140,27 +134,42 @@ class Py3status:
...
+        def deprecate_function(config):
+            # support old thresholds
+            return {
+                    'thresholds': [
+                        (0, 'good'),
+                        (config.get('med_threshold', 40), 'degraded'),
+                        (config.get('high_threshold', 75), 'bad'),
+                        ]
+                    }
+            deprecated = {
```
tobes: This is on the wrong indention level. Should be part of Meta not the function

guiniol: Of course... Should be good now.

ultrabug:
plz rebase mate

guiniol:
@ultrabug done

tobes:
I think one good change (should have been in #541 really) would be to alter the default format to 
```
    format = "[\?color=cpu CPU: {cpu_usage}%], " \
        "[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]"
```

That way users automatically get nicer behavior and I think the change is an improvement to the current output rather than a regression.  But it doesn't stop this being merged

guiniol:
> That way users automatically get nicer behavior and I think the change is an improvement to the current output rather than a regression. But it doesn't stop this being merged

I even added `max_cpu_mem` to get the original behaviour ^^

tobes:
> I even added max_cpu_mem to get the original behaviour ^^

I approve of this approach and generally I do not think output should change.  But I also think users should get the advantage of the new 'works like it should' version without having to do anything in this instance.  The fact they can still have the 'broken' output if they insist is a nice touch.

guiniol:
I can make a "better defaults" PR. Or add it to this one if you want.

tobes:
> I can make a "better defaults" PR. Or add it to this one if you want.

I'm happy either way but maybe do as a separate one just in case @ultrabug doesn't agree

guiniol:
No problem. I'll open another one once this one is in.

guiniol:
@tobes opened #558

PR #555, opened 2016-11-05
Sysdata: deprecate precision and padding

guiniol:
@tobes I used the new Meta thing to deprecate precision and padding.
I have one question. Does that mean that if someone uses any of the deprecated parameters, all the new parameters will be ignored and set from the `deprecate_function` in `Meta`?

EDIT: forgot to mention, this is on top of #553

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -126,7 +126,7 @@ class GetData:
         # Otherwise, results are in gigabytes.
         return total_mem, used_mem, used_mem_p, unit
 
-    def cpuTemp(self, zone):
+    def cpuTemp(self, zone, unit):
         """
         Tries to determine CPU temperature using the 'sensors' command.
         Searches for the CPU temperature by looking for a value prefixed
@@ -134,17 +134,23 @@ class GetData:
         out temperatures of all codes if more than one.
         """
 
-        sensors = None
+        command = ['sensors']
+        if unit.lower() == 'F' or unit == u'°F':
+            command.append('-f')
+        elif unit.lower() not in ['C', u'°C', 'K']:
+            return 'unknown unit'
         if zone:
             try:
-                sensors = self.py3.command_output(['sensors', zone])
+                sensors = self.py3.command_output(command + [zone])
             except:
                 sensors = None
         if not sensors:
-            sensors = self.py3.command_output('sensors')
+            sensors = self.py3.command_output(command)
         m = re.search("(Core 0|CPU Temp).+\+(.+).+\(.+", sensors)
         if m:
             cpu_temp = float(m.groups()[1].strip()[:-2])
+            if unit == 'K':
+                cpu_temp += 273.15
         else:
             cpu_temp = '?'
 
@@ -253,7 +259,7 @@ class Py3status:
 
         # if specified as a formatting option, also get the CPU temperature
         if self.py3.format_contains(self.format, 'cpu_temp'):
-            cpu_temp = self.data.cpuTemp(self.zone)
+            cpu_temp = self.data.cpuTemp(self.zone, self.temp_unit)
             self.values['cpu_temp'] = cpu_temp
             self.py3.threshold_get_color(cpu_temp, 'temp')
 
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -159,7 +159,7 @@ class Py3status:
     format = "[\?color=cpu CPU: {cpu_usage}%], " \
              "[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]"
     mem_unit = 'GiB'
-    temp_unit = '°C'
+    temp_unit = u'°C'
     thresholds = [(0, "good"), (40, "degraded"), (75, "bad")]
     zone = None
 
@@ -175,10 +175,11 @@ class Py3status:
                         ],
                     }
 
-        def update_placeholder_format(config):
-            # padding = config.get('padding', 0)
+        def update_deprecated_placeholder_format(config):
+            padding = config.get('padding', 0)
             precision = config.get('precision', 2)
-            format_vals = ':.{precision}f'.format(precision=precision)
+            format_vals = ':{padding}.{precision}f'.format(padding=padding,
+                                                           precision=precision)
             return {
                     'cpu_usage': format_vals,
                     'cpu_temp': format_vals,
@@ -211,7 +212,15 @@ class Py3status:
                     ],
                 'update_placeholder_format': [
                     {
-                        'function': update_placeholder_format,
+                        'function': update_deprecated_placeholder_format,
+                        'format_strings': ['format']
+                        },
+                    ],
+                }
+
+        update_config = {
+                'update_placeholder_format': [
+                    {
                         'placeholder_formats': {
                             'cpu_usage': ':.2f',
                             'cpu_temp': ':.2f',
@@ -220,15 +229,15 @@ class Py3status:
                             'mem_used_percent': ':.2f',
                             },
                         'format_strings': ['format']
-                        }
-                    ]
+                        },
+                    ],
                 }
 
     def post_config_hook(self):
         self.data = GetData(self)
         self.cpu_total = 0
         self.cpu_idle = 0
-        self.values = {'temp_unit': '°C'}
+        self.values = {'temp_unit': self.temp_unit}
 
     def sysData(self):
         # get CPU usage info
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -135,10 +135,9 @@ class GetData:
         """
 
         command = ['sensors']
-        unit = unit.upper()
-        if unit in ['F', u'°F']:
+        if unit == u'°F':
             command.append('-f')
-        elif unit not in ['C', u'°C', 'K']:
+        elif unit not in [u'°C', 'K']:
             return 'unknown unit'
         if zone:
             try:
@@ -244,7 +243,15 @@ class Py3status:
         self.data = GetData(self)
         self.cpu_total = 0
         self.cpu_idle = 0
-        self.values = {'temp_unit': self.temp_unit}
+        temp_unit = self.temp_unit.upper()
+        if temp_unit in ['C', u'°C']:
+            temp_unit = u'°C'
+        elif temp_unit in ['F', u'°F']:
+            temp_unit = u'°F'
+        elif not temp_unit == 'K':
+            temp_unit = 'unknown unit'
+        self.values = {'temp_unit': temp_unit}
+        self.temp_unit = temp_unit
 
     def sysData(self):
         # get CPU usage info
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -135,9 +135,10 @@ class GetData:
         """
 
         command = ['sensors']
-        if unit.lower() == 'F' or unit == u'°F':
+        unit = unit.upper()
+        if unit in ['F', u'°F']:
             command.append('-f')
-        elif unit.lower() not in ['C', u'°C', 'K']:
+        elif unit not in ['C', u'°C', 'K']:
             return 'unknown unit'
         if zone:
             try:
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -9,7 +9,7 @@ Configuration parameters:
         '[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]')*
     mem_unit: the unit of memory to use in report, case insensitive.
         ['dynamic', 'KiB', 'MiB', 'GiB'] (default 'GiB')
-    temp_unit: unit used for measuring the temperature
+    temp_unit: unit used for measuring the temperature ('C', 'F' or 'K')
         (default '°C')
     thresholds: thresholds to use for color changes
         (default [(0, "good"), (40, "degraded"), (75, "bad")])
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -7,12 +7,14 @@ Configuration parameters:
     format: output format string
         *(default '[\?color=cpu CPU: {cpu_usage}%], '
         '[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]')*
+    format_mem: format for the flat values
+        (default "{value:.2f}")
+    format_percent: format for the values in percent
+        (default "{value:.2f}")
+    format_temp: format for the temperature value
+        (default "{value:.2f}{unit}")
     mem_unit: the unit of memory to use in report, case insensitive.
         ['dynamic', 'KiB', 'MiB', 'GiB'] (default 'GiB')
-    padding: length of space padding to use on the left
-        (default 0)
-    precision: precision of values
-        (default 2)
     thresholds: thresholds to use for color changes
         (default [(0, "good"), (40, "degraded"), (75, "bad")])
     zone: thermal zone to use. If None try to guess CPU temperature
@@ -25,6 +27,16 @@ Format placeholders:
     {mem_unit} unit for memory
     {mem_used} used memory
     {mem_used_percent} used memory percentage
+
+format_mem placeholders:
+    {value} numeric value
+
+format_percent placeholders:
+    {value} numeric value
+
+format_temp placeholders:
+    {unit} temperature unit
+    {value} numeric value
 
 Color thresholds:
     cpu: change color based on the value of cpu_usage
@@ -159,9 +171,10 @@ class Py3status:
     cache_timeout = 10
     format = "[\?color=cpu CPU: {cpu_usage}%], " \
         "[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]"
+    format_mem = "{value:.2f}"
+    format_percent = "{value:.2f}"
+    format_temp = "{value:.2f}{unit}"
     mem_unit = 'GiB'
-    padding = 0
-    precision = 2
     thresholds = [(0, "good"), (40, "degraded"), (75, "bad")]
     zone = None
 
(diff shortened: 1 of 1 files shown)


tobes:
> Does that mean that if someone uses any of the deprecated parameters, all the new parameters will be ignored and set from the deprecate_function in Meta? 

No if users have set a parameter then it is left alone.  Except if a format contains `{}` and needs updating to `{something}` as that is a legacy issue that needs fixing.  So we respect what the user has set.

And the module only ever sees the corrected config.  Hope that helps.

guiniol:
It's really great. I was afraid if everything was in the same function, it'd all get overwritten.

ultrabug:
plz rebase mate

py3status/modules/sysdata.py
```diff
@@ -138,55 +144,90 @@ class Py3status:
     cache_timeout = 10
     format = "[\?color=max_cpu_mem CPU: {cpu_usage}%, " \
         "Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]"
-    padding = 0
-    precision = 2
-    thresholds = None
+    format_flat = "[\?min_length=0 {value:.2f}]"
```
tobes: Can't these just be `{value:.2f}`?

I'm not sure about the name `format_flat` it is not very expressive from what I can tell

guiniol: > Can't these just be `{value:.2f}`?

Yes. As they should.

> I'm not sure about the name `format_flat` it is not very expressive from what I can tell

That's in contrast to `format_percent`. Would `format_value` be better (the others being `format_percent` and `format_temp`)?

tobes:
@guiniol you'll need to rebase again due to your other PR being merged

guiniol:
Done, but wasn't this PR supposed to wait for #520?

tobes:
> Done, but wasn't this PR supposed to wait for #520?

Well both would be good and I don't think the order makes too much difference.  Anyway it'll make things easier either way.

Thanks for the quick response

guiniol:
> Well both would be good and I don't think the order makes too much difference. Anyway it'll make things easier either way.

Ok.

> Thanks for the quick response

I'm just trying to get all my PRs in so that I don't need custom modules ^^

py3status/modules/sysdata.py
```diff
@@ -132,21 +144,31 @@ class Py3status:
...
 
     class Meta:
 
         def deprecate_function(config):
             # support old thresholds
+            padding = config.get('padding', 0)
+            precision = config.get('precision', 2)
+            format_vals = '[\?min_length={padding} {{value:.{precision}f}}'.format(
+                    padding=padding, precision=precision)
+            format_temp = '[\?min_length={padding} {{value:.{precision}f}}{{unit}}'.format(
```
tobes: This is missing a closing `]`.  You want `'[\?min_length={padding} {{value:.{precision}f}}{{unit}}]'` same with `format_vals`.

guiniol: Ah, yes. Nice catch.

py3status/modules/sysdata.py
```diff
@@ -7,10 +7,12 @@
...
         '[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]')*
-    padding: length of space padding to use on the left
-        (default 0)
-    precision: precision of values
-        (default 2)
+    format_percent: format for the values in percent
+        (default "{value:.2f}")
+    format_temp: format for the temperature value
+        (default "{value:.2f}{unit}")
+    format_value: format for the flat values
```
tobes: I think it might be good to call this `format_mem` because with #520 we will really want a memory specific one with `{units}` whereas the loads in #552 might want there own format

As an aside `{mem_free}` might be nice at some point too.

guiniol: Sure. I'll wait for #520 to be in and change it to `format_mem` and move `{mem_unit}` from `format` to `format_mem`.
@pferate you'd be ok with that?

ultrabug:
plz rebase mate

guiniol:
@ultrabug done, but don't merge it quite yet.
@tobes I renamed `format_value` to `format_mem` but I can't move `{mem_unit}` from `format` to `format_mem` until #520. Should we merge now, and move `{mem_unit}` later?

guiniol:
On top of that, I want to see if I can completely remove the `format_*` using #564

guiniol:
@tobes I removed all the `format_*` in the last commit. What do you think?

guiniol:
I'm not a big fan of the result. The `format` string becomes very long.

Also, I `temp_unit` shouldn't be configurable and I have to rewrite the deprecate function.

tobes:
> I'm not a big fan of the result. The format string becomes very long.

Ah but you forget that we can still cheat.  We shouldn't need to add :.2f to the default format because we can do that via the deprecation stuff so we set them to :.2f if not specified either via precision etc or the user explicitly setting them.  So yeah it seems more complicated but actually it is hidden from the user.

> Also, I temp_unit shouldn't be configurable and I have to rewrite the deprecate function.

yeah this is a bit of thought/work.

guiniol:
> Ah but you forget that we can still cheat. We shouldn't need to add :.2f to the default format because we can do that via the deprecation stuff so we set them to :.2f if not specified either via precision etc or the user explicitly setting them. So yeah it seems more complicated but actually it is hidden from the user.

But if we do that, someone who wants to customise the output a bit, and writes a custom `format`, they won't understand why the precision of all numerical values changed.

tobes:
> But if we do that, someone who wants to customise the output a bit, and writes a custom format, they won't understand why the precision of all numerical values changed.

But the beauty is this never happens.  Once we have #565 all the troubles disappear.  Essentially we make `:0.2f` the default unless the user has set it differently either explicitly in the format eg `{cpu_usage:d}` or via the deprecated precision/padding.

guiniol:
> But the beauty is this never happens. Once we have #565 all the troubles disappear. Essentially we make :0.2f the default unless the user has set it differently either explicitly in the format eg {cpu_usage:d} or via the deprecated precision/padding.

In that case. Yes. That's definitely the way. It's the `update_placeholder_format`?

tobes:
> It's the update_placeholder_format?

That's the one, it was inspired by a comment you made in one of these issues

guiniol:
Perfect. I'll update sysdata as soon as I install the `deprecation-expansion` branch.

guiniol:
@tobes I tried, but I'm doing something wrong because it's not working. Also, I'm not sure how to handle padding in this case. I could add it in `update_placeholder_format` but that wouldn't use `[\?min_length`

tobes:
Hmmm, ok so this is working perfectly if the user has a defined format in their config.  But not for the default.  So the options would seem to be.

* add formats to the default (this starts looking a bit messy and isn't user friendly)

* allow updating of the default format.  If we do this it can either be done explicitly by a module in the post_init_hook() or we can do it at the same time as the deprecation.  I'm not sure which would be a better approach.

This patch will fix the issue by updating the default format if none is user supplied for testing etc. but I think I'm leaning towards the more explicit updating of the default
```
index 107cac6..bb77506 100644
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -408,7 +408,10 @@ class Module(Thread):
                             placeholder_formats.update(item['function'](mod_config))
                         format_strings = item['format_strings']
                         for format_param in format_strings:
-                            format_string = mod_config.get(format_param)
+                            format_string = mod_config.get(
+                                format_param,
+                                getattr(class_inst, format_param, None)
+                            )
                             if not format_string:
                                 continue
                             format = Formatter().update_placeholder_formats(
```

with the padding we can just add it to the format like before
``` ':{padding}.{precision}f'.format(padding=padding, precision=precision)```

**Comment by guiniol** [comment]
Would it be possible for the `update_format_modifier` to be executed as a pre-postconfig_hook function? That would be the cleanest I think, but I didn't look at how it works behind the scenes ^^

**Comment by tobes** [comment]
> Would it be possible for the update_format_modifier to be executed as a pre-postconfig_hook function?

That's sort of what that patch does. The thing is it makes sense in this case as we are doing the deprecation stuff, but for some modules I think this would be helpful but doing as part of deprecation wouldn't make sense as it is not really deprecation based more just a helper.  So maybe we need both routes to be available.

I suppose one question is do we always want to do this (also update the default) during deprecation if it is happening or is this something that may not be wanted. Really a few more use-cases would help with my thinking here.  Anyhow the main thing is that we can easily do this the difficulty is about deciding how and when we do it.

Anyhow I think it would be nice to get all the sysdata changes into 3.3 if we can

**Comment by guiniol** [comment]
> I suppose one question is do we always want to do this (also update the default) during deprecation if it is happening or is this something that may not be wanted. Really a few more use-cases would help with my thinking here. Anyhow the main thing is that we can easily do this the difficulty is about deciding how and when we do it.

If we look at the sysdata case, I think the `update_placeholder_format` could be in the Meta class but in `updated` instead of `deprecated`. `updated` would be run as a pre-post-config. In our case, that means in `deprecated` we can replace the old `*_threshold` with the new `threshold` and set the `precision` and `padding`. Then, in `updated`, we can set the default `precision` and `padding`. That way, there are three cases:
1. The user used the old `padding`/`precision`. They are set during the deprecation stage.
2. The user used the placeholders but didn't specify padding or precision. We set it to its default.
3. The user used placeholders with padding and precision. We don't do anything.

> Anyhow I think it would be nice to get all the sysdata changes into 3.3 if we can

I'm not against it, but I don't have a lot of time this week, and even less next week...

**Comment by tobes** [comment]
> What do you think?

I'm thinking this might be a good way to do it.

It separates the deprecation from the update which covers my main concern.  I think I'd do it just slightly differently.

Do the deprecation as we do now so there is still a `update_placeholder_format` in `Meta.deprecated` this would do the precision/padding stuff as before, remember we then remove these from the format so we don't have them later on (but only set the format if padding or precision are actually user provided).  Also they are deprecated so it is where they should be dealt with.  Then we add your new section `Meta.update_config` which also has `update_placeholder_format` which is where we set the defaults.  So this way formats can be simple eg `{value}` and the module can then set reasonable defaults if it needs to eg `{value:.2f}`

would that work?

@ultrabug Do you have any views on this?

The basic situation is that in a module like sysdata we have placeholders eg `{cpu_usage}` this ends up being a value something like `20.542317173377157` which is obviously stupid as the value to use but gives the user the ability to have as much precision as they want.  A module writer may decide that they want this displayed as `20.54` so `{cpu_usage:.2f}` would do this. but having a default format containing that just looks long/silly and the user setting a custom format just wants to do `format = 'CPU: {cpu_usage}%'` and get expected results ie not the full precision.  If they don't like the default formatting of the number they could still do  `format = 'CPU: {cpu_usage:d}%'` etc.

So the proposed solution would be to have such a module do 
```
class Py3status:

    class Meta:

        update_config = {
            'update_placeholder_format': [
                {
                    'placeholder_formats': {   # dict of placeholder:format
                        'cpu_usage': ':.2f',
                    },
                    'format_strings': ['format'],  # config settings to update
                }
            ],
        }
```
This way it is not part of the deprecation stuff (which feels wrong) but can be done before the module becomes 'live'.  It is also self contained and out of the way.  To be fair I think this is a rare thing that will be done in modules but will still have several users. This is my preferred option.

The other option is to make it much more explicit but that would probably end up being a different interface than just sharing the deprecation one.  eg

```
Py3status:

    def post_config_hook(self):
        self.py3.update_placeholder_format(...)
```

**Comment by guiniol** [comment]
> Do the deprecation as we do now so there is still a `update_placeholder_format` in `Meta.deprecated` this would do the precision/padding stuff as before, remember we then remove these from the format so we don't have them later on (but only set the format if padding or precision are actually user provided). Also they are deprecated so it is where they should be dealt with. Then we add your new section `Meta.update_config` which also has `update_placeholder_format` which is where we set the defaults. So this way formats can be simple eg `{value}` and the module can then set reasonable defaults if it needs to eg `{value:.2f}`

That makes a lot of sense. I'm all for it. It even makes sense that they have the same name, since they do the same thing, just not at the same time.

**Comment by tobes** [comment]
#567 adds this functionality.  We'll see if it is accepted

**Comment by guiniol** [comment]
Great. I'll adjust this PR.

**Comment by tobes** [comment]
> We'll see if it is accepted

It's now in master :)

**Comment by guiniol** [comment]
> It's now in master :)

Great. Now I only have to do my work ^^

**Comment by guiniol** [comment]
@tobes I finally have a little time to work on this. I did the `update_placeholder_format` according to what I understood from the doc, and it seems to work. However, the default `format` is being used instead of the one I supplied.

**Comment by guiniol** [comment]
@tobes the tox fail is due to py2 handling utf-8:
```
`temp_unit` (default '\xc2\xb0C') does not match docstring u'\xb0C'
```

I'm pretty sure you told me before how to fix this, but I can't remember and can't find it either...

**Comment by tobes** [comment]
`temp_unit = u'°C'`

**Comment by guiniol** [comment]
Thanks.
Did you see my comment about it using the default `format` instead of the user supplied one? What am I doing wrong?

**Comment by tobes** [comment]
yeah,  I'll take a look when I get some time but the unicode thing didn't need any thinking.

Hope you're enjoying xmas

**Comment by guiniol** [comment]
> yeah, I'll take a look when I get some time but the unicode thing didn't need any thinking.

Great. Didn't want it to be lost in the noise.

> Hope you're enjoying xmas

Sure am. Finally some time to code ^^

Have some nice holidays too!

**Comment by tobes** [comment]
Ok, there is a bug in the update code something like the following patch fixes it.

```
index e527bf5..013aea1 100644
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -471,7 +471,9 @@ class Module(Thread):
                         placeholder_formats = item.get('placeholder_formats', {})
                         format_strings = item['format_strings']
                         for format_param in format_strings:
-                            format_string = getattr(class_inst, format_param, None)
+                            format_string = mod_config.get(format_param)
+                            if not format_string:
+                                format_string = getattr(class_inst, format_param, None)
                             if not format_string:
                                 continue
                             format = Formatter().update_placeholder_formats(
```

**Comment by guiniol** [comment]
> Ok, there is a bug in the update code something like the following patch fixes it.

@tobes I can confirm this fixes it for me.

File: py3status/modules/sysdata.py [hunk 4]
```diff
@@ -9,10 +9,8 @@
         '[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]')*
     mem_unit: the unit of memory to use in report, case insensitive.
         ['dynamic', 'KiB', 'MiB', 'GiB'] (default 'GiB')
-    padding: length of space padding to use on the left
-        (default 0)
-    precision: precision of values
-        (default 2)
+    temp_unit: unit used for measuring the temperature
+        (default '°C')
```
tobes: I don't like this setting.  All it is doing is creating a placeholder which feels like adding no value.

I think it would be better to have temp_unit actually be `C` or `F` for any crazy people who want Fahrenheit (`sensors -f`).  The placeholder would then become `°C` or `°F`.

guiniol: It adds the option to show the unit.
I'll add the code to change the unit based on `temp_unit`.

guiniol: @tobes I added the option for Fahrenheit and Kelvin
`C` or `°C` gives Celsius
`F` or `°F` gives Fahrenheit
`K` gives Kelvin (Celsius + 273.15)

tobes: > I added the option for Fahrenheit and Kelvin

you seem to have added it to #552 not here

guiniol: hum... fixed ^^

py3status/modules/sysdata.py
```diff
@@ -126,25 +126,31 @@ def memory(self, unit='GiB'):
...
         """
         Tries to determine CPU temperature using the 'sensors' command.
         Searches for the CPU temperature by looking for a value prefixed
         by either "CPU Temp" or "Core 0" - does not look for or average
         out temperatures of all codes if more than one.
         """
 
-        sensors = None
+        command = ['sensors']
+        if unit.lower() == 'F' or unit == u'°F':
```
tobes: This is broken. 
```
unit = unit.lower()
if unit in [...]:
    ...
```

would be better.

Personally I'd just have unit as 'C', 'F', or 'K' without the degree but maybe that's just me

guiniol: trying to match `string.lower()` and `F`... sorry

py3status/modules/sysdata.py
```diff
@@ -127,25 +126,31 @@ def memory(self, unit='GiB'):
...
+                sensors = self.py3.command_output(command + [zone])
             except:
                 sensors = None
         if not sensors:
-            sensors = self.py3.command_output('sensors')
+            sensors = self.py3.command_output(command)
         m = re.search("(Core 0|CPU Temp).+\+(.+).+\(.+", sensors)
         if m:
             cpu_temp = float(m.groups()[1].strip()[:-2])
+            if unit == 'K':
```
tobes: just do the `.upper()` once and then you don't get bugs like this one

py3status/modules/sysdata.py
```diff
@@ -190,44 +207,68 @@ def deprecate_function(config):
...
+                        },
+                    ],
                 }
 
     def post_config_hook(self):
         self.data = GetData(self)
         self.cpu_total = 0
         self.cpu_idle = 0
-        self.values = {}
+        self.values = {'temp_unit': self.temp_unit}
```
tobes: This is incorrect imho.

If I have a `temp_unit = 'c'`  then `{temp_unit}` should be `°C` etc

py3status/modules/sysdata.py
guiniol: Yep. Done.

py3status/modules/sysdata.py
guiniol: My idea was that you could write `c` if you wanted it displayed as `c` or `°C` if you wanted it as `°C`.

py3status/modules/sysdata.py
```diff
@@ -127,25 +126,32 @@ def memory(self, unit='GiB'):
...
         Tries to determine CPU temperature using the 'sensors' command.
         Searches for the CPU temperature by looking for a value prefixed
         by either "CPU Temp" or "Core 0" - does not look for or average
         out temperatures of all codes if more than one.
         """
 
-        sensors = None
+        command = ['sensors']
+        unit = unit.upper()
+        if unit in ['F', u'°F']:
```
tobes: this can now just be `if unit == u'°F':`

guiniol: yes indeed. done

py3status/modules/sysdata.py
tobes: This should probably have the options available.

ultrabug:
How are we here plz ? I'm getting there after the battle so I'm not sure what to do

tobes:
I think we are close but not quite.

guiniol:
I rebased and added the options available for `temp_unit`. I think that was it, no?

tobes:
this looks like it is good except #602 introduced a bug that is fixed in #662 so when that is in I'll do a final review.  Sorry for the delays

guiniol:
No worries. Just ping me once you're ready and I'll update the branch.

tobes:
Cool I think we are finally done.  I've tested this with master merged in and it looks good to me

guiniol:
Glad to hear that ^^

PR #558, opened 2016-11-06
Better default colors for sysdata

guiniol:
As discussed in #553, this uses better defaults for sysdata by separating the colors for cpu and memory.


guiniol:
Well, the option to color part of the output really made this possible.

ultrabug:
thanks ;)

PR #560, opened 2016-11-06
Log py3status/python versions and git info

tobes:
When looking in py3status logs it can be hard to tell what version of py3status was running.  this PR adds a few extra lines to the logging.  The git lines are only shown if running in a git repo
```
2016-11-06 20:45:06 INFO ========
2016-11-06 20:45:06 INFO Starting py3status version 3.3rc0 python 3.5.2
2016-11-06 20:45:06 INFO git: branch log-extra-info
2016-11-06 20:45:06 INFO git: dcf3661 Merge pull request #556 from ultrabug/modfix-3.1-vnstat
```

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -8,6 +8,7 @@ import time
 
 from collections import deque
 from json import dumps
+from platform import python_version
 from signal import signal, SIGTERM, SIGUSR1, SIGTSTP, SIGCONT
 from subprocess import Popen
 from threading import Event
@@ -171,7 +172,6 @@ class Py3statusWrapper():
 
         # only asked for version
         if options.version:
-            from platform import python_version
             print('py3status version {} (python {})'.format(config['version'],
                                                             python_version()))
             sys.exit(0)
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks for this, I think it's a great idea indeed.

The only thing about the git detection is that I wouldn't use the git command itself but instead use basic python file reading to check for .git/HEAD and .git/FETCH_HEAD files to get the information you want.

This will avoid all this parsing while being more efficient.

jakubjedelsky:
+1 for not using git command or external commands at all. From point of view of package maintainer of fedora rpm it's useless to add unnecessarily dependencies to the package. Anyway I like the idea.

tobes:
> use basic python file reading 

Ok done now I had to dig a bit more into the logs for the info but we are just doing two file reads now

>  it's useless to add unnecessarily dependencies to the package

Well `git` was not actually needed as a dependency here but I agree with keeping the requirements to an absolute minimum especially as this is for i3wm which is trying to be lightweight.

ultrabug:
thanks !

PR #561, opened 2016-11-06
Py3: _report_exception() internal helper

tobes:
Currently exceptions in Py3 may not always get reported for example an error in a format.

This PR adds the ability for Py3 methods to log exceptions (where they are catching them) eg a bad format may now produce log extract like

```
2016-11-06 21:00:26 WARNING Invalid format `[[{output}]~{button}~` (Exception) frame.py line 126.
2016-11-06 21:00:26 INFO Traceback
Exception: Block not closed
  File "/home/toby/dev/py3status/py3status/modules/frame.py", line 126, in frame
    output = self.py3.build_composite(self.format, composites=composites)
  File "/home/toby/dev/py3status/py3status/py3.py", line 564, in build_composite
    attr_getter=attr_getter,
  File "/home/toby/dev/py3status/py3status/formatter.py", line 318, in format
    raise Exception('Block not closed')
```

We have to do a bit of magic to make it show the error as being caused by the module in this case frame.  The module output remains `invalid format` as before and the user receives no notification.

The logs are limited so as not to be spammed which a bad format could do constantly.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -12,7 +12,7 @@ from signal import signal, SIGTERM, SIGUSR1, SIGTSTP, SIGCONT
 from subprocess import Popen
 from threading import Event
 from syslog import syslog, LOG_ERR, LOG_INFO, LOG_WARNING
-from traceback import extract_tb, format_tb
+from traceback import extract_tb, format_tb, format_stack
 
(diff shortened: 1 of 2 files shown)


ultrabug:
this is nice indeed

PR #563, opened 2016-11-07
sysdata: bugfix for python2 cpu_usage

tobes:
Sysdata module cpu_usage was not working

* python2 broken due to unicode issues.  Fix this by using `Py3.command_output()`

* When a zone was given the output broke.  Fix by running output through the regex.  If the zone is bad we just use the default.  We can change this behaviour but it seems good to me maybe we should return '?'

* When we cannot get a temp we return '?' and allow this to propagate.

@guiniol it would be good if you could review this

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -38,13 +38,15 @@ be available, provided by the `lm-sensors` or `lm_sensors` package.
 from __future__ import division
 
 import re
-import subprocess
 
 
 class GetData:
     """
     Get system status
     """
+    def __init__(self, parent):
+        self.py3 = parent.py3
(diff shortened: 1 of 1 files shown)


guiniol:
@tobes Sorry for the regression. I never think to test with python2.
This looks good to me. I didn't know about (or had forgotten) `py3.command_output`.

tobes:
> I didn't know about (or had forgotten) py3.command_output.

It's new, I keep adding stuff ;p

guiniol:
> It's new, I keep adding stuff ;p

Keep up the good work ^^

ultrabug:
Thanks :)

PR #564, opened 2016-11-07
Formatter: Allow formatting numeric strings

tobes:
So @guiniol has been doing things with sysdata and I was doing some testing of #555 and I was finding it a little frustrating.  If I wanted to do custom formatting of say `{cpu_usage}` then the formatting was done in `format_percent` which I understand from the developer perspective but from the users perspective this is confusing in my mind.  I should be able to do `format = '{cpu_usage:.0f}'` and it should work.  But as `{cpu_usage}` is a string even though it looks like a number this fails.

So anyhow I've made this so it now works.  If we have a value that could be numeric and is having numeric formatting applied it will now work.  It also means we can do `{cpu_usage:.0f}` on a string.  we also can do `{cpu_usage:d}` on a float/string as a quick 'make this an int' shortcut.

These changes also highlighted an issue around checking for placeholders in a format string.  modules commonly do
```
if '{cpu_usage}' in self.format:
    ...
```
but this fails if formats are used, so I've added a helper so modules can now do
```
if self.py3.format_contains(self.format, 'cpu_usage'):
    ....
```
It is more verbose but it does catch formatted placeholders and it also allows wildcards like `'mem_*'` to be queried.

There are also a few formatter tests for the new behaviours.

sysdata has been updated to use the new functionality.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -969,6 +969,23 @@ The following functions can be registered
 >
 > Note: This function should only be used by containers.
 
+__format_contains(format_string, name)__
+
+Determines if `format_string` contains placeholder `name`
+
+`name` is tested against placeholders using fnmatch so the following
+patterns can be used:
+
+    * 	    matches everything
+    ? 	    matches any single character
+    [seq] 	matches any character in seq
+    [!seq] 	matches any character not in seq
+
+This is useful because a simple test like
+`'{placeholder}' in format_string`
+will fail if the format string contains placeholder formatting
+eg `'{placeholder:.2f}'`
+
 __safe_format(format_string, param_dict=None, force_composite=False,
 attr_getter=None)__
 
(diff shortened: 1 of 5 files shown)


guiniol:
That looks really great. I'll have to fix all the `if '{cpu_usage}' in self.format:` I put in modules ^^

tobes:
> That looks really great. 

Thanks

> It's also better to be able to remove all the format_* and only keep format. I'm all for it.

Do you think we could do that?  I've not come up with a better solution yet but maybe you have ideas.  I'll think more too.

guiniol:
> Do you think we could do that? I've not come up with a better solution yet but maybe you have ideas. I'll think more too.

We could just add the things like `{unit}` in format (and rename it to `{temp_unit}` or `{mem_unit}`) like in #520. With this, I think we're covered.

ultrabug:
That formatter stuff is becoming pretty serious, I must admit I lost touch with it few PRs ago

PR #565, opened 2016-11-08
Deprecation expansion

tobes:
This PR is on top of #564 so the second commit is the stuff here.

It allows us to use the deprecation logic to rename placeholders (I'm sure this will come in handy)

It also allows us to add formats to placeholders in format strings if the user has not explicitly set them.  I think @guiniol might find these helpful with the precision stuff in sysdata.


ultrabug:
plz rebase

tobes:
> plz rebase

done

ultrabug:
thank you !

PR #566, opened 2016-11-09
Flake8 fixes and testing improvements

tobes:
This switches testing from flake8 to pytest-flake8 (It seems to be more thorough in its testing)

There are a bunch of whitespace fixes so that the tests pass.

sysdata still need fixes but I was going to wait till the pending changes are merged so currently ignore the current errors.


ultrabug:
:+1: awesome

PR #567, opened 2016-11-13
Allow updating of placeholder formatting of modules

tobes:
In a module like sysdata we have placeholders eg `{cpu_usage}` this ends up
having a value something like `20.542317173377157` which is strange as the
value to use but gives the user the ability to have as much precision as they
want. A module writer may decide that they want this displayed as `20.54` so
`{cpu_usage:.2f}` would do this. Having a default format containing that
just looks long/silly and the user setting a custom format just wants to do
`format = 'CPU: {cpu_usage}%'` and get expected results ie not the full
precision. If they don't like the default formatting of the number they could
still do format = 'CPU: {cpu_usage:d}%' etc.

So using this allows sensible defaults formatting and allows simple
placeholders for user configurations.

There are also a few minor cosmetic documentation fixes that i noticed whilst doing this PR.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -29,6 +29,7 @@ py3status documentation
 * [Composites](#composites)
 * [Module documentation](#docstring)
 * [Deprecation of configuration parameters](#deprecation)
+* [Updating configuration parameters](#update_config)
 * [Module testing](#testing)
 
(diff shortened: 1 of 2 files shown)


ultrabug:
I see thanks, it's good

PR #568, opened 2016-11-13
Disable modules if post-config_hook() raises Exception

tobes:
Some modules eg volume_status raise exceptions in post_config_hook() when things are bad.  Currently we do not handle this very well.  This PR causes such exceptions to disable the module.


ultrabug:
It's clean and will be a good start for the day we want interactive enable/disable of modules

tobes:
> the day we want interactive enable/disable of modules

you saw through my plan

PR #571, opened 2016-11-17
RFC: New feature popups

tobes:
Introducing py3status popups.

What are these I hear you ask.  When the user clicks on a module we can now display a popup menu.
The popups are drawn using dzen2.

example:

```
my_module {
    on_click 1 = "popup hello"  # display some text fairly pointless
}

my_module {
    on_click 1 = "popup Firefox:exec firefox"   # launcher
}

my_module {
    on_click 1 = "popup Firefox:exec firefox\nChrome:exec google-chrome"  # 2 item launcher
}

my_module {
    on_click 1 = "popup Firefox:exec firefox\n"
    on_click 1 += "Chrome:exec google-chrome"  # 2 item launcher just written different
}

my_module {
    on_click 1 = "popup `cal`"  # run a command and output to the popup
}
```

There is also some functionality to allow modules to use popups eg in their on_click()

I need to do a few tidy ups and also write documentation etc but I though it would be good to post this here so people have a chance to make suggestions on how we can improve it etc.

~~I'll also improve how we choose the menu colors as currently hardcoded but we can get these from i3config or our config.~~

Feedback welcome as ever.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -27,6 +27,7 @@ py3status documentation
 * [Example 5: Using color constants](#example_5)
 * [Module methods](#module_methods)
 * [Py3 module helper](#py3)
+* [Py3 popups](#py3_popup)
 * [Composites](#composites)
 * [Module documentation](#docstring)
 * [Deprecation of configuration parameters](#deprecation)
@@ -1160,6 +1161,58 @@ general one that will be used.
 
 ***
 
+
+## <a name="py3_popup"></a>Popups
+
+the Py3 helper contains popup functionality allowing. modules to create popups
+these can be of two types.
+
+`info` these are just plain text.
+
+`menu` these allow the user to click on an item in the popup to trigger an
+action.  It also allows items to be selected or not and the user is able to
+change this selection.
+
+The data for the popup can be one of the following
+
+`string` the data contains the text to display `\n` is used to separate lines.
+
+`'hello\nhow are you?'` would create a popup with two lines.
+
+`list` each line is a separate item in the list.
+
+`['hello', 'how are you?']` would create a popup with two lines.
+
+As well as strings in the list tuples can be provided that contains the the
+text and a value as to if the item is selected.
+
+`['Alarm', ('enabled', true)]` would create a popup an option showing if the
+alarm is enabled or not.
+
+when the user clicks on the popup a callback function will be called if it was
+provided. The function will receive the text of the line that was clicked by
+the user and a dict containing all items as keys and a value corresponding to
+if the item is selected or not.
+
+The following helper functions are defined in the modules `self.py3`.
+
+__popup_open(data, type='info', callback=None)__
+
+Open a popup for the module.
+
+__popup_close()__
+
+Close the modules popup.
+
+__popup_toggle(data, type='info', callback=None)__
+
+Close the menu if it is open else open it.
+
+__popup_update(data)__
+
+Update the popup displaying the new data.
+
+***
 
 ## <a name="composites"></a>Composites
 
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -21,6 +21,7 @@ from py3status.events import Events
 from py3status.helpers import print_line, print_stderr
 from py3status.i3status import I3status
 from py3status.parse_config import process_config
+from py3status.popup import PopupController
 from py3status.module import Module
 from py3status.profiling import profile
 from py3status.version import version
@@ -352,6 +353,9 @@ class Py3statusWrapper():
         if self.config['debug']:
             self.log('i3status thread {} with config {}'.format(
                 i3s_mode, self.config['py3_config']))
+
+        # popup_controller
+        self.popup_controller = PopupController(self)
 
         # setup input events thread
         self.events_thread = Events(self)
--- a/py3status/events.py
+++ b/py3status/events.py
@@ -1,8 +1,10 @@
+import re
 import select
+import shlex
 import sys
 
 from threading import Thread
-from subprocess import Popen, PIPE
+from subprocess import Popen, PIPE, check_output
 from json import loads
 
 from py3status.profiling import profile
@@ -107,8 +109,10 @@ class Events(Thread):
             return
         elif command == 'refresh_all':
             self.py3_wrapper.refresh_modules()
+            return
         elif command == 'refresh':
             self.py3_wrapper.refresh_modules(module_name)
+            return
         else:
             # In commands we are able to use substitutions for the text output
             # of a module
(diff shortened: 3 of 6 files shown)


ultrabug:
I love the idea thanks a lot !

Few remarks since I wanted to try and failed

- if dzen2 is not present and user asked for popups, we should notify him once on start and then disable popups altogether
- we miss some debugging logs on the dzen command that is ran, when py3status starts with -d, we should log the dzen2 cmd
- dzen2 does not like DejaVu like fonts ? says it's unkown to him :(
- the hello example displayed nothing (ran it from command line to bypass the font problem), dunno if it's expected or I missed something ?

tobes:
I'll take a look at your issues tomorrow. The font is a pain as i3 uses a different format than dzen2 what is your i3 font line?

I'm sure we can fix it up. Just hard guessing setups in different distros etc. 

I'll check the examples in case I broke them

tobes:
>   if dzen2 is not present and user asked for popups, we should notify him once on start and then disable popups altogether

Done

>    we miss some debugging logs on the dzen command that is ran, when py3status starts with -d, we should log the dzen2 cmd

Done.

I don't make much use of the debug option due to the amount of noise from module output.  Maybe we can make the debug more streamlined eg choose which modules provide debug output etc.


>    dzen2 does not like DejaVu like fonts ? 

my config uses `font pango: DejaVu Sans Mono 14` and is happy.  does `fc-list : family` run for you?  Odd you should get output even with bad fonts.

> says it's unkown to him :(

where is he saying this?

>    the hello example displayed nothing (ran it from command line to bypass the font problem), dunno if it's expected or I missed something ?

working for me, maybe if you post the dzen2 command we are trying to run it will help me debug.
They sort of need to run via py3status as dzen2 gets its input from stdin and we do odd things so running from cli as we do things is not so simple.

ultrabug:
> I'll take a look at your issues tomorrow. The font is a pain as i3 uses a different format than dzen2 what is your i3 font line?

font pango:DejaVu Sans Mono 13px

TBH, I'd choose a functional font for everyone by default instead of trying to guess it and fail (or provide a general section option to set it up)

> where is he saying this?

When I run it from command line (please note the trailing whitespace, yes I tried to remove it etc)
```bash
dzen2 -x 2480 -y 21 -w 80  -p -ta l -sa l -l 1 -bg "#002b36" -fg "#93a1a1" -fn 'DejaVu Sans Mono-12 ' -h 21 -e "button3=ungrabkeys,exit;button4=scrollup;button5=scrolldown;key_Escape=ungrabkeys,exit;button1=exit:99;onstart=raise;"

dzen: error, cannot load font: 'DejaVu Sans Mono-12 '
```

> working for me

I guess this comes from my multi screen setup, this is something that's not taken into account yet in this RFC

tobes:
> TBH, I'd choose a functional font for everyone by default instead of trying to guess it and fail (or provide a general section option to set it up)

But it is so nice when it uses the same font as the bar.
In your case we have 2 factors in play

1) you have the `px` at the end I wasn't aware this was possible - should be easy to fix  

2)  your dzen2 is likely compiled without pango support.

try using an font like  `font -misc-fixed-medium-r-normal--13-120-75-75-c-70-iso10646-1` in your config and things should work.

I'll make some changes to

a) improve font picker

b) allow setting of popup_font in the i3status.conf

c) catch the error with the bad font and let it just use its default

The trailing whitespace doesn't cause problems for me but i'll remove it.  Thanks for testing.

> I guess this comes from my multi screen setup

This shouldn't be a problem here but I might need to do some stuff around multi-screen setups.

tobes:
@ultrabug 

If you get a chance could you paste your output of `dzen2 -v`

ultrabug:
```bash
➜  ~ dzen2 -v
dzen-0.9.5-svn, (C)opyright 2007-2009 Robert Manea
Enabled optional features:  XINERAMA
```

tobes:
Thanks that looks like it gives the info I need. I'll have a play

tobes:
I updated this branch do you want to see if it now works for you at all.  I still need to do some more work but it'd be good to where it's now at

lasers:
>dzen: error, cannot load font: 'DejaVu Sans Mono-12 '

I do not have this problem. That long dzen2 command in terminal worked okay here. It may be the pango support after all.
```
lasers~ dzen2 -v                                                                                                                                                                                                                              
dzen-0.9.5-svn, (C)opyright 2007-2009 Robert Manea
Enabled optional features:  XFT
```
Other than _my minor hiccup_ where the popup need to go away first for it to exec something afterward. This RFC is pretty nifty.

tobes:
`XFT` is pango support

tobes:
@lasers 

> Other than my minor hiccup where the popup need to go away first for it to exec something afterward. This RFC is pretty nifty.

This is because the popup has an action when you click on it.  Maybe the documentation needs to be clearer.  Or is it that you want different behaviour?  Would a time limited popup (displays for x seconds) have any uses?

There is a separate issue that currently we only allow a single command.  We could extend the events to allow multiple commands eg `on_click 1 ['command 1', 'command 2']` or something similar.  I'm not sure how much demand exists but it would be easy to implement if we did the list method, doing as a string eg `on_click 1 'command 1; command 2'` is actually a much harder problem.

lasers:
@tobes I'm biased on certain behavior because I already got this in my config.
```
# customized dunst notifications
set $msg exec $nsi notify-send -t 3

# toggling layout / fullscreen / splits
bindsym $mod+Shift+f fullscreen global ; $msg "fullscreen global"
bindsym $mod+Shift+space floating toggle ; $msg "floating toggle"
bindsym $mod+w layout toggle all ; $msg "layout toggle all"
bindsym $mod+a [urgent="latest"] focus; focus parent ; $msg "focus"
bindsym $mod+f fullscreen ; $msg "fullscreen"
bindsym $mod+semicolon split h ; $msg "[ H ]"
bindsym $mod+space focus mode_toggle ; $msg "mode toggle"
bindsym $mod+t border toggle ; $msg "border toggle"
bindsym $mod+v split v ; $msg "[ V ]"
...
```
I use this to open applications too.
```
bindsym i exec firefox; $msg "Firefox"; $wq
bindsym Shift+n exec netflix; $msg "Netflix"; $wq
...
```


I don't know what people will do with this RFC, but it's possible that one might want it that way too. I was only following your examples (and under impression that it would do both things at once).
```
on_click 1 = "popup Firefox:exec firefox"   # launcher
```
>Would a time limited popup (displays for 3 seconds) have any uses?

Yes. It could... I just don't know what... Maybe w/ formatter. Get notification/popup when an a new mail arrives or something else.

I have questions about urgent + formatter for all modules (customization)... Will be on IRC.

tobes:
Seeing as this got no traction I'm going to close it

PR #574, opened 2016-11-18
fedora_updates redo security check when number of updates decreases

tobes:
Minor update.

Fixes issue when some packages are updated but some are still outstanding.  In this situation we now recheck if some updates are security related.


ultrabug:
Makes sense, thanks !

PR #575, opened 2016-11-18
spotify: deactivated playback code

Andrwe:
At least with the current version of spotify (1.0.26.125-1) the dbus
interface to get playback status does not respond.
As there is no alternative this PR disables the playback status request
and the colors to display current status.

Signed-off-by: Andrwe Lord Weber <github@andrwe.org>


tobes:
Hi,

What error where you getting from dbus?  Is this something that we can catch?

I am assuming that this used to work, and therefore will be fixed in a future spotify release so can we just ignore the logic if PlaybackStatus is unavailable.

Andrwe:
Hi,

Sorry I just found out that the package name changed an I was using an old version of Spotify.
With version 1.0.42.151 the dbus interface is working fine.

PR #577, opened 2016-11-19
added cmus module

raspbeguy:
It's a module to display current song in cmus. You can also play/pause by clicking.


tobes:
@raspbeguy Thanks for the PR.  This module looks fine apart from a couple of minor things, however py3status already has many different music display modules and there is an attempt to try to improve existing media modules functionality rather than adding more modules.

It looks like cmus has mpris capabilities (but seems like it's a compile option) in which case the `mpris` module might work for you, alternatively `player_control` could 'learn' about cmus, although I'm not sure if it would be so easy to adapt for all the functionality in this module.

raspbeguy:
Actually you're right ! I did a quick research few days ago and for some reason I concluded that cmus didn't have any support for mpris. I was a bit deceived, but well, I coded something. And now I have a useless module :)
I guess mpris will be included in next release of cmus.

raspbeguy:
Closing, useless module

PR #578, opened 2016-11-19
Use pretty printing in logs for data structures

tobes:
This PR makes dicts, lists, sets and tuples more readable in the logfile by formatting them using `pprint.pformat()`.  It does not change the syslog output if that is used.


ultrabug:
Great idea and simple implementation = 💃

PR #579, opened 2016-11-20
IMAP module : added support for STARTTLS

raspbeguy:
Now you have a parameter to define security method.


tobes:
@raspbeguy you might find it easier to make a new branch for each feature.  Having your master branch just track upstream makes things much easier imho.

Additionally starttls is only available in python >= 3.2 and as py3status aims to support version 2.6 it might be good to add that to the documentation.

I'm also not keen on the no security option I suppose this might be just about ok on a local network but I'd be inclined to note that it is a crazy thing to be doing.

One option to consider might be trying the ssl approach first and if that fails then try using starttls.  That way we wouldn't need the setting (if we have the insecure method then that would need to be explicitly set by the user). A context might be useful see https://docs.python.org/3/library/ssl.html#ssl-security

raspbeguy:
I don't really see how it is less secure than a standard mail client such as Thunderbird. I think that the settings are the same.

raspbeguy:
What's less secure in startssl anyway?

tobes:
> What's less secure in startssl anyway?

I never said it was.  Using plain text passwords on a network was my issue (the 'none' option)

raspbeguy:
So all you want is remove the `none` option ? Sure, I can do that. I added the `none` option because most of regular mail client propose it. In my opinion it is up to the user to use security option whenever he can, not to the softwar, but if you want me to remove that option, I will. Please just confirm that's what you actually want.

tobes:
> So all you want is remove the none option ?

Personally I'd remove the whole config option and have things just work, but yeah I don't like the none option (others may think it's great)

There were some other comments in my original response too.

raspbeguy:
> Personally I'd remove the whole config option and have things just work

No, things won't just work, else I wouldn't have done it.

> One option to consider might be trying the ssl approach first and if that fails then try using starttls.

Why do you want to implement mechanisms that doesn't exist in most of the mail clients?

tobes:
> No, things won't just work, else I wouldn't have done it.

Something like this?  Note __this code is terrible__ on many levels but illustrates that this can easily be done,  Real code would probably want to cache the effective method and not have raw exceptions etc

```
try:
    connection = imaplib.IMAP4_SSL(self.imap_server, self.port)
except:
    try:
        connection = imaplib.IMAP4(self.imap_server, self.port)
        connection.starttls()
    except:
        connection = None
```

> Why do you want to implement mechanisms that doesn't exist in most of the mail clients?

Because it can be done and I think config options can often be avoided. If you don't agree that's fine tell me why I'm wrong :)

raspbeguy:
I think the configuration option must be the closest possible to what a regular client would propose. Honestly, It would take me a lot of time to discover that the security protocol is automatically discovered, because every single client I used so far asked me this protocol.
Furthermore, in the future, if another protocol is added to globally accepted mail security standards, it may be possible that protocol detection would be impossible, so you would have to add this option and then break existing configurations.
To conclude, I understand that adding options would sometimes add unwanted complexity compared to what it brings in terms of features, but I really think that this option is necessary and would avoid a lot of confusion.
[Explicit is better than implicit.](https://www.python.org/dev/peps/pep-0020/#id3)

tobes:
ok that seems a good argument

raspbeguy:
Thanks.
So I'll remove the `none` option.

raspbeguy:
So, it is ready to merge now ?

tobes:
> So, it is ready to merge now ?

It'd be nice if it was a clean branch so the unrelated cmus commits were not here etc.  Anyhow we can work around that, but it will add delay.

Also I think it could be written in a more clear way, but I'm not too bothered here as we like to get contributions from new people.  Merging of PRs into master can take a while, also ultimately it's @ultrabug who will make that final call.

raspbeguy:
@tobes I was only asking to know if I did good or if I needed to work more on it. Sure I can wait, no problem.

tobes:
I still think a context would be useful see https://docs.python.org/3/library/ssl.html#ssl-security generally it is a very bad idea to ignore security recommendations.

I'd lose the switch stuff and just use an if/else The switch just makes the code less readable.

Also as I said making a clean PR (no unneeded commits etc) would be very helpful

raspbeguy:
So you want me to close this pull request and open a new one ?
I don't really know how to delete unwanted commits else.
Concerning the switch thing, I'll redo it if you really wants to, but I thought this was a common and standard way to do this kind of stuff.
I'll add SSL context, of course, this is my mistake.

tobes:
> Concerning the switch thing, I'll redo it if you really wants to, but I thought this was a common and standard way to do this kind of stuff.

Maybe if you have many options yes, but for two it is just noise

> So you want me to close this pull request and open a new one ?

That is probably the easiest way. I would really recommend keeping master branch  in sync with upstream and create new branches for development.

raspbeguy:
Okie Dokie

PR #583, opened 2016-11-22
Persistent storage for modules

tobes:
It is useful for modules to be able to maintain state across restarts.  This allows for richer functionality. As an example in this PR, the clock module is updated so that it will remember the format that it was displaying, rather than being reset when py3status is reset. Currently we have a few modules that do this eg rate_counter, but they each do their own thing and this provides a unified approach.

Data is stored in a sqlite database `py3status.sqlite` created in the same directory that the config file is in (maybe we want to add an option to explicitly set a path).  Data is stored for each instance of a module so modules do not conflict. Data is stored as key value pairs. Currently we save the data as json allowing bool, int, float, None datatypes as well as lists and dicts, but this can be expanded in future eg to allow storing of dates (but dates are hard).

The idea is to keep things simple eg if the key requested does not exist we return None rather than raising an exception.


ultrabug:
That's an awesome idea I also had in mind for quite some time thanks !
I also saw it as a sort of state cache between reloads.

Few remarks

- you're adding sqlite as a dependency of py3status for users using a storage enabled module
- I think we could achieve this without sqlite tbh, my initial idea was a creating a simple file using the standard pickle serializer

We could see this storage as a simple dict which keys are the module's name and then add/remove values underneath it

```python
{
'weather_yahoo paris': {'last_output': 'o x y'}
}
```

This storage would be read from disk on start and written on stop, all operations in between would happen in RAM to preserve performances (maybe have a write interval option to be safe)

Does that make sense ?

tobes:
> you're adding sqlite as a dependency of py3status

my understanding is that this is in the standard libs.

pydsigner:
sqlite3 is definitely in the stdlib.

ultrabug:
You're right guys I'm sorry, got mislead by the fact that I can choose to build python with sqlite support or not on Gentoo ;)

tobes:
> I can choose to build python with sqlite support or not on Gentoo ;)

I've updtaed this to work (ie not fail) when sqlite3 is not available.  I'd rather stick with sqlite than try to roll our own as it can do all the atomic stuff etc

ultrabug:
Okay, I gave more thought to this and I still don't like the fixed sqlite schema (yes, I'm one of those NoSQL schemaless fools)

I'd rather we use a simple pickle (de)seralized dict which will allow us a more versatile form of usage and improvement over time.

Would that work for you ?

tobes:
OK I'll have a look, but yeah you nosql fool ;)

tobes:
I've changed this to now write to a file not use sqlite.

I've chosen json over pickle as I prefer it.

The file format is fairly simple each line describes a key/value pair for the module, currently only json types (as with the sqlite version).  Lines can also clear a value.  So we just append changes to the file rather than rewriting it all the time.  When we read the file we just replay the changes.  When we read the file we 'vacuum' it so it only has current values (to keep it small).

I need to check it with python 2/3 unicode etc, add comments, make things more robust eg corrupt files and simultaneous writes from different module threads but the basic idea is here so it'd be good to know if you are happy with the general direction.

ultrabug:
I don't think it's optimal sorry and I don't understand your dislike of pickle which has the advantage of keeping those python objects intact (datetime !)

The cPickle dumps/loads implementation is fast enough for easy enough to implement imho.

I'd go for RAM only set/delete/get (I'm no fan of the clear name) operations on a single dict with the module name as the key. TBH we could almost basically mirror the module object we already have on this dict.

Then I'd spawn a new thread which would read it from disk to RAM on load and persist the object on disk every minute and on stop only. All other operations should be done in RAM only tbh.

I'm not sure if it makes sense.

tobes:
> I don't think it's optimal sorry

ok, part of the way it is is because I like the journalling idea and thought I'd see how it looked


> and I don't understand your dislike of pickle which has the advantage of keeping those python objects intact (datetime !)

It uses binary blobs.  I'm also not sure how it works around the whole python 2/3 str/unicode/bytes stuff.  From my point of view I'd want to have it fully compatible as I often run py3status under both 2/3.

> The cPickle dumps/loads implementation is fast enough for easy enough to implement imho.

I don't think the speed is an issue

> I'd go for RAM only set/delete/get (I'm no fan of the clear name) operations on a single dict with the module name as the key. TBH we could almost basically mirror the module object we already have on this dict.

clear -> delete makes sense

Personally I think the keys is a useful feature as it saves the module needing to care about the full data structure it is saving.

> Then I'd spawn a new thread which would read it from disk to RAM on load and persist the object on disk every minute and on stop only. All other operations should be done in RAM only tbh.

We only need to write it if it is changed so once a minute seems unnecessary.  I'm also totally paranoid about loosing data which is why it does the journal approach so only the last bit of data gets hit if we die during the write.  Modules like rate_counter could use this and that data would be a pain if it got lost/corrupted. To be fair the update can be done so it is synced like in vacuum() etc so we should prevent data loss but then we are writing all the data when anything is updated.

> I'm not sure if it makes sense.

yeah, and it's good to discuss these things

tobes:
I've switched it to use pickle.  Hopefully all the issues around python 2/3 bytes/unicode are resolved.  I suspect we may have some edge cases which is why we log the storage on first use.

ultrabug:
This will be considered in 3.7 since it's related to performance

raspbeguy:
@tobes Can I do something to help this PR to be merged ASAP?

tobes:
@ultrabug can we move this forward this week?

ultrabug:
Ok, but now that I see this I remember why I didn't want to merge it and it was about methods naming AND storage format.

Let's move forward this week @tobes and propose a renewed version to our pals 👍

tobes:
We are using pickle so the format should be acceptable. I think it was just the issue of having keys that was not agreed

ultrabug:
You're right, the pickle thing I said and you corrected already. So we'll get back to my Jan 10 comment and iterate from there.

ultrabug:
@tobes plz review and flag for merge if okay

doc/README.md
```diff
@@ -1159,6 +1160,42 @@ general one that will be used.
...
+__storage_del(key=None)__
+
+Remove the value stored with the key from storage.
+If key is not supplied then all values for the module are removed.
+
+__storage_keys()__
+
+Return a list of the keys for values stored for the module or an empty list.
+
+__storage_keys()__
```
lasers: Duplicate `keys()` OK? or use `items()` ?

tobes:
@ultrabug I added a bug fix, a very brief usage example, and a typo fix.

I think we should be good enough now to merge

lasers:
>Data is stored in a sqlite database py3status.sqlite created in the same directory that the config file is in (maybe we want to add an option to explicitly set a path).

I could see users not liking this as default when they go there to edit their i3config. Is `/tmp` a better place? or `~/.config/py3status`? or `~/.cache/py3status`?

raspbeguy:
Well, as long as the default can be overwritten, with a setting, I see no reason to change this. Maybe a lot of people would not like this path, but let's face it, a lot of people wouldn't like any other default path.

tobes:
> Is /tmp a better place?

You know /tmp gets deleted at boot so not good for persistance.

ultrabug:
@lasers @raspbeguy and most people won't care, so let's start with this.

ultrabug:
Thanks @tobes it's a great start for new features :)

raspbeguy:
Cool :) thanks everybody

tobes:
ha ha only a year :)

 Did you break master too!

PR #584, opened 2016-11-22
Module selector

tobes:
This PR introduces a new module `module_selector` that allows modules to be enabled/disabled whilst py3status is running.  It relies on PRs #585 (data-storage) and #571 (popups) but I thought it would be good to just have it here.  This branch contains those PRs.

Both py3status and i3status modules can be enabled/disabled.

I did consider presenting this as multiple PRs but it started getting a bit messy so I decided to just do a dump here.  It'll be much more readable if the PRs it depends on are merged first.  Anyhow I'm more than happy to split this into many PRs if needed.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -26,6 +26,7 @@ py3status documentation
 * [Example 5: Using color constants](#example_5)
 * [Module methods](#module_methods)
 * [Py3 module helper](#py3)
+* [Py3 storage](#py3_storage)
 * [Composites](#composites)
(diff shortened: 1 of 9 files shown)


ultrabug:
I'd like it in but on its own, is that possible ?

tobes:
I need to clean up/some of my prs. This needs persistent storage/pop-ups
etc so I'll try to them in first

tobes:
I'm closing this till we have the needed features eg persistent storage and popups

PR #585, opened 2016-11-23
imap module: added starttls option

raspbeguy:
See #579.

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -13,6 +13,7 @@ Configuration parameters:
     new_mail_color: what color to output on new mail (default '')
     password: login password (default '<PASSWORD>')
     port: IMAP server port (default '993')
+    security: what authentication method is used: 'ssl' or 'starttls' (default 'ssl')
     user: login user (default '<USERNAME>')
 
 Format placeholders:
@@ -22,6 +23,7 @@ Format placeholders:
 """
 
 import imaplib
+from ssl import create_default_context
 
 
 class Py3status:
@@ -37,6 +39,7 @@ class Py3status:
     new_mail_color = ''
     password = '<PASSWORD>'
     port = '993'
+    security = 'ssl'
     user = '<USERNAME>'
 
     def check_mail(self):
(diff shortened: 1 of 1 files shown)


py3status/modules/imap.py
```diff
@@ -64,11 +67,30 @@ def check_mail(self):
 
         return response
 
+    def _connection_ssl(self):
+        connection = imaplib.IMAP4_SSL(self.imap_server,
+                                       self.port,
+                                       ssl_context=create_default_context()
```
tobes: from python docs
> Changed in version 3.3: ssl_context parameter added.

We support python 2.6 so this will break peoples setups. Either this needs removing or you need to check we can use it.

py3status/modules/imap.py
```diff
@@ -64,11 +67,30 @@ def check_mail(self):
...
             mail_count = 0
             directories = self.mailbox.split(',')
-            connection = imaplib.IMAP4_SSL(self.imap_server, self.port)
+
+            if self.security == "ssl":
+                connection = self._connection_ssl()
+            elif self.security == "starttls":
+                connection = self._connection_starttls()
+            else:
+                raise ValueError("Unknown security protocol")
```
tobes: Either you want to do this test in `post_config_hook()` so the module just won't run (here it will just flood the logs with tracebacks) or just test for 'starttls' else assume 'ssl'

py3status/modules/imap.py
```diff
@@ -13,6 +13,7 @@
     new_mail_color: what color to output on new mail (default '')
     password: login password (default '<PASSWORD>')
     port: IMAP server port (default '993')
+    security: what authentication method is used: 'ssl' or 'starttls' (default 'ssl')
```
tobes: please add a note that 'starttls' needs python 3.2+

py3status/modules/imap.py
tobes: Actually looking a bit more at the code, it will just be ignored and confuse the user so we should still fix it.

py3status/modules/imap.py
raspbeguy: the original module didn't include any `ssl_context` so I think this is safe to remove it.

ultrabug:
Wow this is a cool feature @raspbeguy thank you very much for your contributions :)

raspbeguy:
you're welcome @ultrabug

PR #586, opened 2016-11-24
New RSS module

raspbeguy:
This module aims to display information about unread RSS items in your RSS aggregator. For now, only OwnCloud/NextCloud feed aggregator is supported, I plan to add support to tt-rss in the future.
There are several basic filtering option (by folder and by feed).


py3status/modules/rss.py
```diff
@@ -0,0 +1,112 @@
...
+folders IDs. You can get those ID by clicking on the desired feed or folder and watching
+the URL:
+```
+https://yourcloudinstance.com/index.php/apps/news/#/items/feeds/FEED_ID
+https://yourcloudinstance.com/index.php/apps/news/#/items/folders/FOLDER_ID
+
+```
+If both are left empty, all unread feed items will be counted. Counted items don't need
+to match both feeds and folders lists. Items matching both a given feed ID and a given
+folder ID will be counted only once.
```
tobes: This commentary is in the wrong place in the docstring. And Could at least link to the owncloud github page or similar.

Sections for colors and requirements are missing.

py3status/modules/rss.py
```diff
@@ -0,0 +1,112 @@
...
+                if (
+                    (not self.feeds_id and not self.folders_id) or
+                    feed["id"] in self.feeds_id or
+                    feed["folderId"] in self.folders_id
+                ):
+                    rss_count += feed["unreadCount"]
+
+            return rss_count
+
+        except:
```
tobes: Raw excepts are nasty.  Maybe try to separate the issues eg connections from feed parsing ones

py3status/modules/rss.py
```diff
@@ -0,0 +1,112 @@
+# -*- coding: utf-8 -*-
+"""
+Display the unread feed items in your OwnCloud/NextCloud account.
+
+Configuration parameters:
+    aggregator: feed aggregator used, see note below (default 'owncloud')
+    cache_timeout: how often to run this check (default 60)
+    feeds_id: list of IDs of feeds to watch, see note below (default [])
```
tobes: should this be `feed_ids`? `feeds_id` sounds odd to me. same with the folders

py3status/modules/rss.py
```diff
@@ -0,0 +1,112 @@
...
+    folders_id: list of IDs of folders ro watch, see note below (default [])
+    format: format to display (default 'Feed: {unseen}')
+    hide_if_zero: don't show on bar if False (default False)
+    new_items_color: text color when counter isn't zero (default '')
```
tobes: counter?

py3status/modules/rss.py
```diff
@@ -0,0 +1,112 @@
...
+    password: login password (default '<PASSWORD>')
+    server: IMAP server to connect to (default 'https://yourcloudinstance.com')
```
tobes: IMAP?

raspbeguy: As you can see, I cloned another module that I know well in order to make this one :)
Sorry

py3status/modules/rss.py
raspbeguy: It's just the number displayed in the bar. Okay, I should explain that a bit more.

py3status/modules/rss.py
raspbeguy: Ok, I'll change it

py3status/modules/rss.py
```diff
@@ -0,0 +1,112 @@
...
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    aggregator = 'owncloud'
+    cache_timeout = 60
+    feeds_id = []
+    folders_id = []
+    format = 'Feed: {unseen}'
+    hide_if_zero = False
```
raspbeguy: okay

raspbeguy:
As I said, for now only Owncloud is supported, but that doesn't mean other RSS aggregator won't ever be supported (that's why the `aggregator` option exists, even if it is useless for the moment).
I don't see the point of a simple RS module (without aggregator), because you must have a reference to determine if a RSS item has been read or not. How would the module be able to count unseen items if you only give it a RSS URL ? I don't know if I was clear.

tobes:
> I don't see the point of a simple RS module (without aggregator)

I think there is a valid argument for having one that would notify the user of new items etc.  But it would not be compatible with this module.  Anyhow renaming the module removes this issue - thanks.

raspbeguy:
@tobes I think you didn't understand what I said but it doesn't matter now.

tobes:
> I think you didn't understand what I said 

likewise, new != unread.

The whole rss feed is an interesting notion that I have occasionally consided, however a decent implementation has not yet been possible. Some current PRs #583, #571 would allow a quite functional model to be written.

> but it doesn't matter now.

true :)

py3status/modules/rss.py
raspbeguy: @tobes should I write it directly in the option description?

tobes: https://github.com/ultrabug/py3status/blob/master/doc/README.md#docstring

raspbeguy:
fixed things, let's see what you think about it.

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    # available configuration parameters
+    aggregator = 'owncloud'
+    cache_timeout = 60
+    feed_ids = []
+    folder_ids = []
+    format = 'Feed: {unseen}'
+    password = '<PASSWORD>'
+    server = 'https://yourcloudinstance.com'
+    user = '<USERNAME>'
```
tobes: I'd be inclined to have this and password default to None so you can tell if a user actually set it or not.   Also if not set then in the request just do `auth=None`

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError('%s is not a supported feed aggregator' % self.aggregator)
+        if self.aggregator == "owncloud":
+            self._verify_owncloud()
+
+    def _verify_owncloud(self):
+        api_url = "%s/index.php/apps/news/api/v1-2/" % self.server
+        try:
+            r = requests.get(api_url + "feeds",
+                             auth=(self.user, self.password),
+                             timeout=1)
```
tobes: A timeout of 1 seems a little short, maybe 10 is more reasonable.  Something like a DNS lookup can easily end up talking a second if the first server doesn't respond.

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,131 @@
...
+            response['color'] = self.py3.COLOR_ERROR or self.py3.COLOR_BAD
+        elif rss_count != 0:
+            response['color'] = self.py3.COLOR_NEW_ITEMS or self.py3.COLOR_GOOD
+            response['full_text'] = self.py3.safe_format(
+                self.format, {'unseen': rss_count}
+            )
+        else:
+            response['full_text'] = self.py3.safe_format(
+                self.format, {'unseen': rss_count}
+            )
```
tobes: I think this code could be improved.  You are repeating
```
    self.format, {'unseen': rss_count}
)
```
when this could be avoided.

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,131 @@
...
+
+Format placeholders:
+    {unseen} sum of numbers of unread feed elements
+
+Color options:
+    color_new_items: text color when there is new items (default color_good)
+    color_error: text color when there is an error (default color_bad)
+
+Requires:
+    python-requests: library to make web requests
```
tobes: I'd just call this `requests: python module from pypi https://pypi.python.org/pypi/requests` as that is how other modules refer to it (the name anyhow)

raspbeguy:
This question is relevant. I am only focusing on the RSD feature. Maybe I'll do another module for the main feature of owncloud (file syncing) but it isn't the aim of this module at all.

raspbeguy:
I think this was a bad idea to put all these tests in `post_config_hook`: you won't be able to start py3status without internet access. So some of these exceptions must be caught during runtime.

raspbeguy:
fixed things

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,132 @@
...
+                raise ValueError('Server returning bad content, check credentials')
+
+    def check_news(self):
+        if self.aggregator == "owncloud":
+            rss_count = self._get_count_owncloud()
+
+        response = {'cached_until': self.py3.time_in(self.cache_timeout)}
+
+        if rss_count is None:
+            response['full_text'] = "Unreachable"
```
ultrabug: I find it better to notify (self.py3.notify_user) the user instead of replacing the whole status text with such a generic and uninformative output.

In case it was the first try, you might want to have the `unseen` placeholder set to `?` for example. Else, just leave the previous value displayed while this temporary error passes.

raspbeguy: I'm not sure it would be comfortable for the user to get a big red banner each time internet is unreachable, especially if all the modules using internet have this strategy (IMAP for instance could do it, even if it isn't the case yet)

raspbeguy: The big idea is: if the tests in `post_config_hook` are successfull, the only possible error later is internet access (or the server maintainer just put the service offline, but I'll put it in the category "access problem").
That's why I had the simple message "unreachable".

raspbeguy: But I agree with the cached response idea and the "?" if error at the first check.

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,132 @@
...
+        api_url = "%s/index.php/apps/news/api/v1-2/" % self.server
+        try:
+            r = requests.get(api_url + "feeds",
+                             auth=(self.user, self.password),
+                             timeout=10)
+        except requests.exceptions.MissingSchema:
+            raise ValueError('Unconsistent server URL')
+        except (
+            requests.exceptions.ConnectionError,
+            requests.exceptions.ConnectTimeout
```
tobes: I think explicitly importing these might make the code more readable
```
from requests.exceptions import ConnectionError, ConnectTimeout
```

tobes:
> Maybe I'll do another module for the main feature of owncloud (file syncing) but it isn't the aim of this module at all.

I've not used owncloud, but reading the docs.  If I did use it, then a rich owncloud module would seem good, it would also simplify the code a little for now.

raspbeguy:
> a rich owncloud module would seem good, it would also simplify the code a little for now.
No. This module is for RSS aggregator info only, it is ABSOLUTELY NOT focused on Owncloud at all. I did the Owncloud implementation for now because I had only an Owncloud installation to test with. But I'll soon do Tiny Tiny RSS implementation.
This module is NOT intended to be a Owncloud client.

tobes:
> This module is NOT intended to be a Owncloud client.

Cool, I was just speculating.

You should probably adjust the docstring intro to explain that this is an rss module and that owncloud is currently the only source and that expanding support for other sources is encouraged.  As the first line is very owncloud centric to the casual reader.

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,132 @@
...
+    format = 'Feed: {unseen}'
+    password = None
+    server = 'https://yourcloudinstance.com'
+    user = None
+
+    def post_config_hook(self):
+        if self.aggregator not in ['owncloud']:  # more options coming
+            raise ValueError('%s is not a supported feed aggregator' % self.aggregator)
+        if self.aggregator == "owncloud":
+            self._verify_owncloud()
```
tobes: I think this wants moving as we might not have connection here but might later.  These feel like they could be handled as part of `check_news()`.  If the error is unrecoverable eg not owncloud instance etc we can still effectively disable the module by returning something like `{'full_text': 'Error',  'cached_until': self.py3.CACHE_FOREVER}`

raspbeguy:
fixed things. again.

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,140 @@
...
+        self._cached = {'cached_until': self.py3.CACHE_FOREVER,
+                        'color': self.py3.COLOR_ERROR or self.py3.COLOR_BAD,
+                        'full_text': '?'
+                        }
+        if self.aggregator not in ['owncloud']:  # more options coming
+            self.py3.notify_user('%s is not a supported feed aggregator' % self.aggregator)
+            return
+        if self.aggregator == "owncloud":
+            self._verify_owncloud()
+        if self.user is None or self.password is None:
```
ultrabug: shouldn't this be checked before `self._verify_owncloud() ` ?

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,140 @@
...
+                self.py3.notify_user('Server returning bad content, check credentials')
+
+    def check_news(self):
+        if self.aggregator == "owncloud":
+            rss_count = self._get_count_owncloud()
+
+        response = {'cached_until': self.py3.time_in(self.cache_timeout)}
+
+        if rss_count is None:
+            return self._cached
```
ultrabug: since `self._cached` returns a `self.py3.CACHE_FOREVER` cache time, the module will never try to refresh again after a connection failure occurs at the first try. I guess this is not the intended behaviour

py3status/modules/rss_aggregator.py
tobes: In my opinion either these checks should be fatal (stop the module running at all), or just be part of the whole `check_news()` and just do `self.py3.notify_user(..., rate_limit=None)` for any problems.

raspbeguy: @tobes so you think that a misconfiguration should stop any other module to run ? That's a real question, actually I don't know what behaviour it should have.

tobes: Personally I'd just do everything in `_get_count_owncloud()` but that might be because I'm lazy and it feels like it'd be less code.  It does mean all the owncloud stuff is in one place too for if other feeds are added. The main thing is that things like the network being down shouldn't kill it, and we don't over notify the user when there is a problem.

tobes: As far as mis configuration stopping other modules.  Yes it should but it can do it however it wants the `post_config_hook()` way is new but the CACHE_FOREVER way is just as good really.

ultrabug: > you think that a misconfiguration should stop any other module to run ? That's a real question, actually I don't know what behaviour it should have.

yes, there's no point in running it constantly if we know its config won't make it possible for it to work. we shall save CPU as much as we can

> It does mean all the owncloud stuff is in one place too for if other feeds are added

I agree, let's be straightforward and not overcomplicate things tbh. + I'd like this PR to get in, so if you're okay @raspbeguy let's move this way

Also @raspbeguy I don't know if you're an IRC person, but it would be cool to have you around so we can speak more interactively !

py3status/modules/rss_aggregator.py
```diff
@@ -0,0 +1,113 @@
...
+                self.format, {'unseen': self._cached}
+            )
+            response['color'] = self.py3.COLOR_ERROR or self.py3.COLOR_BAD
+        else:
+            response['full_text'] = self.py3.safe_format(
+                self.format, {'unseen': rss_count}
+            )
+            if rss_count != 0:
+                response['color'] = self.py3.COLOR_NEW_ITEMS or self.py3.COLOR_GOOD
+        self._cached = rss_count
```
tobes: should this be None ever?

`self._cached = rss_count or self._cached`
Also if you do this earlier you only need to set full_text once.

raspbeguy: There is a `if` statement catching the case of `None` value.
```python
if rss_count is None:
        self.format, {'unseen': self._cached}
    )
    response['color'] = self.py3.COLOR_ERROR or self.py3.COLOR_BAD
```

raspbeguy: Oh, I see, nevermind

raspbeguy:
@tobes we discussed on IRC with @ultrabug. I did only what I was asked to.

tobes:
I think all the issues I still have are minor, and can always be improved later if users want to make a PR .

Good work @raspbeguy

raspbeguy:
Thanks.

ultrabug:
Awesome @raspbeguy ! merci 👍

PR #587, opened 2016-11-28
new not_zero format command

tobes:
It seems a common usecase to not want to show format placeholders when the value is 0 eg number of notifications.

Currently this is possible but a bit verbose `format = '[\?if=value {value}]'` this PR adds a format command to make this `format = '[\?not_zero {value}]'` that makes things more readable it also will treat strings like `'0'` and `'0.0'` as zero values, which feels like it makes sense as we allow these to be formatted as numbers.

### DIFF
--- a/py3status/formatter.py
+++ b/py3status/formatter.py
@@ -23,6 +23,7 @@ class BlockConfig:
     has_commands = False
     max_length = None
     min_length = 0
+    not_zero = False
     show = False
 
     def update_commands(self, commands_str):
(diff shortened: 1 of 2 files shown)


ultrabug:
I think we need better docs on this formatter, its own topic wouldn't hurt

tobes:
> I think we need better docs on this formatter, its own topic wouldn't hurt

I agree.  Part of me wants to get all the modules converted to use it so that then the documentation will apply to all modules as currently it doesn't.  Maybe that is just an excuse to avoid the work.

raspbeguy:
That's a great idea. I didn't know how to do it else.

PR #589, opened 2016-12-03
RSS : added Tiny Tiny RSS as available aggregator

raspbeguy:
I know, the previous version was just merged. I initially couldn't tell that I would implement ttrss so early. But I had a lot of time, so I did it. Hope you guys aren't too mad at me.

### DIFF
--- a/py3status/modules/rss_aggregator.py
+++ b/py3status/modules/rss_aggregator.py
@@ -154,10 +154,10 @@ class Py3status:
                     for feed in r.json()['content']:
                         if feed['id'] in self.feed_ids:
                             rss_count += feed['unread']
-            requests.post(api_url, data={
+            requests.post(api_url, data=json.dumps({
                 'sid': sid,
                 'op': "logOut"
-            })
+            }))
 
             return rss_count
 
--- a/py3status/modules/rss_aggregator.py
+++ b/py3status/modules/rss_aggregator.py
@@ -1,24 +1,38 @@
 # -*- coding: utf-8 -*-
 """
-Display the unread feed items in your OwnCloud/NextCloud account.
+Display the unread feed items in your favorite RSS aggregator.
+
+For now, supported aggregators are:
+    * OwnCloud/NextCloud with News application
+    * Tiny Tiny RSS 1.6 or newer
 
 You can also decide to check only for specific feeds or folders of feeds. To use this
 feature, you have to first get the IDs of those feeds or folders. You can get those IDs
-by clicking on the desired feed or folder and watching the URL:
+by clicking on the desired feed or folder and watching the URL.
+
+For OwnCloud/NextCloud:
 ```
 https://yourcloudinstance.com/index.php/apps/news/#/items/feeds/FEED_ID
 https://yourcloudinstance.com/index.php/apps/news/#/items/folders/FOLDER_ID
 
 ```
+For Tiny Tiny RSS:
+
+```
+https://yourttrssinstance.com/index.php#f=FEED_ID&c=0
+https://yourttrssinstance.com/index.php#f=FOLDER_ID&c=1
+```
+
 If both feeds list and folders list are left empty, all unread feed items will be counted.
-Counted items don't need to match both feeds and folders lists. Items matching both a
-given feed ID and a given folder ID will be counted only once.
+You may use both feeds list and folders list, but given feeds shouldn't be included in
+given folders, else unread count number behavior is unpredictable. Same warning when
+aggregator allows subfolders: the folders list shouldn't include a folder and one of its
+subfolder.
 
 Configuration parameters:
-    aggregator: feed aggregator used, see note below. For now, the only supported value is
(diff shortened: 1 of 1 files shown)


py3status/modules/rss_aggregator.py
```diff
@@ -58,15 +74,18 @@ class Py3status:
...
             raise ValueError("user and password must be provided")
 
     def check_news(self):
         if self.aggregator == "owncloud":
             rss_count = self._get_count_owncloud()
-            self._cached = rss_count or self._cached
+        elif self.aggregator == "ttrss":
+            rss_count = self._get_count_ttrss()
+
+        self._cached = self._cached if rss_count is None else rss_count
```
tobes: you can write this as 
`self._cached =  rss_count or self._cached`
which I find more readable

py3status/modules/rss_aggregator.py
```diff
@@ -1,24 +1,38 @@
...
+https://yourttrssinstance.com/index.php#f=FOLDER_ID&c=1
+```
+
 If both feeds list and folders list are left empty, all unread feed items will be counted.
-Counted items don't need to match both feeds and folders lists. Items matching both a
-given feed ID and a given folder ID will be counted only once.
+You may use both feeds list and folders list, but given feeds shouldn't be included in
+given folders, else unread count number behavior is unpredictable. Same warning when
+aggregator allows subfolders: the folders list shouldn't include a folder and one of its
+subfolder.
```
tobes: Isn't this something you could cope with?  Doesn't each feed have a uuid that you can add to a set and then easily get the unique count?

Anyhow I haven't looked at the data that we get so this is just a guess.

py3status/modules/rss_aggregator.py
```diff
@@ -36,11 +50,13 @@
 
 Requires:
     requests: python module from pypi https://pypi.python.org/pypi/requests
+    json: python module from pypi https://pypi.python.org/pypi/json
```
tobes: json is a core lib so does not need to be listed in the requirements.

py3status/modules/rss_aggregator.py
```diff
@@ -100,10 +119,61 @@ def _get_count_owncloud(self):
...
+                        if feed['id'] in self.feed_ids:
+                            rss_count += feed['unread']
+            requests.post(api_url, data=json.dumps({
+                'sid': sid,
+                'op': "logOut"
+            }))
+
+            return rss_count
+
+        except:
```
tobes: I hate these raw excepts (I know why you've done it).

I think I need to make report_exception available to modules for this type of case so that we can log issues nicely.
anyhow I'm ok with it here.

py3status/modules/rss_aggregator.py
```diff
@@ -1,24 +1,38 @@
 # -*- coding: utf-8 -*-
 """
-Display the unread feed items in your OwnCloud/NextCloud account.
+Display the unread feed items in your favorite RSS aggregator.
+
+For now, supported aggregators are:
+    * OwnCloud/NextCloud with News application
+    * Tiny Tiny RSS 1.6 or newer
```
tobes: When was 1.6 released?  I couldn't find out.   Just curious

py3status/modules/rss_aggregator.py
```diff
@@ -100,10 +119,61 @@ def _get_count_owncloud(self):
...
 if __name__ == "__main__":
     """
     Run module in test mode.
     """
+    config = {
+        'aggregator': "ttrss",
+        'server': "http://localhost:8080",
+        'user': "admin",
+        'password': "password"
+    }
```
tobes: you aren't doing anything with this, If it is just for occasional testing then add a note to that effect

py3status/modules/rss_aggregator.py
raspbeguy: According to project page, 4 years ago.

tobes: ok no need to worry about legacy use unless someone really wants to add it.

py3status/modules/rss_aggregator.py
raspbeguy: okay

raspbeguy:
Pushed changes.
btw, I noticed that `README.md` wasn't updated with this new module. Should I update it myself ?

py3status/modules/rss_aggregator.py
tobes: fair answer

tobes:
>  I noticed that README.md wasn't updated with this new module. Should I update it myself ?

you can if you want `py3status docstring update rss_aggregator` should do it, but some people have had issues using this.  Anyhow it is easy to do so not that important.

raspbeguy:
> `py3status docstring update rss_aggregator` should do it

I don't know how to use it. Anyway, not important.

ultrabug:
> Hope you guys aren't too mad at me.

Mate, never apologize for spending some of your free time on an Open Source project.

**Thanks a lot !**

PR #590, opened 2016-12-05
backlight: fixed bug in self._set_backlight_level

raspbeguy:
The previous version was calling `self.py3.command_run()` which doesn't exists.

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -86,7 +86,7 @@ class Py3status:
         self._set_backlight_level(level)
 
     def _set_backlight_level(self, level):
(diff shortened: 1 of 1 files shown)


ultrabug:
@raspbeguy actually there is a `command_run` method for py3 and this module works (at least for me).

Mind you share a bit more about the error you're facing ?
Did you copy the backlight module or are you running the one from your py3status installation ?

raspbeguy:
@ultrabug ok, I worked with the `py3.py` from my running installation (archlinux repo) and the `backlight.py` from upstream. My bad.
But it doesn't stop the module not to work in the distro repo.

PR #591, opened 2016-12-07
cache format string tokens to reduce work done

tobes:
The formatter spends approximately 20% of its time tokenizing the format string.  Seeing as we keep using the same format strings we can cache this and so reduce our overall workload.


ultrabug:
Very interesting finding indeed, thanks a lot !

PR #596, opened 2016-12-12
Fix python2 unicode use in i3status time modules

tobes:
Fixes bug where unicode format strings killed time/tztime


ultrabug:
Thanks @tobes !

PR #598, opened 2016-12-19
New module: Do Not Disturb

maximbaz:
This module has a very simple idea behind it: to be able to turn on and off all the notifications in the system with a single click.

This is easy to achieve if all potentially distracting apps (like email, chat, etc.) use native system notifications rather than their own implementations. This module assumes such setup :wink: 

## Screenshots:

All the notifications are enabled:
image

Need to focus? A single mouse click disables the all system notifications:
image

Done working? A single mouse click, and you have all your notifications back:
image

## Implementation details

I couldn't figure out a smart way to disable notifications, that would work with all available notification managers. But I saw that many people suggest having a shell script that periodically kills the manager's process... So I took the same approach :smile: a bit ugly, but it works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 build
 dist
+tags
 *.egg-info
 .tox/
 .cache/
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.pyc
 build
 dist
-tags
 *.egg-info
 .tox/
 .cache/
--- a/py3status/modules/do_not_disturb.py
+++ b/py3status/modules/do_not_disturb.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+"""
+A simple "Do Not Disturb" module that can turn on and off all system notifications.
+
+A left mouse click will toggle the state of this module.
+
+Configuration parameters:
+    format_off: Display format when the "Do Not Disturb" mode is disabled.
+        (default 'OFF')
+    format_on: Display format when the "Do Not Disturb" mode is enabled.
+        (default 'ON')
+    notification_manager: The process name of your notification manager.
+        (default 'dunst')
+    refresh_interval: Refresh interval to use for killing notification manager process.
+        (default 0.25)
+
+Color options:
+    color_bad: "Do Not Disturb" mode is enabled.
+    color_good: "Do Not Disturb" mode is disabled.
+"""
+
+from time import sleep
+from os import system
+from threading import Thread, Event
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    format_off = 'OFF'
+    format_on = 'ON'
+    notification_manager = 'dunst'
+    refresh_interval = 0.25
+
+    def __init__(self):
+        self.running = Event()
+        self.killed = Event()
+
+        class MyThread(Thread):
+            def run(this):
+                while not self.killed.is_set():
+                    if self.running.wait():
+                        system("killall '{}'".format(self.notification_manager))
+                        sleep(self.refresh_interval)
+
+        MyThread().start()
+
+    def do_not_disturb(self):
+        response = {
+            'cached_until': self.py3.CACHE_FOREVER,
+            'full_text': self.format_on if self.running.is_set() else self.format_off,
+            'color': self.py3.COLOR_BAD if self.running.is_set() else self.py3.COLOR_GOOD
+        }
+        return response
+
+    def on_click(self, event):
+        if event['button'] == 1:
+            if self.running.is_set():
+                self.running.clear()
+            else:
+                self.running.set()
+
+    def kill(self):
+        self.killed.set()
+        self.running.set()  # ensure the thread won't hang
+
+
+if __name__ == "__main__":
+    """
+    Run module in test mode.
+    """
+    from py3status.module_test import module_test
+    module_test(Py3status)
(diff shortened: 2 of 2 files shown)


tobes:
nice idea,

You can send signals to dunst

```
killall -SIGUSR1 dunst # pause
killall -SIGUSR2 dunst # resume
```

It does queue notifications which may not be what you wanted but `killall -SIGSTOP dunst` would kill the queue and re-enable

Now this would probably be dunst specific but as we have to define the notification process we could teach the module about new processes if needed.

maximbaz:
Cool, I didn't know!

I think we should definitely _not_ queue notifications, with managers like `notify-osd` users would then sit for 2 hours seeing all old notifications, and the new ones will be blocked until all the old ones are gone :smile: 

Reading now about `pkill` vs `killall`: 
> `killall` by default it requires the precise process name, whereas `pkill` does basic pattern matching. Thus, `killall` is safer for users to blindly copy and paste.

I'm definitely switching to `killall` then!

I also tried `-SIGSTOP` vs `-SIGTERM` (which is default).
The `notify-osd` for example completely hangs after it receives the `-SIGSTOP` signal, moreover if there already was a visible notification on the screen, it stays visible forever. And if I execute `notify-send ...`, it will also hang.

I'd like to keep the module as simple and as generic as possible, i.e. not return to it for every new notification manager :smile:. Therefore I'd rather keep the simple `killall <process_name>`, which works for all. 

What do you think @tobes ?

tobes:
> Therefore I'd rather keep the simple killall <process_name>, which works for all.

I don't like the fact that it is always running the thread.  You could probably adapt it so we only run the thread when we are in do not disturb mode, which would reduce the load.

The advantage of the specific approach is we really can cut down on the work done.  So maybe a compromise and we have a general `killall <process>` and we can also add specific managers as we see fit (but this can always be done later)

py3status/modules/do_not_disturb.py
```diff
@@ -0,0 +1,75 @@
...
+from threading import Thread, Event
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    format_off = 'OFF'
+    format_on = 'ON'
```
tobes: I'd be tempted to have
```
format = '{state}'
state_on = 'ON'
state_off = 'OFF'
```

but maybe I'm a little insane

maximbaz:
> I don't like the fact that it is always running the thread. You could probably adapt it so we only run the thread when we are in do not disturb mode, which would reduce the load.

The good thing is that thread is not running in an infinite loop when the "do not disturb" is disabled, the execution is stopped on this line and waiting for the event:

```
if self.running.wait():
```

I think such paused thread will consume almost no resources, and I'm not sure this is worse than creating and deleting threads on every click.

py3status/modules/do_not_disturb.py
maximbaz: Up to you, if you like it, I can change :wink: 

I took the current configuration from the example which I saw in my own `.i3status.conf`:

```
wireless wlp3s0 {
  format_up = ""
  format_down = ""
}
```

tobes:
> if self.running.wait():

Yeah I saw that but after I made my comment.  Not noticed that before, always good to learn new things and I still have to take a look.

maximbaz:
Just to confirm, is there any action from my side to be done before this pull request can be merged?

tobes:
hmm,

I still like the state_.. over format_..

I'd also just not start the thread unless we actually need to.  Also I don't like the whole `kill ..` stuff constantly looping.  That just feels dirty.  I don't really like the sending USR1 to dunst either.  I can't find how dunst is actually being started on my system which is annoying.  I'd have thought this is where systemd might be useful to enable/disable services.

So that's where my thoughts are at.

maximbaz:
While `state_` vs `format_` is an easy change, the other comments show a conceptual difference :wink:

I'm also not particularly fond of this approach, but I don't see any better ones for now, and since I use `notify-osd` these days I don't feel like implementing some special logic for `dunst` (that would for example use `SIGSTOP` instead of a thread with `SIGTERM` that works for all).

I'll keep on hosting this module locally then, and if you decide to merge it, ping me and I'll adapt the `format` parameter. Or feel free to take the code and adapt it to your preferences, if this module will still work fine with `notify-osd` I will gladly switch to use it :stuck_out_tongue_winking_eye:

tobes:
I don't have a problem with it being merged as other users may find it useful.

I really like the concept and if your code works then that's a good start

py3status/modules/do_not_disturb.py
```diff
@@ -0,0 +1,78 @@
...
+                    if self.running.wait():
+                        system("killall '{}'".format(self.notification_manager))
+                        sleep(self.refresh_interval)
+
+        MyThread().start()
+
+    def do_not_disturb(self):
+        response = {
+            'cached_until': self.py3.CACHE_FOREVER,
+            'full_text': self.format.format(state=self.state_on if self.running.is_set() else self.state_off),
```
tobes: this should really use `self.py3.safe_format()`

.gitignore
```diff
@@ -1,6 +1,7 @@
 *.pyc
 build
 dist
+tags
```
ultrabug: I'm not sure I understand why you are modifying the gitignore mate ?

maximbaz-microsoft: that's an autogenerated ctags file, I used it at some point, but actually I can just revert this piece, doesn't matter

maximbaz: that's an autogenerated ctags file, I used it at some point, but actually I can just revert this piece, doesn't matter

ultrabug:
Thanks a lot for your contribution once again @maximbaz !

PR #601, opened 2016-12-24
Use color 'bad' instead of 'high' as a default threshold in sysdata

maximbaz:
It doesn't make sense to use "high" as a default color, there is no such color in py3status and without extra configuration it won't be colorized at all. It should have been "bad", which is rendered as red.

Also see line 176 which shows that the default value is "bad" for those who still use the obsolete way of configuring thresholds.


tobes:
I think we should just change this too.  I'd consider it a bug fix.

ultrabug:
thanks @maximbaz idd

PR #603, opened 2016-12-29
volume_status PactlBackend does not read user-specified device

vicyap:
the volume_status module's PactlBackend regex was hardcoded for sink #0 , now the regex uses self.device

I ran into the error because my computer has two output devices (HDMI=0, Stereo=1).


tobes:
@vicyap thanks for this improvement

ultrabug:
thanks a lot @vicyap !

PR #604, opened 2016-12-30
No device fix

### LINKED ISSUE #557: module battery_level: crash when no battery present
The module crashes when there is no battery present, I get the following output:
```
  File "battery_level.py", line 448, in <module>
    module_test(Py3status)
  File "/usr/lib/python3.5/site-packages/py3status/module_test.py", line 51, in module_test
  File "battery_level.py", line 174, in battery_level
    self._refresh_battery_info()
  File "battery_level.py", line 316, in _refresh_battery_info
    battery = battery_list[self.battery_id]
```
Naturally, with no battery installed, battery_list is empty, which leads to the crash.

laxd:
Quick fixes for battery and backlight displays (in particular fixes #557):

- Battery module failed when running on a computer without a battery
- Backlight module failed when running on a computer with a monitor that can't have its brightness adjusted.

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -96,8 +96,11 @@ class Py3status:
         return brightness * 100 // brightness_max
 
     def backlight(self):
-        level = self._get_backlight_level()
-        full_text = self.py3.safe_format(self.format, {'level': level})
+        full_text = ""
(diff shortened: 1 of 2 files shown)


ultrabug:
Awesome @laxd thanks a lot !

Issue #606, opened 2017-01-03
TypeError in PrivateBase64(Private) after recent Python update


green-tea:
Hi,

I've been running py3status for some time on two Arch machines here, with identical configurations. After a recent update on one of the machines, which seems to have brought in Python 3.6 and a host of related updated packages, I get this error when running py3status (either the Arch packaged py3status 3.3-1 or a build of py3status-git):

> TypeError: _catch() missing 1 required positional argument: 's'

The other machine is running fine with the same configuration, and I've tested the latest git checkout there as well.

Full output:

> $ py3status__ 
> Traceback (most recent call last):
>   File "/usr/bin/py3status", line 11, in <module>
>     load_entry_point('py3status==3.3', 'console_scripts', 'py3status')()
>   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 561, in load_entry_point
>     return get_distribution(dist).load_entry_point(group, name)
>   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2631, in load_entry_point
>     return ep.load()
>   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2291, in load
>     return self.resolve()
>   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2297, in resolve
>     module = __import__(self.module_name, fromlist=['__name__'], level=0)
>   File "/usr/lib/python3.6/site-packages/py3status/__init__.py", line 4, in <module>
>     from py3status.core import Py3statusWrapper
>   File "/usr/lib/python3.6/site-packages/py3status/core.py", line 22, in <module>
>     from py3status.parse_config import process_config
>   File "/usr/lib/python3.6/site-packages/py3status/parse_config.py", line 23, in <module>
>     from py3status.private import PrivateHide, PrivateBase64
>   File "/usr/lib/python3.6/site-packages/py3status/private.py", line 124, in <module>
>     class PrivateBase64(Private):
> TypeError: _catch() missing 1 required positional argument: 's'

onny:
I also have this issue, thanks for reporting it!

tobes:
Thanks for reporting this. I'll look at fixing it as soon as I get my interweb back and can get a copy of 3.6

I think Unicode may have a new method, maybe around the format strings that is causing this.

tobes:
@green-tea @onny 

[...]

PR #610, opened 2017-01-05
travis include 3.6

tobes:
This adds 3.6 to the travis tests, also fixes the private issue #606 

It is possible we have other 3.6 issues that have not been triggered


ultrabug:
Thanks a lot, that's something I'd have had a hard time fixing

PR #630, opened 2017-01-08
window_title_async: Fix name

lasers:
This helps with telling the difference between two scripts in the log.


tobes:
I'm not that fussed with this but it does no harm

ultrabug:
Sanity makes sane people 🍀

PR #637, opened 2017-01-10
whoami: Add format

lasers:
This adds format: string to print (default '{username}').


ultrabug:
Thanks !

tobes:
We should document the Format placeholders see https://github.com/ultrabug/py3status/tree/master/doc#docstring

PR #662, opened 2017-01-14
fix update_placeholder_formats not updating

tobes:
#602 introduced a bug whereby we no longer update the config.  This fixes it and is needed for #555


ultrabug:
omg I would love to see #555 merged ! thanks

Issue #595, opened 2016-12-11
MPRIS: won't catch the player if the module is launched before any player


raspbeguy:
Hi,

There is an annoying bug in the MPRIS module:

When I start (or restart) i3 and no MPRIS player is launched, then it won't catch any player. So I have to launch my player and then restart i3 in order for the player to be recognized.

I gave a look to the code and I believe that the function listing the running MPRIS players is launched in the `post_config_hook`, that would explain this behavior. I would gladly fix it in a PR, but I don't understand everything in this module and maybe there is a reason for that.

tobes:
Yeah the starting things in post config hook probably shouldn't be
happening. I'm not sure this actually causes your problems but you may be
right.

I'm out so can't look at the code

tobes:
@raspbeguy I think you'll need to do some debugging on this.  I think maybe you don't have pydbus or gi.repository installed right

@ritze have you any thoughts.  With the post_config_hook() stuff don't worry to much as this needs a good solution for a few other modules too.

ritze:
I know this bug. I already mentioned it in #354 at the end. It always happens when you start a new player after launching the module started. I think the problem was starting a new thread with a dbus listener for the new player.

tobes:
@ritze thanks for replying I can find this in #354 but that sounds a different issue to what @raspbeguy is having.

There are still two issues:

 *   Running a player with a high priority and starting a second player with a lower priority the module do nothing until the second player is closed.
*    Sometimes the module freezes.

raspbeguy:
@tobes I have all the requirements installed.

ritze:
I think we should rewrite the mpris module in the future™. Finding this bug cost me at least three hours. This is a typical example for grown source code.

tobes:
this should now be fixed in master

PR #668, opened 2017-01-15
Detect players starting after py3status

ritze:
The mpris module catches all player which are started after the py3status instance (see Bug #595)

This fix doesn't work with all players, e.g. bomi!

### DIFF
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -310,7 +310,7 @@ class Py3status:
             self._kill = True
 
     def _name_owner_changed(self, *args):
-        player_add = args[5][2]
+        player_add = args[5][0]
         player_remove = args[5][1]
         if player_add:
             self._add_player(player_add)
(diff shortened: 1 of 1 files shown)


ritze:
I think bomi starts not fast enough and we got a race condition. We got an exception when accessing on the dbus in __add_player()_: `self._dbus.get('org.mpris.MediaPlayer2.bomi', SERVICE_BUS_URL)`

Sometimes it works with bomi. With other players like vlc this always works perfectly.

tobes:
@raspbeguy does this fix #595 for you/

ultrabug:
Thanks @ritze , waiting for @raspbeguy to respond

raspbeguy:
@ultrabug @tobes really sorry for the delay, I'm very busy right now. Yes, it fixed the problem, thanks @ritze

ultrabug:
Thanks a lot !

PR #673, opened 2017-01-16
remove i3-py dependencies

tobes:
#620 and #622 revealed some bugs in `window_title` and `scratchpad_counter`

the bugs were the searching functions returning None and i3-py unicode issues.

As i3-py is broken/unmaintained and we only use it to get the tree we can just remove it and use `i3-msg -t get_tree`


ultrabug:
☠️ dead code

PR #674, opened 2017-01-16
rainbow: add multi_color option

tobes:
#618 suggested a great new feature.

So now if you do 

```
rainbow {
    multi_color = True
    ...
}
```

We have even more color.  This is such a great option I'd almost suggest defaulting it to True

### DIFF
--- a/py3status/modules/rainbow.py
+++ b/py3status/modules/rainbow.py
@@ -25,6 +25,8 @@ Configuration parameters:
     gradient: The colors we will cycle through, This is a list of hex values
         *(default [ '#FF0000', '#FFFF00', '#00FF00', '#00FFFF',
         '#0000FF', '#FF00FF', '#FF0000', ])*
+    multi_color: If True then each module the rainbow contains will be colored
+        differently (default True)
     steps: Number of steps between each color in the gradient
         (default 10)
 
@@ -78,6 +80,7 @@ class Py3status:
         '#FF00FF',
         '#FF0000',
     ]
+    multi_color = True
     steps = 10
 
     class Meta:
@@ -176,7 +179,13 @@ class Py3status:
         content = self._get_current_output()
 
         output = []
-        for item in content:
+
+        if content:
+            step = len(self.colors) // len(content)
+        for index, item in enumerate(content):
+            if self.multi_color:
+                offset = (self.active_color + (index * step)) % len(self.colors)
+                color = self.colors[offset]
             obj = item.copy()
             if self.force or not obj.get('color'):
                 obj['color'] = color


lasers:
Aww yiss! Sweet! 🌈 😂 Almost there! I got `IndexError: list index out of range` w/ all modules.

tobes:
@lasers can you post the traceback from the logs

lasers:
Testing it without `external_script` right now. Will do.

tobes:
I think that will fix it and also give you a nicer rainbow

lasers:
I think it was user error because I didn't specific any options in`external_script` and all other modules. Running fine since.... No crash. Will go back to old way to retrieve you logs.

lasers:
Hi. Back. Sorry. Bad X11 or something. Been like this for some time now. Was freezing up with this module running. Went to TTY. Killed everything. Updated 80~ packages including X11 stuffs and python 3.6.0-2.

Should rainbow have an option to turn colors off? That way, I don't have to comment out everything and keep non-rainbow duplicates in i3status? Hard to make sure options are consistent between two versions.

tobes:
> Should rainbow have an option to turn colors off? 

please explain

lasers:
Duplicate options. One for non-rainbow, one for rainbow. http://sprunge.us/giUX

lasers:
Will running everything under rainbow override timeout_cache for all modules? If I use `rainbow` module with `cycle_time =  60`, then all time.... oh I just got an IndexError. Nice!

tobes:
> Will running everything under rainbow override timeout_cache for all modules?

They just do their thing as normal.  They can force rainbow to update but not the other way

lasers:
Just FYI. That IndexError was with old commit. Now running latest `rainbow` commit. Does your latest commit fix that error log I shared?

tobes:
It should also the output should be slightly different

lasers:
```
2017-01-16 10:51:08 WARNING Instance `rainbow`, user method `rainbow` failed (IndexError) rainbow.py line 188.
2017-01-16 10:51:08 INFO Traceback
    response = method()
  File "/home/chris/src/py3status/py3status/modules/rainbow.py", line 188, in rainbow
    color = self.colors[offset]
```

tobes:
I think you have the old version 

line 187 should be 
```
offset = (self.active_color + (index * step)) % len(self.colors)
```

lasers:
Twas `master`. Yes. Sorry. I switched branches forgetting about testing modules. 👍 
I'm not experiencing any issues with rainbow. What do you think about an option to toggle on/off color if I'm putting everything under rainbow?

tobes:
> What do you think about an option to toggle on/off color if I'm putting everything under rainbow?

It's an interesting idea but it should be made to work with all containers really so would be part of the config parsing stuff.  I'll have a think about a nice way to do it.

maybe a disable option that works for all modules

Anyhow I think the rainbow fix is good it was just an operator precedence error

lasers:
Agreed. No issues so far. I believe this is ready to go. Thanks for this. 🌈 

I cleaned my config. Everything under rainbow with 60s.
One minor issue, I think is... I hadn't seen the color yellow yet.

Just got hit with error. Looks same as the one I pasted above.

tobes:
if you do `cache_timeout = 1` it should hit it pretty quickly.

just check that line is correct.  Adding

```
self,py3.log('%s %s' % (len(self.colors), offset)
```
just before the line that errors should help track it.  That you don't see yellow makes me think you are on the none forced commit

lasers:
I changed `cache_timeout` and it happened right away. See EDIT above. Will add that log.

tobes:
also check that line is correct

lasers:
```
2017-01-16 12:07:01 INFO Module `rainbow`: 60 48
2017-01-16 12:07:01 INFO Module `rainbow`: ! HERE !
2017-01-16 12:07:01 INFO Module `rainbow`: 60 52
2017-01-16 12:07:01 INFO Module `rainbow`: ! HERE !
2017-01-16 12:07:01 INFO Module `rainbow`: 60 56
2017-01-16 12:07:01 INFO Module `rainbow`: ! HERE !
2017-01-16 12:07:01 INFO Module `rainbow`: 60 60
2017-01-16 12:07:01 WARNING Instance `rainbow`, user method `rainbow` failed (IndexError) rainbow.py line 190.
2017-01-16 12:07:01 INFO Traceback
    response = method()
  File "/home/chris/src/py3status/py3status/modules/rainbow.py", line 190, in rainbow
    color = self.colors[offset]
```

tobes:
you have the old branch still

lasers:
Okay... on hidden `015aced`. I use some aliases to pull. Does not work well with github's pull requests / non-forced commit. Will look into that. Sorry. No crash now.

tobes:
I've made multi_color default to True as it really makes the module so much more rainbow

@lasers
>: normal_color: If True, turn off colorizing the modules (default False) ?

As I've said before this is not the place to do this.

>    cycle_time... not cycle_timeout? :o

changing config options just for the sake of it is __evil__

lasers:
I commented somewhere else. What about `cache_timeout` to create consistent because the rainbow module by itself does nothing.

lasers:
red, custom, red, red, red, custom, red, red, red, adjusted volume, gpu, dpms.

ultrabug:
Silly :D

tobes:
> Silly :D

you need to work on your English  you meant __spectacular__

PR #678, opened 2017-01-17
Add swap support for sysdata module

valdur55:
Showing memory usage is not enough.


valdur55:
I will make new pull request after upgrading to new codebase.

Issue #534, opened 2016-10-28
Mpris: on_click error when no player


tobes:
If there is no player running and the user clicks on mpris an error occurs.  This only happens if no player is run between py3status starting and mpris being clicked

@ritze do you want to look at this it should be an easy fix.

```
AttributeError: Py3status instance has no attribute '_control_states'
  File "/home/toby/dev/py3status/py3status/module.py", line 401, in click_event
    click_method(event)
  File "/home/toby/dev/py3status/py3status/modules/mpris.py", line 541, in on_click
    if index not in self._control_states.keys():
```

chinmaya1011:
![image](https://user-images.githubusercontent.com/36993291/87882450-b0f61100-ca1d-11ea-8e2e-7ebba8dd6c74.png)

Still happens on the current Arch package

chinmaya1011:
Never mind, was a problem with my the service enabling mpris support for mpd. Sorry!

PR #688, opened 2017-01-21
Fix mouse clicks

ritze:
The order of checking `index` was wrong. So mouse clicks weren't recognized if `index` was `None`.

### DIFF
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -538,9 +538,6 @@ class Py3status:
         index = event['index']
         button = event['button']
 
-        if not index:
-            return
-
         if index not in self._control_states.keys():
             if button == self.button_toggle:
                 index = 'toggle'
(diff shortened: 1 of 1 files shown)


py3status/modules/mpris.py
```diff
@@ -555,9 +552,10 @@ def on_click(self, event):
         elif button != 1:
             return
 
-        control_state = self._control_states[index]
-        if self._player and self._get_button_state(control_state):
-            getattr(self._player, self._control_states[index]['action'])()
+        if index:
```
tobes: I don't think this test is needed but it does no harm.

ritze:
This condition was implemented because of bug #534

tobes:
Ok, well that bug has been reintroduced.

Really you want to do `self._control_states = {}` in both `__init__()` and `mpris()` in the `if self._player is None:` section.

Also use `control_state = self._control_states.get(index)` instead of `control_state = self._control_states[index]` or you'll still hit errors.

ritze:
Of cause you're right. I don't know what I thought. Apparently nothing…

ritze:
Can the access on `_control_states` with  the function `.keys()` crash if _control_states is empty?

tobes:
` {}.keys() == []`  I'll look through it again sometimes these things are hard to spot.  Thanks for the hard work.

tobes:
@ritze Sorry for the delay in reviewing this

ultrabug:
Thanks a lot guys :)

Issue #665, opened 2017-01-14
Not able to change i3-nagbar font as calls to it are hardcoded


randomguy:
A quick grep through the py3status source show that there currently is no way of setting a font for calls to i3-nagbar. I see no trivial way to add a configuration option as the config file has to be compatible to i3status. Maybe it could be added as an command line option to py3status? I'll gladly try my hands on it as soon as I have some feedback as to whether such a feature is desired.

https://github.com/ultrabug/py3status/blob/master/py3status/core.py#L411

tobes:
In theory not a problem as seems useful.

Getting the info is more of an issue.

I'd like to allow this sort of stuff via the config

```
general {
    i3_nagbar_font = ....
}

# or maybe a 
py3status {
    ...
}
```

@ultrabug what are your thoughts, adding command line options for something like this just seems overkill

randomguy:
> @ultrabug what are your thoughts, adding command line options for something like this just seems overkill

I agree, the config would be the appropriate place.

PR #690, opened 2017-01-23
Add ability to set i3-nagbar font

randomguy:
This is a proposal to resolve issue #665.

A section named 'custom' has been added as a valid py3status configuration section. If it contains a key named 'nagbar_font', its value will be used as an argument to the i3-nagbar -f option.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -32,6 +32,7 @@ DBUS_LEVELS = {'error': 'critical', 'warning': 'normal', 'info': 'low', }
 CONFIG_SPECIAL_SECTIONS = [
     '.group_extras',
     '.module_groups',
+    'custom'
     'general',
     'i3s_modules',
     'on_click',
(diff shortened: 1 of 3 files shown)


tobes:
Looks nice and clean.  My only thought is about the section name.  I think I'd prefer `py3status` over `custom`, maybe even leave it in general but that probably causes more work due to i3status also using that section.

randomguy:
You are right, custom was to generic. I adjusted it and added some documentation.

randomguy:
So how is the documentation now? Any more feedback?

py3status/parse_config.py
```diff
@@ -563,6 +563,8 @@ def parse_config(config):
         general_defaults.update(config_info['general'])
     config['general'] = general_defaults
 
+    if 'py3status' in config_info:
+        config['py3status'] = config_info['py3status']
```
tobes: How about just

`config['py3status'] = config_info.get('py3status', {})`

that way config['py3status'] will always exist and be a dict

py3status/core.py
```diff
@@ -408,7 +409,12 @@ def notify_user(self, msg, level='error', rate_limit=None, module_name=''):
                 cmd = ['notify-send', '-u', DBUS_LEVELS.get(level, 'normal'),
                        '-t', '10000', 'py3status', msg]
             else:
-                cmd = ['i3-nagbar', '-m', msg, '-t', level]
+                config = self.i3status_thread.config
+                nagbar_font = config.get('py3status').get('nagbar_font')
```
tobes: `nagbar_font = config.get('py3status').get('nagbar_font')`

will fail if `config['py3status']` is `None`. `nagbar_font = config.get('py3status', {}).get('nagbar_font')` would be fine but my other comment fixes this issue

randomguy:
Changed as proposed.

tobes:
@ultrabug I think that adding the new `py3status` block to the config is a good plan.  I can see it having other uses.

tobes:
@ultrabug I think the `pystatus` section would be really helpful for other settings too.  The nagbar font is a bonus.

ultrabug:
Interesting point, I agree that it can be useful, thanks a lot @randomguy !!

PR #693, opened 2017-01-25
New module: tor_rate

fmorgner:
The tor_rate module shows the incoming and outgoing data rates of a Tor
daemon instance.

### DIFF
--- a/py3status/modules/tor_rate.py
+++ b/py3status/modules/tor_rate.py
@@ -71,7 +71,7 @@ class Py3status:
             try:
                 self._register_event_handler()
             except ProtocolError:
-                text = 'Error: Failed to register even handler!'
+                text = 'Error: Failed to register event handler!'
             except SocketError:
                 text = 'Error: Failed to establish control connection!'
             except AuthenticationFailure:
--- a/py3status/modules/tor_rate.py
+++ b/py3status/modules/tor_rate.py
@@ -1,0 +1,132 @@
+# -*- coding: utf-8 -*-
+"""
+Display the current transfer rates of a tor instance
+
+Configuration parameters:
+    cache_timeout: An integer specifying the cache life-time of the modules
+        output in seconds (default 2)
+    control_address: The address on which the Tor daemon listens for control
+        connections (default "127.0.0.1")
+    control_password: The password to use for the Tor control connection
+        (default None)
+    control_port: The port on which the Tor daemon listens for control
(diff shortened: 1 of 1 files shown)


py3status/modules/tor_rate.py
```diff
@@ -0,0 +1,132 @@
...
+    control_password: The password to use for the Tor control connection
+        (default None)
+    control_port: The port on which the Tor daemon listens for control
+        connections (default 9051)
+    format: A string describing the output format for the module
+        (default "↑ {up} ↓ {down}")
+    rate_unit: The unit to use for the transfer rates
+        (default "B/s")
+    si_units: A boolean value selecting whether or not to use SI units
+        (default 0)
```
tobes: `False` would be a better default

py3status/modules/tor_rate.py
```diff
@@ -0,0 +1,132 @@
...
+    value_format: A string describing how to format the transfer rates
+        (default "[\?min_length=12 {rate:.1f} {unit}]")
```
tobes: `format_value` would be a better name.  You should document the placeholders for this format too.

py3status/modules/tor_rate.py
```diff
@@ -0,0 +1,132 @@
...
+
+Example:
+
+```
+tor_rate {
+    cache_timeout = 10
+    format = "IN: {down} | OUT: {up}"
+    control_port = 1337
+    control_password = "TertiaryAdjunctOfUnimatrix01"
+    si_units = 1
```
tobes: `si_units = True` yes we can use `0` and `1` but a boolean looks nicer.

py3status/modules/tor_rate.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self):
+        self._down = 0
+        self._handler_active = False
+        self._up = 0
+
+    def tor_rate(self, outputs, config):
+        text = ''
+        if not self._handler_active:
+            try:
+                self._register_event_handler()
```
tobes: is this something we want to keep doing if it fails?  for some it feels ok eg no connection, but for authentication fail should we just give up?

tobes:
I haven't been able to test this but it looks good to me.

@fmorgner thanks for this

ultrabug:
Thanks a lot for your contribution @fmorgner ! :)

PR #698, opened 2017-01-27
Allow modules with multiple files inside folders

randomguy:
With a few changes it is possible to have modules inside subfolders allowing multiple files in this subfolder. Docstrings are working. Is a feature like this desired? An example module can be found in the refactor-simple-pomodoro branch in my fork.

### DIFF
--- a/py3status/docstrings.py
+++ b/py3status/docstrings.py
@@ -3,6 +3,7 @@ import ast
 import re
 import os.path
 import difflib
+from glob import glob
 
 from py3status.helpers import print_stderr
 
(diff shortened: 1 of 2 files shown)


tobes:
This is an interesting idea.  I quite like the flat all the modules are in one directory approach that we currently have but we are now at ~75 so maybe we need to start thinking about splitting them.  I also like the idea of categorizing the modules and this would help there.

tobes:
> With a few changes it is possible to have modules inside subfolders allowing multiple files in this subfolder. 

Maybe I misunderstood before but I looked at https://github.com/randomguy/py3status/tree/refactor-simple-pomodoro/py3status/modules/simple_pomodoro

Personally I think it is much cleaner/clearer if modules are self contained and a single file.  In your simple_pomodoro module I'm not sure you gain very much by splitting the module into pieces.  Modules should really be fairly simple things and so I'm not sure that splitting will ever really be needed. 

However I do like the idea that maybe we should consider splitting the modules up by type at some point in the future.  But that's not really what this PR is and I think we should consider it when/if we hit the 100 module mark

randomguy:
After letting it sit for a while I too come to the conclusion that it doesn't help much. It is mostly a subjective taste thing, I guess. Thanks for the feedback and feel free to close this or whatever.

tobes:
@randomguy Thanks for this but yeah I think for now it is not something we need,  I may steal some of the code at a later date but it's good to clean the PR queue

PR #733, opened 2017-02-03
allow check_commands() to take a string

tobes:
`py3.check_commands()` takes a list of commands.  It is easy to pass a string instead.  this PR allows us to pass a string and it will be treated as expected.  It also makes it cleaner to just check for a single command.

`py3.check_commands(['foo'])` ->  `py3.check_commands('foo')`


ultrabug:
Good point, thanks a lot !

PR #737, opened 2017-02-08
Do not treat workspace name as window name.

valdur55:
It happened when you focused empty workspace and then reloaded i3. (always_show is true)


tobes:
cool, thanks @valdur55

ultrabug:
Thanks @valdur55 !

PR #739, opened 2017-02-09
Cleanup i3status config access

tobes:
This PR just cleans up the `py3_config` this is what I'm calling our parsed config file.  I'm easy on the name.  The aim it to make it part of the py3status config currently it is I3status and we keep having to reach into it.

As well as improving clarity it also means that the i3status stuff running becomes optional rather than being vital We can probably remove the whole mock thread etc.


ultrabug:
It makes sense indeed, thanks mate

ultrabug:
Tested OK

PR #741, opened 2017-02-11
allow improved access to settings from py3status section

tobes:
This PR improves support for the py3status section by making it available via `Py3statusWrapper.get_config_attribute()`.  This means that we can easily get at the settings and inherit first from `py3status` and then `general`.  This is really useful for many features eg the popup font could be set at the application or module level.

So it's not about giving modules access but for global features.

```
general {
    color_good = '#00FF00'
}

py3status {
    color_good = '#00AA00'  # overrides general setting for py3status modules
    nagbar_font = 'pango:Ubuntu Mono 12'  # global setting
    some_new_setting = 'wow'  # module specific setting
}

my_module {
    color_good = '#00CC00'
    some_new_setting = None  # overrides
}
```

Edit:  I did a bit more playing with this.  Adding `py3._get_config_setting(name)`. It adds a bit of noise but we make things more general (less color_xx specific) and moves some stuff into better places.

Again, this does not add much direct functionality, more the capacity to do new things that I have plans for.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -96,14 +96,21 @@ imap {
 ```
 
 #### <a name="py3status_section"></a>py3status configuration section
-This special section holds py3status specific configuration. As of now, the
-only possible key is 'nagbar_font'. It will be used as an argument to
-`i3-nagbar -f`, thus setting its font. Example usage:
-```
-py3status {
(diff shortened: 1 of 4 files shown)


ultrabug:
I like the implementation & idea, thanks a lot

PR #742, opened 2017-02-11
Clock plugin: Add %s, %T for time_delta calculation

ShareDVI:
Strftime formats %s and %T include number of seconds, but time_delta is equal to 60, which leads to the seconds counter to be stuck.


tobes:
@ShareDVI Thanks, nice bugfix

ultrabug:
🎩 thanks @ShareDVI !

PR #743, opened 2017-02-11
Diskdata: add clarification to docs

ShareDVI:
Add clarification to docs, that disk /dev name, and not mountpoint, is required


tobes:
Thanks,

Yeah, looking at this the documentation could be clearer.

`modules/README.md` is an auto-generated file the correct place to fix this is the modules docstring https://github.com/ultrabug/py3status/blob/master/py3status/modules/diskdata.py#L8

tobes:
@ShareDVI this fix is has been added via PR #753

PR #758, opened 2017-02-18
support unicode wifi SSID

blahgeek:
For unicode SSID name, iw will output: `SSID: \xe5\xaf\x8c\xe5\xbc\xba\xe5\xba\xb7\xe4\xb9\x90`, it should be properly decoded.

### DIFF
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -125,7 +125,7 @@ class Py3status:
         if ssid_out:
             ssid = ssid_out.group(1)
             ssid = ssid.encode('latin-1').decode('unicode_escape')
-            ssid = ssid.encode('latin-1').decode()
+            ssid = ssid.encode('latin-1').decode('utf-8')
         else:
             ssid = None
 
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -124,6 +124,8 @@ class Py3status:
         ssid_out = re.search('SSID: (.+)', iw)
         if ssid_out:
             ssid = ssid_out.group(1)
+            ssid = ssid.encode('latin-1').decode('unicode_escape')
+            ssid = ssid.encode('latin-1').decode()
         else:
             ssid = None
 
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -124,6 +124,9 @@ class Py3status:
         ssid_out = re.search('SSID: (.+)', iw)
         if ssid_out:
             ssid = ssid_out.group(1)
+            # `iw` command would prints unicode SSID like `\xe8\x8b\x9f`
+            # the the `ssid` here would be '\\xe8\\x8b\\x9f' (note the escape)
+            # it needs to be decoded using 'unicode_escape', to '苟'
             ssid = ssid.encode('latin-1').decode('unicode_escape')
             ssid = ssid.encode('latin-1').decode('utf-8')
         else:


tobes:
@blahgeek thanks for reporting this problem

are you expecting output like `SSID: 富强康乐`

could you paste the output of `py3status -v` to help me reproduce the issue.  I think you want utf-8 not latin-1 but this should be being done automatically.

blahgeek:
My versions:

```
$ py3status -v
py3status version 3.4 (python 3.6.0)
$ iw --version
iw version 4.9
```

My `iw dev wlp2s0 link` would output:

```
Connected to 78:d3:8d:de:75:50 (on wlp2s0)
	SSID: \xe5\xaf\x8c\xe5\xbc\xba\xe5\xba\xb7\xe4\xb9\x90
	freq: 2432
        ....
```

So py3status would display `\xe5\xaf\x8c\xe5\xbc\xba\xe5\xba\xb7\xe4\xb9\x90` as SSID. I would expect `富强康乐` so it needs decoding.

tobes:
Your patch does not work with python2 but I confirmed the issue by faking your iw output.  I'm not sure why the encoding is `latin-1` rather than `utf-8` I'd need to test by setting up a wifi access point to test more.

this patch works for me in both python2/3 
```
diff --git a/py3status/modules/wifi.py b/py3status/modules/wifi.py
index fc9fac4..cd562ca 100644
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -124,6 +124,8 @@ class Py3status:
         ssid_out = re.search('SSID: (.+)', iw)
         if ssid_out:
             ssid = ssid_out.group(1)
+            if not self.py3.is_python_2():
+                ssid = ssid.encode('latin-1').decode()
         else:
             ssid = None
 
index d5ed51e..3914613 100644
--- a/py3status/py3.py
+++ b/py3status/py3.py
```


could you test if it works for you?  `python2 py3status/modules/wifi.py` etc

you may need to change the test (at the bottom of the module) like this to pickup your device, The auto device selector code is broken it seems

```
if __name__ == "__main__":
    """
    Run module in test mode.
    """
    config = {'device': 'wlp2s0'}
    module_test(Py3status, config)
```

the output I get is here.  The `?` is due to my faked iw output

```
$ python2 py3status/modules/wifi.py
{'cache_until': 1487436549.0, 'color': '#00FF00', 'full_text': u'W: ? MBit/s ?% \u5bcc\u5f3a\u5eb7\u4e50'}

$ python3 py3status/modules/wifi.py
{'color': '#00FF00', 'cache_until': 1487436542.0, 'full_text': 'W: ? MBit/s ?% 富强康乐'}
```

blahgeek:
Hi @tobes I pushed a new commit that should work in Python 2 (in python3, `decode()` is equal to `decode('utf8')`)

I think you may got me wrong. The output of `iw dev wlp2s0 link` is LITERALLY `\xe5\xaf\x8c\xe5\xbc\xba\xe5\xba\xb7\xe4\xb9\x90`, and in python, the `ssid` variable is `"\\xe5\\xaf\\x8c\\xe5\\xbc\\xba\\xe5\\xba\\xb7\\xe4\\xb9\\x90"` (note the escapes).

So it needs to be decoded using `unicode_escape`, the `latin-1` is only used to convert `str` to `bytes` in python.

blahgeek:
See the codes below:

tobes:
@blahgeek Thanks for the explanation.  I've now done some testing and so can confirm the fix.

Could you add a comment to just explain why we are doing this. Something like
```
# unicode characters need to be converted
# e.g.  \xe5\xaf\x8c should be converted to富
```

blahgeek:
Done!

tobes:
@blahgeek Thank you for all your work on this.  Full unicode support is important to us.

PR #766, opened 2017-02-20
universal gradients

tobes:
This is something that i3pystatus had that I though was a nice feature.

Instead of using sharp color changes on thresholds ie either good/degraded/bad we allow gradients between the colors.

Gradients are disabled by default to enable add `gradients = True` to `py3status` section or to individual module in config.

Any modules that use thresholds will then work.

The code creating the gradients is a little as we need to use hue, saturation, value colorspace rather than rgb so that we can have really good gradients. but it is self-contained.

There may be issues if modules have very small threshold ranges eg `[(0, 'good'), (0.5, 'bad')]` but I don't think modules currently do this and we can always fix if we have some.

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -14,6 +14,8 @@ from time import time
 from py3status import exceptions
 from py3status.formatter import Formatter, Composite
 from py3status.request import HttpResponse
+from py3status.util import Gradiants
+
 
 PY3_CACHE_FOREVER = -1
 PY3_LOG_ERROR = 'error'
@@ -85,6 +87,7 @@ class Py3:
 
     # Shared by all Py3 Instances
     _formatter = None
+    _gradients = Gradiants()
     _none_color = NoneColor()
 
     # Exceptions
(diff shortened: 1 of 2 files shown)


py3status/py3.py
```diff
@@ -86,6 +88,7 @@ class Py3:
     # Shared by all Py3 Instances
     _formatter = None
     _none_color = NoneColor()
+    _gradients = Gradiants()
```
ultrabug: bad kitty 🐈

py3status/py3.py
```diff
@@ -101,9 +104,10 @@ def __init__(self, module=None, i3s_config=None, py3status=None):
         self._format_placeholders = {}
         self._format_placeholders_cache = {}
         self._i3s_config = i3s_config or {}
-        self._module = module
         self._is_python_2 = sys.version_info < (3, 0)
+        self._module = module
         self._report_exception_cache = set()
+        self._threshold_gradients = {}
```
ultrabug: 🐈

py3status/py3.py
```diff
@@ -949,13 +956,34 @@ def threshold_get_color(self, value, name=None):
...
-                    break
+        thresholds = self._thresholds.get(name_used)
+        if color is None and thresholds:
+            # if gradients are enabled then we use them
+            if self._get_config_setting('gradients'):
+                try:
+                    colors, minimum, maximum = self._threshold_gradients[name_used]
+                except KeyError:
+                    colors = self._gradients.make_threshold_gradient(self, thresholds)
+
```
ultrabug: useless empty line

tobes:
updated.

ultrabug:
NOTE: this should be documented with an example ;)

PR #767, opened 2017-02-20
Py3.format_placeholders() helper

tobes:
This adds a helper to get all the placeholders from a format_string.

It is helpful for a number of potential modules.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1005,6 +1005,22 @@ This is useful because a simple test like
 will fail if the format string contains placeholder formatting
 eg `'{placeholder:.2f}'`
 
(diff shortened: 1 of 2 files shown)


ultrabug:
The method name is not clear to me, shouldn't it be named `get_placeholders_list` or something instead ?

tobes:
is `format_list_placeholders()` acceptable? If not I'll go with your suggestion.

ultrabug:
Well I may be stubborn but this looks like a getter function to me, so I'd stick with my previous statement sorry

tobes:
stubborn wins :)

PR #775, opened 2017-02-25
Prevent an attribute error

### LINKED ISSUE #774: group cycle/click events fail
Since happy-endings we get an AttributeError L154 of events.py

tobes:
this should fix #774 

I'm not sure if there is also another issue related to clicks but I need to test more first.


ultrabug:
Thanks !

tobes:
Always a pleasure.

We need to work on our test coverage :)

PR #776, opened 2017-02-26
[diskstats module] Fix stats for fs mounted multiple times

eturner:
When the same block device has multiple lines in df output, like btrfs subvolumes, the disk usage is multiplied.
There's probably a better way to do this, but not without rewriting it.

### DIFF
--- a/py3status/modules/diskdata.py
+++ b/py3status/modules/diskdata.py
@@ -143,6 +143,7 @@ class Py3status:
                 data = line.split()
                 if data[0] in devs:
                     # Make sure to count each block device only one time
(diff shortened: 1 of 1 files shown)


tobes:
@eturner thanks for this.

Could you add a comment explaining the duplicates issue so it's clear what we are doing.

eturner:
Glad to help.
Does that suffice? I haven't gotten the hang of this yet.

tobes:
> Does that suffice? 

I amended it slightly

> I haven't gotten the hang of this yet.

You seem to be doing ok, and once again thanks for the patch

PR #778, opened 2017-03-03
Py3 provide http request functionality

tobes:
Seeing as #522 seems to have stalled here is an alternative.

it adds `py3.request()` which returns a simple object that can provide the http status code, text and json.

It allows basic authentication.
request is POST if data is supplied otherwise it will be a GET
query string params can be given in url or as a dict
headers can be supplied.
timeout can be given

I did add proxy support but it ended up getting so complicated (both as code but also testing - https is not supported well) so I removed it all.  For complex things we should just use requests (I'd like to eventually do that via this method so that for example proxies would be usable for all requests)

I've updated all the modules I was able to test.

If it seems good I'll update the documentation

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -10,8 +10,9 @@ from pprint import pformat
 from subprocess import Popen, PIPE
 from time import time
 
+from py3status import exceptions
 from py3status.formatter import Formatter, Composite
-
+from py3status.request import HttpResponse
 
 PY3_CACHE_FOREVER = -1
 PY3_LOG_ERROR = 'error'
@@ -66,6 +67,12 @@ class Py3:
     # Shared by all Py3 Instances
     _formatter = Formatter()
     _none_color = NoneColor()
+
+    # Exceptions
+    Py3Exception = exceptions.Py3Exception
+    RequestException = exceptions.RequestException
+    RequestTimeout = exceptions.RequestTimeout
+    RequestURLError = exceptions.RequestURLError
 
     def __init__(self, module=None, i3s_config=None, py3status=None):
         self._audio = None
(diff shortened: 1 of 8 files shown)


py3status/request.py
```diff
@@ -0,0 +1,97 @@
...
+            return self._text
+        except AttributeError:
+            if IS_PYTHON_3:
+                encoding = self._response.headers.get_content_charset('utf-8')
+            else:
+                encoding = self._response.headers.getparam('charset')
+            self._text = self._response.read().decode(encoding)
+        return self._text
+
+    def json(self):
```
ultrabug: why is text a property while json is not ?

py3status/py3.py
```diff
@@ -787,3 +794,39 @@ def threshold_get_color(self, value, name=None):
...
+        # The aim of this function is to be a lmited lightweight replacement
+        # for the requests library but using only pythons standard libs.
+
+        # IMPORTANT NOTICE
+        # This function is excluded from private variable hiding as it is
+        # likely to need api keys etc which people may have obfuscated.
+        # Therefore it is important that no logging is done in this function
+        # that might reveal this information.
+
+        #  py3 = self
```
ultrabug: remove me ?

ultrabug:
A long time discussed need, thanks a lot for working on this !

py3status/request.py
tobes: it is the same way that requests does things.  It also meant less changes to modules.  I'm happy to change it if you'd prefer a property

py3status/py3.py
tobes: sorry it is from my debugging

py3status/request.py
ultrabug: okay makes sense

ultrabug:
Thanks a lot !

PR #788, opened 2017-03-05
netdata: optimization no.2

lasers:
I have nothing to say.


tobes:
@lasers nice bit of cleanup

PR #790, opened 2017-03-06
improve docstring tests to prevent reserved config parameters

tobes:
Some configuration parameters have special meaning in the py3status config. This is mainly because they have meaning to i3bar.

`min_width`
`separator`
`separator_block_width`
`align`

This PR adds tests to prevent them being used in modules.

Currently `group` uses `align` and I can't think how to easily remove it so I gave it a free pass

### DIFF
--- a/tests/test_module_doc.py
+++ b/tests/test_module_doc.py
@@ -21,6 +21,17 @@ from py3status.docstrings import core_module_docstrings
 IGNORE_MODULE = [
 ]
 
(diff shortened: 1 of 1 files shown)


ultrabug:
This is pure favouritism over your group baby 😱

tobes:
> pure favouritism

True, anyone else's baby would have been thrown on the fire 

👶 ➡️ 🔥 = 💀

PR #794, opened 2017-03-10
add -- after pactl call

DextersHub:
If the volume is preceded by a '-' pactl will raise the following error:
'pactl: invalid option -- '

We need to pass '--' right after pactl:

$ pactl -- set-sink-volume -5%

Otherwise pactl tries to parse the number as an option, which it isn't. 

reference: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=686667


tobes:
@DextersHub thank you for this contribution.

PR #796, opened 2017-03-11
fix weird behavior when disabling button_up/button_down

Tadly:
When `button_up` and `button_down` are both disabled, there's now reason to give`self._set_backlight_level(level)` a call.
This just causes weird side-effects.
e.g. for me, even though I disabled both, the brightness **always** decreases no matter the direction I scroll in.
Having set `on_click 2 = ...` combined with two-finger scrolling on a touchpad makes this really awkward.

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -73,6 +73,9 @@ class Py3status:
         if not self.xbacklight:
             return None
 
+        if not self.button_up and not self.button_down:
+            return None
+
         level = self._get_backlight_level()
         button = event['button']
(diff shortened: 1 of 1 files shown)


tobes:
@Tadly thanks for this PR.

Having looked at the issue I've created #797 that fixes the more fundamental problem that we are getting the wrong brightness, which is causing your dimming issue.  If you could test that that PR fixes it for you that would be great, it should allow you to change the brightness as expected too - if enabled ;)

I'm going to close this PR because of that fix, but this has been very helpful in getting it sorted.

PR #797, opened 2017-03-12
backlight fixes

tobes:
This is in response to PR #796 by @Tadly

It does 5 things

1. only calls `_set_backlight_level()` when needed rather than on every event.

2. if we have `xbacklight` installed we use it to get the brightness rather than `/sys/class/backlight`.  THe reason for this is that we can be looking at the wrong device eg `intel_backlight` rather than `acpi_video0` which causes problems.

3. The `brightness_delta` changed from `5` to `8` the reason for this is that some displays only have 15 brightness levels and so misbehave when the delta is 5

4. slightly simplifies the button checks, as there is no need to check for `self.button_up` if it is `None`/`0` then it will not have the value `event['button']`

5. calls `xbacklight -set` with option `-time 0` this causes the update to be immediate and makes the module much more responsive.

It might change the output slightly but this is probably the correct behaviour.

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -4,7 +4,7 @@ Adjust screen backlight brightness.
 
 Configuration parameters:
     brightness_delta: Change the brightness by this step.
-        (default 5)
+        (default 8)
     brightness_initial: Set brightness to this value on start.
         (default None)
     brightness_minimal: Don't go below this brightness to avoid black screen
@@ -50,7 +50,7 @@ class Py3status:
     """
     """
     # available configuration parameters
-    brightness_delta = 5
+    brightness_delta = 8
     brightness_initial = None
     brightness_minimal = 1
     button_down = 5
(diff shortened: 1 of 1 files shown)


Tadly:
Looks good to me, thanks for fixing it properly :)

Scrolling with a touchpad still is quite brutal but that's a different issue
(amount of scroll events + the time it takes to change the level for one event)

tobes:
> Scrolling with a touchpad still is quite brutal but that's a different issue

Yeah I agree the following patch makes things more responsive
```
diff --git a/py3status/modules/backlight.py b/py3status/modules/backlight.py
index 697cffc..a6f758b 100644
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -87,7 +87,7 @@ class Py3status:
             self._set_backlight_level(level)
 
     def _set_backlight_level(self, level):
-        self.py3.command_run(['xbacklight', '-set', str(level)])
+        self.py3.command_run(['xbacklight', '-time', '0', '-set', str(level)])
 
     def _get_backlight_level(self):
         if self.xbacklight:
```

part of the problem is that the scroll wheel does lots of quick fire events.  If you set the up/down buttons to ones you can have greater control with eg left/right click it will be easier.

@ultrabug how do you feel about the patch?

Tadly:
@tobes didn't know xbacklight allowes to pass a `-time` option. Didn't even bother checking as I expected it to be as fast as possible from the get-go (seems kinda silly to me it doesn't but I'm sure there's a reason for it)

The scroll wheel send so many events is what I expected.
In fact, using e.g. `libinput-debug-events` shows just how many there are.

In other words, the patch supplied would be nice to have (works great and finally feels responsive) :)

tobes:
> the patch supplied would be nice

I've added it to the PR

Tadly:
@tobes while configuring my setup I just realized that `brightness_delta = 1` means you can increase the brightness level anymore via `button_up`.
Haven't had time to investigate yet but wanted to let you know either way.

ultrabug:
Thanks a lot for taking the time to do this, I was not happy with it either (being one of this module's users), and I'm very happy with the new responsiveness

tobes:
> I just realized that brightness_delta = 1 means you can increase the brightness level anymore via button_up

do you mean you **cannot** change the brightness.

If so it is because the number of brightness steps supported, hence me setting the default to 8.  We could probably make this so that it is detected (via xrandr)

Tadly:
>do you mean you cannot change the brightness

sort of. down by one works just fine, only up doesn't.
Using a delta `>2` obviously solves the problem

But I feel like I'm somewhat nitpicking right now. Just wanted to mention it :)

ultrabug:
Just tested, you're right indeed with delta 1 the brightness counter only goes down and never up.

tobes:
I suspect this is a rounding issue as the brightness will be say 73.333333 and we then cast as an `int`.  So we could probably fix this too if we care enough.

Tadly:
Hm... messing with the way we round up/down probably will cause other issues.
What about using `xbacklight -inc/-dec <value>` rather than `-set`?

PR #798, opened 2017-03-12
skip empty modules in group

tobes:
Fix for #779

when the `group` module shows empty modules it probably should not.  In the default mode you cannot change to the next module.  When an empty group is detected we try the 'next' module.

We make sure we are not endlessly cycling and if there is no module with output we try again every 5 seconds and switch if one has content.

Although this changes the behaviour slightly I think the improvement makes sense.

### DIFF
--- a/py3status/modules/group.py
+++ b/py3status/modules/group.py
@@ -82,6 +82,8 @@ group disks {
 
 from time import time
 
+RETRY_TIMEOUT_NO_CONTENT = 5
+
 
 class Py3status:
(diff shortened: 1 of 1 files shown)


ultrabug:
It's nice, my only comment will be that I would **explicitly** pass the delta and remove the default of 1

Thanks a lot

tobes:
ok I'll fix up

ultrabug:
Cool thx !

PR #799, opened 2017-03-12
show urgent on closed containers

tobes:
This is a fix for #750 

There are lots of options but this one feels the simplest and least user invasive.

When a container is closed (so we don't see the urgent module) we just make the 'closed' container urgent.  In practical terms this means that the button will show as urgent.  The user can the decide if they wish to open the container.

This prevents the need for multiple options for what behaviour we want.  This feel a sensible default.

Currently we do not care if the module stops being urgent.  This is quite simple but requires `Module.urgent` to be exposed to modules which it is not currently.  This would trivial but then makes this PR overly complex.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -352,6 +352,7 @@ class Module(Thread):
 
         # update all components
         color = response.get('color')
+        urgent = response.get('urgent')
         for index, item in enumerate(response['composite']):
             # validate the response
             if 'full_text' not in item:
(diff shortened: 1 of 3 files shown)


ultrabug:
What is that new option you're talking about here ? From a quick review it seems you just honor and pass along any urgent response flag set by a module. Am I missing something ?

tobes:
> What is that new option you're talking about here ? 

I meant as in there are choices of how we could implement this eg auto open the container.  The whole point is **not** to add any user options

ultrabug:
Okay then it's clear, thanks !

PR #802, opened 2017-03-13
group/frame: use safe_format() not build_composite

tobes:
These modules were using the old way of dealing with composites.  Changed them to use the preferred methods.


ultrabug:
🐈 bad kitty, but okay anyway

PR #803, opened 2017-03-13
TRIVIAL remove some un-needed ignored parameters

tobes:
When checking module config params we ignore some specific ones.  These are no longer needed.


ultrabug:
Good spot

ultrabug:
Good spot !

PR #810, opened 2017-03-14
documentation IV (autodoc modules)

tobes:
Introducing `autodoc.py` this is used to auto create documentation.

It can be run independently via `python autodoc.py`

In this PR it learns about the modules and creates `doc/modules-info.inc` documenting modules

### DIFF
--- a/py3status/autodoc.py
+++ b/py3status/autodoc.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+from py3status.docstrings import core_module_docstrings
+
+
+def markdown_2_rst(lines):
+    """
+    Convert markdown to restructured text
+    """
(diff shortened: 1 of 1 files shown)


py3status/autodoc.py
```diff
@@ -0,0 +1,57 @@
...
+                out.append('\n')
+        else:
+            if code and line.strip():
+                line = '    ' + line
+            out.append(line)
+    return out
+
+
+def create_module_docs():
+    '''
```
ultrabug: " plz, not '

tobes: Updated.  I should write some tests to enforce your harsh standards

ultrabug:
Thanks !

PR #811, opened 2017-03-14
documentation V (sphinx)

tobes:
This PR adds the sphinx setup

it currently builds the documentation base, output here http://py3status.readthedocs.io

build instructions
```
pip install sphinx

cd doc
make html
```
output is in `doc/_build/html`

Mostly it is sphinx boiler plate however we have a few lines of custom stuff at the end of `doc/conf.py` so that `version` and `release` will be auto-generated


ultrabug:
Great, thanks !

PR #812, opened 2017-03-14
documentation VI (general pages)

tobes:
The main hand written documentation.

Most of this is just the current pages converted to .rst

`intro.rst` is probably the only one to really care about as it is 'new' stuff.

I think we can improve the docs from this base.

as with the documentation I-V this is independent of the other PRs


ultrabug:
intro is good, it's a great start to improve the docs afterwards

PR #813, opened 2017-03-14
documentation VII (autodoc py3)

tobes:
Ok,

So this one is the worst of this entire series.

What we do is look in Py3 and get all methods, exceptions, and constants and create a documentation include file for each.

We get the methods via looking in Py3, we need to use `inspect` to get the calling parameters

Exceptions we also get via looking in Py3

For constants we need to look at the ast for the file so we can find 'docstrings' for the constants.

A couple of methods log() and notify_user() have constants for defaults so we need `CONSTANT_PARAMS` to allow us to map this.

Hopefully it is broken out enough to be reviewable without too much pain.

### DIFF
--- a/py3status/autodoc.py
+++ b/py3status/autodoc.py
@@ -1,6 +1,17 @@
 # -*- coding: utf-8 -*-
 
+import ast
+import inspect
+
 from py3status.docstrings import core_module_docstrings
+from py3status.py3 import Py3
+
+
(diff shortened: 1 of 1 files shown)


ultrabug:
I get it, this is crazy but I guess it's what we should do for now.

Obviously we'll need to go for proper docstrings to have sphinx pick them up

PR #817, opened 2017-03-18
[whatismyip] Add `countryCode`, use COLOR_GOOD always when online

maximbaz:
Adjustments to the `whatismyip` module:

- Add `countryCode` format, as the whole country name is too large for my laptop screen 😉
- Ensure the COLOR_GOOD is always used when online, not only in one of the two possible states (indentation typo?). In my setup I rely a lot on colors, so all non-good colors quickly stand out.

### DIFF
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -4,7 +4,7 @@ Display public IP address and online status.
 
 Configuration parameters:
     cache_timeout: how often we refresh this module in seconds (default 30)
-    format: available placeholders are {ip} and {country}
+    format: available placeholders are {ip}, {country} and {countryCode}
             (default '{ip}')
     format_offline: what to display when offline (default '■')
     format_online: what to display when online (default '●')
@@ -20,6 +20,7 @@ Configuration parameters:
 
 Format placeholders:
     {country} display the country
+    {countryCode} display the country code
     {ip} display current ip address
 
 Color options:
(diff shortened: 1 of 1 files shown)


py3status/modules/whatismyip.py
```diff
@@ -20,6 +20,7 @@
 
 Format placeholders:
     {country} display the country
+    {countryCode} display the country code
```
tobes: how about `{country_code}`?  This is how we seem to do this rather than camelCase

maximbaz:
hey @tobes, thanks for looking into this! 

I actually have a bigger idea for this module, instead of allowing only `ip`, `country` and `countryCode` I want to allow any format, that is available in the JSON of `url_geo`. The current code is very bound to the specific format of `http://ip-api.com/json` (where `ip` is available under the `query` key, `countryCode` is camel case, etc), what I want instead is to allow using any URL and using any format that is available in the JSON of that url.

For example, I found out that `http://ip-api.com/json` have DNS issues and it is not always available for me, but the alternative site `https://freegeoip.net/json/` works great. However, they have different format keys than `http://ip-api.com/json`:

- `ip` instead of `query`
- `country_name` instead of `country`
- `country_code` instead of `countryCode`

So really even though currently plugin allows using different URLs, you can't use any other URL because `query`, `country` and now `countryCode` are hardcoded. So I want to allow any format options.

What do you think of this? I can send the PR with this idea soon.

maximbaz:
@tobes let me know what you think of #819, if you like it I suggest we close this PR in favor of that one.

maximbaz:
We clearly went for the #819, so I'm closing this PR as obsolete.

PR #819, opened 2017-03-18
[whatismyip] Support any format placeholder, always use `url_geo` with a better default

maximbaz:
A better way to approach #817.

Adjustments to the `whatismyip` module:

- Support any format placeholder that is available in JSON response of `url_geo`, not only hardcoded ones.
- Use a better default value for `url_geo`, which doesn't have DNS issues.
- Remove `url` and always use `url_geo` to simplify the code.
- Ensure the COLOR_GOOD is always used when online.

@tobes please have a look and let me know what you think, I'll test more later today and fix lint issues if they occur.

### DIFF
--- a/py3status/modules/getjson.py
+++ b/py3status/modules/getjson.py
@@ -94,7 +94,7 @@ class Py3status:
             resp = self.py3.request(self.url, timeout=self.timeout)
             status = resp.status_code == 200
             resp = resp.json()
-        except self.Py3.RequestException:
+        except self.py3.RequestException:
             resp = None
             status = False
         return resp, status
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -4,7 +4,7 @@ Display public IP address and online status.
 
 Configuration parameters:
     cache_timeout: how often we refresh this module in seconds (default 30)
-    format: available placeholders are {ip} and {country}
+    format: any key in JSON fetched from `url_geo` is available as a placeholder
             (default '{ip}')
     format_offline: what to display when offline (default '■')
     format_online: what to display when online (default '●')
@@ -13,14 +13,8 @@ Configuration parameters:
         (default 'ip')
     negative_cache_timeout: how often to check again when offline (default 2)
     timeout: how long before deciding we're offline (default 5)
-    url: change IP check url (must output a plain text IP address)
-        (default 'http://ultrabug.fr/py3status/whatismyip')
     url_geo: IP to check for geo location (must output json)
-        (default 'http://ip-api.com/json')
-
-Format placeholders:
-    {country} display the country
-    {ip} display current ip address
+        (default 'https://freegeoip.net/json/')
 
 Color options:
     color_bad: Offline
@@ -49,8 +43,17 @@ class Py3status:
     mode = 'ip'
     negative_cache_timeout = 2
     timeout = 5
-    url = 'http://ultrabug.fr/py3status/whatismyip'
-    url_geo = 'http://ip-api.com/json'
+    url_geo = 'https://freegeoip.net/json/'
+
+    class Meta:
+        deprecated = {
+            'remove': [
+                {
+                    'param': 'url',
+                    'msg': 'obsolete parameter, use `url_geo` instead',
+                },
+            ],
+        }
 
     def on_click(self, event):
         """
@@ -61,43 +64,37 @@ class Py3status:
         else:
             self.mode = 'ip'
 
-    def _get_my_ip_and_location(self):
+    def _get_my_ip_info(self):
         """
         """
         try:
-            if self.py3.format_contains(self.format, 'country'):
-                resp = urlopen(self.url_geo, timeout=self.timeout).read()
-                resp = json.loads(resp)
-                country = resp['country']
-                ip = resp['query']
-            else:
-                country = None
-                ip = urlopen(self.url, timeout=self.timeout).read()
-                ip = ip.decode('utf-8')
+            resp = urlopen(self.url_geo, timeout=self.timeout).read()
+            resp = json.loads(resp)
+            info = {}
+            for placeholder in self.py3.get_placeholders_list(self.format):
+                if placeholder in resp:
+                    info[placeholder] = resp[placeholder]
+            return info
         except Exception:
-            country = None
-            ip = None
-        return country, ip
+            return None
 
     def whatismyip(self):
         """
         """
-        country, ip = self._get_my_ip_and_location()
+        info = self._get_my_ip_info()
         response = {
             'cached_until': self.py3.time_in(self.negative_cache_timeout)
         }
 
-        if ip is None and self.hide_when_offline:
+        if info is None and self.hide_when_offline:
             response['full_text'] = ''
-        elif ip is not None:
+        elif info is not None:
             response['cached_until'] = self.py3.time_in(self.cache_timeout)
             if self.mode == 'ip':
-                response['full_text'] = self.py3.safe_format(self.format, {
-                    'country': country,
-                    'ip': ip})
+                response['full_text'] = self.py3.safe_format(self.format, info)
             else:
                 response['full_text'] = self.format_online
-                response['color'] = self.py3.COLOR_GOOD
+            response['color'] = self.py3.COLOR_GOOD
         else:
             response['full_text'] = self.format_offline
             response['color'] = self.py3.COLOR_BAD
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -86,7 +86,7 @@ class Py3status:
             for old, new in self.substitutions.items():
                 info[old] = resp.get(new)
             return info
-        except Exception:
+        except self.py3.RequestException:
             return None
 
     def whatismyip(self):
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -101,11 +101,11 @@ class Py3status:
             response['full_text'] = ''
         elif info is not None:
             response['cached_until'] = self.py3.time_in(self.cache_timeout)
+            response['color'] = self.py3.COLOR_GOOD
             if self.mode == 'ip':
                 response['full_text'] = self.py3.safe_format(self.format, info)
             else:
                 response['full_text'] = self.format_online
-            response['color'] = self.py3.COLOR_GOOD
         else:
             response['full_text'] = self.format_offline
             response['color'] = self.py3.COLOR_BAD
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -15,7 +15,7 @@ Configuration parameters:
     negative_cache_timeout: how often to check again when offline (default 2)
     timeout: how long before deciding we're offline (default 5)
     url_geo: IP to check for geo location (must output json)
-        (default 'https://freegeoip.net/json')
+        (default 'https://freegeoip.net/json/')
 
 Format placeholders:
     {country} display the country
@@ -29,15 +29,8 @@ Color options:
 @author ultrabug
 """
 
-import json
-try:
-    # python3
-    from urllib.request import urlopen
-except:
-    from urllib2 import urlopen
-
 URL_GEO_OLD_DEFAULT = 'http://ip-api.com/json'
-URL_GEO_NEW_DEFAULT = 'https://freegeoip.net/json'
+URL_GEO_NEW_DEFAULT = 'https://freegeoip.net/json/'
 
 
 class Py3status:
@@ -66,10 +59,11 @@ class Py3status:
 
     def post_config_hook(self):
         # Backwards compatibility
+        self.substitutions = {}
         if self.url_geo == URL_GEO_NEW_DEFAULT:
-            self.format = self.format.replace('{country}', '{country_name}')
+            self.substitutions['country'] = 'country_name'
         elif self.url_geo == URL_GEO_OLD_DEFAULT:
-            self.format = self.format.replace('{ip}', '{query}')
+            self.substitutions['ip'] = 'query'
 
     def on_click(self, event):
         """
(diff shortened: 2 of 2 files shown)


maximbaz:
Hmm, I've just realised that this module is starting to look very similar to the `getjson` module 🙂

Differences:
- It has a special mode where it just shows online/offline symbol
- It has colors

tobes:
> Hmm, I've just realised that this module is starting to look very similar to the getjson module

Whilst this is true, I think it is no bad thing to have 'specialist modules', they allow for quick setups without needing to know much.  eg here I can just add the module to my i3bar and see my ip.  A small config change and I can add my country.  Yes I could use `getjson` but then I'd need to know the json format plus even that the service is available.

So I don't think this is a bad thing.  I'll try to review this PR at some point soon

py3status/modules/whatismyip.py
```diff
@@ -13,14 +13,8 @@
...
     negative_cache_timeout: how often to check again when offline (default 2)
     timeout: how long before deciding we're offline (default 5)
-    url: change IP check url (must output a plain text IP address)
-        (default 'http://ultrabug.fr/py3status/whatismyip')
     url_geo: IP to check for geo location (must output json)
-        (default 'http://ip-api.com/json')
-
-Format placeholders:
-    {country} display the country
-    {ip} display current ip address
```
maximbaz: Should I keep this list? I'm not sure. 

Also, in the new default `url_geo`, the placeholders are `{ip}` and `{country_name}` (plus many others), in the old default `url_geo` the placeholders were `{query}` and `{country}`, and in the old code it was hardcoded to replace `{ip}` with `{query}`. But we can't play smart and do such hidden replacements anymore, if we really want people to allow entering any format placeholders. 

TLDR: in the default configuration `{country}` changes to `{country_name}`, but we shouldn't do this replacement by ourselves, because if user redefines `url_geo`, on that another website the placeholder might be indeed `{country}`.

tobes: > TLDR: in the default configuration {country} changes to {country_name}

can't we be clever and if the url is `freegeoip.net/json` then just create country in the dict as well as country_name then the placeholder will still work.

We could also do query -> ip for `ip-api.com/json` too so that remains supported

And yes I think we should keep the placeholders but you can add ... if other items are returned then auto placeholders ...

maximbaz: oo, I like this idea! pushed the code.

py3status/modules/whatismyip.py
```diff
@@ -13,14 +14,13 @@
         (default 'ip')
     negative_cache_timeout: how often to check again when offline (default 2)
     timeout: how long before deciding we're offline (default 5)
-    url: change IP check url (must output a plain text IP address)
-        (default 'http://ultrabug.fr/py3status/whatismyip')
```
tobes: i'm not sure how @ultrabug will be about you removing his wonderful service.  I think if we only are looking for `{ip}` then we could continue to use this.

How about:

* deprecate url_geo and default `url = None`

* if user supplies url we use it

* no supplied url then default to ultrabug.fr if only ip is wanted else freegeoip.net

* when we call request the url we can see if it is json or just an ip and act accordingly

py3status/modules/whatismyip.py
```diff
@@ -13,14 +14,13 @@
...
     url_geo: IP to check for geo location (must output json)
-        (default 'http://ip-api.com/json')
+        (default 'https://freegeoip.net/json')
```
tobes: out of interest what was the dns issue?

py3status/modules/whatismyip.py
```diff
@@ -36,6 +36,9 @@
 except:
     from urllib2 import urlopen
```
tobes: we now have [self.py3.request()](https://github.com/ultrabug/py3status/blob/master/py3status/py3.py#L873) which returns a [HttpResponse](https://github.com/ultrabug/py3status/blob/master/py3status/request.py#L25)

```
response = self.py3.request(url)
data = response.json()
```

py3status/modules/whatismyip.py
maximbaz: to clarify @tobes, it was not enough to just extend the dict with the `country` as well as `country_name`, because later the exact value is used to format the string - but the adjustment of the `format` itself worked flawlessly.

py3status/modules/whatismyip.py
```diff
@@ -49,8 +52,24 @@ class Py3status:
...
+                },
+            ],
+        }
+
+    def post_config_hook(self):
+        # Backwards compatibility
+        if self.url_geo == URL_GEO_NEW_DEFAULT:
+            self.format = self.format.replace('{country}', '{country_name}')
+        elif self.url_geo == URL_GEO_OLD_DEFAULT:
+            self.format = self.format.replace('{ip}', '{query}')
```
tobes: although this is a nice approach it is flawed eg `'{country:^20}'` is a valid format string.

I think it might just be easier to do something like

```
def post_config_hook(self):
    if self.url_geo == URL_GEO_NEW_DEFAULT:
        self.substitutions = [('country_name', 'country')]

def _get_my_ip_info(self):
    ...
    for field, sub in self.substitutions:
        info[sub] = info.get(field)
```

I'm sure this can be improved

py3status/modules/whatismyip.py
maximbaz: When I'm on VPN, half of the time this address does not resolve to IP and so this module reports me as offline. The new default is always resolving fine.

py3status/modules/whatismyip.py
```diff
@@ -4,7 +4,8 @@
 
 Configuration parameters:
     cache_timeout: how often we refresh this module in seconds (default 30)
-    format: available placeholders are {ip} and {country}
+    format: available placeholders are {ip} and {country},
+            as well as any other key in JSON fetched from `url_geo`
             (default '{ip}')
     format_offline: what to display when offline (default '■')
     format_online: what to display when online (default '●')
```
lasers: Unrelated: Rename them to `icon_off`, `icon_on` (along w/ deprecation fix)

py3status/modules/whatismyip.py
maximbaz: We could... but why having two ways to detect IP address when we could just have one? 🙂 
It is a wonderful and simple service, no doubt, but this algorithm becomes quite complicated if we want to support both URLs :)

py3status/modules/whatismyip.py
tobes: sure but as a separate PR please

py3status/modules/whatismyip.py
maximbaz: nice 🙂

py3status/modules/whatismyip.py
maximbaz: I didn't even know about such magic with format strings 😃

I like your approach, I used it with a slight modifications.

maximbaz:
All right guys, unless I missed something, I think I fixed all the comments 🙂

@tobes please have a final look when you get a minute.

@lasers I made the changes you requested in a separate branch, I can't send PR yet until this is merged, but if you want to have an early look at the diff, here is is: https://github.com/maximbaz/py3status/pull/1

py3status/modules/whatismyip.py
```diff
@@ -61,43 +74,38 @@ def on_click(self, event):
...
             response['cached_until'] = self.py3.time_in(self.cache_timeout)
             if self.mode == 'ip':
-                response['full_text'] = self.py3.safe_format(self.format, {
-                    'country': country,
-                    'ip': ip})
+                response['full_text'] = self.py3.safe_format(self.format, info)
             else:
                 response['full_text'] = self.format_online
-                response['color'] = self.py3.COLOR_GOOD
+            response['color'] = self.py3.COLOR_GOOD
```
lasers: I think this should be moved to line 104 because I feel if we keep it here, someone eventually will view this as a mis-indentation in the future.

maximbaz: Very valid point :+1:

py3status/modules/whatismyip.py
```diff
@@ -61,43 +74,38 @@ def on_click(self, event):
...
+                if placeholder in resp:
+                    info[placeholder] = resp[placeholder]
+            for old, new in self.substitutions.items():
+                info[old] = resp.get(new)
+            return info
         except Exception:
-            country = None
-            ip = None
-        return country, ip
+            return None
```
lasers: No idea if we want this or to use `except self.Py3.RequestException:` instead.... or to use `except (self.py3.RequestException):` like bunch of other modules. My noob review here is done and @tobes said he went out in other issue... just a heads up. Fix exception bit and we'll get this in probably tomorrow. 👍

maximbaz: Not sure what is better, current code keeps us safe 100% of the time, i.e. also if exception happens while parsing JSON, or potentially if something happens while getting the placeholders list.

lasers: My guess. `try/except self.Py3.RequestException:` the resp (Line 81). Leave everything else exposed. If something potentially happened, py3status should stop the module, turn the module red, pass the exception to the red `whatismyip` module where you can click on it and it'll expand the exception/error message... That and one can find the issue in the log. From there, one can figure out what the issue was... 

It's possibly better than keep requesting API only to get something always failed in the json/placeholder parser... and still continue returning None... which would keep this loop going.

maximbaz: I didn't know that the module can show an exception, that's rather cool! Then I'm totally convinced 😉

maximbaz:
The `except self.Py3.RequestException` (with the capital `P`) in `getjson` module is a typo, fixed while at it.

image

tobes:
Thanks for the `getjson` fix

I think `url` is a nicer than `url_geo` but not that bothered

I'm going to let @ultrabug make the judgement call on removing his service.  I'd still prefer to allow it for the `{ip}` only case as I think the needed logic is minimal.

I think it also is important for backward compatibility that we at least support these type of service.

maximbaz:
I like the idea of using only one parameter instead of two, that's what we already have. For backwards compatibility it is easier to keep `url_geo` as someone could already define a custom URL for it.

I don't like supporting two formats of the URL that return either IP in plaintext or JSON, as this is extra code (extra code is more complicated to read) and could lead to confusion (users might provide URL that returns IP in plaintext, but then specify a format that contains `{ip}` and `{country}`, and be confused that it doesn't work). 
 
Historically it made sense to use URL that returns IP and nothing more, but since then the module grew, it now can do so much more, I think it is worth cleaning the code in favor of the new functionality. My respect to the author doesn't reduce by not using their service to determine IP address 😉

With this change, we deprecate the plaintext format and unite all users, and if anyone defined `url` this param is now just ignored and everyone is using the URL that returns JSON response.

----

I understand why you are hesitant to accept this change, I don't mind waiting for @ultrabug to respond. I only wish it doesn't take too long, as I have two more PR that are coming for this module 🎉, but I can't send them until this is merged.

tobes:
> I like the idea of using only one parameter instead of two, that's what we already have. For backwards compatibility it is easier to keep url_geo as someone could already define a custom URL for it.

All good points here I just think `url` is nicer than `url_geo` but it may prove too much of a pain to use `url` Doing this means we need to support the plaintext ip only url which you are opposed to so let's see how we end up.

> I don't like supporting two formats of the URL that return either IP in plaintext or JSON, as this is extra code (extra code is more complicated to read) and could lead to confusion (users might provide URL that returns IP in plaintext, but then specify a format that contains {ip} and {country}, and be confused that it doesn't work).

The added code is pretty minimal, basically we can try to parse the json if that fails then see if output looks like an ip and set `info['ip']` then just go as normal and it'd be all good.  This is really about backward compatibility but we can wait for @ultrabug.  In general I agree that extra code is bad and should be avoided wherever we can

lasers:
@tobes For clarification, do we want only to `try:except` Line 81 (resp), then leave others exposed to allows py3status picking up specific exceptions (if any) or to keep it as-is?

If we keep it as-is, will it mean both thing just goes to same place?

tobes:
@lasers from what I can see here we are doing the correct thing.

So if we can't get the url then we are offline and don't really want to see an error in the output.  But if we have bad json then we will show an error.  I haven't actually tested the new code yet.  Maybe we should try to catch the it's a bad url but that probably needs work on the py3.request thing and is outside the scope of this PR.

py3status/modules/getjson.py
```diff
@@ -94,7 +94,7 @@ def _query_url(self):
             resp = self.py3.request(self.url, timeout=self.timeout)
             status = resp.status_code == 200
             resp = resp.json()
-        except self.Py3.RequestException:
+        except self.py3.RequestException:
```
tobes: @maximbaz good catch catch.

If you add this as a separate PR I'll merge it straight away as a bugfix

maximbaz:
Hey guys, what do you think, is there a chance to merge it this month? 😜

py3status/modules/getjson.py
maximbaz: I somehow missed this comment, #849

tobes:
@ultrabug would be nice to know your thoughts here.  specifically the removal of `http://ultrabug.fr/py3status/whatismyip` as a usable source.

py3status/modules/getjson.py
lasers: That PR went through. We can skip this change + file.

maximbaz: I'll rebase this branch on master to do a final test when it will seem like we are close to merging it, and this will automatically be resolved by git 😉

py3status/modules/whatismyip.py
ultrabug: I don't really care tbh. I created my service because all the ones I used to use were sloppy and unreliable in terms of availability. If you trust freegeoip, let's go for it

ultrabug:
As I said in my comment, I'm okay, let's proceed !

PR #822, opened 2017-03-18
#820 round emoji clock to the nearest 15 or 45 min

### LINKED ISSUE #820: [clock] It is more intuitive to round clock emoji to the nearest 15 or 45 minutes
The emoji clock is being used in many apps these days, and I noticed that py3status implementation is not consistent with the rest and is less intuitive to me.

Currently the clock emoji is not being updated until the hour exactly reaches the next step, so for example:
- 14:59 🕝

Instead, it is more intuitive to choose icon for the time rounded to the nearest 15 or 45 minutes, that is:

- 14:01 🕑
- 14:14 🕑
- 14:15 🕝
- 14:44 🕝
- 14:45 🕒
- 14:59 🕒

/cc @tobes as you created this module.

maximbaz:
Fixes #820.

/cc @tobes


py3status/modules/clock.py
```diff
@@ -227,12 +226,19 @@ def clock(self, i3s_output_list, i3s_config):
...
+                    elif m >= 0.25 and m < 0.75:
+                        m = 0.5
+                    else:
+                        h += 1
+                        m = 0
                     # make 12 hourly etc
                     h = h % self.block_hours
-                    idx = int(math.floor(h / self.block_hours * (len(
-                        self.blocks))))
+                    idx = int((h + m) / self.block_hours * len(self.blocks))
```
tobes: This isn't working in python2 I think a `from __future__ import division` is the best solution.

maximbaz:
Hey @tobes, out of curiosity, is there anything else you want to see in this PR to get it merged? 🙂

tobes:
@maximbaz sorry, I'm busy this weekend.

I was thinking of the case where `blocks = '123456789AB0'` where it would make sense for it to change on the hour.  Saying that I think the general idea of this PR makes sense for the case you have given.  I should be able to do a proper review tomorrow.

tobes:
@maximbaz I wonder if you've had time to think about the how we deal if each icon is for an  hour, because it feels to me that we would be showing the wrong one in that instance.  Maybe we need a setting about if it changes on the 'on the time' (current behaviour) or 'too the nearest' (your patch)

Because to me depending on the blocks I choose both behaviours are the correct one.

maximbaz:
I think the "to the nearest" is a behavior very specific to the unicode clock icons, for other use cases people are probably expecting "on the time". Maybe just change the logic if and only if the `blocks` parameter has been overwritten by user in their config?

tobes:
@maximbaz So just a special hack for the default. I could live with that 

@ultrabug you have a view?

maximbaz:
Alternative idea: use "on the time" if length of BLOCKS is 12, otherwise use "to the nearest" because it will show a more intuitive approximation. Or can you think of a counter-example?

tobes:
`len(blocks) / block_hours` is probably useful.  It is a hard problem as you are guessing what the user wants.

like what would we do here?  `blocks = '◷◶◵◴'` `block_hours = 1`

maximbaz:
This module is too generic 😃 

So the preferred choice so far is a special hack for the default, ping me when this idea is approved for implementation.

tobes:
I think the only way we can really do this is with a config option much though I'd like to avoid one.  I'd still like to see this in the 3.5 release.

tobes:
ok, here I think we should add a param let's call it `nearest_block` or better name.  Anyhow when set it does like here and will use the nearest block if not we have the old behaviour.  I'm happy to have the `nearest_block = True` as default.

We could also make the calculations based on `len(blocks) / block_hours`to work with all sort of crazy schemes.

maximbaz:
I called it `round_to_nearest_block`, have a look :+1:

tobes:
Thanks,

As we are effectively just moving the time forward by half a block could we not just do this, which should work with different block settings eg `blocks = '◷◶◵◴' block_hours = 1`
```
h = t.hour + (t.minute / 60)
if self.round_to_nearest_block:
    h += (self.block_hours / len(self.blocks)) / 2
```

maximbaz:
This works, added :+1:

ultrabug:
Should we proceed here mates ?

tobes:
> Should we proceed here mates ?

yeah I'm happy with this

maximbaz:
Since everyone is happy, you merge it @tobes?

tobes:
> you merge it @tobes?

yeah I will do, soon :)

PR #825, opened 2017-03-22
bugfix prevent terminated modules from being enabled

tobes:
A module is terminated because it failed to load or raised an exception in `post_config_hook()`

However `killall -USR1 py3status` will cause the module to get started when it shouldn't.

This trivial patch fixes the issue.


ultrabug:
good catch

PR #828, opened 2017-03-23
ensure normal click behaviour

tobes:
If a module errors and then starts working correctly we still catch some clicks and behave like we are in error eg button 3 hides the module.  This PR fixes this so things behave as expected.


ultrabug:
I think that was the problem I failed to deep dive into 👍 thanks a lot

PR #853, opened 2017-03-27
prevent NoneColor breaking formatter

tobes:
@valdur55 found a bug in the formatter around NoneColors this fixes it and adds a new test to the formatter


ultrabug:
Thanks @valdur55 ! and u @tobes :)

PR #860, opened 2017-03-29
Do not ask application name/full_name when we know  this player already.

valdur55:
Last minute improvements.


tobes:
Thanks for this update

PR #862, opened 2017-03-30
better module error handeling prevent reappearance when dismissed

tobes:
With a module error if it was dismissed by a user and then an error occurred again the error would be re-shown.  With this PR the error will not show.

However if the module starts sending 'good' output and then an error happens the error will be shown.  This feels the right approach and is wanted (and tested) in #722


ultrabug:
Thanks mate

PR #864, opened 2017-03-31
allow - in config token

tobes:
for #861 

tokens must start with `a-z` or `_` following chars can be `a-z` `0-9` `_` and `-`


ultrabug:
I think you need to escape it since it's a keywords for a range of symbols

tobes:
from python docs

> If - is escaped (e.g. [a\-z]) or if it’s placed as the first or last character (e.g. [a-]), it will match a literal '-'.

but made it explicit.

ultrabug:
didn't know ! :) thanks

PR #865, opened 2017-03-31
github: prevent unneeded updates on clicks

tobes:
In #826 @lasers highlighted github doing more updates than needed.

This prevents the updates unless the `button_refresh` is clicked by default this is `2`


ultrabug:
Thanks

PR #883, opened 2017-04-06
[whatismyip] Add 'expected' param

maximbaz:
I was running this locally for quite a while, and I must say it is very useful feature.

With it I don't have to visually compare all the time if my IP / country changed or not, I can use py3status to do this for me and colors to quickly bring the mismatch to my attention.

Unless, this somehow can be done with magical `format`? 😉

### DIFF
--- a/py3status/modules/whatismyip.py
+++ b/py3status/modules/whatismyip.py
@@ -6,7 +6,8 @@ Configuration parameters:
     cache_timeout: how often we refresh this module in seconds (default 30)
     expected: define expected values for format placeholders,
               and use `color_degraded` to show the output of this module
-              if any of them does not match the actual value
+              if any of them does not match the actual value.
+              This should be a dict eg {'country': 'France'}
               (default None)
     format: available placeholders are {ip} and {country},
             as well as any other key in JSON fetched from `url_geo`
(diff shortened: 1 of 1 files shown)


tobes:
`if self.expected is not None and self.expected != response['full_text']:`

to me this seems really complicated. especially if we have a complex format.

How about expected is a dict and we check that any values in it match  eg

`expected = {'ip': '127.0.0.1'}`   or `expected = {'country': 'ru'}`

maximbaz:
This is definitely more scalable, yes. I'll have a look. 

> How about expected is a dict and we check that any values in it match

We check that **all** values match to show GREEN state, and YELLOW otherwise - agreed?

tobes:
> We check that all values match to show GREEN state, and YELLOW otherwise - agreed?

sounds good

maximbaz:
I love how it works now!

I simplified the logic in `_get_my_ip_info`, because this allows to use all placeholders in the `expected` dict, even those that are not used in the `format` itself. 

For example, with this change I can do this, and the IP address will be shown, but it will be GREEN or YELLOW based on the country I'm in:

```
format = '{ip}'
expected = {'country': 'France'}
```

py3status/modules/whatismyip.py
```diff
@@ -47,6 +52,7 @@ class Py3status:
     """
     # available configuration parameters
     cache_timeout = 30
+    expected = {}
```
tobes: cool,  I'd change this default to `None` and then set it to `{}` in `post_config_hook()`

I think that's a little nicer.

maximbaz: hm, not sure if I agree, but I'll do the change ;)

PR #892, opened 2017-04-10
raise exception on invalid JSON in request

tobes:
if we do `py3.request(url).json()` and there is invalid JSON we get a nasty exception.

This PR make it so we now raise  `RequestInvalidJSON` so now such exceptions are easy to catch.  This is useful for example in PR #877


ultrabug:
Thanks

PR #901, opened 2017-04-12
fix for clock module due to returned composite

tobes:
@lasers found a bug that came to light with the formatter changes.

This is due to a composite being returned rather than a string.  This is a bug in clock that the formatter changes triggered.


ultrabug:
Thanks guys

Issue #548, opened 2016-10-31
battery_level: measurement_mode 'sys' can end up over 100% full


tobes:
Error with battery_level module

```
IndexError: string index out of range
  File "/home/toby/dev/py3status/py3status/module.py", line 444, in run
    response = method()
  File "/home/toby/dev/py3status/py3status/modules/battery_level.py", line 157, in battery_level
    self._update_icon()
  File "/home/toby/dev/py3status/py3status/modules/battery_level.py", line 383, in _update_icon
    (len(self.blocks) - 1)))]
```

Seems that my battery thinks it is over 100% full.  We should probably just limit percent full to 100%
```
cat /sys/class/power_supply/BAT0/uevent 
POWER_SUPPLY_NAME=BAT0
POWER_SUPPLY_STATUS=Unknown
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_TECHNOLOGY=Li-ion
POWER_SUPPLY_CYCLE_COUNT=0
POWER_SUPPLY_VOLTAGE_MIN_DESIGN=11100000
POWER_SUPPLY_VOLTAGE_NOW=12399000
POWER_SUPPLY_POWER_NOW=0
POWER_SUPPLY_ENERGY_FULL_DESIGN=73260000
POWER_SUPPLY_ENERGY_FULL=44550000
POWER_SUPPLY_ENERGY_NOW=45820000
POWER_SUPPLY_CAPACITY=102
POWER_SUPPLY_CAPACITY_LEVEL=Normal
POWER_SUPPLY_MODEL_NAME=08K8193
POWER_SUPPLY_MANUFACTURER=JingYi
POWER_SUPPLY_SERIAL_NUMBER= 1353
```

PR #908, opened 2017-04-16
Battery level icons now work with capacity over 100%.

docwalter:
This is how I fixed https://github.com/ultrabug/py3status/issues/548 on my Dell notebook where the battery level can go above 100% . Simply limits the icon character index to the string length, so the rightmost character will be also be shown for levels > 100%.


tobes:
@docwalter thanks for this elegant solution

PR #910, opened 2017-04-17
Fix net_iplist to work with master

guiniol:
This fixes `net_iplist` by using the new `composite_join`. Previously, it would do a `str.join` but that doesn't work with composites.

Additionally, the regexp has been improved so that it can display ip addresses of peered interfaces (such as VPNs) too.

EDIT: this is similar to #901 I think.


tobes:
@guiniol thanks for this bugfix

PR #912, opened 2017-04-19
bugfix yahoo_weather and clock

tobes:
couple of fixes


ultrabug:
Thanks, as I said on IRC I think this composite_join and manipulation is overly complex for a simple module like weather_yahoo

I'll merge it but I guess we'll need to talk about that

PR #687, opened 2017-01-20
i3blocks support

tobes:
i3blocks https://github.com/vivien/i3blocks is a popular alternative to py3status.

In the spirit of _keep your friends close but your enemies closer_ I've written a module to allow i3blocks blocklets (and compatible scripts) to be run directly in py3status.

### <a name="i3block"></a>i3block

Support i3blocks blocklets in py3status.

[i3blocks](https://github.com/vivien/i3blocks) is a project to allow simple
scripts to provide output to the i3bar.  This module allows these blocklets to
run under py3status.  The configuration of the blocklets is similar to how they
are configured in i3blocks.

Some blocklets may return pango markup eg `<span ...` if so set
`markup = pango` in the config for that module.

```
# i3blocks config

[time]
command=date '+%D %T'
interval=5

[wifi]
instance=wls1
label='wifi:'
command=~/i3blocks/wifi.sh
interval=5

# py3status config

order += 'i3block time'

i3block time {
    command = "date '+%D %T'"
    interval = 5
}

order += 'i3block wifi'

i3block wifi {
    instance = wls1
    label = 'wifi:'
    command = '~/i3blocks/wifi.sh'
    interval = 5
}
```

Note: `format` configuration parameter is used as is standard in py3status, not
as in i3blocks configuration.  Currently blocklets must provide responses in
the standard i3blocks manner of one line per value (not as json).

Configuration parameters:
  - `cache_timeout` How often the blocklet should be called (in seconds).
    This is similar to cache_timeout used by standard modules.  However it
    can also take the following values; `once` the blocklet will be called
    once, `repeat` the blocklet will be called constantly, or `persist`
    where the command will be expected to keep providing new data.  If this
    is not set or is `None` then the blocklet will not be called unless
    clicked on.
    Note: To simplify i3block compatability this configuration parameter
    can also be provided as `interval`.
    *(default None)*
  - `command` Path to blocklet or command *(default None)*
  - `format` What to display on the bar *(default '{output}')*
  - `instance` Will be provided to the blocklet as $BLOCK_INSTANCE *(default '')*
  - `label` Will be prepended to the blocklets output *(default '')*
  - `name` Name of the blocklet - passed as $BLOCK_NAME *(default '')*

Format placeholders:
  - `{output}` The output of the blocklet

**This PR has got very long - the upshot is that several issues were discovered and resolved and this module now appears to run i3blocklets as they would be via i3blocks**

PR #915, opened 2017-04-29
external_script module: add support for colors

enguerrand:
Read a color code from the second line of output.

If the external script has only one line of output the behavior
is unchanged.

### DIFF
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -3,7 +3,10 @@
 Display output of a given script.
 
 Display output of any executable script set by `script_path`.
-Pay attention. The output must be one liner, or will break your i3status !
+Only the first two lines of output will be used. The first line is used
+as the displayed text. If the output has two or more lines, the second
+line is set as the text color (and should hence be a valid hex color
+code such as #FF0000 for red).
 The script should not have any parameters, but it could work.
 
 Configuration parameters:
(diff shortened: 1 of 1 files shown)


tobes:
@enguerrand thanks for this PR.  I'm slightly worried by the change in behaviour.  A script that worked previously might no longer.  So maybe it would be better to add an option for this.

The behaviour also seems similar to that of https://github.com/ultrabug/py3status/pull/687 although that would need some adaptation of your script (colour is 3rd line) and is over-complicating things.

enguerrand:
@tobes The only scenario I can think of right now where a previously working script would now stop working is if that script gave a second line of output which isn't a valid hex color code. (Or am I overlooking something?).
I am happy to build in a check for that and make it fall back to the previous behavior (no color) if the 2nd line is not a valid color code. 
Let me know what you think.

tobes:
> I am happy to build in a check for that and make it fall back to the previous behavior (no color) if the 2nd line is not a valid color code.

that seems a good solution and also no extra user config params to confuse

enguerrand:
There it is. Not a 100% secure check for a valid hex color, but should be good enough to avoid breaking old scripts?

tobes:
@enguerrand I think this is good enough and a nice feature.  Thanks

enguerrand:
@tobes You are welcome & thanks for merging it!

PR #917, opened 2017-04-30
color_degraded support in status mode for whatismyip module

aethelz:
As of now status mode in whatismyip only shows either color_good or color_bad and as such ignores user "expected" dictionary, showing color_good even when there is an IP/country mismatch. This following commit fixes this and adds color_degraded support for status mode.


aethelz:
I simply moved relevant code up a bit, now there is no unnecessary code duplication, as requested.

tobes:
@aethelz thanks for this improvement

PR #923, opened 2017-05-13
spaceapi: prevent 'used before assignment' error

fmorgner:
The 'lastchange' field in the 'state' object is optional according to the SpaceAPI specification. If it is not present, the lastchanged variable was never set, causing a 'used before assignment' error during formatting.


tobes:
@fmorgner thanks for the bugfix

PR #929, opened 2017-06-01
mpris: raise keyboard interrupt

ritze:
When you ran the _mpris_ module in test mode you couldn't abort it with ^C. The exception was 
caught in the module and wasn't raise up to _module_test_.


tobes:
@ritze Thanks,  makes this module much easier to work with

PR #931, opened 2017-06-02
spotify: sanitize album/track title

vvoland:
A lot of spotify album/track titles have this rubbish Remastered or Stereo suffix (eg. "Ziggy Stardust - Remastered 2012 Version"), which is an absolute waste of space on the screen, looks very unaesthetic and also is very useless.

This PR introduces new config options
sanitize_titles: true/false 
sanitize_words: list of words that indicate meta data

Preview

### DIFF
--- a/py3status/modules/spotify.py
+++ b/py3status/modules/spotify.py
@@ -9,6 +9,12 @@ Configuration parameters:
         (default 'Spotify not running')
     format_stopped: define output if spotify is not playing
         (default 'Spotify stopped')
+    sanitize_titles: whether to remove meta data from album/track title
+        (default True)
+    sanitize_words: which meta data to remove
+        (default ['bonus', 'demo', 'edit', 'explicit',
+                  'extended', 'feat', 'mono', 'remaster',
+                  'stereo', 'version'])
 
 Format placeholders:
     {album} album name
@@ -47,6 +53,7 @@ stopped
 
 from datetime import timedelta
 import dbus
+import re
 
 
 class Py3status:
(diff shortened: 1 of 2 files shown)


py3status/modules/spotify.py
```diff
@@ -102,6 +113,24 @@ def _get_text(self):
...
+        """
+        expression = expression.replace('META_WORDS_HERE', self.sanitize_words)
+        return re.compile(expression, re.IGNORECASE)
+
+    def _sanitize_title(self, title):
+        """
+        Remove redunant meta data from title and return it
+        """
+        inside_brackets = self._compile_re(r"([\(\[][^)\]]*?(META_WORDS_HERE)[^)\]]*?[\)\]])")
+        after_delimiter = self._compile_re(r"([\-,;/])([^\-,;/])*(META_WORDS_HERE).*")
```
tobes: You could just compile these once in `post_config_hook()` rather than every time they get used.

I'm interested in your regexs it might be worth commenting on what they are doing eg.  remove content in parenthesis if one of sanitize_words is inside.

I note you don't care about matching so `...( featuring ]..` will be removed.  This is probably not a problem.

You can also end up with trailing whitespace from what I can see.

py3status/modules/spotify.py
```diff
@@ -9,6 +9,10 @@
         (default 'Spotify not running')
     format_stopped: define output if spotify is not playing
         (default 'Spotify stopped')
+    sanitize_titles: whether to remove meta data from album/track title
+        (default True)
+    sanitize_words: which meta data to remove separated by |
+        (default 'stereo|mono|remaster|edit|bonus|extended|demo|explicit|version|feat')
```
tobes: you could have this as a list eg `sanitize_words = ['stereo', 'mono', 'remaster'...]`

py3status/modules/spotify.py
vvoland: Good point with the post_config_hook, will fix it.

Matching the same kind of braces in regexps may be too tricky for me (I'm a beginner when it comes to regular expressions) and I don't think there would be much benefit in that, so as you've said, that should not be a problem.

After processing title with regexes in _sanitize_title I do title.strip(),1 so I think that actually there should not be any trailing whitespace ;)

py3status/modules/spotify.py
vvoland: I'm trying to do that but I encounter a problem with default value in docstring
```
    sanitize_words: which meta data to remove
        (default ['stereo', 'mono', 'remaster', 'edit', 'bonus', 'extended', 'demo', 'explicit', 'version', 'feat'])
```
Results in 

> E501 line too long (116 > 99 characters)

and splitting it like that
```
        (default ['stereo', 'mono', 'remaster', 'edit',
                  'bonus', 'extended', 'demo', 'explicit',
                  'version', 'feat'])
```

Results in 
>`sanitize_words` default not in docstring

Can I somehow work this around without adding it to IGNORE_ITEM?

tobes: yes https://github.com/ultrabug/py3status/blob/master/tests/test_module_doc.py#L37 is the place to do it.

The docsting parsing is a bit primitive in the tests.

vvoland: Done

ultrabug:
I'm okay to add this defaulting to True since space saving is important on the bar.

As much as I like it, I would like to use this moment to point out @vvoland that we love to order stuff alphabetically in py3status code guidelines.

If you check out your PR, you will see that elements of lists, variables placements etc are far from being alphabetically ordered.

If you don't mind, I'd love you to change this, else I will permit myself to do it after it's merged. Your call.

vvoland:
@ultrabug Okay, I changed the order of variables, methods and list elements to alphabetical. Except the post_config_hook which I left on top, because I saw it that way in other modules and it feels right. If it's not okay or somehow I did miss something, feel free to change it to follow the project guildelines :).
Also squashed commits into one.

ultrabug:
Thanks for your efforts and understanding @vvoland, it's greatly appreciated!

This is find like this, only one is not done but I'll fix it myself. Will link the commit here for reference ;)

ultrabug:
^ mention commit

PR #932, opened 2017-06-04
Fix typo in pomorodo module doc

valdur55:


ultrabug:
Thanks mate ! good spot

PR #933, opened 2017-06-04
volume_status: fix PactlBackend on non english system

vvoland:
In order to get volume/mute state this module uses regexp containing hardcoded english words Mute, Sink, Volume. This fails on systems with language set to some other language than English, because this regular expression does not match any line which results in exception.

This commit forces english pactl output by setting LC_ALL environment variable to 'C' in spawned subprocess

### DIFF
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -89,7 +89,7 @@ mute
 """
 
 import re
-from os import devnull
+from os import devnull, environ as os_environ
 from subprocess import check_output, call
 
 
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks a lot for this neat implementation and fix @vvoland , I like your style 👍 

I guess you tested it okay on your side.

@tobes anything to add before I merge this ?

tobes:
@ultrabug Ideally we'd do this in Py3 but I think given that this doesn't yet use the helpers, and fixes a current issue, we should merge this and then fix it up 'properly' at a later time.

I should also be back in the land of the living in the next day or two.  Both our locations seem to be having interesting times today :(  For me it is just a short walk away.

ultrabug:
Ok let's proceed this way and move forward, thanks.

ultrabug:
PS @tobes : glad to have you back soon, safe travels ;)

PR #936, opened 2017-06-08
Round brightness values (so that 94.96% is rendered as 95% and not as 94%)

maximbaz:
```
❯  xbacklight -get
100.000000

❯  xbacklight -get
94.947121
```

In these cases the module was outputting `100%` and `94%`. I want to see `95%` in the latter case 🙂


tobes:
@maximbaz this seems a reasonable change, as ever thanks for your contribution

PR #939, opened 2017-06-24
mpris: dbus signal fix

ritze:
There is a bug with the dbus signals. We get callbacks for all running players if only one player change the playback state.

To catch the false signals I check the playback state with the dbus method additionally. Let me know if you've an better idea.

### DIFF
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -377,6 +377,13 @@ class Py3status:
             status = data.get('PlaybackStatus')
             if status:
                 player = self._mpris_players[player_id]
+ 
+                # Note: Workaround.
+                # Since all players get noted if playback status has been
+                # changed we have to check if we are the choosen one
+                if status != player['_dbus_player'].PlaybackStatus:
+                    return
+
                 player['status'] = status
                 player['_state_priority'] = WORKING_STATES.index(status)
             self._set_player()
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -398,16 +398,18 @@ class Py3status:
                     p['name'] = self._mpris_names[p['identity']]
                     p['full_name'] = u'{} {}'.format(p['name'], p['index'])
 
-        status = player.PlaybackStatus
-        state_priority = WORKING_STATES.index(status)
         identity = player.Identity
+        name = self._mpris_names.get(identity)
+        if name not in self.player_priority and '*' not in self.player_priority:
+            return False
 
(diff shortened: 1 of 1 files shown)


ritze:
With this code changes we don't recognize any title change. I'll work on it.

py3status/modules/mpris.py
```diff
@@ -375,10 +376,23 @@ def player_on_change(*args):
...
+            player = self._mpris_players[player_id]
+
+            # Note: Workaround. Since all players get noted if playback status
+            #       has been changed we have to check if we are the choosen one
+            try:
+                dbus_status = player['_dbus_player'].PlaybackStatus
+            except GError:
+                # Prevent errors when calling methods of deleted dbus objects
+                return
+            if status != dbus_status:
```
ritze: With this if statement we stop an title change. But how can we find out if we are the right player here?

ritze:
Works now. ;-)

ritze:
`player = self._dbus.get(player_id, SERVICE_BUS_URL)` in `_add_player` throws sometimes an error:
> GLib.Error: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.UnknownObject: No such object path '/org/mpris/MediaPlayer2' (41)

Mostly when you start a new player after py3status/mpris was loaded.

tobes:
When I use `d-feet` to look at dbus whilst running `vlc` I see the following:

* name `org.mpris.MediaPlayer2.vlc` unique name `:1.159` object path `/org/mpris/MediaPlayer2`

* name `:1.159` unique name `:1.159` object path `/org/mpris/MediaPlayer2`

* name `:1.160` unique name `:1.160` object path **Error Unknown Path**

they all have the same unix path/pid so it feels like it may be related.  I can't trigger your bug though.  So maybe try returning False when we hit this and see if that fixes it.  I'd still like to understand it more as we already check the name.

Additionally in `_add_player()` where we do the `return False` we should probably avoid this if `self.player_priority == []` which is the default, and treat it like `['*']` - this took me a little while to work out why nothing was working for me.

ritze:
I think this bug is only happen with [bomi](https://bomi-player.github.io/). I don't have this behaviour with vlc.

> Additionally in `_add_player()` where we do the `return False` we should probably avoid this if `self.player_priority == []` which is the default, and treat it like `['*']` - this took me a little while to work out why nothing was working for me.
Merge state
Sounds good. I'll change this.

tobes:
@ritze thanks for your ongoing support for this great module

PR #942, opened 2017-07-02
Fix usage of cower in arch_updates module

maximbaz:
There is no `-b` flag, using it leads to zero results.

### DIFF
--- a/py3status/modules/arch_updates.py
+++ b/py3status/modules/arch_updates.py
@@ -114,7 +114,7 @@ class Py3status:
 
         pending_updates = b""
         try:
-            pending_updates = str(subprocess.check_output(["cower", "-bu"]))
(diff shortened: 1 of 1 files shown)


tobes:
hmm, it seems it used to have a `-b` (brief) option but no longer does (since release 17, on  24 April) .  I'm going to assume that removing it will cause issues with old versions of cower over reporting the number of updates.  However seeing as this is arch specific we should assume that users are on the bleeding edge and so will have the latest version.

@maximbaz do you think that is a fair assessment?

maximbaz:
I agree that we should assume that everyone has the latest version by now.

I actually think the old version will not break, because we are not parsing the output, we are just counting number of lines, I'm guessing that the `brief` parameter was just producing less output per line, but the same number of lines.

tobes:
@maximbaz cool thanks for this

Issue #732, opened 2017-02-03
battery_level: ZeroDivisionError


rixx:
Hi there, when booting up after my Laptop shutting down due to both batteries being empty, I got

```
py3status: Instance `battery_level` failed (ZeroDivisionError) battery_level.py line 364. Please try to fix this and reload i3wm (Mod+Shift+R)
```

The line mentioned has moved up a few lines (I looked it up, seems I'm not up to date with `master`), it corresponds to https://github.com/ultrabug/py3status/blob/master/py3status/modules/battery_level.py#L357 at this time of writing (05b970651931e1598623ae90bbd9e9b70ae5d614).

tobes:
@rixx thanks for reporting this bug.

tablet-mode:
Is there anything speaking against just putting _rate_second_per_mah_ and _time_remaining_seconds_ assignment (lines 360-367) in a try except block catching _ZeroDivisionError_ and setting both values to 0 in this case?

tobes:
> Is there anything speaking against just putting rate_second_per_mah and time_remaining_seconds assignment (lines 360-367) in a try except block catching ZeroDivisionError and setting both values to 0 in this case?

That would probably fix things.  imho the code could be made a little more readable but it'd be nice to just have a quick fix for this issue

tablet-mode:
I created a pull request. Considering that it's [easier to ask for forgiveness than permission](https://docs.python.org/3.4/glossary.html#term-eafp) I think this is an appropriate solution.

tobes:
fixed via #943

PR #943, opened 2017-07-04
Catch ZeroDivisionError for #732

tablet-mode:
Quickfix for #732. Catches ZeroDivisionError which could happen if either the active or inactive battery is reporting 0% capacity.


tobes:
@tablet-mode thanks for this

PR #946, opened 2017-07-09
Parse and use GitHub pagination headers (closes #945)

### LINKED ISSUE #945: GitHub: Show correct count of unread notifications
If you have more than 50 unread notifications in total, and you want to see that total notification count, the GitHub module still only shows 50 due to pagination:

https://github.com/ultrabug/py3status/blob/master/py3status/modules/github.py#L155

rixx:


tobes:
really we should make headers available via `py3.request()` but I'm happy to merge this and then fix it up later.

@ultrabug you happy with that?

rixx:
Yeah, I was a bit surprised at having to search for the headers, but I was mostly focussed on getting that pesky "50" to display the true post-vacation number. ;)

ultrabug:
Oh yeah thanks a lot @rixx !!

I'm so lagging behind that I just realized that I'm affected by this bug and your fix works like a charm 👍

PR #949, opened 2017-07-11
Unless this is a float, truncation will cause quality to be == 0 if <…

markrileybot:
… 100


tobes:
@markrileybot thanks.  The python 2 division fix looks good.

The wifi patch is failing our tests,  Travis is down so I can't see the actual issue.  It makes life easier if you make each issue a separate PR.  I can always manually merge things if that proves to painful.

PR #950, opened 2017-07-12
Unless this is a float, truncation will cause quality to be == 0 if <…

markrileybot:
… 100


tobes:
@markrileybot thanks for finding this

PR #951, opened 2017-07-12
Feature/pipe vpn format

markrileybot:


py3status/modules/vpn_status.py
```diff
@@ -125,8 +125,11 @@ def return_status(self):
                 name = vpn
                 color = self.py3.COLOR_GOOD
 
-        # Format and create the response dict
-        full_text = self.py3.safe_format(self.format, {'name': name})
+        if name is None:
+            full_text = self.py3.safe_format(self.format)
+        else:
+            full_text = self.py3.safe_format(self.format, {'name': name})
+
```
tobes: This change is not needed.  Because `{'name': None}` will cause the desired effects so we can just keep things simple.  Other than this I think it is a cool improvement to the module

ultrabug:
Thanks indeed @markrileybot 

Please remove the change mentioned by @tobes so we can merge this 👍

py3status/modules/vpn_status.py
markrileybot: Great! I didn't know that format_save would do this for me!

ultrabug:
Thanks @markrileybot sorry to be a bit stubborn, but those last lines where you remove comments & add a blank line are not necessary either, can we keep it to the minimum ?

Thanks for your patience

ultrabug:
Thanks a lot!

PR #952, opened 2017-07-13
Make headers available via py3.request() and update github module

tobes:
Allows modules to access headers from `py3.request()`.

Updates `github` module to use them.

Cleaned up the import of `urlparse`

Fixed logical bug where we would always check for links  `''.split(', ') == ['']`

Fixed typo in `Py3.request()` comment


ultrabug:
Clean & simple, thanks mate

Issue #948, opened 2017-07-10
py3status depends on i3status


Cypher1:
Why does py3status need i3status installed? Can we remove this dependency or at least fix the error message which currently gives: (FileNoteFound) i3status.py (line 341).

Thanks

tobes:
i3status should only be needed if you are using one of it's modules.  Do you have any i3status modules defined in your config?

Cypher1:
Bingo! I didn't realise that we hadn't re-implemented the "time" module.
Can we get a better error message for it still?

tobes:
yeah, there is no reason that we can't handle this more gracefully.  Thanks for reporting it.

There is the `clock` module that may help

Cypher1:
Thanks I'm trying to fix my preferred formatting with it now.
Cheers

ultrabug:
Hi @Cypher1 

> Bingo! I didn't realise that we hadn't re-implemented the "time" module.
I like that you say **WE** 👍 

What message would you propose for this mate? Your user experience would be of great help in providing the best message for this case!

Cypher1:
Hi @ultrabug 

I'm thinking something like "Error: Dependency i3bar not installed. Please install it." is all that is needed. 

If I'm not busy tomorrow morning I'll see if I can put together a pull request for it.

_Sent from my Sony F8331 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

ultrabug:
Thanks @Cypher1 , please check https://github.com/ultrabug/py3status/pull/953 for this

Cypher1:
Ack. Thanks

tobes:
fix now in master

PR #953, opened 2017-07-13
i3status missing better reporting

tobes:
from issue #948 this now reports a better error to the user when i3status is missing.

It is very simplistic we could do a `which i3status` to confirm but that feels overkill for now

Better logging of i3status spawning (when it has)

removed some unneeded bits from mocking

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -329,9 +329,6 @@ class I3status(Thread):
         try:
             with NamedTemporaryFile(prefix='py3status_') as tmpfile:
                 self.write_tmp_i3status_config(tmpfile)
-                self.py3_wrapper.log(
-                    'i3status spawned using config file {}'.format(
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks, I'm just asking myself if we couldn't use the `check_command` helper to make sure about i3status presence ?

Also, I'm wondering if this check could be done ONLY if i3status modules are used ?

tobes:
>  I'm just asking myself if we couldn't use the check_command helper

In many ways this might be good.  I think it will need a bit of code movement though.  We only try to start i3status if we need to so we wouldn't check otherwise.

ultrabug:
Okay, I'd like ACK from @Cypher1 before merging this. Cheers!

Cypher1:
Looks good to me. Thanks

PR #960, opened 2017-07-19
Adding apt_updates module and updating Insync module for better formatting

Cypher1:
Insync module previously didn't catch pause state and used keywords rather than neat icons, all formatting is still able to be formatted to fit the old style.

Apt_updates module uses `apt list --upgradable` to list the number of packages that are available for update.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,15 +1,23 @@
 version 3.6 (milestone)
 * battery_level module: fix bug with over 100% full icon, by @docwalter
+* battery_level module: fix divide by zero bug, by @tablet-mode
 * backlight module: combine device/device_path params, by @lasers
+* backlight module: improve rounding of brightness value, by @maximbaz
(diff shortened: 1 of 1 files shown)


tobes:
@Cypher1 Thanks for this PR.  It will be easier to work with, if you make this two PRs one for the apt_updates and one for the insync module.  You'll need to have them as separate branches.

With the insync it is unlikely that the icon changes will be accepted as we try to not 'break' things for existing users, but the pause bug looks like a good fix.

Cypher1:
Thanks. I was wondering about how to do that. GitHub wouldn't let me make a
pull request from a specific commit. I'll try breaking it into two branches
and PRing from each.

Cypher1:
Thanks @tobes. I've broken this PR into https://github.com/ultrabug/py3status/pull/965 and https://github.com/ultrabug/py3status/pull/966

PR #962, opened 2017-07-20
make hide_if_zero work for net_rate

tobes:
From #935 make `hide_if_zero` work.

A new py3 helper is added to allow us to get the needed info.

PR #964, opened 2017-07-20
Add hide_if_zero to sysdata module

maximbaz:
Depends on https://github.com/ultrabug/py3status/pull/962, do not merge before that one.

-------------------

Why I did it: I monitor CPU, RAM and swap. I don't want to see the output for swap if it is not used at the moment.

I wish I could define this in a format, but as far as I know it is not possible. What I did instead is I defined `sysdata` module 3 times in my `.i3status.conf`, one for each sensor. Now my `sysdata swap` module only outputs swap percentage, and using `hide_if_zero` I can hide this module entirely when swap is not used.

Hope that makes sense what I'm trying to accomplish?

/cc @tobes

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -7,6 +7,8 @@ Configuration parameters:
     format: output format string
         *(default '[\?color=cpu CPU: {cpu_usage}%], '
         '[\?color=mem Mem: {mem_used}/{mem_total} GB ({mem_used_percent}%)]')*
+    hide_if_zero: hide output if all values are zero.
+        (default False)
     mem_unit: the unit of memory to use in report, case insensitive.
         ['dynamic', 'KiB', 'MiB', 'GiB'] (default 'GiB')
     swap_unit: the unit of swap to use in report, case insensitive.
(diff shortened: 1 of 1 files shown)


tobes:
`format = '<CPU, RAM>[\?not_zero  {swap_used_percent}%]'` might be close to what you are after.

It is not perfect because it will show if the percent was 0.0001 but that can be easily fixed by a small change in the formatter.  Does this work if so I'll look at a patch to fix the formatter.

I'd like to avoid adding `hide_if_zero` type params.  In the future I'd like to make it so that `0` is generally hidden as that is usually the wanted behaviour.

maximbaz:
That is certainly a better fix @tobes! I agree that hide `0` by default is a good idea in general.

Let's fix the formatter so that 0.00001 doesn't make `0.00%` be displayed. Another thing that I discovered, is that the order of these formatting markers matters, so that:

```
format = '[\?not_zero [\?color=swap {swap_used_percent}%]]'   # this doesn't work, I'll see 0.00%
format = '[\?color=swap [\?not_zero {swap_used_percent}%]]'   # this works, it is hidden when swap is not used
```

I wonder if long-term both options should be allowed.

tobes:
`format = '[\?color=swap&not_zero {swap_used_percent}%]'` should be best, the reason is that the commands are local within the square brackets, you can join multiple ones though.  I should see if we can inherit without causing problems.

The formatting needs better documenting too.

maximbaz:
I see, that makes sense. Thanks :)

I'll close this PR and let you handle the formatter separately.

PR #965, opened 2017-07-20
Added apt updates module

Cypher1:
This adds a module for monitoring updates available as listed by apt. It is based on the updates module for pacman / aur for arch linux.

### DIFF
--- a/py3status/modules/apt_updates.py
+++ b/py3status/modules/apt_updates.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Display number of pending updates for Ubuntu based Distros.
+Display number of pending updates for Debian based Distros.
 
 Thanks to Iain Tatch <iain.tatch@gmail.com> for the script that this is based on.
 
@@ -9,8 +9,7 @@ This will display a count of how many 'apt' updates are waiting to be installed.
 Configuration parameters:
     cache_timeout: How often we refresh this module in seconds (default 600)
     format: Display format to use
-        (default 'UPD: {apt}')
-    no_updates: Display format to use if there are no updates
+        (default 'UPD[\?not_zero : {apt}]')
 
 Format placeholders:
     {apt} Number of pending apt updates
@@ -28,19 +27,12 @@ SAMPLE OUTPUT
 import subprocess
 import sys
 
-FORMAT = 'UPD: {apt}'
-NO_UPDATES = 'UPD'
 LINE_SEPARATOR = "\\n" if sys.version_info > (3, 0) else "\n"
-
 
 class Py3status:
     # available configuration parameters
     cache_timeout = 600
-    format = FORMAT
-    no_updates = NO_UPDATES
-
-    def post_config_hook(self):
-        self.include_apt = self.py3.format_contains(self.format, 'apt')
+    format = 'UPD[\?not_zero : {apt}]'
 
     def check_updates(self):
         apt_updates = self._check_apt_updates()
@@ -48,14 +40,12 @@ class Py3status:
         color = self.py3.COLOR_DEGRADED
         if apt_updates == 0:
             color = self.py3.COLOR_GOOD
-            full_text = self.no_updates
-        else:
-            full_text = self.py3.safe_format(
-                self.include_apt,
-                {
-                    'apt': apt_updates
-                }
-            )
+        full_text = self.py3.safe_format(
+            self.format,
+            {
+                'apt': apt_updates
+            }
+        )
         return {
             'color': color,
             'cached_until': self.py3.time_in(self.cache_timeout),
@@ -68,10 +58,9 @@ class Py3status:
         to determine how many updates are waiting to be installed via
         'apt list --upgradeable'.
         """
-        pending_updates = str(subprocess.check_output(["apt", "list", "--upgradeable"]))
-        pending_updates = pending_updates.split(LINE_SEPARATOR)
-        pending_updates = pending_updates[1:-1]
-        return len(pending_updates)
+        output = str(subprocess.check_output(["apt", "list", "--upgradeable"]))
+        output = output.split(LINE_SEPARATOR)
+        return len(output[1:-1])
 
 
 if __name__ == "__main__":
--- a/py3status/modules/apt_updates.py
+++ b/py3status/modules/apt_updates.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+"""
+Display number of pending updates for Ubuntu based Distros.
+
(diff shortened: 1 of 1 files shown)


py3status/modules/apt_updates.py
```diff
@@ -0,0 +1,82 @@
...
+    def _check_apt_updates(self):
+        """
+        This method will use the 'checkupdates' command line utility
+        to determine how many updates are waiting to be installed via
+        'apt list --upgradeable'.
+        """
+        pending_updates = str(subprocess.check_output(["apt", "list", "--upgradeable"]))
+        pending_updates = pending_updates.split(LINE_SEPARATOR)
+        pending_updates = pending_updates[1:-1]
+        return len(pending_updates)
```
tobes:
```
output = self.py3.command_output('apt list --upgradeable')
return  len(output.splitlines()[1:-1])
```

Would use the new functionality.  `arch_updates` should get upgraded too.

Then we don't need to import `sys`, `subprocess`, line separators etc.

py3status/modules/apt_updates.py
```diff
@@ -0,0 +1,82 @@
...
+
+FORMAT = 'UPD: {apt}'
+NO_UPDATES = 'UPD'
+LINE_SEPARATOR = "\\n" if sys.version_info > (3, 0) else "\n"
+
+
+class Py3status:
+    # available configuration parameters
+    cache_timeout = 600
+    format = FORMAT
```
tobes: you could just do `format = 'UPD: {apt}'`

`arch_updates` does the odd stuff because of the `aur` stuff.

py3status/modules/apt_updates.py
```diff
@@ -0,0 +1,82 @@
...
+
+SAMPLE OUTPUT
+{'full_text': 'UPD: 5'}
+"""
+
+import subprocess
+import sys
+
+FORMAT = 'UPD: {apt}'
+NO_UPDATES = 'UPD'
```
tobes: we can get the same effect by just using a format of `'UPD[\?not_zero : {apt}]'` this will the output `UDP`, `UDP: 5` etc.

This reduces the number of config params we need and gives the user flexibility

py3status/modules/apt_updates.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+"""
+Display number of pending updates for Ubuntu based Distros.
```
tobes: should this be debian based rather than ubuntu based?

Cypher1: Yep, forgot my linux history

py3status/modules/apt_updates.py
Cypher1: We don't tend to do this in other modules (python string formatting knowledge is not assumed). Should these other modules also be updated? I'll make the change in the mean time.

py3status/modules/apt_updates.py
Cypher1: @tobes what is the way to ignore stderr output in `self.py3.command_output`? It seems that command_output doesn't listen to return status but takes any stderr output to be a failure.
It also relies on the CommandError exception and that functionality is not yet in the released version of py3status. Should I keep using the old method until this has been released?

Thanks

py3status/modules/apt_updates.py
tobes: The formatting is newer than many of the modules so updating things takes time.

py3status/modules/apt_updates.py
tobes: > It seems that command_output doesn't listen to return status but takes any stderr output to be a failure.

ok maybe this is something to improve.  What is apt sending to stderr?  It is always nice to find these odd situations.  We could add an option to `command_output` to redirect stderr > stdout

If you are running py3status 3.5 rather than using master then I'm happy to keep this module compatible with 3.5 so you can use it.

tobes: would a patch (untested) like this give you the functionality needed

```
index 4c20682..b6d7a9d 100644
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -8,7 +8,7 @@ import shlex
 from fnmatch import fnmatch
 from math import log10
 from pprint import pformat
-from subprocess import Popen, PIPE
+from subprocess import Popen, PIPE, STDOUT
 from time import time
 
 from py3status import exceptions
@@ -861,7 +861,7 @@ class Py3:
             msg = "Command '{cmd}' {error}".format(cmd=command[0], error=e.errno)
             raise exceptions.CommandError(msg, error_code=e.errno)
 
-    def command_output(self, command, shell=False):
+    def command_output(self, command, shell=False, capture_stderr=False):
         """
@@ -871,8 +871,11 @@ class Py3:
         # convert the command to sequence if a string
         if isinstance(command, basestring):
             command = shlex.split(command)
+
+        stderr = STDOUT if capture_stderr else PIPE
+
         try:
-            process = Popen(command, stdout=PIPE, stderr=PIPE, close_fds=True,
+            process = Popen(command, stdout=PIPE, stderr=stderr, close_fds=True,
                             universal_newlines=True, shell=shell)
         except Exception as e:
             msg = "Command '{cmd}' {error}".format(cmd=command[0], error=e)
```

Cypher1: Stderr from apt
`WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`
Which is fair, but should be fairly easy to fix.
The above fix seems like overkill for this module, we just need to be able to get the full output from CommandError (which seems to be possible via `error.output`. I'm just not sure that approach is reasonable as CommandError isn't in my local version yet and so I imagine that it isn't in most other people, making this module useless until py3status updates.

tobes:
looking good you just have a linting minor issue.

```
apt_updates.py:32:1: E302 expected 2 blank lines, found 1
class Py3status:
^
```

py3status/modules/apt_updates.py
```diff
@@ -0,0 +1,72 @@
...
+    {apt} Number of pending apt updates
+
+Requires:
+    apt: Needed to display pending 'apt' updates
+
+@author Joshua Pratt <jp10010101010000@gmail.com>
+@license BSD
+
+SAMPLE OUTPUT
+{'full_text': 'UPD: 5'}
```
tobes: `{'color': '#FFFF00', 'full_text': 'UPD: 5'}` is a better reflection of the output.

Apart from that I think we're done

Cypher1:
Thanks @Tobes

tobes:
@Cypher1 thanks to you

ultrabug:
Thanks a lot @Cypher1 :)

PR #966, opened 2017-07-20
Updated Insync module to recognise SYNCED state 

Cypher1:
Also a small fix to make error messages more readable when used in custom formatting.


Cypher1:
Looks like there are config errors with Travis at the moment. pypy isn't installed.
Thanks

I've left the symbol update out of this PR. I thought I could add it to the doc as a suggested setting if that would be alright. It requires font support but most people have full unicode fonts installed already and so it shouldn't be a problem.

tobes:
If you rebase on master the travis issue has been resolved

py3status/modules/insync.py
```diff
@@ -85,14 +87,17 @@ def insync(self):
             }
 
         color = self.py3.COLOR_DEGRADED
+        format = self.format
         if status == "Insync doesn't seem to be running. Start it first.":
             color = self.py3.COLOR_BAD
-            status = STRING_ERROR
+            format = STRING_ERROR
```
Cypher1: I've changed it as I realised that formating like "I{number of syncing files}" would turn into I{error message}" which was harder to read. This was intended as a bug fix.

tobes: one option would be to use `self.py3.error()` here so it is treated as an error rather than as output.

Cypher1:
I'll try to check in my AM that it is actually piped via stderr. I'm not
certain from memory.

lasers:
(Question) We might not want to call this module every `10` seconds because this could require an access to some remote service. If that's true, then this module could somewhat becomes useless... since we wouldn't be able to check status often without annoying somebody's API rules. Is that what's happening here?

(Possible solution) Use new thread / watchdog / asyncio to monitor path (if we can get them from somewhere like `insync` command and/or config... then we monitor for changes... and when we detect them, we rapidly update the module every second for next 5 seconds?

(Possible solution) We just use their `insync` tray instead?

Cypher1:
Thanks Lasers,

I've been using this for the past few weeks now that I've come back to
linux (having used it last time I was on Linux without any problems) and
I'm just using the get_status etc. commands from Insync. It is possible
that they update insync's state but that would contradict the names of the
commands as I understand them.

Have we seen them call a web API and/or get rate limited. I'm happy to have
the default update turned down and set mine higher in the meantime to see
if it breaks for me.

About the traybar version, it's not bad but doesn't show as much status
information in my opinion. Not to mention that I personally use py3status
blocks instead of traybar icons.

Thanks

Cypher1:
Additionally, though I don't think we currently have a problem, your
solution is pretty neat.
I'm concerned that we'd start making an py3status Google drive client
though.

py3status/modules/insync.py
tobes: if we change this to 
```
if status == "Insync doesn't seem to be running. Start it first.":
    self.py3.error(STRING_ERROR)
```

then I think we can merge this PR.  That way the error gets handled consistently across modules.  This is relatively new functionality so some existing modules don't yet use it.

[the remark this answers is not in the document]
ultrabug: @Cypher1 do you want to go on the proposed solution so we can finally merge this nice contribution?

ultrabug:
merging, gonna add the comment from @lasers in myself, thanks !

PR #967, opened 2017-07-20
formatter not zero fix for displayed values

tobes:
@maximbaz found an issue around values that get rounded to zero not being treated as zero in some situations. eg #964

Moving the zero check till after the value has been formatted fixes the issue.

Tests added to prevent regressions


ultrabug:
Good catch @maximbaz thanks for the fix @tobes :)

PR #968, opened 2017-07-20
set pypy version - fix build tests

tobes:
Travis is now using `trusty` rather than `precise` for builds which broke pypy.

This sets a pypy version.  Which fixes things.  We should look into what pypy versions we want to/cn support via travis


ultrabug:
hmm indeed, py3status on pypy 👍 nice!

PR #969, opened 2017-07-20
formatter allow some commands to be inherited

tobes:
This PR enables `not_zero` and `show` formatter commands to be inherited.  Previously `color` ended up propagating.

This means we can now write a format like `\?not_zero Output:[ X {x}][ Y {y}]` rather than `Output:[\?not_zero  X {x}][\?not_zero  Y {y}]` which should help us simplify formats a bit

eg uptime `'up [\?if=weeks {weeks} weeks ][\?if=days {days} days ][\?if=hours {hours} hours ][\?if=minutes {minutes} minutes ]'`  == `'\?not_zero up [ {weeks} weeks][ {days} days][ {hours} hours][ {minutes} minutes] '`

Due to the inheritance changes, color now uses the same code rather than it's old method.

Some tests added for luck

### DIFF
--- a/py3status/formatter.py
+++ b/py3status/formatter.py
@@ -291,6 +291,7 @@ class BlockConfig:
     """
 
     REGEX_COLOR = re.compile('#[0-9A-F]{6}')
+    INHERITABLE = ['color', 'not_zero', 'show']
 
     # defaults
     _if = None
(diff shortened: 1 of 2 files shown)


ultrabug:
Thanks a lot, nice implementation!

PR #970, opened 2017-07-22
Only show first line of error in i3bar

tobes:
If a runtime error is more than one line it displays incorrectly in i3bar.  This fixes the issue by only showing the first line.


ultrabug:
Thanks!

Another option would have been to join them using space but I guess that won't fit on the bar width

PR #971, opened 2017-07-22
command_output allow piping of STDERR to STDOUT

tobes:
Some commands provide valid output via STDERR this PR allows this to be piped to STDOUT.

This is an issue for `apt` and `dnf`


tobes:
Another option would be to change `Py3.command_output()` to work more similar to `subprocess.check_output()` that is always pipe stderror > stdout and not care about exit codes. so would just return `''` in worst situation.

ultrabug:
TBH I'd go for the second option yea?

tobes:
> TBH I'd go for the second option yea?

I'll redo the PR.  It makes the most sense

PR #972, opened 2017-07-23
cmus, moc: fix screenshots

tobes:
New `cmu` and `moc` modules had errors in their screenshot data.  This also broke the documentation build.

* fix modules screenshot data

* improve screenshot building so that if this happens again the build will not die

**edit**

fixed additional docstring errors causing build warnings/formatting issues


ultrabug:
Thanks mate, I shall check the doc build better

PR #973, opened 2017-07-23
Fix py3-cmd link in docs

tobes:
When we renamed `py3-cmd` we missed this reference in the documentation.


ultrabug:
good spot, thx!

PR #974, opened 2017-07-23
Docs: add new screenshot directive

tobes:
To allow screenshots to be embedded in the documentation, this PR adds a `screenshot` directive to sphinx.  This will be useful for documenting formatting.

```
.. screenshot::

        {'color': '#00FF00', 'full_text': 'Example output'}
```

The changes to `screenshots.py` are to allow single images to be created, but don't really change how things are done.  There is a bit of nastiness around caching `font` and `glyph_data` as that helps performance.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -2,7 +2,11 @@
 # -*- coding: utf-8 -*-
 import re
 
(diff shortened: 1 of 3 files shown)


ultrabug:
We're going too far :) thanks a lot!

PR #975, opened 2017-07-25
share code so module_test improved

tobes:
#806 called attention to problems with `module_test` not using the same code paths as when `py3status` is running. This is due to changes regarding getting config settings and the formatter updates. This causes issues

master (color missing and Composite causing confusion)
```
$ python modules/dpms.py
{'color': None, 'full_text': <Composite [{'full_text': u'DPMS'}]>, 'cached_until': 1500839733.0}
```

this branch (output as we send to i3bar - excluding cached_until which is added for user debugging)
```
$ python modules/dpms.py
{'color': '#00FF00', 'full_text': u'DPMS', 'cached_until': 1500839670.0}
```

Although it looks large there are few changes to how the code works.

__core.py__

We create a new class `Common` which holds common code `report_exception()` and `get_config_attribute()` so that the code can be accessed without `Py3statusWrapper`

Long term I'd like to improve on this so we have better separation between different parts of `Py3status Wrapper` eg cli parsing, running modules etc so we can use it for both running a single module and running all in the config.

__module.py__

This now can either take a module name (as currently) or an actual instance (testing).

A few extra bits to add `cached_until` into test output and to handle errors correctly for testing eg show full traceback on errors and to stop testing when this happens.

__py3.py__

Since we now have module for all uses we can clean up some of this code and make it more universal.

__module_test.py__

Rewrite to use new ability to run module correctly rather than mirroring finding methods etc.

__Longer term__

I'd like to replace the `MockPy3statusWrapper` with a common `Py3StatusWrapper` that supports both testing and running as normal and also allow us to do `py3status --test <module_name/path>` or similar

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -54,6 +54,115 @@ class NoneSetting:
         return 'None'
 
 
+class Common:
+    """
+    This class is used to hold core functionality so that it can be shared more
+    easily.  This allow us to run the module tests through the same code as
+    when we are running for real.
+    """
+
+    def __init__(self, py3_wrapper):
+        self.py3_wrapper = py3_wrapper
+        self.none_setting = NoneSetting()
+        self.config = py3_wrapper.config
+
+    def get_config_attribute(self, name, attribute):
+        """
+        Look for the attribute in the config.  Start with the named module and
+        then walk up through any containing group and then try the general
+        section of the config.
+        """
+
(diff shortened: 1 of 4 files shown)


ultrabug:
Thx, trusting you on this, the diff scared me to hell

tobes:
>  the diff scared me to hell

yeah just looking at it again I feel your pain.  thanks

PR #979, opened 2017-08-01
py3.command_output normalized and enhanced error reporting

ultrabug:
This one could use a better report on the output of the command if available

+ some minor QA

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -858,7 +858,7 @@ class Py3:
                 command, stdout=PIPE, stderr=PIPE, close_fds=True
             ).wait()
         except Exception as e:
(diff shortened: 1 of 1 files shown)


tobes:
hmm,

I think This is something that we could discuss in much detail, but if we go for a simpler `command_output()` as discussed in #971 then it doesn't matter as the exception will be removed.

So I'm happy to merge this and then make that change.

I'm not entirely sure why you are doing the one line stuff as really the exception should have the actual output imho.  also replacing `\n` with a space would probably create more readable output.

ultrabug:
Okay then maybe we need this conversation yea 👍 

Amended commit with space instead of empty string replacement of newline, thanks

tobes:
I've looked at this again and it seems like a fine solution.

PR #982, opened 2017-08-08
volume_status allow device as a int

tobes:
thanks to yabbes for reporting


ultrabug:
Thanks yabbes on IRC 👍

PR #984, opened 2017-08-08
use wait() instead of sleep() in main loop

tobes:
use `wait()` with a timeout rather than `sleep()`

this reduce the amount of overhead and also makes us more responsive as we do the updates exactly when they are needed.

i3status is special so it needs to tell us when it wants to update the times which we still do in the main loop.

i3status stuff can be much improved but we get a nice benefit here and keep the PR easy to review.  later we can do that cleanup.

the test is rather crude we just run py3status for 10 mins but it shows we save cycles

```
time timeout 600 py3status > /dev/null

# branch
real	10m5.061s
user	0m0.800s
sys	0m0.125s

# master
real	10m5.056s
user	0m1.151s
sys	0m0.232s
```

### DIFF
--- a/py3status/i3status.py
+++ b/py3status/i3status.py
@@ -195,19 +195,29 @@ class I3status(Thread):
         self.i3status_pipe = None
         self.time_modules = []
         self.tmpfile_path = None
+        self.update_due = 0
 
     def update_times(self):
         """
         Update time for any i3status time/tztime items.
+        Returns the time till next update needed.
         """
-        updated = []
-        for module in self.i3modules.values():
-            if module.is_time_module:
-                if module.update_time_value():
-                    updated.append(module.module_name)
-        if updated:
-            # trigger the update so new time is shown
-            self.py3_wrapper.notify_update(updated)
+        now = time()
+        if now > self.update_due:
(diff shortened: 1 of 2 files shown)


py3status/i3status.py
```diff
@@ -195,19 +195,29 @@ def __init__(self, py3_wrapper):
...
+            updated = []
+            for module in self.i3modules.values():
+                if module.is_time_module:
+                    if module.update_time_value():
+                        updated.append(module.module_name)
+            if updated:
+                # trigger the update so new time is shown
+                self.py3_wrapper.notify_update(updated)
+                # time we next need to do an update
+            self.update_due = int(now) + 1
```
ultrabug: this is not true that i3status needs a 1s refresh time, we can and should get it from the config I think

tobes: This is because we need to update the i3status `time` module seconds.

This needs to be fixed properly but doing this just keeps things working without confusing this PR .

Once this is merged I can start fixing that up.  But it needs a fair bit of refactoring of `i3status.py`.

ultrabug:
Okay thanks, this will be a great base to start from!

PR #990, opened 2017-08-10
better error catching for missing {

tobes:
the following config broke py3status see #987

```
order += static_string

static_string 
    format = 'py3status test'
}
```


tobes:
updated with better escaping of chars to i3bar

ultrabug:
Awesome, thanks!

PR #991, opened 2017-08-10
allow report exception to be available early

tobes:
Whilst looking at #987 I found that `report_exception()` is not available as early as it is needed.  this PR allows us to use it much earlier in the process. To be honest I think this predates the Common code changes.

We also output the traceback to `stderr` if logging options are not set so that we can see the traceback in the cli


ultrabug:
Thanks, is clean

PR #1002, opened 2017-08-15
xrandr: clean docstring

tobes:
Update xrandr docstring.

To both improve the display and for upcoming documentation work.  Fixes glitches in readthdocs output

To enable `<OUTPUT>_workspaces` etc to be recognized the re used for documentation creation now allows parameter names starting with `<`


ultrabug:
Thanks!

PR #1006, opened 2017-08-17
new profiling abilities

tobes:
Profiling py3status is painful we need it to be nicer for 3.7

a) we have to turn it on manually by editing a file

b) profile is swamped by sleep, poll calls which we do not generally care about

This PR adds a `Profiler` which can be enabled via the `--profile` switch.  We can now mark the code we are interested in profiling using `enable` `disable` calls.  Profile info is converted into a suiable format for `kcachegrind` if module `pyprof2calltree` is installed.

Currently the output is to fixed locations to keep things simple but this can be improved.

Having the enable/disable calls has a slight overhead when profiling is not enabled, however I think that this is acceptable as the advantages of being able to do easy profiling outweigh the cost.

I have another branch that adds a `performance_check` utility that add a standardised config etc so that we can get repeatable output to aid profiling.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -23,7 +23,7 @@ from py3status.helpers import print_line, print_stderr
 from py3status.i3status import I3status
 from py3status.parse_config import process_config
 from py3status.module import Module
-from py3status.profiling import profile
+from py3status.profiling import profile, Profiler
 from py3status.version import version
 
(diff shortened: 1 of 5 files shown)


tobes:
I've worked out how to deal with the sleep/poll issue, and although there is still a problem of not being able to easily enable/disable profiling,  I'm closing this

I should have a new PR that will give improved profiling info but based on the current `@profile` approach.  It means we don't have to change the existing core code to allow better profiling.

It's been good to see where we are being bad and we can definitely improve stuff.

PR #1015, opened 2017-08-23
remove Timers and spend less time in threading

tobes:
This PR reduces the amount of time we spend in threading module by ~50%,  we do increase the work done in core by ~40% but this is a small increase compared to the saved time.

Overall it is about a ~15% saving

on my 10 minute test based on pprofile line profiling.

**master**

py3status module updates: 1595
total cpu time: 1,285,655
threading: 411,426
core time: 86,658

**branch**

py3status module updates: 1622
total cpu time: 1,109,407
threading: 208,338
core time: 123,383

* `Py3StatusWrapper.queue` renamed to `Py3StatusWrapper.update_queue` for readability

* `Module` no longer a `Thread`

* remove the `Timers` in `module.py`

* replace `Py3StatusWrapper.lock` with `Py3StatusWrapper.running` where possible to reduce overhead.

* We now run modules using a simple `Runner` thread when updates needed.

* We store when each module is due for an update and create a `Runner` as needed

Although we are still creating a `Runner` thread every time we get module output we end up with fewer threads running.  In future we can probably make better use of these by reusing them but for now it keeps things simple and we get our big saving.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -12,7 +12,7 @@ from platform import python_version
 from pprint import pformat
 from signal import signal, SIGTERM, SIGUSR1, SIGTSTP, SIGCONT
 from subprocess import Popen
-from threading import Event
+from threading import Event, Thread
 from syslog import syslog, LOG_ERR, LOG_INFO, LOG_WARNING
 from traceback import extract_tb, format_tb, format_stack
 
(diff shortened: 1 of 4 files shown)


ultrabug:
That indeed looks very promising, thanks a lot!

Quick (and late) test shows that:

- Overall click events are very slow
- Grouped modules updating/execution is somewhat hazardous

tobes:
Thanks for the test.  Seems I forgot to add the module to the queue in `Module.force_update()` click events and also updates in groups should now work as expected.

ultrabug:
Thanks, it's indeed better but you can still see clearly the difference between instant module update on click / mouse wheel between master and this branch :(

tobes:
Ok should be good now.  when the update should be instant ie the cache_time we send to ` timeout_queue_add_module()` in core.  we now 'set()` the `Event` so that the the main loop in py3statuswrapper runs straight away.

ultrabug:
Tested awesome, thanks! Let's iterate from there now.

ultrabug:
After suspending and resuming my laptop and since this merge I got

```
Aug 25 18:16:52 localhost /__init__.py: Runtime error (KeyError) core.py line 278.
```

PR #1017, opened 2017-08-24
Cleanups to i3status including reducing workload

tobes:
Some minor cleanups to `i3status.py`

* setup things like `I3StatusModules` in advance rather than in the loop checking `i3status` output.

* only check timezone info on first run and then on the half hour rather than every minute.  This also improves checking after sleep/suspend (timezone checking is costly due to `strptime()`)

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -215,10 +215,10 @@ class Py3statusWrapper:
         # these are used to schedule module updates
         self.timeout_update_due = deque()
         self.timeout_queue = {}
+        self.timeout_queue_lookup = {}
         self.timeout_keys = []
-        self.timeout_queue_members = set()
 
-    def timeout_queue_add_module(self, module_name, cache_time=0):
+    def timeout_queue_add_module(self, module, cache_time=0):
         """
         Add a module to the timeout_queue if it is scheduled in the future or
         if it is due for an update imediately just trigger that.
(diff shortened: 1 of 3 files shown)


ultrabug:
Test NOK due to i3status clock behaving strangely: seconds pass very fast or slower, the tick is not consistent somehow.

At first glance, I can't understand why but this is what I can see from switching between master and this branch :(

I experienced this a lot when I started working on py3status, this is almost a nightmare :)

tobes:
I've update this branch.  We now get super smooth time.

the timeout queue now gets the module rather than it's name that makes things easier.  I've also cleaned things up so hopefully they are simpler.  Also fixed the bug from #1015 #1021
If it is easier I can split this PR.

i3status time tztime modules now add themselves to the timeout queue so they get updated when needed.

ultrabug:
Do we agree that if a user module is queued right before a time module, the user module's time to complete will affect the time module ?

If my current understanding of this is right, I guess we need a thread dedicated to time modules

tobes:
Yeah we do. But the time is minimal so the thread cost is probably more. I
can put it in a thread if you you prefer. I'll check if it makes a huge
difference. If not I'll do that.  I need to add my 'test suite' too

ultrabug:
Okay, I will try to create a module that sleeps on update to check the behavior

Meanwhile, I've noticed a bug with my cycling TZ group
```
group tz {
	cycle = 10
	format = "{output}"

	tztime la {
		format = "LA %H:%M"
	}

	tztime ny {
		format = "NY %H:%M"
		timezone = "America/New_York"
	}

	tztime du {
		format = "DU %H:%M"
		timezone = "Asia/Dubai"
	}
}
```

If you scroll this with the mouse, you'll see the raw output of the tztime module parsing instead of the formatted value ! weird isn't it ?

tobes:
> you'll see the raw output of the tztime module

Ok this was hard to find, but was due to the value from i3status being saved as well as the 'corrected/fake' value.  I've changed things in  `I3statusModule.update_from_item()` so that the time values are not saved which fixes the issue.

ultrabug:
Thanks mate, hard to find indeed

> Do we agree that if a user module is queued right before a time module, the user module's time to complete will affect the time module ?

I have tried to produce this, but failed, if a module sleeps it does not block the time from being updated. I guess I'm missing the core concept of why this is working when I thought it would be an issue?

Example module
```python
# -*- coding: utf-8 -*-
"""
"""

from time import sleep
from random import randint


class Py3status:
    """
    """
    cache_timeout = 10

    def on_click(self, event):
        """
        """
        time_slept = randint(1, 5)
        sleep(time_slept)

    def _do_work(self):
        with open('/dev/shm/test', 'w') as fp:
            for _ in range(1000000):
                fp.write('%s this is a line\n' % _)

    def sleeper(self):
        """
        """
        time_slept = randint(1, 5)
        sleep(time_slept)
        # self._do_work()
        response = {
            'cached_until': self.py3.time_in(time_slept),
            'full_text': 'zz %s' % time_slept
        }
        return response


if __name__ == "__main__":
    """
    Run module in test mode.
    """
    module_test(Py3status)
```

tobes:
Py3status modules run in a thread so they should not block things whilst
they run. The i3status time is just updated as it is very quick. The output
from i3status is still in its own thread so it should not affect the main
loop either.

ultrabug:
Different thread for i3status & modules, ofc! /me facepalms

Thanks mate

PR #1022, opened 2017-08-28
[WIP] Implement graph primitives using characters

guiniol:
This is still a work in progress. I'm opening the PR for discussions.

I added 2 methods + 1 helper to `py3.py`.

The helper transforms a value (usually a percent, but could be something else) to a block of the appropriate length. There are 8 different blocks and the partition is as follows (for percents): `[0,12.5[`, `[12.5, 25[`, `[25, 37.5[`, `[37.5, 50[`, `[50, 62.5[`, `[62.5, 75[`, `[75, 87.5[`, `[87.5, 100]`.

The base case for `concurrent_bar_graph` is to show the used percent per CPU core. You give it a list of values, and it gives you a composite with each value transformed to a block.

The base case for `history_bar_graph` is to show the historic network usage. So you give it a value and a list to store the history in. By default it will start by filling the history to the max length with the default value and will then add the new value, and will then remove old values to keep within the length.

The attached test.zip has a pretend module for each of the functions.
What do you guys think? Is the idea useful? Is the code written the way it should? Do you have better name ideas?

EDIT: the ideas came from a friend showing me polybar, and I realised they are just using characters. The battery module is already kind of doing that, but by adding it in `py3.py`, it makes it easier for every module to use it. Also, I'm planning on adding horizontal graphs.

[test.zip](https://github.com/ultrabug/py3status/files/1258344/test.zip)


py3status/py3.py
```diff
@@ -1008,3 +1008,73 @@ def request(self, url, params=None, data=None, headers=None,
...
+
+        :param history: history of the variable
+        :param value: value to add to the history
+        :param length: length of the history
+        :param default_value: default value to pad with
+        :param min_val: minimum value possible
+        :param max_val: maximum value possible
+        """
+
+        if default_value is not None and length is not None:
```
lasers: Can you explain this part? Could `default` and `length` be `None`? The only way those get `None` is if we give them one... What will happen if we give them `None`? Same thing for `if length is not None:` below.

I think it makes sense if we defaults them `0` and `1`, then we don't have to worry about parameters on other modules if we add them... eg single value = single bar. Maybe that's the plan already. Idk.

guiniol:
Thanks for the support, and thank you so much for the gifs!!

I'll put the pretend modules with the other modules for the tests.

py3status/py3.py
```diff
@@ -1008,3 +1008,73 @@ def request(self, url, params=None, data=None, headers=None,
...
+        bars = ['▁', '▂', '▃', '▄', '▅', '▆', '▇', '█']
+
+        idx = int((value - min_val) / (max_val - min_val) * len(bars))
+        if idx == len(bars):
+            idx -= 1
+        full_text = bars[idx]
+        color = self.threshold_get_color(value)
+        return {'full_text': full_text, 'color': color}
+
+    def concurrent_bar_graphs(self, values, min_val=0, max_val=100):
```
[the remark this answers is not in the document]
guiniol: The idea with this function and the other is just to make it really easy to use the bars. No need to think about how you want to handle them. Just send a bunch of values to the function and it just works!

Prime candidates are CPU/RAM/Network, but the `concurrent_bar_graphs` could be used for the volume if you want to show the volume of several outputs at once for example.

py3status/py3.py
```diff
@@ -1008,3 +1008,73 @@ def request(self, url, params=None, data=None, headers=None,
...
+                                             min_val=min_val,
+                                             max_val=max_val))
+
+        return self.composite_create(val_bars)
+
+    def history_bar_graph(self, history, value, length=5, default_value=0, min_val=0, max_val=100):
+        """
+        Transform a value into a bar graph with history.
+
+        :param history: history of the variable
```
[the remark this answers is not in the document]
guiniol: I tried both, but figured it would be better to have the module handle the history. That way, they can tinker with it, and if the module has several such graphs, we're sure there won't be any collision.

I'm also unsure whether we want to store the bar or the raw value. If we store the bar, that's less computation, but if we store the value, then we could change the min/max halfway through and recompute the right bar (this could be used for networks, where there isn't really a maximum, so we could make it dynamic depending on the last few values or something). It also makes it easier for the module to tinker with the history if they have access to the raw values.

py3status/py3.py
guiniol: This is just in case something goes wrong (i.e. someone sets them to `None`). It doesn't make much sense, TBH. I agree with the default to `0` and `1`.

py3status/py3.py
```diff
@@ -1008,3 +1008,73 @@ def request(self, url, params=None, data=None, headers=None,
...
+
+        if default_value is not None and length is not None:
+            while len(history) < length:
+                history.append(self._val_to_bar(default_value,
+                                                min_val=min_val,
+                                                max_val=max_val))
+
+        history.append(self._val_to_bar(value,
+                                        min_val=min_val,
+                                        max_val=max_val))
```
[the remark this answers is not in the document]
guiniol: I usually use the names of named arguments. I find it better in case their order change at some point. But I don't really care if we want to remove them.

Same thing for abbrevs, I can change that.

guiniol:
@lasers so, I added `cpu_concurrent_bar` and `cpu_history_bar` to the `sysdata` module if you want to try with real values.

Also, I was thinking, we could add the possibility to add a space between each bar or something. We could either change the `bars` (I haven't added that possibility to `sysdata`), or we could add a subformat. I don't see any advantages to adding the subformat, so I'm all for the first solution. `bars` can default to `None` in which case it will use the defaults from `py3.py`.

lasers:
>I don't see any advantages to adding the subformat, so I'm all for the first solution. bars can default to None in which case it will use the defaults from py3.py.

I think @tobes would like to have this done right (eg second solution). If we do the first solution, we can get ` ` or `, ` at the end of the `concurrent_bar` and the another irk issue is that we would have to add ` ` or `, ` to all items in the `bars = []`. That will look unkempt easily. Also, more spaces = more unkempt.

We can just do this right the first time... even if nobody want to use it... or if somebody want spaces to make things a bit more readability.
* Volume slider adjusting... `---|-------`  // `- - - | - - - - - - -`   (30%)

Btw, that's a second idea for you... Horizontal graph... and now a slider. 👍 `Volume ───|───────`

if `separator=None (default)`, skip... else... we use `composite_join`.

py3status/py3.py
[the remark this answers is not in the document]
lasers: >I'm also unsure whether we want to store the bar or the raw value...

Could be decided based on `history`. If `history=None`, use bar. Otherwise, use raw.

guiniol:
> think @tobes would like to have this done right (eg second solution). If we do the first solution, we can get or , at the end of the concurrent_bar and the another irk issue is that we would have to add or , to all items in the bars = []. That will look unkempt easily. Also, more spaces = more unkempt.
> 
> We can just do this right the first time... even if nobody want to use it... or if somebody want spaces to make things a bit more readability.
> 
>     Volume slider adjusting... ---|------- // - - - | - - - - - - - (30%)
> 

I really like that separator idea. Plus, I imagine `composite_join` does exactly what we need. I'll add it tonight.

> Btw, that's a second idea for you... Your horizontal graph... and now a slider. 👍 Volume ───|───────
> 
Exactly! That was my last idea. I have a few questions. The idea is to have single method with which we can obtain something like `────┼───` or `|||||||    ` by specifying different characters. Basically, have one character for the left part, one for the indicator and one for the right part (we could also add ones for the extremities to get something like `├────┼───┤`, but that may be overdoing it). Then we have the problem of the colours. For each part, we could want either a static colour (say, white, red, white), or we could want to use thresholds instead, or any combination thereof, like that for example screenshot from 2017-08-30 09-15-57). Any ideas on how to specify either a static colour or the thresholds for each part?

py3status/py3.py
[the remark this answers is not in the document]
guiniol: if `history=None` where do we store it? I think storing the raw value is fine, creating the composites each time shouldn't be too taxing.

lasers:
Uhh... I'm a n00b... I don't know if this would be the way `py3status` would want to go because I don't know if they like `one format to rule them all` policy. Here, we would be doing something not so simple... for the users. More customization and/or rainbows though. 🌈 

This simply break the friendly user interface, but I imagine that we would stop using `threshold_get_color` and start colorizing them ourselves using `safe_format` on `bar = [idx]`. ;-)

Bar replaceable... (untested)

```
    bars = [
        '\?color=#000000  \u2581',
        '\?color=threshold \u2581',
        '\?color=threshold \u2582',
        '\?color=threshold \u2583',
        '\?color=threshold \u2584',
        '\?color=threshold \u2585',
        '\?color=threshold \u2586',
        '\?color=threshold \u2587',
        '\?color=threshold \u2588',
    ]
```
Rainbow bars... (untested)
```
    bars = [
        '\?color=None \u2581',
        '\?color=#CE93D8 \u2581',
        '\?color=#B39DDB \u2582',
        '\?color=#9FA8DA \u2583',
        '\?color=#90CAF9 \u2584',
        '\?color=#81D4FA \u2585',
        '\?color=#80DEEA \u2586',
        '\?color=#80CBC4 \u2587',
        '\?color=#A5D6A7 \u2588',
    ]
```
What you're asking for... (partially completed ; untested)
```
    bars = [
        '\|[\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '\?color=good \|[\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '\?color=good \|\|[\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '\?color=good \|\|\|[\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '\?color=good \|\|\|\|[\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '\?color=good \|\|\|\|\|[\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '[\?color=good \|\|\|\|\|][\?color=degraded&show \|][\?color=#000&show \|\|\|\|\|\|\|\|\|\|\|\|\|\|]',
        '[\?color=bad&show fill in the blanks]'
    ]
```

guiniol:
@lasers i wasn't very clear when asking for help sorry ^^
It's actually easy to colour according to threshold with `threshold_get_color`, and since we have a composite per character each character can be coloured to match its value. The problem is how to apply a static colour. That, I do not know what to do, but I should be able to figure it out by looking at the battery module. I'll let you know how it goes tonight!

py3status/py3.py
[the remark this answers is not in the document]
lasers: In same place you put `self._default_bars`. Call it `self._index_cache, I guess.

I think in the long run, we might want to keep `bars` inside the modules. (Only 1 so far)... so maybe no more `self._default_bars`... kinda like `clock.py` and its blocks.

tobes:
@guiniol nice idea.  I did an implementation previously I'll try to dig it up but used a class to keep things self contained.  I'm not saying that is a better method just what I did. 

I should have some time at the weekend to look at your code in detail.

lasers:
>The problem is how to apply a static colour

In my examples... The replaceable bars are full of thresholds. The rainbow bars are full of static colors. They both starts off with static color using `\?color=None`. If you can't figure it, try `safe_format` and my examples. It should get you something you're asking for.

guiniol:
@lasers I'll give it a shot tonight

@tobes I'm open to suggestions. `py3.py` seems to be a helper class, so that's why I put it there, but it can be put in other places too (I actually wondered if it should be in `Formatter` at the beginning)

tobes:
@guiniol util.py might be a good home with the methods available via Py3

guiniol:
@tobes alright, I'll move them there.

guiniol:
@tobes looking at `util.py`, if I follow your drift, you want me to create a class like `Gradients` but called `BarGraphs` or something and put the various helper methods in there. Then, I put a few high level functions (like the current `concurrent_bar_graph` or `history_bar_graph`) which will use this `BarGraphs` class. Is that what you were thinking ?

guiniol:
@tobes I was almost through with that until I realised I needed `threshold_get_color` which is defined `py3.py`, so I'm leaving everything there for now.

guiniol:
@lasers the progress bar has been implemented. Easily tested with the `pretend_ram` module.

Add this to your config to completely change the way the progress bar looks:
```
pretend_ram {
	left_char = '|'
	right_char = '|'
	middle_color = "#bb88dd"
	right_color = "#d0d0d0"
	length = 20
}
```

What do you think ?

lasers:
With `timeout_cache = 0` and `length = 20`...
r0
From left to right... `Original bar`, `experiment bar`, `experiment full block`.

This is nice. If you need `threshold_get_color` inside py3, then all of your methods so far maybe should have `threshold_get_color`'s `name=None` parameter too... so users can pass something different. 

With `timeout_cache = 0` and `length = 150`...
r1

py3status/py3.py
```diff
@@ -1008,3 +1009,165 @@ def request(self, url, params=None, data=None, headers=None,
...
+        :param separator: character to insert between the bars
+
+        :returns: Composite list
+        """
+
+        if bars is None:
+            bars = self._default_bars
+
+        while len(history) < length:
+            history.append(default)
```
lasers: `history = [default for i in range(length - 1)]` works?

py3status/py3.py
```diff
@@ -1008,3 +1009,165 @@ def request(self, url, params=None, data=None, headers=None,
...
+
+        history.append(value)
+
+        while len(history) > length:
+            history = history[1:]
```
lasers: `del history[length - 1:]` works?

py3status/py3.py
```diff
@@ -1008,3 +1009,165 @@ def request(self, url, params=None, data=None, headers=None,
...
+
+        return self.composite_join(separator,
+                                   [self._value_to_bar(x,
+                                                       minimum,
+                                                       maximum,
+                                                       bars)
+                                    for x in history])
```
lasers: Easy on the eyes.
```
         return self.composite_join(separator, [self._value_to_bar(
            x, minimum, maximum, bars) for x in history])
```

py3status/py3.py
```diff
@@ -1008,3 +1009,165 @@ def request(self, url, params=None, data=None, headers=None,
...
+            i_val = self._index_to_value(i, length, minimum, maximum)
+            if left_color is None:
+                color = self.threshold_get_color(i_val)
+            else:
+                color = left_color
+            progress.append({'full_text': left_char, 'color': color})
+
+        if middle_color is None:
+            color = self.threshold_get_color(value)
+        else:
```
lasers: We waste 20 loops doing `i_val = self._index_to_value(i, l, m, m)` if `left_color` is not `None`.
We waste 20 loops doing `i_val = self._index_to_value(i, l, m, m)` if `right_color` is not `None`.

Let's check the colors first before we do anything with `i_val`.

py3status/py3.py
guiniol: hum. This one is actually tricky. We want to pad the beginning of the history, not the end, but yes, we should just compute how many we need and add them, not loop. That's sloppy.

py3status/py3.py
guiniol: it should (without the `- 1`, I think)

guiniol:
Thanks for the gifs again.

I'm out for the night, but your comments are accurate, I'll fix that tomorrow.

py3status/py3.py
```diff
@@ -1008,3 +1009,161 @@ def request(self, url, params=None, data=None, headers=None,
...
+        :param bars: characters to use for the different levels of the bar
+        :param separator: character to insert between the bars
+
+        :returns: Composite list
+        """
+
+        if bars is None:
+            bars = self._default_bars
+
+        history += [default] * (length - len(history)) + history
```
guiniol: the `+=` here is needed so that the reference stays the same. Otherwise, `history` starts pointing to a new object and the original `history` is never updated. Any better way to do that?

guiniol:
@lasers addressed your concerns (there's a comment where I still have one...)
I also added the optional threshold names.

Also, I forgot that I changed the defaults in `pretend_ram`, so my `Add this to your config to completely change the way the progress bar looks` comment must have looked weird, but if you retry `pretend_ram` now that I removed the default values, you'll understand ^^

py3status/py3.py
lasers: We can stick with the old one if we can't optimize it better. The old one also works fine and is well-readable.

lasers:
More gifs for others to enjoy.
s0

guiniol:
@lasers found a good way to do prepend the default value.

As far as the helper functions are concerned, I think we're good to go. What do you guys think?

Do you also want me to add the bars to a few modules ?
I will personnaly use it for the network and the per cpu-core bars. But I can also see people using it for every measure in sysdata, the volume, disk and battery (actually, I think battery already does that, so it would be a matter of converting it maybe?). So, I can put bars in each of these modules. Do we leave it at that and let others progressively add bars where they want them to be?

lasers:
>As far as the helper functions are concerned, I think we're good to go. What do you guys think?

Up to others. My mere opinion... I think if we were to make something like `format_index`... then we can make our own bars/blocks/clocks/others that could be just added in all modules and we wouldn't have to write any new methods... and not having to add no more than 1 new config and N* placeholders. I am not even sure if this idea would work or even if it makes sense.

>Do you also want me to add the bars to a few modules ?

If you want to... As it stands right now, it could be a while before this PR and many other PRs to get merged. No worries. Just keep putting out the new and the fun stuffs. This might get merged in `3.8` as they want to focus on other things: [3.7 milestone](https://github.com/ultrabug/py3status/milestone/6). Same with my new modules. :(

>Do we leave it at that and let others progressively add bars where they want them to be?

Maybe they won't. It's up to us... to enhance things... for fun.

>EDIT: bonus question: `blocks`, `block`, `bars`, `bar`

Whatever sounds good. Change it if you change your mind. And again. And again. Tweak away.

tobes:
> As far as the helper functions are concerned, I think we're good to go. What do you guys think?

I think there are some great ideas here but I think that the code needs a bit of work.

It neeeds to be Python 2 friendly which it isn't but that is not a big deal for now.

It'd be nice to have some tests.  We really should be having tests for all our code but I think it is good to have for new code.

I'd like there to be better encapsulation, maybe that's just me but I don't like the passing of history etc,  why should the module writer need to care about that?  I think that the graph having state is useful.

Do we need to have all these different functions or are they just different styles eg should the module user have more say in things than the module writer.  eg if sysdata tracks cpu then as a user should I not be able to chose what sort of graph I see (and how do we do that nicely/ for multiple graphs in the same module).

So imho we are still at a how do we want the interface to the functionality phase.

> Do you also want me to add the bars to a few modules ?

I think it is best to get this PR sorted first (will take a while)

As I said I did some work on this style of idea before I've pushed the branch here https://github.com/ultrabug/py3status/compare/graphs?expand=1 but as you can see it is also a work in progress especially the demo modules.

guiniol:
@lasers @tobes I understand what you mean.

So, new idea: currently, we can do something like `{cpu:2.1f}`, what if we could replace the `2.1f` part with a `hist` or `vbar` or `hbar` ? I'll look into it, but it should be doable in `safe_format`. The question is, how do we specify which characters/colors to use for the bars (and this needs to be done per bar because we could have a bunch of different bars in a single module e.g. sysdata). This also makes it mandatory that the history is handled transparently. Question: how do we handle concurrent bars (eg one bar per core) ? Do we have `vbar` be a single bar if the value is a float and multiple bars if the value is a list? Is that possible?

Also, @tobes, I see you made a segmented horizontal bar, so that's something I should add (but after we figure out the rest of the details).

guiniol:
@lasers @tobes 
So this has fallen a bit to the wayside, but I have been trying things, and they have not worked.
My most promising attempt is to use the functions. So, basically, add one function per type of graph. The only issue I have with that is colours. If you guys agree that functions are the way to go, I'll make it work.

tobes:
>  If you guys agree that functions are the way to go

I don't think they are as then the module has to worry about state and I think that should be the responsibility of the graph itself

guiniol:
^^
OK. That's one the things I hit a wall on...
Going back to the drawing board.

lasers:
Rebase, plz. I want to look at this again. :fire: :sunglasses:

guiniol:
I must admit I hadn't touched anything for a long while.
I rebased and after a few tweaks to make it more similar to the currently existing code, it works.
However, I have not addressed any issues. I am not even sure which issues are still worth addressing, so maybe we could start anew on the comments.
In any case, here it is:
- it works
But:
- there are currently 3 functions for 3 types of graphs.
- there are no tests.

lasers:
@guiniol There is a conflict. Can you fix it? This PR is too cool to let it die. Thx. :-)

guiniol:
@tobes totally agree, but I don't know that I will have the time to do that
@lasers I haven't had a lot of time these last few months and will have even less for a few more months. Do you want to pick up this CL and actually make it happen?

tobes:
@lasers

> I'll give it a shot. 

The code is the least important part.  I hope you know this

guiniol:
> @guiniol I'll give it a shot. What is `CL`?

Sorry, meant PR. Too much time at work 😀

ultrabug:
Timeout, closing, reopen if needed mate; Thanks!

PR #1024, opened 2017-08-31
xrandr: use post_config_hook

lasers:


ultrabug:
Clean, pertinent and simple PR, thanks!

PR #1029, opened 2017-09-02
make timeout queue thread safe

tobes:
#1021 seems to be due to a race.

We now use a lock to prevent the race.

We try to avoid holding the lock if we can

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -12,7 +12,7 @@ from platform import python_version
 from pprint import pformat
 from signal import signal, SIGTERM, SIGUSR1, SIGTSTP, SIGCONT
 from subprocess import Popen
-from threading import Event, Thread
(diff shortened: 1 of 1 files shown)


ultrabug:
Quick look at this and the theory defeats me. Since all of the proposed lock logic lies in the core and that the core runs on its own "thread" I don't see what parallel threading we're trying to solve here?

To my quick understanding, there's no parallelism within the core methods here.

tobes:
That is what I initially thought but I can easily create a race by adding a sleep in `timeout_queue_add_module()` not the same as affecting you but close enough that I'm sure it is the issue.  That function is called by modules so even thought the code is in `core` it is not in the main thread.

tobes:
closing in favour of #1064

PR #1031, opened 2017-09-02
move thread checks out of core main loop

tobes:
Currently we check event and i3status threads are alive in the main loop.  This moves these checks into the actual modules.  Also remove some other cruft in the loop that is no longer needed.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -922,7 +922,6 @@ class Py3statusWrapper:
         signal(SIGTERM, self.terminate)
 
         # initialize usage variables
-        i3status_thread = self.i3status_thread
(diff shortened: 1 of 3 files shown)


ultrabug:
schedule every 5s the i3status thread aliveness plz

ultrabug:
friendly bump @tobes

tobes:
@ultrabug should be good to go now

ultrabug:
Awesome, if I get it right we always restart i3status now, but every 5s

tobes:
> if I get it right we always restart i3status now, but every 5s

ahh I just thought this was the merge conflict I made #1120 for this :)

PR #1054, opened 2017-09-03
exchange_rate: allow for bad json being returned

tobes:
fix for #1052 

Sometimes we get bad json back in a response.  Now we will just keep the old values when this happens.


tobes:
I've found the real problem and working on a much better fix

Issue #1079, opened 2017-09-16
Py3status Battery_level notifications formatted incorrectly


Cypher1:
I'm having an issue where the string sent by the battery_level module (to my notifications daemon) has been mangled.
When I click to get a notification containing something like:
```
<Composite [{'full_text': '00:53 till fully charged. (47*)'}]>
```
Rather than just
```
00:53 till fully charged. (47*)
```

I'm running: **py3status version 3.6** and **python 3.5.2**

Here is the relevant section of my i3status.conf:
```
battery_level {
  format_notify_charging = "{time_remaining} till fully charged. ({percent}%)"
  format_notify_discharging = "{time_remaining} remaining ({percent}%)"
  notification = True
  notify_low_level = True
}
```

lasers:
Confirmed. @tobes I think cb63abf8c broke this because `notify_user` is getting a composite instead of a `string`. Do we have something to simplify a composite into a string? Should `notify_user` even check for a composite and break it down into a string... or I think we can just do that ourselves.

Also, if not possible, do we rename `format_notify_charging` and `format_notify_discharging` since there might be no gain in safeformatting the values? Other than length and such...

@Cypher1 This module definitely could be updated... New thresholds, remove hardcoded strings, simplifications, and whatnot.

Cypher1:
I'm all for safer formatting etc. I do have a concern that the formatting rules are not well known and that they make the configs harder to read but the flexibility that they offer is nice.
I've changed this in my fork to just use format.format(<kwargs>) rather than safe format.
It's not as nice in practice but it works for my everyday usage.

tobes:
@Cypher1 thanks for reporting this.  I've made a PR #1081 to fix the issue.

Cypher1:
Thanks @tobes 

Closed in #1081

lasers:
@tobes I think if we want to deprecate the thresholds in this module, we might need something like `self.py3.threshold_get_name` or a new parameter for `self.py3.threshold_get_color` where it can return `bad`, `degraded` or `full` after we give it a number to determine.

Current
```
     threshold_bad: a percentage below which the battery level should be
         considered bad
         (default 10)
     threshold_degraded: a percentage below which the battery level should be
         considered degraded
         (default 30)
     threshold_full: a percentage at or above which the battery level should
         should be considered full
         (default 100)
```
Ideal
```
thresholds = [(10, 'bad'), (30, 'degraded'), (100, 'good')]
```
[...]

PR #1081, opened 2017-09-17
only show text of Composite when notifying user

tobes:
A simple fix for #1079

`Py3.notify_user()` only shows the text if it gets a Composite as input

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -496,6 +496,8 @@ class Py3:
         rate_limit is the time period in seconds during which this message
         should not be repeated.
         """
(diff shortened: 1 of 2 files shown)


py3status/py3.py
```diff
@@ -496,6 +496,8 @@ def notify_user(self, msg, level='info', rate_limit=5):
         rate_limit is the time period in seconds during which this message
         should not be repeated.
         """
+        if isinstance(msg, Composite):
+            msg = msg.text()
         # force unicode for python2 str
```
Cypher1: This is awesome. Thanks

PR #1088, opened 2017-09-26
Clear clipboard onclick and optional log (xsel)

JohnAZoidberg:
I don't like old string lying around in my clipboard visible all the time and if it is a password even more so. That is why you can now click the string to clear it out.
To not lose the old copied strings one can optionally turn on logging.

### DIFF
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -22,9 +22,6 @@ Requires:
 @author Sublim3 umbsublime@gamil.com
 @license BSD
 
-SAMPLE OUTPUT
-{'full_text': 'selected text'}
-
 Example:
 xsel {
   max_size = 50
@@ -32,6 +29,9 @@ xsel {
   on_click 1 = "exec xsel --clear --clipboard"
   log_file = "~/.local/share/xsel/clipboard_log"
 }
+
+SAMPLE OUTPUT
+{'full_text': 'selected text'}
 """
 
 
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -13,8 +13,7 @@ Configuration parameters:
     max_size: strip the selection to this value (default 15)
     symmetric: show beginning and end of the selection string
         with respect to configured max_size. (default True)
-    log_file: where to log the clipboard, default doesn't log at all
-        (default None)
+    log_file: specify the clipboard log to use (default None)
 
 Format placeholders:
     {selection} output from clipboard command
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import re
 import time
 """
 Display X selection.
@@ -55,7 +54,7 @@ class Py3status:
                 f.write("{}\n{}\n".format(datetime, selection))
         self.selection_cache = selection
 
-        selection = re.sub(r"\s+", " ", selection)  # merge whitespace into one
+        selection = ' '.join(selection.split())
         if len(selection) >= self.max_size:
             if self.symmetric is True:
                 split = int(self.max_size / 2) - 1
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -30,6 +30,7 @@ xsel {
   max_size = 50
   command = "xsel --clipboard --output"
   on_click 1 = "exec xsel --clear --clipboard"
+  log_file = "~/.local/share/xsel/clipboard_log"
 }
 """
 
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -6,7 +6,7 @@ Display X selection.
 
 Configuration parameters:
     cache_timeout: refresh interval for this module (default 0.5)
-    command: the clipboard command to run (default 'xsel -output')
+    command: the clipboard command to run (default 'xsel --output')
     click_command: the clipboard command to run on click
         (default 'xsel --clear --clipboard')
     format: display format for this module (default '{selection}')
@@ -38,7 +38,7 @@ class Py3status:
     """
     # available configuration parameters
     cache_timeout = 0.5
-    command = 'xsel -output'
+    command = 'xsel --output'
     click_command = 'xsel --clear --clipboard'
     format = '{selection}'
     max_size = 15
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -6,8 +6,6 @@ Display X selection.
 Configuration parameters:
     cache_timeout: refresh interval for this module (default 0.5)
     command: the clipboard command to run (default 'xsel --output')
-    click_command: the clipboard command to run on click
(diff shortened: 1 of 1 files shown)


py3status/modules/xsel.py
```diff
@@ -32,13 +38,25 @@ class Py3status:
...
     cache_timeout = 0.5
-    command = 'xsel -o'
+    command = 'xsel -output'
+    click_command = 'xsel --clear --clipboard'
     format = '{selection}'
     max_size = 15
     symmetric = True
+    log_file = None
+
+    selection_cache = None
```
lasers: This should go under `post_config_hook`. See `twitch_streaming` (broken module) for an example.

Also, we could add `self.log_file = os.path.expanduser(self.log_file)` too to deal with `~/xsel.log` or `~/.config/py3status/xsel.log`.

py3status/modules/xsel.py
```diff
@@ -32,13 +38,25 @@ class Py3status:
...
-        selection = self.py3.command_output(self.command)
+        selection = self.py3.command_output(self.command).strip()
+
+        if self.log_file and selection and selection != self.selection_cache:
+            with open(self.log_file, "a") as f:
+                datetime = time.strftime("%Y-%m-%d %H:%M:%S")
+                f.write("{}\n{}\n".format(datetime, selection))
+        self.selection_cache = selection
+
+        selection = re.sub(r"\s+", " ", selection)  # merge whitespace into one
```
lasers: `selection = ' '.join(selection.split())` works? See if we can avoid `import re`. I have this PR too.

py3status/modules/xsel.py
```diff
@@ -1,14 +1,20 @@
...
     cache_timeout: refresh interval for this module (default 0.5)
-    command: the clipboard command to run (default 'xsel -o')
+    command: the clipboard command to run (default 'xsel -output')
+    click_command: the clipboard command to run on click
+        (default 'xsel --clear --clipboard')
     format: display format for this module (default '{selection}')
     max_size: strip the selection to this value (default 15)
     symmetric: show beginning and end of the selection string
         with respect to configured max_size. (default True)
+    log_file: where to log the clipboard, default doesn't log at all
```
lasers: (Opinion) Avoid `default doesn't log at all`. Use `specify the clipboard log to use (default None)`

py3status/modules/xsel.py
```diff
@@ -32,13 +38,25 @@ class Py3status:
...
     symmetric = True
+    log_file = None
+
+    selection_cache = None
 
     def xsel(self):
-        selection = self.py3.command_output(self.command)
+        selection = self.py3.command_output(self.command).strip()
+
+        if self.log_file and selection and selection != self.selection_cache:
```
lasers: (Optional) For every `i3-msg restart`, the log will get a date with `(selection)`. Sometimes same `(selection)`. We could be a bit clever by reading the last `(selection)` in the file instead of using `self.selection_cache = None` in the `post_config_hook`.

py3status/modules/xsel.py
```diff
@@ -32,13 +38,25 @@ class Py3status:
     """
     # available configuration parameters
     cache_timeout = 0.5
-    command = 'xsel -o'
+    command = 'xsel -output'
+    click_command = 'xsel --clear --clipboard'
```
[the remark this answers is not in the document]
lasers: @tobes (Nitpicking) We could support `on_click any = 'xsel --clear --clipboard'`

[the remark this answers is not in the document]
tobes: We could but I'm not a fan.  Maybe `on_click 1 2 3 4 5 = 'xsel --clear --clipboard'` could be done at some point as more flexible.  I don't see this as a priority.

I do agree that this should be done via the config rather than as part of the module.

ultrabug:
Thanks a lot for your contribution @JohnAZoidberg 

Could you please tell us what you think about the comments above?
Also, please rebase your PR so we can move forward

Cheers!

py3status/modules/xsel.py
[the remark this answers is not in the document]
JohnAZoidberg: Ohh nice, I was not aware of that functionality.

py3status/modules/xsel.py
JohnAZoidberg: Great idea!

py3status/modules/xsel.py
JohnAZoidberg: But I would still keep `self.selection_cache = None` to avoid file reads all the time. I would just add another case to the if at the end to check whether the current `selection` is already present. This should only read the file when py3status starts up.

tobes:
I think it'd be really nice if we had an example that used the file logging

JohnAZoidberg:
Absolutely

py3status/modules/xsel.py
```diff
@@ -22,8 +25,13 @@
...
 
-example
-{'full_text': 'rrrrr > wtf is a warlock doing in here'}
+Example:
+xsel {
+  max_size = 50
+  command = "xsel --clipboard --output"
+  on_click 1 = "exec xsel --clear --clipboard"
+  log_file = "~/.local/share/xsel/clipboard_log"
+}
```
tobes: @JohnAZoidberg 

One last thing can you move you Example above the `SAMPLE OUTPUT` line and replace the
```
example
{'full_text': 'rrrrr > wtf is a warlock doing in here'}
```

These are used for creating the documentation see https://py3status.readthedocs.io/en/3.6/modules.html#xsel

lasers: `{'full_text': '<grrrrr > wtf is a warlock doing in here'}` pls

py3status/modules/xsel.py
lasers: If I restart `i3` occasionally to play with `py3status` modules, my log would looks like this.
```
2017-10-23 06:48:00
~
2017-10-23 06:48:18
~
2017-10-23 06:48:27
~
2017-10-23 06:48:36
~
2017-10-23 06:48:45
~
2017-10-23 06:48:59
~
```

py3status/modules/xsel.py
JohnAZoidberg: What do you want to do with `{'full_text': '<grrrrr > wtf is a warlock doing in here'}`? Add it back in?
Does this really add value?

lasers: >Does this really add value?

Yes. The images in that link comes from the modules themselves.

We generate the screenshots with that. It is a good real-time example for `xsel` too as you had this in your clipboard and probably in your log too. All modules have a `SAMPLE OUTPUT` strings.

tobes: @JohnAZoidberg use your judgement the main thing is to move your example above the sample output

tobes:
@JohnAZoidberg Thanks for this.  There are a couple of small fixes that were needed, which I've done as #1144

JohnAZoidberg:
Great! Thanks :)

PR #1110, opened 2017-10-14
Add YubiKey module (show a visible indication that YubiKey is waiting for a touch)

maximbaz:
This module is designed to show a visible indication that YubiKey is waiting for a touch.

Basically it uses a bunch of hacks to solve [this issue](https://forum.yubico.com/viewtopic.php?f=35&t=2397). 
Owners of YubiKey, especially YubiKey 4 Nano would be extremely glad for this module 🙂

How it works:
- It requires [maximbaz/yubikey-touch-detector](https://github.com/maximbaz/yubikey-touch-detector) to run in the background and subscribes to updates from it via unix sockets.

**DEMO**:

yubikey_indicator

### DIFF
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -19,11 +19,10 @@ Control placeholders:
 SAMPLE OUTPUT
 {'color': '#00FF00', 'full_text': 'YubiKey'}
 
-Dependencies:
-    gpg: to check for pending gpg access request
-    inotify-simple: to check for pending gpg and u2f requests
-    github.com/maximbaz/pam-u2f: a fork that adds watch capability to pam-u2f module
-    subprocess32: if after all these years you are still using python2
+Requires:
+    inotify-simple: (for gpg and u2f) to check for requests
+    github.com/maximbaz/pam-u2f: (for u2f) a fork that adds watch capability to pam-u2f module
+    subprocess32: (for gpg) needed only if after all these years you are still using python2
 
 @author Maxim Baz (https://github.com/maximbaz)
 @license BSD
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -10,7 +10,7 @@ Configuration parameters:
         (default 0.1)
     gpg_check_watch_paths: A list of files to watch, start gpg check if any one was opened.
         (default ['~/.gnupg/pubring.kbx', '~/.ssh/known_hosts'])
-    sudo_check_watch_paths: A list of files to watch, toggle sudo check if any one was opened.
+    u2f_check_watch_paths: A list of files to watch, toggle u2f check if any one was opened.
         (default ['~/.config/Yubico/u2f_keys'])
 
 Control placeholders:
@@ -21,7 +21,7 @@ SAMPLE OUTPUT
 
 Dependencies:
     gpg: to check for pending gpg access request
-    inotify-simple: to check for pending gpg and sudo requests
+    inotify-simple: to check for pending gpg and u2f requests
     github.com/maximbaz/pam-u2f: a fork that adds watch capability to pam-u2f module
     subprocess32: if after all these years you are still using python2
 
@@ -42,6 +42,96 @@ else:
     import subprocess
 
 
+class GpgWatchingThread(threading.Thread):
+    """
+    A thread watchng if YubiKey is waiting for a touch for gpg access
+    """
+
+    def __init__(self, parent):
+        super(GpgWatchingThread, self).__init__()
+        self.parent = parent
+
+    def _check_card_status(self):
+        return subprocess.Popen(
+            'gpg --no-tty --card-status',
+            shell=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            preexec_fn=os.setsid)
+
+    def _set_pending(self, new_value):
+        self.parent.pending_gpg = new_value
+        if not self.parent.pending_u2f:
+            self.parent.py3.update()
+
+    def run(self):
+        inotify = inotify_simple.INotify()
+        watches = []
+        for path in self.parent.gpg_check_watch_paths:
+            if os.path.isfile(path):
+                watch = inotify.add_watch(path, inotify_simple.flags.OPEN)
+                watches.append(watch)
+
+        while not self.parent.killed.is_set():
+            for event in inotify.read():
+                # e.g. ssh doesn't start immediately, try several seconds to be sure
+                for i in range(20):
+                    with self._check_card_status() as check:
+                        try:
+                            # if this doesn't return very quickly, touch is pending
+                            check.communicate(
+                                timeout=self.parent.gpg_check_timeout)
+                            time.sleep(0.25)
+                        except subprocess.TimeoutExpired:
+                            self._set_pending(True)
+                            os.killpg(check.pid, signal.SIGINT)
+                            check.communicate()
+                            # wait until device is touched
+                            with self._check_card_status() as wait:
+                                wait.communicate()
+                                self._set_pending(False)
+                                break
+
+                # ignore all events caused by operations above
+                for _ in inotify.read():
+                    pass
+
+        for watch in watches:
+            inotify.rm_watch(watch)
+
+
+class U2fWatchingThread(threading.Thread):
+    """
+    A thread watchng if YubiKey is waiting for a touch for u2f access
+    """
+
+    def __init__(self, parent):
+        super(U2fWatchingThread, self).__init__()
+        self.parent = parent
+
+    def _set_pending(self, new_value):
+        self.parent.pending_u2f = new_value
+        if not self.parent.pending_gpg:
+            self.parent.py3.update()
+
+    def run(self):
+        inotify = inotify_simple.INotify()
+        watches = []
+        for path in self.parent.u2f_check_watch_paths:
+            if os.path.isfile(path):
+                watch = inotify.add_watch(path, inotify_simple.flags.OPEN)
+                watches.append(watch)
+
+        while not self.parent.killed.is_set():
+            for event in inotify.read():
+                self._set_pending(True)
+                for event in inotify.read():
+                    self._set_pending(False)
+
+        for watch in watches:
+            inotify.rm_watch(watch)
+
+
 class Py3status:
     """
     """
@@ -49,103 +139,29 @@ class Py3status:
     format = '\?if=is_waiting YubiKey'
     gpg_check_timeout = 0.1
     gpg_check_watch_paths = ['~/.gnupg/pubring.kbx', '~/.ssh/known_hosts']
-    sudo_check_watch_paths = ['~/.config/Yubico/u2f_keys']
+    u2f_check_watch_paths = ['~/.config/Yubico/u2f_keys']
 
     def post_config_hook(self):
         self.gpg_check_watch_paths = [
             os.path.expanduser(p) for p in self.gpg_check_watch_paths
         ]
-        self.sudo_check_watch_paths = [
-            os.path.expanduser(p) for p in self.sudo_check_watch_paths
+        self.u2f_check_watch_paths = [
+            os.path.expanduser(p) for p in self.u2f_check_watch_paths
         ]
 
         self.killed = threading.Event()
 
         self.pending_gpg = False
-        self.pending_sudo = False
-
-        class GpgThread(threading.Thread):
-            def _check_card_status(this):
-                return subprocess.Popen(
-                    'gpg --no-tty --card-status',
-                    shell=True,
-                    stdout=subprocess.DEVNULL,
-                    stderr=subprocess.DEVNULL,
-                    preexec_fn=os.setsid)
-
-            def _set_pending(this, new_value):
-                self.pending_gpg = new_value
-                if not self.pending_sudo:
-                    self.py3.update()
-
-            def run(this):
-                inotify = inotify_simple.INotify()
-                watches = []
-                for path in self.gpg_check_watch_paths:
-                    if os.path.isfile(path):
-                        watch = inotify.add_watch(path,
-                                                  inotify_simple.flags.OPEN)
-                        watches.append(watch)
-
-                while not self.killed.is_set():
-                    for event in inotify.read():
-                        # e.g. ssh doesn't start immediately, try several seconds to be sure
-                        for i in range(20):
-                            with this._check_card_status() as check:
-                                try:
-                                    # if this doesn't return very quickly, touch is pending
-                                    check.communicate(
-                                        timeout=self.gpg_check_timeout)
-                                    time.sleep(0.25)
-                                except subprocess.TimeoutExpired:
-                                    this._set_pending(True)
-                                    os.killpg(check.pid, signal.SIGINT)
-                                    check.communicate()
-                                    # wait until device is touched
-                                    with this._check_card_status() as wait:
-                                        wait.communicate()
-                                        this._set_pending(False)
-                                        break
-
-                        # ignore all events caused by operations above
-                        for _ in inotify.read():
-                            pass
-
-                for watch in watches:
-                    inotify.rm_watch(watch)
-
-        class SudoThread(threading.Thread):
-            def _set_pending(this, new_value):
-                self.pending_sudo = new_value
-                if not self.pending_gpg:
-                    self.py3.update()
-
-            def run(this):
-                inotify = inotify_simple.INotify()
-                watches = []
-                for path in self.sudo_check_watch_paths:
-                    if os.path.isfile(path):
-                        watch = inotify.add_watch(path,
-                                                  inotify_simple.flags.OPEN)
-                        watches.append(watch)
-
-                while not self.killed.is_set():
-                    for event in inotify.read():
-                        this._set_pending(True)
-                        for event in inotify.read():
-                            this._set_pending(False)
-
-                for watch in watches:
-                    inotify.rm_watch(watch)
+        self.pending_u2f = False
 
         if len(self.gpg_check_watch_paths) > 0:
-            GpgThread().start()
+            GpgWatchingThread(self).start()
 
-        if len(self.sudo_check_watch_paths) > 0:
-            SudoThread().start()
+        if len(self.u2f_check_watch_paths) > 0:
+            U2fWatchingThread(self).start()
 
     def yubikey(self):
-        is_waiting = self.pending_gpg or self.pending_sudo
+        is_waiting = self.pending_gpg or self.pending_u2f
         format_params = {'is_waiting': is_waiting}
         response = {
             'cached_until': self.py3.CACHE_FOREVER,
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -39,17 +39,21 @@ class YubiKeyTouchDetectorListener(threading.Thread):
 
     def _connect_socket(self):
         try:
+            self.parent.error = None
             self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
             self.socket.connect(self.parent.socket_path)
         except:
             self.socket = None
+            self.parent.error = Exception(
+                "Cannot connect to yubikey-touch-detector")
 
     def run(self):
         while not self.parent.killed.is_set():
             self._connect_socket()
 
             if self.socket is None:
-                # Socket is not available, try again soon
+                # Socket is not available, show error and try again soon
+                self.parent.py3.update()
                 time.sleep(60)
                 continue
 
@@ -84,11 +88,14 @@ class Py3status:
             'is_gpg': False,
             'is_u2f': False,
         }
+        self.error = None
 
         self.killed = threading.Event()
         YubiKeyTouchDetectorListener(self).start()
 
     def yubikey(self):
+        if self.error:
+            raise self.error
         response = {
             'cached_until': self.py3.CACHE_FOREVER,
             'full_text': self.py3.safe_format(self.format, self.status),
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -10,7 +10,7 @@ Configuration parameters:
         (default 0.1)
     gpg_check_watch_paths: A list of files to watch, start gpg check if any one was opened.
         (default ['~/.gnupg/pubring.kbx', '~/.ssh/known_hosts'])
-    sudo_check_watch_paths: A list of files to watch, start sudo check if any one was opened.
+    sudo_check_watch_paths: A list of files to watch, toggle sudo check if any one was opened.
         (default ['~/.config/Yubico/u2f_keys'])
 
 Control placeholders:
@@ -22,6 +22,7 @@ SAMPLE OUTPUT
 Dependencies:
     gpg: to check for pending gpg access request
     inotify-simple: to check for pending gpg and sudo requests
+    github.com/maximbaz/pam-u2f: a fork that adds watch capability to pam-u2f module
     subprocess32: if after all these years you are still using python2
 
 @author Maxim Baz (https://github.com/maximbaz)
@@ -114,22 +115,12 @@ class Py3status:
                     inotify.rm_watch(watch)
 
         class SudoThread(threading.Thread):
-            def _restart_sudo_reset_timer(this):
-                if this.sudo_reset_timer is not None:
-                    this.sudo_reset_timer.cancel()
-
-                this.sudo_reset_timer = threading.Timer(
-                    5, lambda: this._set_pending(False))
-                this.sudo_reset_timer.start()
-
             def _set_pending(this, new_value):
                 self.pending_sudo = new_value
                 if not self.pending_gpg:
                     self.py3.update()
 
             def run(this):
-                this.sudo_reset_timer = None
-
                 inotify = inotify_simple.INotify()
                 watches = []
                 for path in self.sudo_check_watch_paths:
@@ -140,8 +131,9 @@ class Py3status:
 
                 while not self.killed.is_set():
                     for event in inotify.read():
-                        this._restart_sudo_reset_timer()
                         this._set_pending(True)
+                        for event in inotify.read():
+                            this._set_pending(False)
 
                 for watch in watches:
                     inotify.rm_watch(watch)
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -93,7 +93,7 @@ class Py3status:
             'cached_until': self.py3.CACHE_FOREVER,
             'full_text': self.py3.safe_format(self.format, self.status),
         }
-        if any([waiting for waiting in self.status.values()]):
+        if any(self.status.values()):
             response['urgent'] = True
         return response
 
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -4,7 +4,7 @@ Show an indicator when YubiKey is waiting for a touch.
 
 Configuration parameters:
     format: Display format for the module.
-        (default '[\?if=is_gpg YubiKey][\?if=is_u2f YubiKey]')
+        (default '[YubiKey[\?if=is_gpg ][\?if=is_u2f ]]')
     socket_path: A path to the yubikey-touch-detector socket file.
         (default '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket')
 
@@ -80,7 +80,7 @@ class Py3status:
     """
     """
     # available configuration parameters
-    format = '[\?if=is_gpg YubiKey][\?if=is_u2f YubiKey]'
+    format = '[YubiKey[\?if=is_gpg ][\?if=is_u2f ]]'
     socket_path = '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket'
 
     def post_config_hook(self):
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -53,7 +53,6 @@ class Py3status:
 
         self.pending_gpg = False
         self.pending_sudo = False
-        self.sudo_reset_timer = None
 
         class GpgThread(threading.Thread):
             def _check_card_status(this):
@@ -65,47 +64,69 @@ class Py3status:
                     preexec_fn=os.setsid)
 
             def run(this):
+                inotify = inotify_simple.INotify()
+                gpg_watch = inotify.add_watch(
+                    os.path.expanduser('~/.gnupg/pubring.kbx'),
+                    inotify_simple.flags.OPEN)
+                ssh_watch = inotify.add_watch(
+                    os.path.expanduser('~/.ssh/known_hosts'),
+                    inotify_simple.flags.OPEN)
+
                 while not self.killed.is_set():
-                    with this._check_card_status() as check_process:
-                        try:
-                            # if this doesn't return very quickly, touch is pending
-                            check_process.communicate(timeout=0.1)
-                            self.pending_gpg = False
-                        except subprocess.TimeoutExpired:
-                            self.pending_gpg = True
-                            os.killpg(check_process.pid, signal.SIGINT)
-                            check_process.communicate()
-                            with this._check_card_status() as wait_process:
-                                # wait until device is touched
-                                wait_process.communicate()
-                                self.pending_gpg = False
-                    time.sleep(self.cache_timeout)
+                    for event in inotify.read():
+                        # e.g. ssh doesn't start immediately, try several seconds to be sure
+                        for i in range(20):
+                            with this._check_card_status() as check:
+                                try:
+                                    # if this doesn't return very quickly, touch is pending
+                                    check.communicate(timeout=0.1)
+                                    time.sleep(0.25)
+                                except subprocess.TimeoutExpired:
+                                    self.pending_gpg = True
+                                    os.killpg(check.pid, signal.SIGINT)
+                                    check.communicate()
+                                    # wait until device is touched
+                                    with this._check_card_status() as wait:
+                                        wait.communicate()
+                                        self.pending_gpg = False
+                                        break
+
+                        # ignore all events caused by operations above
+                        for _ in inotify.read():
+                            pass
+
+                inotify.rm_watch(gpg_watch)
+                inotify.rm_watch(ssh_watch)
 
         class SudoThread(threading.Thread):
+            def _restart_sudo_reset_timer(this):
+                def reset_pending_sudo():
+                    self.pending_sudo = False
+
+                if this.sudo_reset_timer is not None:
+                    this.sudo_reset_timer.cancel()
+
+                this.sudo_reset_timer = threading.Timer(5, reset_pending_sudo)
+                this.sudo_reset_timer.start()
+
             def run(this):
+                this.sudo_reset_timer = None
+
                 inotify = inotify_simple.INotify()
                 watch = inotify.add_watch(self.u2f_keys_path,
                                           inotify_simple.flags.ACCESS)
+
                 while not self.killed.is_set():
                     for event in inotify.read():
-                        self._restart_sudo_reset_timer()
+                        this._restart_sudo_reset_timer()
                         self.pending_sudo = True
+
                 inotify.rm_watch(watch)
 
         GpgThread().start()
 
         if os.path.isfile(self.u2f_keys_path):
             SudoThread().start()
-
-    def _restart_sudo_reset_timer(self):
-        def reset_pending_sudo():
-            self.pending_sudo = False
-
-        if self.sudo_reset_timer is not None:
-            self.sudo_reset_timer.cancel()
-
-        self.sudo_reset_timer = threading.Timer(5, reset_pending_sudo)
-        self.sudo_reset_timer.start()
 
     def yubikey(self):
         is_waiting = self.pending_gpg or self.pending_sudo
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -51,9 +51,11 @@ class YubiKeyTouchDetectorListener(threading.Thread):
         while not self.parent.killed.is_set():
             self._connect_socket()
 
+            # Refresh once to show or clear error as needed
+            self.parent.py3.update()
+
             if self.socket is None:
-                # Socket is not available, show error and try again soon
-                self.parent.py3.update()
+                # Socket is not available, try again soon
                 time.sleep(60)
                 continue
 
@@ -95,7 +97,8 @@ class Py3status:
 
     def yubikey(self):
         if self.error:
-            raise self.error
+            self.py3.error(str(self.error), self.py3.CACHE_FOREVER)
+
         response = {
             'cached_until': self.py3.CACHE_FOREVER,
             'full_text': self.py3.safe_format(self.format, self.status),
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -56,21 +56,29 @@ class Py3status:
         self.sudo_reset_timer = None
 
         class GpgThread(threading.Thread):
+            def _check_card_status(this):
+                return subprocess.Popen(
+                    'gpg --no-tty --card-status',
+                    shell=True,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                    preexec_fn=os.setsid)
+
             def run(this):
                 while not self.killed.is_set():
-                    with subprocess.Popen(
-                            'gpg --no-tty --card-status',
-                            shell=True,
-                            stdout=subprocess.DEVNULL,
-                            stderr=subprocess.DEVNULL,
-                            preexec_fn=os.setsid) as process:
+                    with this._check_card_status() as check_process:
                         try:
-                            process.communicate(timeout=0.1)
+                            # if this doesn't return very quickly, touch is pending
+                            check_process.communicate(timeout=0.1)
                             self.pending_gpg = False
                         except subprocess.TimeoutExpired:
                             self.pending_gpg = True
-                            os.killpg(process.pid, signal.SIGINT)
-                            process.communicate()
+                            os.killpg(check_process.pid, signal.SIGINT)
+                            check_process.communicate()
+                            with this._check_card_status() as wait_process:
+                                # wait until device is touched
+                                wait_process.communicate()
+                                self.pending_gpg = False
                     time.sleep(self.cache_timeout)
 
         class SudoThread(threading.Thread):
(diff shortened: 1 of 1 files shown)


py3status/modules/yubikey.py
```diff
@@ -0,0 +1,128 @@
...
+
+Configuration parameters:
+    cache_timeout: How often to detect a pending touch request.
+        (default 1)
+    format: Display format for the module.
+        (default '{state}')
+    state_off: Message when the touch is not needed.
+        (default '')
+    state_on: Message when the YubiKey is waiting for a touch.
+        (default 'Y')
```
lasers: We could remove configs by doing this...
```
format = '\?if=is_waiting Y'  # Default
```
```
format = '\?if=is_waiting Y|N'
```

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,128 @@
...
+        self.sudo_reset_timer.start()
+
+    def yubikey(self):
+        pending_touch = self.pending_gpg or self.pending_sudo
+        state = self.state_on if pending_touch else self.state_off
+        response = {
+            'cached_until': self.py3.time_in(self.cache_timeout),
+            'full_text': self.py3.safe_format(self.format, {'state': state}),
+            'color': self.py3.COLOR_BAD
+            if pending_touch else self.py3.COLOR_GOOD
```
lasers: We could remove color too.
```
format = '\?if=is_waiting&color=good Y'  # Default
```
```
format = '\?if=is_waiting&color=good Y|\?color=bad N'
```

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,128 @@
...
+    cache_timeout: How often to detect a pending touch request.
+        (default 1)
+    format: Display format for the module.
+        (default '{state}')
+    state_off: Message when the touch is not needed.
+        (default '')
+    state_on: Message when the YubiKey is waiting for a touch.
+        (default 'Y')
+    u2f_keys_path: Full path to u2f_keys if you want to monitor sudo access.
+        (default '')
```
lasers: I think `(default None)` would look more nice... and then do `if None: ''` thing in `post_config_hook`. We could add `os.path.expanduser(path)` in `post_config_hook` too.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,128 @@
...
+Configuration parameters:
+    cache_timeout: How often to detect a pending touch request.
+        (default 1)
+    format: Display format for the module.
+        (default '{state}')
+    state_off: Message when the touch is not needed.
+        (default '')
+    state_on: Message when the YubiKey is waiting for a touch.
+        (default 'Y')
+    u2f_keys_path: Full path to u2f_keys if you want to monitor sudo access.
```
lasers: This could be generic. `key_path`.

maximbaz: I'd rather keep it like this, because this is the file name used by [pam-u2f](https://github.com/Yubico/pam-u2f) and all online tutorials use this name. If I saw `key_path` option I would never guess that I need to provide path to `u2f_keys` file

lasers: No problem. The `pam-u2f` link seems to use `authfile`. If we parse `key_path` option, then we probably did parse the description too. Also... Using `auth_file` or `auth_key` would put it on top of the config list. Just stating things.

I wonder... I see that `authfile` points to `$HOME/.config/Yubico/u2f_keys`. Is that your path too? And if it is useful to default `$HOME/.config/Yubico/u2f_keys`?

py3status/modules/yubikey.py
maximbaz: I like it!

py3status/modules/yubikey.py
maximbaz: Yes I use the default, and thanks for the hint with `os.path.expanduser`, I thought that it will expand `~` to `/root`, but no, it expands to my own user - sweet! `$HOME` is not being expanded, only `~` by the way.

Although they use `authfile` in their README, the rest of the internet refers to this as `u2f_keys`, so I'll still keep the configuration name. Thought hopefully most people don't need to configure this anymore :)

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,118 @@
...
+Configuration parameters:
+    cache_timeout: How often to detect a pending touch request.
+        (default 1)
+    format: Display format for the module.
+        (default '\?if=is_waiting&color=bad Y')
+    u2f_keys_path: Full path to u2f_keys if you want to monitor sudo access.
+        (default '~/.config/Yubico/u2f_keys')
+
+SAMPLE OUTPUT
+{'color': '#FF0000', 'full_text': 'Y'}
```
lasers: I think I'd like something more readable like...
* `Waiting`
* `Key Waiting`
* `YubiKey`
* `Key` <-- @lasers's pick 
* `KEY`
* `KEY!`

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,118 @@
...
+    def post_config_hook(self):
+        self.u2f_keys_path = os.path.expanduser(self.u2f_keys_path)
+
+        self.killed = threading.Event()
+
+        self.pending_gpg = False
+        self.pending_sudo = False
+        self.sudo_reset_timer = None
+
+        class GpgThread(threading.Thread):
```
lasers: I don't know anything about threading, but I ponder if we need classes here... And if this could be done by using methods alone. @tobes can give you better feedbacks.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def yubikey(self):
+        is_waiting = self.pending_gpg or self.pending_sudo
+        response = {
+            'cached_until':
+            self.py3.time_in(self.cache_timeout),
+            'full_text':
+            self.py3.safe_format(self.format, {'is_waiting': is_waiting})
+        }
+        return response
```
lasers: Do we want to include `'urgent': is_waiting` here too?  If users does not like it, they can disable this by adding `allow_urgent = False` in the module config section. This might help with your ADHD too. (jk)

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,118 @@
...
+        self.sudo_reset_timer = threading.Timer(5, reset_pending_sudo)
+        self.sudo_reset_timer.start()
+
+    def yubikey(self):
+        is_waiting = self.pending_gpg or self.pending_sudo
+        response = {
+            'cached_until':
+            self.py3.time_in(self.cache_timeout),
+            'full_text':
+            self.py3.safe_format(self.format, {'is_waiting': is_waiting})
```
lasers: (Nit)
```
    return {
                self.format, {'is_waiting': is_waiting})
        }
```

maximbaz: I have `yapf` automatically formatting the code here

py3status/modules/yubikey.py
maximbaz: Huh, that's an interesting idea, I didn't know about this property. Looks neat, will add:

image

py3status/modules/yubikey.py
maximbaz: Yeah, I'm personally always using some fancy icons, so coming up with a default text message was a tough choice...

Not to confuse what key is the key that is waiting, how about a more specific `Touch YubiKey`?

image

lasers: I feel it's a bit on the nose and I'd like to stay with with subtle approach for QA purposes. Maybe just `YubiKey` or `Key` as I believe people will pick this up quicker than our expectations... And I feel that seeing `Touch YubiKey` repeatedly could be tiresome. It is akin to seeing `Check Gmail` repeatedly too... when `Gmail` or `Mail` would work just fine here.

Same thing with `key_path`. Make `yubikey` look clean in the config. (All my QA opinions).

py3status/modules/yubikey.py
maximbaz: Extracted params in a separate variable, this lead `yaps` to decide to change to a more pleasant styling 😉

py3status/modules/yubikey.py
maximbaz: I'm more concerned by the approaches that this code takes, these are very hacky tricks... I hope someone from Yubico would give a hint on how to make it better. Especially the check for `gpg`...

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def yubikey(self):
+        is_waiting = self.pending_gpg or self.pending_sudo
+        format_params = {'is_waiting': is_waiting}
+        response = {
+            'cached_until': self.py3.time_in(self.cache_timeout),
+            'full_text': self.py3.safe_format(self.format, format_params),
+            'urgent': is_waiting
+        }
+        return response
```
lasers: Okay. Now I know you like this. Can we do this instead?
```
response = {
     'full_text': self.py3.safe_format(self.format, format_params),
}
if is_waiting:
    response['urgent'] = True
return response
```
This prevents us from alway seeing `urgent: False` in the log / testing. It's all about simple pleasures.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,118 @@
...
+Show an indicator when YubiKey is waiting for a touch.
+
+Configuration parameters:
+    cache_timeout: How often to detect a pending touch request.
+        (default 1)
+    format: Display format for the module.
+        (default '\?if=is_waiting Touch YubiKey')
+    u2f_keys_path: Full path to u2f_keys if you want to monitor sudo access.
+        (default '~/.config/Yubico/u2f_keys')
+
```
lasers: Missing.
```
Control placeholders:
    {is_waiting} a boolean based on gpg content and sudo access
```

py3status/modules/yubikey.py
maximbaz: `YubiKey` it is then. Config name `u2f_keys_path` will leave for the reasons described in another comment.

py3status/modules/yubikey.py
maximbaz: Sure

maximbaz:
It just got much more performant, but more complicated as well. The long explanation is posted [here](https://github.com/Yubico/yubikey-manager/issues/26), where I'm also hoping to get some ideas on how to improve/simplify stuff here.

I'll be testing a lot more during the following days, but so far I like it a lot, it is _almost_ perfect. I'm also listening to [this ticket](https://github.com/Yubico/pam-u2f/issues/77) on how to get rid of `sudo_reset_timer` and truly detect when Yubikey was pressed. 

Regardless, this PR doesn't have to wait for any of those tickets, review and merge when you have time, and if I hear anything interesting I'll always make a separate PR.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,170 @@
...
+            GpgThread().start()
+
+        if len(self.sudo_check_watch_paths) > 0:
+            SudoThread().start()
+
+    def yubikey(self):
+        is_waiting = self.pending_gpg or self.pending_sudo
+        format_params = {'is_waiting': is_waiting}
+        response = {
+            'cached_until': self.py3.time_in(self.cache_timeout),
```
maximbaz: @tobes, @lasers: I have to put a small number here, `1` or even smaller, but most of the time it would be wasting cycles and just redrawing the same output over and over again. I would really love to put here `self.py3.CACHE_FOREVER`, and instead just manually force module refresh when at least one checks is triggered in other threads - is this possible to achieve?

lasers: Maybe you're looking for `self.py3.update()`? `scratchpad_async` and `window_title_async` might be interesting to you as both modules uses the function and the threading.

maximbaz: Exactly, thanks! Now it is instantly reacting only when the value changes - awesome!!

maximbaz:
I did not get a support from folks at pam-u2f (because that project is being rewritten), but I made my own fork that adds this functionality.

Now I'm fully happy with the code, ready to be merged. I'll revisit the approaches here if I figure out better ways to run a detection.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,167 @@
...
+                        for event in inotify.read():
+                            this._set_pending(False)
+
+                for watch in watches:
+                    inotify.rm_watch(watch)
+
+        if len(self.gpg_check_watch_paths) > 0:
+            GpgThread().start()
+
+        if len(self.sudo_check_watch_paths) > 0:
```
lasers: (Nit). `if len(self.sudo_check_watch_paths):` would work okay here too.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,167 @@
...
+        self.sudo_check_watch_paths = [
+            os.path.expanduser(p) for p in self.sudo_check_watch_paths
+        ]
+
+        self.killed = threading.Event()
+
+        self.pending_gpg = False
+        self.pending_sudo = False
+
+        class GpgThread(threading.Thread):
```
lasers: (Useful?) If you look at `volume_status`, you'll find several classes outside of `class Py3status`. If you are able to do the same here, then you could reclaim some indentations... and some things might be more readable... I recall we have to use `parent` for py3 stuffs... eg `self.parent.py3.update()`. Hopefully that's useful.

Something to try if we don't like how it's sitting inside `post_config_hook` instead of a line or two.

py3status/modules/yubikey.py
maximbaz: yeah, but I like it explicit

py3status/modules/yubikey.py
maximbaz: It started as two very simple classes with just few lines, but since then they grew... Now it is time to extract them indeed, makes total sense to me.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,183 @@
+# -*- coding: utf-8 -*-
+"""
+Show an indicator when YubiKey is waiting for a touch.
+
+Configuration parameters:
+    format: Display format for the module.
+        (default '\?if=is_waiting YubiKey')
+    gpg_check_timeout: Time to wait for gpg check response.
+        Use value as small as possible that doesn't introduce false positives.
+        (default 0.1)
```
lasers: (Q) Is it useful for users to have `gpg_check_timeout` config if the default is `0.1`? The message also say `Use value as small as possible that doesn't introduce false positives`. Another thing... I see a hardcoded `time.sleep(0.25)` too... right after that config too... so what can they do with this config?

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,183 @@
...
+    {is_waiting} a boolean indicating whether YubiKey is waiting for a touch.
+
+SAMPLE OUTPUT
+{'color': '#00FF00', 'full_text': 'YubiKey'}
+
+Dependencies:
+    gpg: to check for pending gpg access request
+    inotify-simple: to check for pending gpg and u2f requests
+    github.com/maximbaz/pam-u2f: a fork that adds watch capability to pam-u2f module
+    subprocess32: if after all these years you are still using python2
```
lasers: Modules uses `Requires:` and this `gpg` line might not be needed (installed everywhere).

Not sure if we can use (optional) thing like this.
```
python-inotify-simple: a simple wrapper around inotify
github.com/maximbaz/pam-u2f: (is this optional?) a fork that adds... 
python2-subprocess32: (optional) backported subprocess module for python2
```

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,183 @@
...
+import os
+import signal
+import sys
+import threading
+import inotify_simple
+
+if os.name == 'posix' and sys.version_info[0] < 3:
+    import subprocess32 as subprocess
+else:
+    import subprocess
```
lasers: We import `sys` for this. Wanting to see if we can live w/o `sys`.
```
try:
    import subprocess  # py3
except:
    import subprocess32 as subprocess  # py2
```

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,183 @@
...
+
+        if len(self.gpg_check_watch_paths) > 0:
+            GpgWatchingThread(self).start()
+
+        if len(self.u2f_check_watch_paths) > 0:
+            U2fWatchingThread(self).start()
+
+    def yubikey(self):
+        is_waiting = self.pending_gpg or self.pending_u2f
+        format_params = {'is_waiting': is_waiting}
```
lasers: Do more with formatting tricks by replacing `is_waiting` with `gpg` and `u2f`, eg...
```
format = '[\?color=#fa0 YubiKey [\?if=gpg&color=#aaf GPG][\?if=u2f&color=#faf U2F]]'
format = '[\?if=gpg&color=degraded YubiKey GPG][\?if=u2f&color=#fa0 YubiKey U2F]]'
format = '[\?color=good YubiKey [\?if=gpg GPG][\?if=u2f U2F]]'
format = '[\?color=good [\?if=gpg GPG][\?if=u2f U2F]]'
format = '[\?color=good&u2f U2F]'  # ignore GPG
```

py3status/modules/yubikey.py
maximbaz: Not really, because `subprocess` exists in both py2 and py3, it's just that the module in py2 doesn't contain timeout functionality that was introduced in py3.

py3status/modules/yubikey.py
maximbaz: good ideas:
- Yubikey is bought primarily for gpg operations, so gpg is definitely installed on the system 🙂
- only patched pam-u2f is optional, because that's the only dependency for u2f, will mark it so.

py3status/modules/yubikey.py
maximbaz: This is interesting thought, but in practice this would be just too much. Users don't care whether the module is flashing because of `gpg` or `u2f`, once it flashes for whatever reason they need to figure out manually what caused this and decide on the spot, touch or not touch.

What could have been interesting is to see the process name that requests the touch, e.g. I can configure using both `gpg` or `u2f` technology that `ssh`'ing somewhere will ask me to touch the device for confirmation, and in real life I don't care which technology is being used, instead I want to know which process wants this confirmation. But I don't know how to detect that from the module, we will leave this to a different time 🙂

lasers: >Users don't care whether the module is flashing because of gpg or u2f, once it flashes for whatever reason they need to figure out manually what caused this and decide on the spot, touch or not touch.

Possibly all the more reasons to do... as GPG is related to security, keys, encryption, logins, credentials, etc. Some users may like that option or may use nice icons to differentiate.

>once it flashes for whatever reason they need to figure out manually what caused this

It at least differentiates `Something to do with GPG`. Hopefully it will grow more from there... to support `{process_name}`, `{pid}`, Idk what else.

>maximbaz (in another thread): Of course it would be awesome if one of the yubikey CLI tools provided such information out of the box, it is a really useful feature in my mind...

If that happen, then we should use it to expose everything for users... just like `cmus`, `moc` and `coin_market` as they all have placeholders the users may not care about. We included them all because the data is there.

py3status/modules/yubikey.py
lasers: Flip the `import subprocess` lines could be sufficent to omit `sys`... then I wonder if we should explicitly throw an exception on subprocess32 not imported and is python2... `yubikey: subprocess32 not installed`. py3status support python2 until or after https://pythonclock.org and I think `py3status` is a wordplay on `i3wm` which is also a wordplay on `wmii`. ;-)

py3status/modules/yubikey.py
maximbaz: Let's first find at least one person who needs that 🙂

py3status/modules/yubikey.py
maximbaz: The exception will happen with current implementation, right? Python3 will import subprocess and nothing else, python2 will import subprocess32 and throw an exception if it cannot do so.

lasers: >The exception will happen with current implementation, right?

Yup. Sharing the findings.

Packages: `inotify_simple`, `subprocess32`
```
2017-10-17 INFO loading module "yubikey" from .../yubikey.py
2017-10-17 INFO Module `yubikey` could not be loaded
```
No exceptions... The bar will report:
```
yubikey // yubikey: Import Error, No module named 'inotify_simple'
yubikey // yubikey: Import Error, No module named 'subprocess32'
```
It's just a matter of if we should need `sys`... and/or if we should make this nice like `module: not installed`... eg `yubikey: subprocess32 not imported` or `yubikey: inotify_simple not imported`. Anyhow, I leave this to others. Same with the formatting.

ultrabug:
I'm just quickly coming around this PR to say big **thanks** to you @maximbaz since I'm also a Yubikey lover and user and I dreamed of this very module ! 👍 

I'll do my best to try it asap

maximbaz:
Looking forward 🙂

maximbaz:
@ultrabug when you get to test the module, let me know what you think of the default `gpg_check_timeout` value and whether you have touch policy for gpg operations enabled on your device (e.g. whether you have to touch Yubikey to confirm the `gpg --sign` operation). I have the touch policy enabled and the value of `0.1` is just perfect, but by default the touch policy is disabled on Yubikey, and in such cases on my laptop the values below `1` produce false positives. So I'm considering to increasing the default to `1`.

ultrabug:
@maximbaz I wanted to try this out and got stuck in the setup process. What are those options supposed to refer to and are used for?

```
    gpg_check_watch_paths: A list of files to watch, start gpg check if any one was opened.
        (default ['~/.gnupg/pubring.kbx', '~/.ssh/known_hosts'])
    u2f_check_watch_paths: A list of files to watch, toggle u2f check if any one was opened.
        (default ['~/.config/Yubico/u2f_keys'])
```
The GPG ones I got, the u2f_keys file, I don't. What should be in there?

maximbaz:
The defaults will work just work fine, if you don't have the U2F you probably don't use [pam-u2f](https://github.com/Yubico/pam-u2f), the code will ignore all missing files.

Why these files are needed to be configured I'll explain a bit later.

ultrabug:
I understand, but your patch is applied upon v1.0.5 of pam_u2f which has not been released so it's hard for users to get your commit into their pam_u2f lib.

Can you provide a patch on v1.0.4 ?

maximbaz:
v1.0.4 is really old, from 2015, people should use [aur/pam_u2f-git](https://aur.archlinux.org/packages/pam_u2f-git/) anyway. Also, there will never be v1.0.5 because the project is not maintained anymore.

ultrabug:
Not everybody uses Arch mate ;) I know there won't be 1.0.5.

Do you mean that you can't patch on 1.0.4?

maximbaz:
Heh 😄 

https://github.com/maximbaz/pam-u2f/tree/pam_u2f-1.0.4-indicator

maximbaz:
#### What does the `gpg_check_watch_paths` parameter do:

Detecting potential touch request is a heavy gpg operation, if we just run it constantly forever, the Yubikey will be constantly used, will never stop blinking and will heat up. So instead of running it constantly I needed a "trigger", some indication that right now Yubikey is _maybe_ waiting for a touch. This indicator happened to be OPEN event on any of the `gpg_check_watch_paths` files - if one them is opened by some process on the system, quite likely that _maybe_ Yubikey is waiting for a touch. In such case, this module will do the heavy gpg operation to actually check if Yubikey is waiting for a touch or not, and then will go back to sleep.

#### What does the `u2f_check_watch_paths` parameter do:

Same as above, except that there is no other indication available, so that is our _only way_ to detect when pam_u2f will request Yubikey touch - it will open the file from `u2f_check_watch_paths`.

#### Why is patched `pam-u2f` needed?

Contrary to GPG checks, with `u2f` there is no indication when user actually touched the Yubikey, so the patch provides such an indicator.

ultrabug:
Thanks @maximbaz, I think it maybe good to add this explanation on the doc.

> Yubikey touch - it will open the file from u2f_check_watch_paths
Opening (access time) or touching (modification time) ? because a lot of SSD users (like me) use the `noatime` mount option so we will fail to detect such event I guess.

I have a patched pam_u2f now here, I guess I need to reboot to test because for now nothing happens.

maximbaz:
It will use inotify to see if other tools are opening files from `u2f_check_watch_paths`. To simulate, run `inotifywait -m ~/.config/Yubico/u2f_keys` in one tab and do something with U2F in another time (request sudo, if that is what you configured?), and see if this tool captures some events - the module is looking for OPEN access.

I also have `noatime` mount option, so shouldn't be a problem 🙂

ultrabug:
Ok so for pam_u2f I guess I need to reboot for it to work... because the file isn't opened when using it on pam.d/sudo (cant reboot atm)

For GPG, I dunno how you use it yourself but I use gpg-agent to ask for my PIN when I connect to a SSH host. When the PIN request pops in, the Yubikey signal disappears which makes sense I guess since this is not a touch it's looking for.

I thus think that you should provide options to activate the GPG and/or the U2F watchers so that we don't waste 100ms cycles when using GPG+PIN.

maximbaz:
By providing options you mean to deactivate GPG and/or U2F watch entirely? Just setting `u2f_check_watch_paths` and/or `gpg_check_watch_paths` to an empty list should do the trick I think.

maximbaz:
Just so that I understand, is your Yubikey configured to request a touch from you on any action? 

I configured it in the following way: I need to enter PIN only once per session, but then every time I do one of the following operations Yubikey is waiting for a physical touch before allowing this operation:

- SSH somewhere
- Decrypt a GPG file
- Sign a file with GPG
- Request `sudo`

ultrabug:
> Just setting u2f_check_watch_paths and/or gpg_check_watch_paths to an empty list should do the trick I think.

True, fair enough :)

> Yubikey is waiting for a physical touch before allowing this operation:

Only for request sudo. I'm interested in how you made it that it requires a touch for the other operations after you unlocked it with your PIN!

maximbaz:
Install [ykman](https://github.com/Yubico/yubikey-manager/) and use some of these depending on your taste:

```
$ ykman openpgp touch aut on # for ssh
$ ykman openpgp touch sig on # for signing
$ ykman openpgp touch enc on # for encryption
```

maximbaz:
I'm considering to extract the whole detection logic in a separate application, and make this module only use that application to get events and display them nicely. There are a few reasons, first it would benefit more people as not everyone is using our beloved py3status, but also it would be easier for me to keep improving the tool. 

For example, right now I have a case when I ssh to HOST_A (with forwarding ssh agent), from there I ssh to HOST_B which also requires a touch, but since the request is happening on a remote HOST_A and not on my laptop the current detection doesn't capture this event. I already have a working POC that covers this case, but it requires some further polishing and testing first.

Suppose I have the tool ready, it is likely be an always running process, for example wrapped in a systemd service. Which communication protocol would you recommend me to use between py3status module and the tool? Definitely not periodic polling. Unix sockets?

ultrabug:
I think this is a sensible approach indeed. Unix sockets look the right way to move forward yes (at least that's what comes to my mind as well).

Thanks!

maximbaz:
@ultrabug it's time to give it another try 😉 I extracted the tool into [maximbaz/yubikey-touch-detector](https://github.com/maximbaz/yubikey-touch-detector) and made this module subscribe to notifications from it via unix sockets. The message is always of a fixed length of 5 bytes, so the parsing code here is very simple. 

This module is quite resilient, it will detect if the `yubikey-touch-detector` is not running and will retry the connection every minute (useful for testing, upgrading `yubikey-touch-detector` without restarting `i3`, etc).

Finally I also added the `is_waiting_gpg` and `is_waiting_u2f` to the list of control placeholders, this is something @lasers suggested long time ago 🙂 Now it's just super easy because the host app is sending separate messages for GPG and U2F.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,121 @@
...
+    format = '\?if=is_waiting YubiKey'
+    socket_path = '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket'
+
+    def post_config_hook(self):
+        self.socket_path = os.path.expanduser(self.socket_path)
+        self.socket_path = os.path.expandvars(self.socket_path)
+
+        self.killed = threading.Event()
+
+        self.waiting_gpg = 0
```
lasers: (Nit) I see 3 blank lines. See if we can kill one or two.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,121 @@
+# -*- coding: utf-8 -*-
+"""
+Show an indicator when YubiKey is waiting for a touch.
```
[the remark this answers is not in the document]
maximbaz: This way it's not very obvious, let's have it as it is to spell out what exactly this indicator is doing.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,121 @@
...
+Configuration parameters:
+    format: Display format for the module.
+        (default '\?if=is_waiting YubiKey')
+    socket_path: A path to the yubikey-touch-detector socket file.
+        (default '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket')
+
+Control placeholders:
+    {is_waiting} a boolean, True if YubiKey is waiting for a touch for any reason.
+    {is_waiting_gpg} a boolean, True if YubiKey is waiting for a touch due to a gpg operation.
+    {is_waiting_u2f} a boolean, True if YubiKey is waiting for a touch due to a pam-u2f request.
```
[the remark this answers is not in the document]
maximbaz: you mean to remove the `is_waiting` and have the default format be something like this?

```
format = '\?if=is_waiting_gpg OR is_waiting_u2f'
```

what's the correct syntax for this?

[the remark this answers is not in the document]
lasers: `format = '[\?if=is_gpg KEY][\?is_u2f KEY]'`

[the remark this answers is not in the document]
maximbaz: @lasers just stumbled upon an issue with this format, it is possible that there are two commands issued for Yubikey at the same time (e.g. pacman is updating packages and requesting sudo via u2f, while at the same time on a different tab you ssh somewhere and this triggers gpg check). This results in module output "YubiKeyYubiKey", because `is_gpg` and `is_u2f` are both active at the same time. I want a true OR, not two separate ifs - how can we achieve that?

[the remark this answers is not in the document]
maximbaz: I need this for a case when I don't want to see `GPG` or `U2F`, I just want to see `YubiKey` (or in my case, I use a symbol of key).

But using your example I think I solved it, this seems to work like a true OR:

```
format = '[[\?if=is_gpg ][\?if=is_u2f ] Yubikey]'
```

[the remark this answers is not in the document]
lasers: Nice. `format = '[YubiKey[\?if=is_gpg ][\?if=is_u2f ]]'` seems to work too.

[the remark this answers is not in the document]
maximbaz: Indeed, I'll use your way as a default, it's easy to understand how to go from it to additionally showing `GPG` or `U2F`.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,112 @@
...
+    format: Display format for the module.
+        (default '[YubiKey[\?if=is_gpg ][\?if=is_u2f ]]')
+    socket_path: A path to the yubikey-touch-detector socket file.
+        (default '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket')
+
+Control placeholders:
+    {is_gpg} a boolean, True if YubiKey is waiting for a touch due to a gpg operation.
+    {is_u2f} a boolean, True if YubiKey is waiting for a touch due to a pam-u2f request.
+
+SAMPLE OUTPUT
```
lasers: This should come after the author.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,112 @@
...
+                    # Connection dropped, need to reconnect
+                    break
+                elif data == b"GPG_1":
+                    self.parent.is_waiting_gpg = True
+                elif data == b"GPG_0":
+                    self.parent.is_waiting_gpg = False
+                elif data == b"U2F_1":
+                    self.parent.is_waiting_u2f = True
+                elif data == b"U2F_0":
+                    self.parent.is_waiting_u2f = False
```
lasers: (Nit) Maybe we get rid of all `is_waiting`. I barely can see the difference, eg `self.parent.is_u2f`.

maximbaz: true that

py3status/modules/yubikey.py
maximbaz: this should be part of travis check

ultrabug:
Awesome @maximbaz I will try tomorrow 👍

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self.is_gpg = False
+        self.is_u2f = False
+
+        YubikeyTouchDetectorListener(self).start()
+
+    def yubikey(self):
+        format_params = {
+            'is_gpg': self.is_gpg,
+            'is_u2f': self.is_u2f,
```
lasers: (Suggestion). We can make `self.operation = {'is_gpg': None, 'is_u2f' None}` in `post_config_hook` then use `self.operation['is_gpg'] = True` to update the values. No need to make new dict on loops. This removes few lines too.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,112 @@
...
+        (default '[YubiKey[\?if=is_gpg ][\?if=is_u2f ]]')
+    socket_path: A path to the yubikey-touch-detector socket file.
+        (default '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket')
+
+Control placeholders:
+    {is_gpg} a boolean, True if YubiKey is waiting for a touch due to a gpg operation.
+    {is_u2f} a boolean, True if YubiKey is waiting for a touch due to a pam-u2f request.
+
+Requires:
+    github.com/maximbaz/yubikey-touch-detector: tool to detect when YubiKey is waiting for a touch
```
lasers: (Nit) Three lines here are over column 80. I don't have 4K monitors.  ;__;

maximbaz: the line length is enforced by Travis, it was long but I compressed it enough so that Travis is green and happy now.

you should by a 4K screen, and a yubikey :stuck_out_tongue_winking_eye:

maximbaz:
A friendly ping 😜 

You should consider giving @lasers push access to this repo, he is very active at making improvements and code reviews 😄

lasers:
Maybe let's not. I'm a noob. Having more eyes on the code is always good. I enjoy the efforts of fixing and/or improving things, but not so much in the deprecation department. If they're occupied with things, then I guess we wait patiently since they're not noobs. :-)

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,109 @@
...
+            self.socket = None
+
+    def run(self):
+        while not self.parent.killed.is_set():
+            self._connect_socket()
+
+            if self.socket is None:
+                # Socket is not available, try again soon
+                time.sleep(60)
+                continue
```
lasers: (Thought) Do we need `time.sleep(60)` here?
1) Would the thread not keep going and failing until it's finally made?
2) What happen if ssh or gpg is made, but we're in this time.sleep period? I think it could not show up... or even could show up several seconds late! 😡

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,109 @@
...
+    def _connect_socket(self):
+        try:
+            self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+            self.socket.connect(self.parent.socket_path)
+        except:
+            self.socket = None
+
+    def run(self):
+        while not self.parent.killed.is_set():
+            self._connect_socket()
```
lasers: We could put everything here for better readability. Less lines and we skip `= None` too.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,109 @@
...
+    socket_path = '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket'
+
+    def post_config_hook(self):
+        self.socket_path = os.path.expanduser(self.socket_path)
+        self.socket_path = os.path.expandvars(self.socket_path)
+
+        self.status = {
+            'is_gpg': False,
+            'is_u2f': False,
+        }
```
lasers: (Nit) Maybe just `self.status = {'is_gpg': False, 'is_u2f': False}` here and kill a line?

tobes: for me it is more readable as it is.

Readability is better than compactness every time

py3status/modules/yubikey.py
maximbaz: 1. It's not a good idea to completely avoid `sleep`, if the socket connection cannot be established, there is likely an issue with the backend service and manual intervention is needed, and until then connection will probably fail next time as well. Trying to reconnect too often will only result if wasting lots of cycles. One minute is a sweet spot that I hardcode on purpose, not too often and yet fast enough for an automatic recovery - in real life YubiKey is used much rarely than once per minute.

However your question lead me to the following thought, since manual intervention is likely needed, perhaps silently sleeping is not enough, and we need to show an error message in red color. I saw other modules do this when they fail. Could you point me to how to achieve this?

2. It will not show up indeed, whether the problem is on the backend side (not publishing events) or on the module side (not listening to the events), no connection means no events received and no messages shown. However the notification will not show up several seconds late, when connection is finally established, it is starting from the clean state, only next even will trigger a message to be shown.

py3status/modules/yubikey.py
maximbaz: I don't think we can skip `= None` even if we inline the code here, the exception is likely happening during the `connect()`, and I want to cleanup the dangling socket immediately in that case. I like it as it is.

py3status/modules/yubikey.py
lasers: `self.py3.error`. I think it'll keep going behind the scene until https://github.com/ultrabug/py3status/issues/1114 gets addressed. @tobes also said something about extending the `self.py3.error` exceptions to threads. 
```
./bluetooth.py:130:            self.py3.error(STRING_NOT_STARTED)
./yandexdisk_status.py:58:            self.py3.error(STRING_ERROR)
./coin_market.py:170:                self.py3.error('bad markets')
./external_script.py:76:            self.py3.error(output)
```

lasers: I don't know this well, but quick stab...
```
diff --git a/py3status/modules/yubikey.py b/py3status/modules/yubikey.py
index 621a3c90..0d82667c 100644
--- a/py3status/modules/yubikey.py
+++ b/py3status/modules/yubikey.py
@@ -22,7 +22,6 @@ SAMPLE OUTPUT
 {'full_text': 'YubiKey', 'urgent': True}
 """
 
-import time
 import os
 import socket
 import threading
@@ -41,17 +40,16 @@ class YubiKeyTouchDetectorListener(threading.Thread):
         try:
             self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
             self.socket.connect(self.parent.socket_path)
-        except:
             self.socket = None
+            self.error = e
 
     def run(self):
         while not self.parent.killed.is_set():
             self._connect_socket()
 
-            if self.socket is None:
-                # Socket is not available, try again soon
-                time.sleep(60)
-                continue
+            if self.error:
+                self.parent.py3.update()
 
             while not self.parent.killed.is_set():
                 data = self.socket.recv(5)
@@ -89,6 +87,8 @@ class Py3status:
         YubiKeyTouchDetectorListener(self).start()
 
     def yubikey(self):
+        if self.error:
+            raise self.error
         response = {
             'full_text': self.py3.safe_format(self.format, self.status),
```

maximbaz: awesome, thanks, it is very beautiful and descriptive now, clear to the user when this happens and what to investigate 🙂

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,116 @@
...
+        self.error = None
+
+        self.killed = threading.Event()
+        YubiKeyTouchDetectorListener(self).start()
+
+    def yubikey(self):
+        if self.error:
+            raise self.error
```
lasers: Maybe we use `self.py3.error(str(self.err), timeout=self.py3.CACHE_FOREVER)` here instead due to py3ness... and If I understand this okay... that we don't need timeout anymore too due to error? How come we don't use `str(self.err)` -- a potentially better exception message(s) via `except Exception as err`-- over a hardcoded `self.error` message?

ultrabug:
Ok I finally tested the module fully and it works good 👍 

As for the error report that @lasers is pointing out, I think @maximbaz did this as-is because py3status will be handling the error report on the bar. The drawback is that the user will have to click on the red yubikey string that will appear to display the error message. It could be misleading because the actual behaviour of the module is to display a red yubikey string to alert of a touch requirement...

So I'd rather the module outputs the error itself in the full_text field than relying on py3status to do so, a bit like @lasers proposes.

py3status/modules/yubikey.py
```diff
@@ -0,0 +1,116 @@
...
+
+class Py3status:
+    """
+    """
+    # available configuration parameters
+    format = '[YubiKey[\?if=is_gpg ][\?if=is_u2f ]]'
+    socket_path = '$XDG_RUNTIME_DIR/yubikey-touch-detector.socket'
+
+    def post_config_hook(self):
+        self.socket_path = os.path.expanduser(self.socket_path)
```
ultrabug: isn't this overridden by the (re)definition just below on line 85?

maximbaz:
I will play with py3.error, `CACHE_FOREVER` for it definitely makes sense, as well as reducing one click to see the error 🙂

The actual exception is very cryptic, I would never guess what it means, that's why I put my custom one - at least people will know what to investigate next if they see it.

py3status/modules/yubikey.py
maximbaz: yep, on purpose - first `expanduser` expands `~` in the path if it is present, then `expandvars` expands environment variables (again, if they are present). We need a full path, and this allows to define the path both as `$XDG_RUNTIME_DIR/bla` and `~/bla`.

lasers:
>So I'd rather the module outputs the error itself in the full_text field than relying on py3status to do so,

That old thing? We switched away from that... to this new thing. I'd suggest we stay with the new thing for QA / consistency. We also could just make error message much much shorter... like `broken socket` or `no connection`

We should see if it's possible to expand the exceptions right away using global / module config. Maybe make it a default too for the users. That way, we can keep ours toggled off... or even disappear... on account of switching modules non-stop. Just a thought.

maximbaz:
I used py3.error to refresh module less often, this is cool. Visually it looks exactly the same, you see "yubikey" in red, you click on it and see full error message.

I like to keep the error message as it is to make it clear what to investigate, "no socket" or "no connection" aren't explanatory enough to me.

ultrabug:
Impressive work @maximbaz congratulations and thank you! I'm using it 👍

PR #1111, opened 2017-10-14
[do_not_disturb] Add better support for dunst

maximbaz:
No need to run a killer thread if dunst is so nicely reacting on custom signals.

This was proposed by @tobes [a long long time ago](https://github.com/ultrabug/py3status/pull/598#issuecomment-267867768) 🙂 

Since then dunst found a new maintainer and was drastically improved, so I'm back on dunst 😄 

----

I can't make it forget history, @tobes, `-SIGSTOP` just hangs my dunst - if possible I would very much like to clear history - could you try as well and maybe give me a hint?

-----

Deliberately not doing any other improvements so that this can be merged faster.

### DIFF
--- a/py3status/modules/do_not_disturb.py
+++ b/py3status/modules/do_not_disturb.py
@@ -25,6 +25,9 @@ SAMPLE OUTPUT
 
 off
 {'color': '#FF0000', 'full_text': 'ON'}
+
+@author Maxim Baz (https://github.com/maximbaz)
+@license BSD
 """
 
 from time import sleep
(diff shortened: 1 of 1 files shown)


tobes:
Much nicer thanks.  The less theads the better imho

> I can't make it forget history, @tobes

Why are you using `SIGSTOP`?  would `SIGTERM` be the one to use.  I'll leave this as a challenge for you

 Now that it id being worked on again maybe it can learn to be controlled over dbus which would make more sense

maximbaz:
> Why are you using SIGSTOP? would SIGTERM be the one to use.

Because you suggested so last time 😄 `SIGTERM` works indeed, hence #1122 :+1:

tobes:
> Because you suggested so last time 

and you'd take my advice ;p

PR #1123, opened 2017-10-22
task to monitor i3status thread

tobes:
I forgot I'd done this.

It is slightly complicated as it adds a generic way to add random a user defined tasked to the queue.  If accepted I'll make the i3time update one of these too.

Maybe we should make them run in a thread like the py3status modules to simplify stuff too

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -50,6 +50,7 @@ class Runner(Thread):
         self.daemon = True
         self.module = module
         self.py3_wrapper = py3_wrapper
+        self.start()
 
     def run(self):
         try:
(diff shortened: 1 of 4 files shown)


ultrabug:
So, we reviewed this together if I recall correctly but didn't merge, so that means there was something about this we were unsure about?

tobes:
I think this is good.  

* now use `run()` as the method so it looks like a `Thread`.
* now run in our own thread not in the main one.
* removed some hacks so it is much cleaner
* `timeout_queue_add_module()` renamed to `timeout_queue_add()` as now more generalized

PR #1168, opened 2017-11-22
fix summertime bug (hopefully) and some related time module cleanups

tobes:
` update_times()` was dead code so removed

we need to also need to know the update interval to know if we are late or not.


ultrabug:
Awesome, dead code I love too!

Issue #1171, opened 2017-11-24
OpenBSD file: unknown option -- mime-encoding


meutel:
On OpenBSD (6.2-current), py3status fails with an error:
```
file: unknown option -- mime-encoding
usage: file [-bchiLsW] file ...
```
The `file` command on OpenBSD does not have the `--mime-encoding` option. The command is used in: `parse_config.py` ine 536
```
encoding = check_output(['file', '-b', '--mime-encoding', '--dereference', config_path])
encoding = encoding.strip().decode('utf-8')
```
As a workaround, I replaced those lines with hardcoded value:
```
encoding = 'utf-8'
```
then the program runs OK.

Maybe encoding guess should use a python library (chardet?) instead of using the file command. I will try to propose a patch (never coded anything in python, but this is a nice way to learn).

tobes:
Thanks for reporting this and your analysis of the issue

> chardet

We try to keep py3status as light as possible as far as needed packages are concerned.  I've created a fix in PR #1173 that should fix the problem for you.  If you can confirm the fix that would be great

meutel:
Yes, it's fixed. Thanks

tobes:
fixed via  #1173

PR #1173, opened 2017-11-24
assume utf-8 if we cannot guess file encoding

tobes:
#1171 reports an issue on OpenBSD due to config file encoding guessing.

This fixes the issue by assuming utf-8 if guessing fails.  It allows users who could not run to use py3status but does not change existing behaviour for those that can run py3status already

I think it should get in 3.7 as it is trivial and allows us testing on OpenBSD


ultrabug:
You're right, thanks!

PR #1176, opened 2017-11-29
Make POST requests work

raspbeguy:
POST requests couldn't work, because the data parameter was just ignored. Now it works.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,15 @@
-version 3.7 (milestone)
+version 3.7 (2017-11-30)
+* NOTE: this is a performance focused release
+* deep rework of the usage and scheduling of threads to run modules: now py3status does not keep one thread per module running permanently but instead uses a queue to spawn a thread to execute the module only when its cache expires, by @tobes
+* memory footprint of py3status got largely reduced thanks to the threads modifications and thanks to a nice hunt on ever growing and useless variables done by @tobes
+* the next release will contain logic to switch from threads to greenlets and go further in optimizing py3status
 * conversion of many modules to use post_config_hook, by @lasers
 * improve spelling in documentation, by @lasers
 * let Py3.contains_format() accept a list of placeholder names
+* py3: print full command in the error logs, by @lasers
+* air_quality module: fix attributeerror exception, by @lasers
 * apt_updates module: improved error if apt not installed, by @lasers
+* backlight module: allow slower value change after a threshold, by @maximbaz
 * backlight module: detect and bypass xbacklight limitations, by @lasers
 * diskdata module: new {total_space} placeholder, by @pcewing
 * do_not_disturb module: better support for dunst, by @maximbaz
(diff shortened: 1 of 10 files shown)


tobes:
@raspbeguy Thanks for this,  I don't think we have any current POST users so the implementation was never checked.

I'm keen to add this at the start of the next release cycle so it has time to bed and get some testing.

raspbeguy:
Well, I guess this is a pretty critical bugfix, I mean that the feature was announced to work and yet it wasn't.

Also I need this feature to work in order to propose another PR for a new module. I did test it, I know that only one tester is few but I think the fix is very minimalist and pretty obvious and understandable, and in the end of the day this module in only a wrapper for some python lib that is more than reliable.

tobes:
>  I did test it

In which python versions? we support both 2.7 and 3.x also I'd like some coverage with the different modules using the existing functionality.

> pretty critical bugfix

It is a bug fix but as it doesn't break core functionality I'd not deem it critical.  I agree that it is bad for anyone wanting to create a module that needs the functionality.

Whilst it looks good I've learnt to be wary.  Hopefully 3.7 will be released very shortly and then we can look at merging this asap

raspbeguy:
Hum, okay. I still think we should include it in next release because it makes my heart bleed to know that a documented but non-working feature is deliberately left bugged in a significant release. But you're the boss.

tobes:
Is there a reason you changed the headers stuff? because that is my main reservation.

The adding the post data I'm not so worried about

raspbeguy:
Well, I did that because you can't use a `add_header` method on a `Request` object that doesn't exist. I moved the `Request` object creation because we need to tread the `data` parameter first. You will then say that I could have moved the treatment on `data` first, then create the `Request`, then treat the `auth` condition, but it seemed more readable and logical the way I did.

raspbeguy:
I just f***ed up my branch, if you want I can create another pull request.

raspbeguy:
By the way I managed to test it on python 2.7 and it works great.

tobes:
> I just f***ed up my branch, if you want I can create another pull request.

A clean branch would be nice but I'd be keen to merge this soon so if there are any issues we get to test it and find them quickly but a quick glance suggests it's good.

Thanks for the 2.7 testing

raspbeguy:
Hold on, actually I might not have tested the feature properly, I need to do some more tests.

Sorry about the extra delay.

raspbeguy:
Ok, I have now the proof:

```python
test = self.py3.request("https://httpbin.org/post",data={"foo":"bar","yaddy":"yadda"},headers={'User-Agent': 'Mozilla/5.0'})
print(test.text)
```

What I got from that:

```json
{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {
    "foo": "bar", 
    "yaddy": "yadda"
  }, 
  "headers": {
    "Accept-Encoding": "identity", 
    "Connection": "close", 
    "Content-Length": "30", 
    "Content-Type": "application/x-www-form-urlencoded", 
    "Host": "httpbin.org", 
    "User-Agent": "Mozilla/5.0"
  }, 
  "json": null, 
  "origin": "<not.shown.here>", 
  "url": "https://httpbin.org/post"
}
```

Test achieved on both python2 and python3. I believe that's enough to show it is successful.

I will soon close this PR and do a new one, thanks for your patience.

PR #1182, opened 2017-12-02
Small refactor for parse_config.py

alexoneill:
Duplicate code in two places of parse_config can be simplified into one helper function!


tobes:
@alexoneill nice refactor

PR #1186, opened 2017-12-03
add option to enable support for green threads using gevent.

ultrabug:
This allows to run py3status using the gevent event loop which effectively reduces the load of the system by running everything in a concurrent and non blocking way.

Note that py3status' setup now supports extra requires to allow users to choose their dependencies flavours.

### DIFF
--- a/py3status/__init__.py
+++ b/py3status/__init__.py
@@ -16,6 +16,16 @@ except NameError:
 
 
 def main():
+    # detect gevent option early because monkey patching should be done before
+    # everything else starts kicking
+    if '-g' in sys.argv or '--gevent' in sys.argv:
+        try:
+            from gevent import monkey
+            monkey.patch_all()
+        except ImportError:
+            print('Requested gevent, but it is not installed!')
+            sys.exit(4)
+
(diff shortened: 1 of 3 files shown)


py3status/__init__.py
```diff
@@ -16,6 +16,16 @@
...
 def main():
+    # detect gevent option early because monkey patching should be done before
+    # everything else starts kicking
+    if '-g' in sys.argv or '--gevent' in sys.argv:
+        try:
+            from gevent import monkey
+            monkey.patch_all()
+        except ImportError:
+            print('Requested gevent, but it is not installed!')
+            sys.exit(4)
```
tobes: This seems very harsh!

We should allow py3status to run at a minimum just to display a nice error to the user.  Personally I'd like use to run but without the gevent, but add a small warning 'module' that can be clicked to hide.

ultrabug:
@tobes as discussed with you, I updated to not die like this and report properly instead..

tobes:
Thanks for the contribution ;p

It will be nice to see some documentation for this new feature

PR #1193, opened 2017-12-10
Handle cookies within request

raspbeguy:
Some websites needs cookies in order to work. With this modification, you can specify a `CookieJar` or subclass object to `py3.request` in order to exploit cookies.

### DIFF
--- a/py3status/request.py
+++ b/py3status/request.py
@@ -52,9 +52,7 @@ class HttpResponse:
             headers['Authorization'] = 'Basic %s' % auth_str.decode('utf-8')
         if data:
             data = urlencode(data).encode()
-        print(cookiejar)
         if cookiejar is not None:
-            print("prout")
             self._cookiejar = cookiejar
             opener = build_opener(HTTPCookieProcessor(cookiejar))
             install_opener(opener)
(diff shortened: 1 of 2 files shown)


py3status/request.py
```diff
@@ -123,3 +129,20 @@ def headers(self):
...
+        """
+        try:
+            return self._cookiejar
+        except AttributeError:
+            return None
+
+    @cookiejar.setter
+    def cookiejar(self, cj):
+        """
+        Set the cookie jar in care we want to change it after object creation
```
tobes: s/care/case/

raspbeguy: Eeyep

tobes:
I'd like things to be easier for the module writer maybe having CookieJar available via `self.py3` ping me when you've had a chance to do the updates you wanted

raspbeguy:
@tobes I was thinking about adding a file `cookie.py` implementing our own CookieJar subclass, so that wouldn't change anything in this file (because `cookiejar` parameter accepts any CookieJar subclass already).

tobes:
>  I was thinking about adding a file cookie.py

cool but module writers won't get access unless we give it to them.  modules should get everything they need via `self.py3` and not do `from py3status import ...` etc

raspbeguy:
Yes of course I will make it available via `py3` but I mean, this could be the object of a new PR as it won't break the current version of the cookie feature.

ultrabug:
I'm fine with this @raspbeguy and I agree that having it available through self.py3 would be great so please keep up the good work

PR #1194, opened 2017-12-10
Add module for gazelle-based private trackers

raspbeguy:
2018-02-13-232744_249x18_scrot

**NOTE: This PR depends of #1193 and #1278 so please merge it in the right order.**

This new module enable to watch one's ratio on gazelle-based torrent private trackers.

### DIFF
--- a/py3status/modules/gazelle.py
+++ b/py3status/modules/gazelle.py
@@ -18,6 +18,10 @@ Configuration parameters:
         by API, so you are just likely not to fill it) (default None)
     limit_warning: custom warning ratio (default 1)
     password: Account password (default '')
+    thresholds: specify color thresholds to use
+        *(default {'ratio': [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')],
+        'messages': [(0, 'degraded'), (1, 'good')],
+        'notifs': [(0, 'degraded'), (1, 'good')]})*
     username: Account username (default '')
 
 Format placeholders:
@@ -26,12 +30,6 @@ Format placeholders:
     {upload} account upload amount
     {messages} number of unread messages
     {notifs} number of new notifications
-
-Color options:
-    color_ratio_ok: color when ration is bigger than 1 (default color_good)
-    color_ratio_warning: color when ration is between required ratio and 1
-        (available only when tracker has a dynamic required ratio (default color_warning)
-    color_ratio_ko: color when ratio below required ratio (default color_bad)
 
 @author raspbeguy <raspbeguy@hashtagueule.fr>
 @license MIT
@@ -51,6 +49,10 @@ class Py3status:
     limit_required = None
     limit_warning = 1
     password = ''
+    thresholds = {
+            'ratio': [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')],
+            'messages': [(0, 'degraded'), (1, 'good')],
+            'notifs': [(0, 'degraded'), (1, 'good')]}
     username = ''
 
     def _readable_size(self, size):
@@ -120,17 +122,26 @@ class Py3status:
             'messages': result['notifications']['messages'],
             'notifs': result['notifications']['notifications']
             }
-        if data['ratio'] >= self.limit_warning:
-            color = self.py3.COLOR_RATIO_OK or self.py3.COLOR_GOOD
-        elif data['ratio'] >= (self.limit_required or data['requiredratio']):
-            color = self.py3.COLOR_RATIO_WARNING or self.py3.COLOR_DEGRADED
-        else:
-            color = self.py3.COLOR_RATIO_KO or self.py3.COLOR_BAD
+
+        # Ok, so if you think that what I'll do next stinks a bit, remember that
+        # the color threshold design in py3status is a gigantic turd.
+        # Main reason of this ugliness is the impossibility to natively define a
+        # variable threshold, so I have to replace keyword by the actual value.
+        for n, item in enumerate(self.thresholds['ratio']):
+            if item[0] == "requiredratio":
+                lst = list(item)
+                lst[0] = data['requiredratio']
+                self.thresholds['ratio'][n] = tuple(lst)
+        # End of horrible code
+
+        self.py3.threshold_get_color(data['ratio'], 'ratio')
+        self.py3.threshold_get_color(data['messages'], 'messages')
+        self.py3.threshold_get_color(data['notifs'], 'notifs')
+
         full_text = self.py3.safe_format(self.format, data)
         return {
             'full_text': full_text,
             'cached_until': self.py3.time_in(self.cache_timeout),
-            'color': color
         }
 
 
--- a/py3status/modules/gazelle.py
+++ b/py3status/modules/gazelle.py
@@ -14,9 +14,6 @@ Configuration parameters:
         (default 'https://mytracker.stuff')
     cache_timeout: How often we refresh this module in seconds (default 60)
     format: See placeholders below (default 'Ratio: {ratio}')
-    limit_required: minimum ratio tolerated by tracker (if defined, overrides value given
-        by API, so you are just likely not to fill it) (default None)
-    limit_warning: custom warning ratio (default 1)
     password: Account password (default '')
     thresholds: specify color thresholds to use
         *(default {'ratio': [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')],
@@ -46,8 +43,6 @@ class Py3status:
     baseurl = 'https://mytracker.stuff'
     cache_timeout = 60
     format = 'Ratio: {ratio}'
-    limit_required = None
-    limit_warning = 1
     password = ''
     thresholds = {
             'ratio': [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')],
--- a/py3status/modules/gazelle.py
+++ b/py3status/modules/gazelle.py
@@ -13,9 +13,6 @@ Configuration parameters:
     baseurl: Base URL of the tracker (with protocol and no trailing slash)
         (default 'https://mytracker.stuff')
     cache_timeout: How often we refresh this module in seconds (default 60)
-    cookiepath: Path of the file containing identification cookie, can be None if
-        you want to create a new session at each load of the module
-        (default '/var/tmp/gazelle_ratio_cookie.txt')
     format: See placeholders below (default 'Ratio: {ratio}')
     limit_required: minimum ratio tolerated by tracker (if defined, overrides value given
         by API, so you are just likely not to fill it) (default None)
@@ -57,7 +54,6 @@ class Py3status:
     # available configuration parameters
     baseurl = 'https://mytracker.stuff'
     cache_timeout = 60
-    cookiepath = '/var/tmp/gazelle_ratio_cookie.txt'
     format = 'Ratio: {ratio}'
     limit_required = None
     limit_warning = 1
@@ -87,20 +83,11 @@ class Py3status:
                     level=self.py3.LOG_ERROR
                     )
             raise
-        try:
-            self._cookiejar.save(ignore_discard=True)
-        except IOError:
-            self.py3.log(
-                    "Cannot write cookie at %s" % self.cookiepath,
-                    level=self.py3.LOG_WARNING
-                    )
-        except AttributeError:
-            pass  # That means that no cookie pass has been set
-        else:
-            self.py3.log(
-                    "Created cookie at %s" % self.cookiepath,
-                    level=self.py3.LOG_INFO
-                    )
+        self._cookiejar.save(self.baseurl)
+        self.py3.log(
+                "Saved cookie in storage",
+                level=self.py3.LOG_INFO
+                )
 
     def _index(self):
         resp = self.py3.request(
@@ -114,25 +101,17 @@ class Py3status:
         return resp.json()
 
     def post_config_hook(self):
-        if self.cookiepath:
-            self._cookiejar = MozillaCookieJar(self.cookiepath)
-            try:
-                self._cookiejar.load(ignore_discard=True)
-                self._index()  # just to try if saved cookie is good
-            except IOError:
-                self.py3.log(
-                        "Cannot read cookie at %s" % self.cookiepath,
-                        level=self.py3.LOG_INFO
-                        )
-                self._login()
-            except RequestException:
-                self.py3.log(
-                        "Cookie invalid or outdated. Re-logging.",
-                        level=self.py3.LOG_INFO
-                        )
-                self._login()
-        else:
-            self._cookiejar = CookieJar()
+        #self._cookiejar = MozillaCookieJar(self.cookiepath)
+        self._cookiejar = self.py3.storage_new_cookiejar()
+        try:
+            self._cookiejar.load(self.baseurl)
+            self._index()  # just to try if saved cookie is good
+        except RequestException:
+            self.py3.log(
+                    "Cookie invalid, outdated or missing. Re-logging.",
+                    level=self.py3.LOG_INFO
+                    )
+            self._login()
 
     def gazelle(self):
         index = self._index()
--- a/py3status/modules/gazelle.py
+++ b/py3status/modules/gazelle.py
@@ -6,19 +6,20 @@ This module can get various useful infos about you on a private tracker.
 You can display your ratio, upload size, download size.
 The only requirements are that the private tracker is based on the Gazelle framework,
 for instance Apollo or Redacted, and that the sysops didn't block API access somehow.
-This module can save session cookies, and it is advised to avoid too much opened sessions
-that could potentially get you troubles due to a suspect activity.
+This module will save session cookies to avoid too much opened sessions which could
+potentially put you into troubles due to a suspect activity.
 
 Configuration parameters:
-    baseurl: Base URL of the tracker (with protocol and no trailing slash)
-        (default 'https://mytracker.stuff')
+    baseurl: Base URL of the tracker (example: https://what.cd) (default None)
     cache_timeout: How often we refresh this module in seconds (default 60)
     format: See placeholders below (default 'Ratio: {ratio}')
     password: Account password (default '')
     thresholds: specify color thresholds to use
-        *(default {'ratio': [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')],
-        'messages': [(0, 'degraded'), (1, 'good')],
-        'notifs': [(0, 'degraded'), (1, 'good')]})*
+        (default [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')])
+    urgent_new_messages: set output urgent mode when there are new messages
+        (default False)
+    urgent_new_notifs: set output urgent mode when there are new notifications
+        (default False)
     username: Account username (default '')
 
 Format placeholders:
@@ -34,20 +35,21 @@ Format placeholders:
 
 from py3status.exceptions import RequestException
 
+import re
+
 
 class Py3status:
     """
     """
 
     # available configuration parameters
-    baseurl = 'https://mytracker.stuff'
+    baseurl = None
     cache_timeout = 60
     format = 'Ratio: {ratio}'
     password = ''
-    thresholds = {
-            'ratio': [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')],
-            'messages': [(0, 'degraded'), (1, 'good')],
-            'notifs': [(0, 'degraded'), (1, 'good')]}
+    thresholds = [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')]
+    urgent_new_messages = False
+    urgent_new_notifs = False
     username = ''
 
     def _readable_size(self, size):
@@ -91,6 +93,15 @@ class Py3status:
         return resp.json()
 
     def post_config_hook(self):
+        # Check if given URL is really an URL
+        if not isinstance(self.baseurl, str) or not re.search("^https?://", self.baseurl):
+            self.py3.error("Invalid URL")
+        self.baseurl = self.baseurl.strip('/')
+        if not isinstance(self.username, str) or not isinstance(self.password, str):
+            self.py3.error("Invalid login info")
+        # Make a copy of thresholds list with keywords
+        self._thresholds_orig = list(self.thresholds)
+
         self._cookiejar = self.py3.storage_new_cookiejar()
         try:
             self._cookiejar.load(self.baseurl)
@@ -103,11 +114,15 @@ class Py3status:
             self._login()
 
     def gazelle(self):
-        index = self._index()
-        if index['status'] != "success":
-            msg = "Tracker failed to give me user info"
-            self.py3.log(msg, level=self.py3.LOG_ERROR)
-            raise RequestException(msg)
+        try:
+            index = self._index()
+            if index['status'] != "success":
+                msg = "Tracker failed to give me user info"
+                self.py3.log(msg, level=self.py3.LOG_ERROR)
+                raise RequestException(msg)
+        except RequestException:
+            self.py3.error("Network error")
+
         result = index['response']
         data = {
             'ratio': result['userstats']['ratio'],
@@ -122,21 +137,22 @@ class Py3status:
         # the color threshold design in py3status is a gigantic turd.
         # Main reason of this ugliness is the impossibility to natively define a
         # variable threshold, so I have to replace keyword by the actual value.
-        for n, item in enumerate(self.thresholds['ratio']):
+        for n, item in enumerate(self._thresholds_orig):
             if item[0] == "requiredratio":
                 lst = list(item)
                 lst[0] = data['requiredratio']
-                self.thresholds['ratio'][n] = tuple(lst)
+                self.thresholds[n] = tuple(lst)
         # End of horrible code
 
-        self.py3.threshold_get_color(data['ratio'], 'ratio')
-        self.py3.threshold_get_color(data['messages'], 'messages')
-        self.py3.threshold_get_color(data['notifs'], 'notifs')
+        self.py3.threshold_get_color(data['ratio'])
 
         full_text = self.py3.safe_format(self.format, data)
         return {
             'full_text': full_text,
             'cached_until': self.py3.time_in(self.cache_timeout),
+            'urgent':
+                (self.urgent_new_messages and bool(data['messages'])) or
+                (self.urgent_new_notifs and bool(data['notifs']))
         }
 
 
--- a/py3status/modules/gazelle.py
+++ b/py3status/modules/gazelle.py
@@ -39,13 +39,6 @@ Color options:
 
 from py3status.exceptions import RequestException
 
-try:
-    # Python 3
-    from http.cookiejar import (CookieJar, MozillaCookieJar)
-except ImportError:
-    # Python 2
-    from cookielib import (CookieJar, MozillaCookieJar)
-
 
 class Py3status:
     """
@@ -101,7 +94,6 @@ class Py3status:
         return resp.json()
 
     def post_config_hook(self):
-        #self._cookiejar = MozillaCookieJar(self.cookiepath)
         self._cookiejar = self.py3.storage_new_cookiejar()
         try:
             self._cookiejar.load(self.baseurl)
--- a/py3status/modules/gazelle.py
+++ b/py3status/modules/gazelle.py
@@ -1,0 +1,171 @@
+# -*- coding: utf-8 -*-
+"""
+Show your ratio on your favorite Gazelle-based private tracker.
+
+This module can get various useful infos about you on a private tracker.
+You can display your ratio, upload size, download size.
+The only requirements are that the private tracker is based on the Gazelle framework,
+for instance Apollo or Redacted, and that the sysops didn't block API access somehow.
+This module can save session cookies, and it is advised to avoid too much opened sessions
+that could potentially get you troubles due to a suspect activity.
+
+Configuration parameters:
+    baseurl: Base URL of the tracker (with protocol and no trailing slash)
+        (default 'https://mytracker.stuff')
+    cache_timeout: How often we refresh this module in seconds (default 60)
+    cookiepath: Path of the file containing identification cookie, can be None if
+        you want to create a new session at each load of the module
+        (default '/var/tmp/gazelle_ratio_cookie.txt')
+    format: See placeholders below (default 'Ratio: {ratio}')
+    limit_required: minimum ratio tolerated by tracker (if defined, overrides value given
+        by API, so you are just likely not to fill it) (default None)
+    limit_warning: custom warning ratio (default 1)
+    password: Account password (default '')
+    username: Account username (default '')
+
+Format placeholders:
+    {ratio} account ratio
+    {download} account download amount
+    {upload} account upload amount
+    {messages} number of unread messages
+    {notifs} number of new notifications
+
+Color options:
+    color_ratio_ok: color when ration is bigger than 1 (default color_good)
+    color_ratio_warning: color when ration is between required ratio and 1
+        (available only when tracker has a dynamic required ratio (default color_warning)
+    color_ratio_ko: color when ratio below required ratio (default color_bad)
+
+@author raspbeguy <raspbeguy@hashtagueule.fr>
+@license MIT
+"""
+
+from py3status.exceptions import RequestException
+
+try:
+    # Python 3
+    from http.cookiejar import (CookieJar, MozillaCookieJar)
+except ImportError:
+    # Python 2
+    from cookielib import (CookieJar, MozillaCookieJar)
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    baseurl = 'https://mytracker.stuff'
+    cache_timeout = 60
+    cookiepath = '/var/tmp/gazelle_ratio_cookie.txt'
+    format = 'Ratio: {ratio}'
+    limit_required = None
+    limit_warning = 1
+    password = ''
+    username = ''
+
+    def _readable_size(self, size):
+        num = size
+        for unit in ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB']:
+            if abs(num) < 1024.0:
+                return "%3.1f %s" % (num, unit)
+            num /= 1024.0
+        return "%.1f%s" % (num, 'YiB')
+
+    def _login(self):
+        self.py3.request(
+                self.baseurl+"/login.php",
+                data={'username': self.username, 'password': self.password},
+                headers={'User-Agent': 'Mozilla/5.0'},
+                cookiejar=self._cookiejar
+                )
+        try:
+            self._index()  # just to try if login info are good
+        except RequestException:
+            self.py3.log(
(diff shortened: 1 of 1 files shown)


tobes:
To me this module is doing too much of the work with the cookie jar.  Can we not push more of this back into the request stuff in #1193?

Maybe we could even look at storing them via #583 (when we get the merged)

But the simpler we can make it for module writers the better - but it is good to see this example of cookie usage

raspbeguy:
I had this debate with myself earlier. I let the module deal with cookies for 2 main reasons:
* The request part must remain simplistic, as it is precised in the comments.
* The ratio flexibility/complexity is quite good in my opinion. I think it is essential to have the possibility to specify your own type of cookiejar (I'm using here MozillaCookieJar for maybe some people would prefer regular non-persistent CookieJar in other modules). To keep that flexibility, it would require to add more ugly options, something like `usecookies=True` and `persistantcookies=True`.

Concerning #583, I think this is a good idea too, my idea is to implement our own subclass of `CookieJar` that wound be persistent in the sqlite database using this system. That way we could simplify the work of the module developer, keep everything compatible and keep the flexibility which is very needed in my opinion.

raspbeguy:
Also, a cookiejar is useless is this can be used just for one request. So the cookiejar must be shared with the multiple requests done in a module, therefore the module has to manage the cookiejar.

tobes:
>  So the cookiejar must be shared with the multiple requests done in a module, therefore the module has to manage the cookiejar.

we can do all this for the module very easily (each module has it's own `Py3` instance that is persistant).  I'm happy to allow modules to handle the cookie jars if they really need to but I don't see that need here.  To my mind we could just do `request(url ..., cookie_jar=True)` or similar and have py3status handle everything.  Modules could always specify their own `request(url ..., cookie_jar=my_cookie_jar)`

raspbeguy:
Well, I personally find ugly to pass either a boolean either an object as a parameter, and those dirty hack need to be very precisely documented to be usable, except that generally they aren't.

Anyway, if we admit that we let that happen, I think this could work, the only thing that annoys me is that, by using the storage thing defined in #583, every single cookie would be persistent (and by persistent, I mean survive to a restart and be written to the disk), and yet some people might not like that, as it would induce unnecessary disk IO for cookies that are only temporary. I'm not sure that I'm clear on that point.

Anyway, it would be nice to add a system not to use the `save` method or your `storage` object on some values, because they wouldn't need to be saved.

tobes:
>  can I ask why it stalled?

Two main reasons

a) It is quite a long bit of code and this has complicated the review process (we keep pushing it back a release due to time pressures)

b) @ultrabug is not so keen on the keys for `storage_set(key, value)` and just wanted a single `storage_set(value)` but I find this makes using the storage more painful and you can do this by just doing `storage_set(None, value)` maybe I'll see if `storage_set(value, key=None)` would make him happier but to me that parameter order feels wrong

I'll push on getting it merged as it has been pending for a year now

> Well, I personally find ugly to pass either a boolean either an object as a parameter, 

how about I do this in a module and we forget the boolean 

```
def post_config_hook(self):
    self.jar = self.py3.CookieJar(persist=False)

def update(self):
    result = self.py3.request(url, cookie_jar=self.jar)
```

raspbeguy:
Yep, I like it way better, that's what I was talking about when I said we could implement our own `CookieJar` subclasses. I think it's way more elegant and that it keeps flexibility.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,171 @@
...
+            color = self.py3.COLOR_RATIO_OK or self.py3.COLOR_GOOD
+        elif data['ratio'] >= (self.limit_required or data['requiredratio']):
+            color = self.py3.COLOR_RATIO_WARNING or self.py3.COLOR_DEGRADED
+        else:
+            color = self.py3.COLOR_RATIO_KO or self.py3.COLOR_BAD
+        full_text = self.py3.safe_format(self.format, data)
+        return {
+            'full_text': full_text,
+            'cached_until': self.py3.time_in(self.cache_timeout),
+            'color': color
```
lasers: For more flexible customization in coloring, we should remove this `color` line and use `thresholds` instead. The new `format` should be `'\?color=ratio {ratio}'` when it's done correctly.

Also, omit blank line `line 56` for consistency/QA purpose.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,171 @@
...
+            self.py3.log(msg, level=self.py3.LOG_ERROR)
+            raise RequestException(msg)
+        result = index['response']
+        data = {
+            'ratio': result['userstats']['ratio'],
+            'requiredratio': result['userstats']['requiredratio'],
+            'download': self._readable_size(result['userstats']['downloaded']),
+            'upload': self._readable_size(result['userstats']['uploaded']),
+            'messages': result['notifications']['messages'],
+            'notifs': result['notifications']['notifications']
```
lasers: This seems like you're cherry-picking some data. Is there more data? Maybe we should use `flatten_dict` and expose everything? We can get more new placeholders without changing anything. Can you output `index`? :-)

Also, what's the difference between `messages` and `notifications`? I think with more work, we could add `urgent` for them... and if `index` contains an url, we might be able to use that too for `on_click` clicking purposes. ;-)

The new placeholders might look like this.
```
{userstats_requiredratio}
{userstats_downloaded}
{userstats_uploaded}
{notifications_messages}
{notifications_notifications}
etc...
```
If you don't like that, we might be able to skip prefixes too if we can guarantee no clashes.

I have been doing the good job of converting few cherrypicked placeholders to exposing all possible placeholders in lot of modules, eg cmus, moc, deadbeef, nvidia_smi, coin_market, and more... so I'd like for this module to do the same thing too.... It can be useless information, but still good information somehow.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,171 @@
...
+                        level=self.py3.LOG_INFO
+                        )
+                self._login()
+        else:
+            self._cookiejar = CookieJar()
+
+    def gazelle(self):
+        index = self._index()
+        if index['status'] != "success":
+            msg = "Tracker failed to give me user info"
```
lasers: (Guess) I think `index['status']` may fail here on no Internet / error.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,171 @@
...
+                cookiejar=self._cookiejar
+                )
+        return resp.json()
+
+    def post_config_hook(self):
+        if self.cookiepath:
+            self._cookiejar = MozillaCookieJar(self.cookiepath)
+            try:
+                self._cookiejar.load(ignore_discard=True)
+                self._index()  # just to try if saved cookie is good
```
lasers: We should skip this and do it in `gazette` instead because this could lead to slow startup for all modules. You can try and test this offline... and maybe just do simple `first_run` loop (see `coin_market` again). Not sure about others.

lasers: @tobes I kept mentioning this for Internet modules. I take it it is not doable to fix that inside py3status? To have all modules return `''` on first run regardless?

tobes: The problem is that currently we start each module in the main thread.  We need to move to a situation where each module is started in a thread therefore slow modules do not slow the whole process.  I think fixing that will be easier than enforcing clean/quick starting modules.  It may also make the logic of modules simpler if they can do slow startups.

We need to be careful here as container modules need to have their 'children' started before themselves, but that is not too hard to do (it may also be something that can be removed as a dependency but currently it is an issue)

We now have the structures in place to do this so I may look into it.  One problem is that really we need to make some structural changes and these can be painful to review and merge, but it is something we need to do to make py3status more slim and agile.

I created #1197 for this

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,171 @@
...
+This module can save session cookies, and it is advised to avoid too much opened sessions
+that could potentially get you troubles due to a suspect activity.
+
+Configuration parameters:
+    baseurl: Base URL of the tracker (with protocol and no trailing slash)
+        (default 'https://mytracker.stuff')
+    cache_timeout: How often we refresh this module in seconds (default 60)
+    cookiepath: Path of the file containing identification cookie, can be None if
+        you want to create a new session at each load of the module
+        (default '/var/tmp/gazelle_ratio_cookie.txt')
```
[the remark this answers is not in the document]
raspbeguy: I think having `cookiepath` instead of `cookie` is better. Remember the second rule of the Zen of Python: "Explicit is better than implicit."

py3status/modules/gazelle.py
raspbeguy: Well, guess I have to protect `self._index` then.

py3status/modules/gazelle.py
raspbeguy: See [here](https://github.com/WhatCD/Gazelle/wiki/JSON-API-Documentation#index) for a list of mandatory json terms prefent in the response. I'll see about that, dunno if that's a good idea though.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,171 @@
...
+    {download} account download amount
+    {upload} account upload amount
+    {messages} number of unread messages
+    {notifs} number of new notifications
+
+Color options:
+    color_ratio_ok: color when ration is bigger than 1 (default color_good)
+    color_ratio_warning: color when ration is between required ratio and 1
+        (available only when tracker has a dynamic required ratio (default color_warning)
+    color_ratio_ko: color when ratio below required ratio (default color_bad)
```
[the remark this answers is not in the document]
raspbeguy: Actually that's impossible here, because the thresholds are partly dynamic because given by the tracker response. Actually you can be all static if you use `limit_required` but I'm gonna remove that because I realize it's totally useless to override value given by tracker. So nope, can't do.

py3status/modules/gazelle.py
raspbeguy: See my previous comment about thresholds.

py3status/modules/gazelle.py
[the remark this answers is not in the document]
raspbeguy: Well, I can use the threshold mechanic, user would be able to use it it the `format` variable, but won't be able to define it as a nice threshold dict.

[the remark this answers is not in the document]
lasers: From my perspective, it sounds like a `threshold` setting that users can configure to make things working correctly (with the site). A nice example easily can show them how to. Otherwise, `gazelle` will always be fully painted in `bad`, `degraded`, or `good` color with no way to turn it off.

With thresholds, users will be able to paint things their way instead of no choice.
* `\?color=ratio Ratio {ratio}`  <---- This is yours. No customization.
* `Ratio [\?color=ratio {ratio}]` 
* ` [\?color=ratio {ratio}]` 

_`New!`_ Individual colors...
* `Ratio [\?color=ratio {ratio}] Download [\?color=download {download}] Upload [\?color=upload {upload}]`

Heck, we even could colorize messages too...
* `\?color=message {message}`
* `\?color=notif {notifs}`

@tobes and I love rainbows. 🌈  We can rainbow your ratio with this.
```
# bad example
format = '\?color=ratio {ratio}
thresholds = [(0.1, '#fcf'), (0.2, '#fff'), (0.3, '#ffz'), (0.4, '#asdasd')]
```
Et cetera.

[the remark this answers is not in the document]
lasers: Try `coin_market` with `markets = 5` config. An example of what it could look like.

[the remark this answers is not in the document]
raspbeguy: What I said is that here, threshold value depends of the response of the API. Some users will have a minimum ratio of, let's say, 0.1 and some will have a minimum ratio of 0.9. This value is given by the API. So, except if you can write something like this in the config:
```python
thresholds = [
    (0, "bad")
    (ratio_minimum, "degraded")
    (1, "good")
]
```
with   `ratio_minimum` internally calculated at runtime, I'm afraid it won't do it.
Maybe giving abstract variables like this is possible, and in that's case it's great.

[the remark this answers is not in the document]
lasers: The `ratio_minimum` value from API shouldn't change? Rarely?
If yes, then for this site, I find out what `ratio_minimum` is, then I can configure the threshold to...
```
thresholds = [
    (0, "bad")
    (0.75, "degraded")
    (1, "good")
]
```

What if users does not want any colors? Just plain white or grey? I think using `color = '#fff'` wouldn't work for your module because it's going to internally use `response['color']`. So that could be an issue too.

I acknowledge your issue and I'm just saying this because that's the usual way to do things w/ colors, formatters, etc even if it's an issue somewhere else. Maybe @tobes can find a solution by feeding `self.py3.thresholds` some dict too.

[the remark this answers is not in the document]
raspbeguy: `ratio_minimum` depends of your download amount and of the proportion of torrents you're seeding. So yeah, it can change quickly.

Anyway, it would be unacceptable to manually set a value when the application can get it for you (this is the case not only for py3status of course).

I was thinking that I could use threshold formatted text anyway, but the threshold dict wouldn't be defined directly by the user, the colors would be set with traditional parameters variable. Then the threshold dict would be defined internally on the module.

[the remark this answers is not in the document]
lasers: Maybe leave the threshold dict exposed for other placeholders, eg `download`, `upload`, etc... but overwrite the `ratio` or `requiredratio`.... or I think better yet, make up a new `threshold` name like `auto` or `dynamic` for this purpose.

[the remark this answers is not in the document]
raspbeguy: I'll test to put internal variables in the dict, maybe that'll work, what do I know...

py3status/modules/gazelle.py
raspbeguy: In fact, can't do here: some values in the json are boolean and you can't give booleans in the py3status response data, right?

lasers: We can deal with boolean. New module `input` contains over 100 boolean (keys).

See `hueshift` for a good example. There are two boolean in the `format`. One will be used to disable the module when we detect `redshift` process via `pgrep` and the other one is to determine when to hide `is_scrolling`.

Not only that, but we can do some conditions too like...
* `'\?if=percent>50 More than halfway.'`
* `'\?if=percent>95 Almost done!'`

lasers: Knowing you made your own `cmus` module long time ago... The `cmus` I made have this format too.
```
    format: display format for this module
        *(default '[\?if=is_started [\?if=is_playing > ][\?if=is_paused \|\| ]
        [\?if=is_stopped .. ][[{artist}][\?soft  - ][{title}]
        |\?show cmus: waiting for user input]]')*
```

raspbeguy: I dodn't say you couldn't use booleans in `format` field. I said that they can't be used as placeholder (which may only be strings, integers or floats if I'm not wrong) and so they can't be passed in the `full_text` data dict.

lasers: They can be used as placeholders. It'd show `True` or `False`. Useless. In this case, we almost always will want to use `\?if=placeholder_name` instead. Also, check out the `Control placeholders` for module `moc` and `cmus`. All of that is boolean.

raspbeguy: Also on another topic, I wonder why cmus got finally it's own module whereas back in the days it was refused because it was preferable to adapt the existing music module to support cmus in order to avoid creating modules for specific program instead of modules for specific functions. I understood and agreed with this principle.

Also now cmus supports MPRIS wich wasn't the case back then, which makes it entirely compatible with the mpris module.

But we might talk about that in another place.

raspbeguy: OK then, never mind about placeholders. I'll think about that.

raspbeguy: Don't misunderstand, I'm not angry or anything, especially not against you, I'm just curious about how the rules changed and I want to understand those rules in order to make my PRs in this objective and thus spend much less time on fixing philosophical stuff.

@ultrabug @tobes the documentation contains all the technical knowledge to create a well functioning module, and that's great. Maybe you should also consider writing a contribution manifest explaining for instance which conditions are needed in order to create a new module, your vision of module improvement, what is the general direction of the project, what are the good practices...
I know that a little of this info is spread in the whole *writing custom py3status modules* section, but I really think that those "philosophical" considerations should have their own subsection, or even their own entire section.

lasers: I'll try and explain this the best I can.

I think @tobes wanted us to go in a direction where we don't script anything in the code by exposing everything (if possible) as placeholders then users can manipulate the `if`, `else`, etc in the `format` instead of making pull requests. See issue... https://github.com/ultrabug/py3status/issues/752

If you want to, try `window_manipulation`.... That module contains an amount of placeholders and I was able to manipulate lot of things.... With more placeholders, we can do more neat things... without touching the code.

The key is essentially to get all placeholders we can get our hands on in all modules... and have @tobes become a farmer where he can grow the formatters this summer.

Users want colors. Users don't want colors. We can make both things possible.
Users want urgent. Users don't want urgent. We can make both things possible.
Users want floats or ints... We can make both things possible.
Users want to hide or scrub things. We're going to try and make that possible.
Users... et cetera.

This requires lot of changes though... For instance, all modules should stop returning `"color": color` because `module { color = '#ffccff'}` wouldn't work with that since it's hardcoded... Like I said, lot of changes.
* Remove hardcoded color returns.
* Merge separate formats into one `format`. 
* Undo any `int(floating_value)`.
* How to deprecate things properly? I hate this.
* We can compare basic words. `\?if=class=VIP Good Boy!`
* Remove configs like `hide_if_zero` in favor of `'\?if=apt {apt} update}'` or `'\?not_zero {apt} update'`. 

Users does not know about those things too... due to lack of docs and whatnot. Users may not understand that amount of customization too... so I have been trying to come up with decent examples for users to copy and paste into their formats... and make small changes from there.

One problem I think is that @ultrabug may be wanting to keep things stable and make it easy for users to change settings via configs instead of scriptable `format`.

Another problem, I make many PRs. It scares and/or annoys them.
Another problem, they're slow on reviewing and/or merging. It annoys me.

Disclaimer: There are lot of things I don't know about py3status or python... and most of time, I'm just guessing things... and how things work around here... and I'm always doing something wrong too. I'm just a contributor here. Same as you.

raspbeguy: Thanks for your explanations. I still think that it must be formalized somewhere "officially" but this is the idea.

By the way, I find this format language very uncomfortable and unintuitive, but I guess it cannot be helped and this is the best that could have been done. Also this would be so much easier to use **if this was documented somewhere**. People can't think to use stuff it they are not documented. I don't know how long ago this feature exists (I don't remind such a thing from back in the days) and how long ago the `color` field method has been deprecated, but **it's urgent and capital to write something about it in the official documentation**. Else casual contributors like me will continue coding with their feet and everybody will spend more time on debugging stupidities, which leads to less time to important thing like documentation and "good" coding, which is a vicious circle.

I know you're also a (much less) casual contributor (than me) so that's why it needs to be formalized and acknowledged by an official core developer.

I'm not mad, I just express my feeling about what I objectively think to be the good way to progress, so please excuse my strong sentences with **bold text**, if it offend someone that surely wasn't my intention.

tobes: @raspbeguy 

> I just express my feeling about what I objectively think to be the good way to progress, 

Expressing your opinions is always encouraged here.  And both you and @lasers raise some good points.  There are definitely some things that need improving with the project.  Some documentation is desperately needed such as the formatting and some better guidance on module development.

From my point of view I want things to be easy for user and I think we should develop for 2 distinct users.

a) Users who just want things to work.  To me a good example would be the `volume_status` as a user I should be able to just do `order += 'volume_status'` and should get a sensible working volume control (I don't think we are here as the volume changing controls isn't by default)

b) expert users who want to set things up exactly as they want full control of colors etc.  This also to an extent includes people who just want to make small changes say to the format.

Because of this I want to keep things like default formats really simple.  @lasers for example likes to make lots of format usage (different colors for different states)  I resist all attempts for complex default formats (@ultrabug is more forgiving here)

for me simplicity is key so I will always struggle with things like

```
-    format = '{caps} {num} {scr}'

+    format = ('[\?if=caps_lock&color=good CAPS|\?color=bad CAPS] '
+              '[\?if=num_lock&color=good NUM|\?color=bad NUM] '
+              '[\?if=scroll_lock&color=good SCR|\?color=bad SCR]')
```

mainly because to me it shows we need to improve the formatter to make this simpler

I'm also keen to have as few modules as we can that do the same thing (music players being an example)  I'd much prefer one really solid mpris module rather than many different modules one for each player.

@lasers 
>  I make many PRs. It scares and/or annoys them.

I'd say that it is more it makes it harder to manage things as reviewing all open PRs is not easily possible when there are so many of them.  Things can get lost etc.  Sometimes things are just slow.  You also like to make many changes in a PR and so a small part can easily block the entire PR.  I know this is frustrating for you.

raspbeguy: I must agree with @tobes on the "keep configuration simplicity" thing. That format language is dreadful, and a lot of regular user won't even try to understand it. Also, I have no alternative to propose, but I know this will be a real pain for me to use that.

raspbeguy: @tobes Maybe that contribution manifest I was speaking about should include a line like "Every new core feature should be documented in the official documentation" so that a PR adding features to core code (not a module) shouldn't be accepted until the contributor (or anybody) takes the time to write in the official doc what this new feature is and how to use it. I had the case with the Request code that contained a lot of non-documented stuff so I wasn't even able to write down my own feature documentation.

I had plans to code stuff, but I guess I'll spend a few PR to write some documentation (about what I can understand in the code I touched).

raspbeguy:
Fixed it to use storage module instead of cookie file. So now it needs #1278 to be merged before.

lasers:
//P//L//E//A//S//E//  //R//E//B//A//S//E//

raspbeguy:
Huh, that's what I did, that's why there are so many commits.

raspbeguy:
OK, not a git guru.
If you know what to do to make this clean again, I'd glad to hear you.

raspbeguy:
I don't think I deserved those punches :hurtrealbad:
Anyway it worked, thanks.

py3status/modules/gazelle.py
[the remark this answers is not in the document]
raspbeguy: @lasers when you define multiple thresholds, how do you specify the threshold to use in the `format` variable?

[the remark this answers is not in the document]
lasers: `format = '\?color=volume Volume: {percent}`
`self.py3.threshold_get_color(data['percent'], 'volume')`

`sysdata.py`
>    format: output format string
        *(default '[\?color=cpu CPU: {cpu_usage}%], '
    thresholds: thresholds to use for color changes


`netdata.py`
>    format: display format for this module
        *(default '{nic} [\?color=down LAN(Kb): {down}↓ {up}↑]
        [\?color=total T(Mb): {download}↓ {upload}↑ {total}↕]')*
        *(default {'down': [(0, 'bad'), (30, 'degraded'), (60, 'good')],
        'total': [(0, 'good'), (400, 'degraded'), (700, 'bad')]})*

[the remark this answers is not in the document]
raspbeguy: ok, didn't know about `threshold_get_color`.
I feel the headache coming. Honestly that color threshold system is so awfully designed I want to eat my own limbs.

raspbeguy:
I finally added threshold stuff, I did something to support variable threasholds.
I think it's ready for merge.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,153 @@
...
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    baseurl = 'https://mytracker.stuff'
+    cache_timeout = 60
+    format = 'Ratio: {ratio}'
+    limit_required = None
+    limit_warning = 1
```
lasers: We don't need `limit_*` anymore.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,153 @@
...
+        for n, item in enumerate(self.thresholds['ratio']):
+            if item[0] == "requiredratio":
+                lst = list(item)
+                lst[0] = data['requiredratio']
+                self.thresholds['ratio'][n] = tuple(lst)
+        # End of horrible code
+
+        self.py3.threshold_get_color(data['ratio'], 'ratio')
+        self.py3.threshold_get_color(data['messages'], 'messages')
+        self.py3.threshold_get_color(data['notifs'], 'notifs')
```
lasers: (Q) Is it more useful to use `urgent` instead of thresholds for `messages` and/or `notifs`?

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Ok, so if you think that what I'll do next stinks a bit, remember that
+        # the color threshold design in py3status is a gigantic turd.
+        # Main reason of this ugliness is the impossibility to natively define a
+        # variable threshold, so I have to replace keyword by the actual value.
+        for n, item in enumerate(self.thresholds['ratio']):
+            if item[0] == "requiredratio":
+                lst = list(item)
+                lst[0] = data['requiredratio']
+                self.thresholds['ratio'][n] = tuple(lst)
```
lasers: (Small nit) This is used only on first loop. You said your required ratio could change later meaning `requiredratio` could be changed anytime... or maybe much faster if you suck at seeding. `i3-msg restart` would fix it (gives it new ratio). Just pointing out a small thing.

raspbeguy: Yes, I need to keep the position of the variable threshold in a variable somewhere.

py3status/modules/gazelle.py
raspbeguy: dunno, maybe.

raspbeguy:
I think the network exceptions should be catched by the main thread. A standard way to display offline modules is needed.

tobes:
> I think the network exceptions should be catched by the main thread. A standard way to display offline modules is needed.

I think something like this would be good.  Also such modules auto refreshing when the network is available etc.  But I think it will be a while before we get this.

py3status/modules/gazelle.py
raspbeguy: I'll set 2 new boolean variables `urgent_new messages` and `urgent_new_notifs`

raspbeguy:
All right, next code review please.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,164 @@
...
+    thresholds = [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')]
+    urgent_new_messages = False
+    urgent_new_notifs = False
+    username = ''
+
+    def _readable_size(self, size):
+        num = size
+        for unit in ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB']:
+            if abs(num) < 1024.0:
+                return "%3.1f %s" % (num, unit)
```
lasers: We could use `self.py3.format_units` instead. Expose units as placeholders too for more flexible customization, eg `[\?if=max_length 1 {download_unit}]` to print `K`, `M`, `G`, etc. Also, the cutoff is at 1000. File a bug / PR if you want to change that. File manager `nautilus` (K->M) cutoff seems to be at 977 for me... saying it's `1.0 MB`.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+
+    # available configuration parameters
+    baseurl = None
+    cache_timeout = 60
+    format = 'Ratio: {ratio}'
+    password = ''
+    thresholds = [(0, 'bad'), ('requiredratio', 'degraded'), (1, 'good')]
+    urgent_new_messages = False
+    urgent_new_notifs = False
```
lasers: Thoughts on using per-module configuration `allow_urgent = False` to turn them off? I assume users won't get too much messages / notifications to have a need to turn one off? I'm wondering if this is a superflous config and if we can avoid it... Also, Idk the diff between `messages` and `notifications`.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,164 @@
...
+        if not isinstance(self.username, str) or not isinstance(self.password, str):
+            self.py3.error("Invalid login info")
+        # Make a copy of thresholds list with keywords
+        self._thresholds_orig = list(self.thresholds)
+
+        self._cookiejar = self.py3.storage_new_cookiejar()
+        try:
+            self._cookiejar.load(self.baseurl)
+            self._index()  # just to try if saved cookie is good
+        except RequestException:
```
lasers: (?) `self.py3.RequestException` (?)

P.S. I wonder if it is better to test/correct `index` in regular `gazelle` loop instead of `post_config_hook`? One less call on every restart... and if something went haywire, the `login` only can be happening by restarting i3bar because this `login` code is in `post_config_hook`?

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,164 @@
...
+                    'User-Agent': 'Mozilla/5.0'
+                    },
+                cookiejar=self._cookiejar
+                )
+        return resp.json()
+
+    def post_config_hook(self):
+        # Check if given URL is really an URL
+        if not isinstance(self.baseurl, str) or not re.search("^https?://", self.baseurl):
+            self.py3.error("Invalid URL")
```
lasers: We `import re` for one-time `post_config_hook` use. Could use `http not in self.url` or `self.url.startswith('http')`.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,164 @@
...
+        self.baseurl = self.baseurl.strip('/')
+        if not isinstance(self.username, str) or not isinstance(self.password, str):
+            self.py3.error("Invalid login info")
```
lasers: Good checking. The usual behavior / consistency for py3status modules is to start them off with `None` then check to make sure they're not `None`.

py3status/modules/gazelle.py
```diff
@@ -0,0 +1,164 @@
...
+                    )
+            self._login()
+
+    def gazelle(self):
+        try:
+            index = self._index()
+            if index['status'] != "success":
+                msg = "Tracker failed to give me user info"
+                self.py3.log(msg, level=self.py3.LOG_ERROR)
+                raise RequestException(msg)
```
lasers: Does this trigger when users have no Internet? If yes, maybe this is a bit harsh... Better behavior, I think, would be to fail silently and keep trying... Earlier review indicates that this can fail. `index.get('status')` may be a safe move.

py3status/modules/gazelle.py
raspbeguy: I really need the 1024 base and I'm not willing to wait for another PR to be accepted. Anyway, this can be done later.
I think that having a placeholder for units is overkill.

py3status/modules/gazelle.py
raspbeguy: * Notifications: On the tracker frontend, you can define some torrents filters. Notifications are the numbers of unseen torrents that match those filters. Frequency is variable depending of your filters.
* Messages: private messages between users. Frequency is kind of low, consider it like a spam-free personal mailbox.

Because those two counters are very different, we need to differentiate the urgent behavior for each of them.

py3status/modules/gazelle.py
raspbeguy: You probably missed this line at the beginning of the module:
```python
from py3status.exceptions import RequestException
```
I don't understand what you're saying about the call to `_index()`. It isn't necessary to test it at each loop.

py3status/modules/gazelle.py
raspbeguy: Testing that the string begins with `http` is incomplete. I really need the regular expression with the optional `s` between `http` and `://`.

py3status/modules/gazelle.py
raspbeguy: Well, people might enter a list, a dict, a number... which would be incorrect.

py3status/modules/gazelle.py
raspbeguy: That's why we nee a standard way to treat offline modules. I know this won't happen for now, so I'll disable this error raise.

py3status/modules/gazelle.py
lasers: It is all fun and games until someone can't customize that one damn thing.

P.S. Your placeholders also messes with the rounding. We can't do this: `{download:d}`, `{download:.0f}`, `{upload:.1f}`, or *New!* `{download:g}`.

py3status/modules/gazelle.py
lasers: Maybe `use_urgent = ['messages', 'notifs']` instead.

py3status/modules/gazelle.py
lasers: Not about testing it at each loop... I'm wondering about using two different requests for similar purpose versus trying for a good request that might be able to take care of everything gracefully.

py3status/modules/gazelle.py
raspbeguy: what's the purpose of that, except making the config even uglier?

py3status/modules/gazelle.py
raspbeguy: So what? the placeholder bullshit syntax makes already the `format` parameter unreadable, I'm not willing to make it even less readable by adding some useless placeholders.

py3status/modules/gazelle.py
raspbeguy: No a big fan of that. I like the idea of being sure to have everything in place at the end of the `post_config_hook`, including the cookie. The main loop shouldn't handle a bad cookie exception nor write a new one in the storage.

py3status/modules/gazelle.py
lasers: Idk. None other than simplifying and/or aiming for minimal number of configs. This is your baby and I'm just messing around with it. Based on your description, the sensible default could be just `use_urgent = ['messages']` as `notifs` could be too noisy (and/or too repeatedly). IIRC we can't use `urgent = []` because it's a reserved word.

Ugly config? Huh... I thought it looks nice. Maybe in the future, we could have `[\?if=messages&urgent ]`... Ugly but could be very flexible with urgent. Idk if that'll even happen. :-) Anyhow, I defer this suggestion to you/others.

Maybe it's better to just work with `messages` (it's more of a good purpose for `urgent` rather than noisy `notifis`)... then we might not need those `urgent` configs.

raspbeguy: I really don't see the need for making it a list instead of 2 separate vars. It seems way clearer for me. Anyway I would like to avoid using python syntax in a config file as much as possible (that also explain my hate for the thresholds system).

If notifs are too noisy, then the user will decide to turn `urgent` off for them. In my case, I want to get special event when I have some notifications, which aren't that noisy.

lasers:
@raspbeguy Hey. Not long ago, I made this https://github.com/ultrabug/py3status/pull/1428. I needed it for `lm_sensors`. You can use it for this module too. It should look something like this...

```python
    format = '\?color=auto.ratio Ratio: {ratio}'
```
```python
    ratio = [(0, 'bad'), (data['requiredratio'], 'degraded'), (1, 'good')]
    self.py3.threshold_get_color(data['ratio'], ('auto.ratio', None, ratio))
```

`\?color=ratio` is reserved for ratio where users can specify threshold... Users probably won't use it because of `auto.ratio`, but we need to let them to do what they want... just in case. Let me know how that works out for you... or if you need help with this part. Thanks.

ultrabug:
Closing as discussed with @raspbeguy

PR #1208, opened 2017-12-21
move arg parsing to own file to help decouple stuff

tobes:
This is the start of some general cleanups.  As can be seen there are some FIXMEs but I want to keep this PR as simple as possible with the minimum of code change.


ultrabug:
👍 perfect, thanks

PR #1229, opened 2018-01-14
exchange_rates: use new feed

tobes:
New feed as Yahoo api no longer available see #1174

Also makes `update_placeholder_formats()` available to modules

precission has been lowered due to the data being old and so less accurate so this seems appropriate

### DIFF
--- a/py3status/modules/exchange_rate.py
+++ b/py3status/modules/exchange_rate.py
@@ -2,7 +2,7 @@
 """
 Display foreign exchange rates.
 
-The exchange rate data comes from Yahoo Finance.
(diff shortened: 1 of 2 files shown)


ultrabug:
Please update the py3 documentation since you add a new user facing functionality of py3

tobes:
> Please update the py3 documentation

But it is auto documented ;p  Just merge this and it'll be in the docs

ultrabug:
/me wishes the formatter documented itself as well

tobes:
> /me wishes the formatter documented itself as well

haha

PR #1236, opened 2018-01-14
wifi: set after, not before

lasers:
Closes https://github.com/ultrabug/py3status/issues/1231. Thanks for the report! @maximbaz


tobes:
@maximbaz thanks for the bug report/testing
@lasers thanks for the quick response and the fix

PR #1246, opened 2018-01-19
start modules in threads to parallelize

tobes:
Currently modules are loaded in series and so one slow module running it's `post_config_hook()` blocks starting all modules.

This PR allows each module to be started in a thread.

We need to ensure that the module has been setup before it updates. hence `self.enabled`


ultrabug:
Thanks, neat & simple

PR #1247, opened 2018-01-19
github: don't treat first run as special

tobes:
Simplify the code.

This slightly changes the behaviour as we now do not display until we get data.  Before we showed ?.

I think this makes things cleaner and is worth the minor change

This needs #1246 as otherwise it slows down startup of py3status.


ultrabug:
good call, thx

Issue #1209, opened 2017-12-21
gevent unicode issue


tobes:
@ultrabug If I run with the `-g` option under python 2.7 I get the following error.  Without `-g` everything runs as expected.  I'd really like to understand what is happening here, rather than just patch the issue via a try/except

```
2017-12-21 19:41:58 INFO Starting py3status version 3.8_pre python 2.7.14
2017-12-21 19:41:58 INFO git branch: master
2017-12-21 19:41:58 INFO git commit: 96931e6 Fast-forward
2017-12-21 19:41:58 INFO gevent monkey patching is active

<snip>

2017-12-21 19:41:58 WARNING Instance `scratchpad_counter`, user method `scratchpad_counter` failed (UnicodeEncodeError) py3.py line 900.
2017-12-21 19:41:58 INFO Traceback
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb7' in position 4999: ordinal not in range(128)
  File "/home/toby/dev/py3status/py3status/module.py", line 729, in run
    response = method()
  File "/home/toby/dev/py3status/py3status/modules/scratchpad_counter.py", line 54, in scratchpad_counter
    tree = loads(self.py3.command_output('i3-msg -t get_tree'))
  File "/home/toby/dev/py3status/py3status/py3.py", line 900, in command_output
    output = output.decode('utf-8')
  File "/usr/lib64/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
```

ultrabug:
It comes from the command_output method of py3 when we decode the output...

```python
        if self._is_python_2:
            output = output.decode('utf-8')
            error = error.decode('utf-8')
```

But why I dunno yet. If I simply remove the decode there, everything works fine (so I don't even know why we try to decode anyway).

tobes:
Yeah I didn't have the time to look at it in detail, or to test if other modules are affected.

ultrabug:
It comes from somewhere in here: https://github.com/gevent/gevent/blob/master/src/gevent/subprocess.py#L265

ultrabug:
sorry, since we implement the check_output ourselves, must come from somewhere in here instead: https://github.com/gevent/gevent/blob/master/src/gevent/subprocess.py#L373

either way, our way of (re)implementing check_output is breaking the subprocess monkey patching made by gevent, that's granted

tobes:
I think it maybe related to `six`

[...]

PR #1250, opened 2018-01-19
add mellowplayer module, fork from spotify module

cyrinux:
Hi,

Just a little module, forked from spotify, to display track infos from service supported by mellowplayer.
Infos depend on the service.

### DIFF
--- a/py3status/modules/mellowplayer.py
+++ b/py3status/modules/mellowplayer.py
@@ -1,0 +1,169 @@
+# -*- coding: utf-8 -*-
+"""
+Display song currently playing in mellowplayer. (fork from spotify module)
+
+Configuration parameters:
+    cache_timeout: how often to update the bar (default 5)
+    format: see placeholders below (default '{artist} : {title}')
+    format_down: define output if mellowplayer is not running
+        (default 'mellowplayer not running')
+    format_stopped: define output if mellowplayer is not playing
+        (default 'mellowplayer stopped')
+    sanitize_titles: whether to remove meta data from album/track title
+        (default True)
+    sanitize_words: which meta data to remove
+        (default ['bonus', 'demo', 'edit', 'explicit',
+                  'extended', 'feat', 'mono', 'remaster',
+                  'stereo', 'version'])
+
+Format placeholders:
+    {album} album name
+    {artist} artiste name (first one)
+    {time} time duration of the song
+    {title} name of the song
+
+Color options:
+    color_offline: mellowplayer is not running, defaults to color_bad
+    color_paused: Song is stopped or paused, defaults to color_degraded
+    color_playing: Song is playing, defaults to color_good
+
+i3status.conf example:
+
+```
+mellowplayer {
+    format = "{title} by {artist} {time}"
+    format_down = "no mellowplayer"
+}
+```
+
+Requires:
+    mellowplayer (>= 3.3.1.0, depend on service)
+
+@author Pierre Guilbert, Jimmy Garpehäll, sondrele, Andrwe, Cyril Levis
+
+SAMPLE OUTPUT
+{'color': '#00FF00', 'full_text': 'Rick Astley : Never Gonna Give You Up'}
+
+paused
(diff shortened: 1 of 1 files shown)


lasers:
The (only real) change is:
```
<             self.__bus = bus.get_object('org.mpris.MediaPlayer2.spotify',
---
>             self.__bus = bus.get_object('org.mpris.MediaPlayer2.MellowPlayer',
```

This may already work in `mpris`. Nice if you could confirm this. With clicks and everything.

I think would be better to add a fix in `spotify` to support `mellowplayer` too, but at a glance, it seems that `mellowplayer` is much more than `spotify` as it also supports other streaming music sites too... and is open source.

Maybe `mpris` works okay with `spotify` too. If `mpris` works okay with `mellowplayer`,
then the real fix may be just to fix `mpris` some more as it would not always behave correctly with vlc, gmpdp, clementine... and possibly spotify/mellowplayer next.

@tobes For fun, I leak I have `newmodule-rhythmbox` branch, but may be absolutely unnecessary due to `mpris`. Anyhow, what do you want to do with this `mellowplayer` PR?

tobes:
@lasers I think if it works with mpris then that is the way to go.  Having one really solid module is where I think we want to be.  Unfortunately mpris is quite complicated but if it works it is my preferred option.  Hopefully we can simplify it over time and get it super stable.

It is similar to the updates issue and having lots of modules becomes hard to maintain, especially if they lack users.  Mpris relies on dbus, which it seems that this module needs that too so the requirement makes no difference.

cyrinux:
Hi,
I would prefer i global mpris module too. I have recheck, and I have no "{time}' info with service I have tested. 
- I can't get working spotify service for the moment on my archlinux, so I don't know if "time" works only only with spotify service?
- Youtube service display any informations.
If a global module is done, how do you think manage several player running in the same time?

tobes:
details of the mpris module are here

https://py3status.readthedocs.io/en/latest/modules.html#mpris

you can specify the player(s) that an instance prioritieses, does it meet your needs?

cyrinux:
Oh thanks @tobes, I wil try this, yes I think it is what i want (on the
paper) ;)

cyrinux:
Hi,
Could you say me how I can help you with my mpris problem?
Started it with 

 /usr/bin/python /usr/bin/py3status --dbus-notify --gevent --debug
{"version": 1, "click_events": true, "stop_signal": 20}
[[]
then it is stuck.
If I control+c it,
 ^C,[{"name": "path_exists", "instance": "PIA", "color": "#fb4934", "markup": "none", "full_text": "PIA: no"}]
...
I start to have all my other modules running but not mpris
Any idea?

lasers:
See if `mpris` will work without `--gevent`. I can't try `mpris` because I don't have `pydbus` available.

If `mpris` is still broken without `--gevent`, then maybe some other commits between py3status versions might had broken `mpris` (again) so we should see if that was the problem here. Otherwise, I have no idea where to start looking... Also, `mpris` might be a hard module to fix too.

cyrinux:
@lasers I confirm this is working without gevents. What should I do? Do you need more information? Thanks for your help

lasers:
Don't use `--gevent` for now... Or try and solve this.

`py3status` added `gevent` not long ago... Last month or so.

Initially, there was a problem with `subprocess` https://github.com/ultrabug/py3status/issues/1209. I don't see `mpris` using any `subprocess` or `Popen` so I assume the problem is related to threading/Thread this time.
* https://github.com/gevent/gevent/blob/master/src/gevent/thread.py
* https://github.com/gevent/gevent/blob/master/src/gevent/threading.py

Does `mellowplayer`/`spotify` work fully with `mpris`? With clicks and everything?

cyrinux:
Mellowplayer seems to work well without gevent. I have test tidal and
soundcloud service inside mellowplayer, no problem, I can play/pause,
next/prev

lasers:
@cyrinux Is it okay for you to close this in favor of working `mpris`?

cyrinux:
Yes @lasers, I will be back if necessary ;)

PR #1254, opened 2018-01-20
Document availability as a Debian package (#1253)

sdelafond:


sdelafond:
I still think adding a note to README.rst (even though it'll be deprecated at some point) would be a good idea...

tobes:
@sdelafond 

Thanks for this.  Yes please add a note on `README.rst` too as that is where most users will look first

PR #1255, opened 2018-01-20
Add documentation about availability in Debian to README.dst (Closes:…

sdelafond:
… #1253)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,6 +117,16 @@ Using dnf:
 
     $ dnf install py3status
 
+Debian
+------
+Packaged by @sdelafond, and available via apt-get:
+::
+
+    $ apt-get install py3status
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -117,6 +117,16 @@ Using dnf:
 
     $ dnf install py3status
 
+Debian
```
tobes: Debian/Ubuntu?

tobes:
Thanks for this 👍

ultrabug:
Thanks a lot @sdelafond !

ultrabug:
btw, there's a mention for pip and debian users:

> Note: Debian users should use pypi-install from the python-stdeb package instead of pip.

shall we remove it or at least drop the bold font?

PR #1263, opened 2018-01-21
allow test modules to use storage

tobes:
#1261 running modules in test mode failed if they used storage.  This is a quick fix for that issue where we cannot get the `i3status_config_path`.

Really we need a proper fix where we load test modules via the full stack.  We will have to consider if test modules can write to our datastore as that may not be something we want them to do.  But that is for another day.


ultrabug:
okay, thank you

PR #1265, opened 2018-01-27
Async script (low-prio)

girst:
I  know y'all busy, but now that I've finished IMAP_IDLE (#1175) I quickly whipped this one up (no hard feelings if you don't want to merge this, as long as my IDLE gets merged sometime ;) )

## What?
This module is similar to `external_script`, but keeps the called script running in the background and updates whenever a new line comes in. It can also update exactly when the script wants instead of whatever py3status' cache_timeout might be.

I've refrained from adding this feature into `external_script` as they are requiring two different scripts and have different use cases. 

## Why

if the _external script_ opens a network connection, it cannot be reused (easily). this way, the script can be kept running and even update not when py3status thinks it needs to, but when the script does, meaning not having to wait for the timeout to call when something happens.

### more
 * I've called it `async_scirpt` -- better ideas?
 * it will switch the color, if a line contains only `#XXXXXX` (can be force-disabled)
 * for testing, I've been using this script to print the date in a random color: 
```
#!/bin/bash
C=(00 7F FF)
while sleep 1
do	date
	echo \#${C[((RANDOM%3))]}${C[((RANDOM%3))]}${C[((RANDOM%3))]}
done
```


girst:
interesting. it works on my machine™

```
order += "async_script"
async_script {
    script_path='ping 127.0.0.1'
}
```

have you tried specifying the full path and wrapping it in a script (so there are no arguments to it)?

lasers:
Sorry, you're right. It is working okay now. I have a messy 700~ config lines. It was using old `async_script` config somewhere far down with different `script_path` pointing to non-existent `/tmp/whatever.sh` script.

P.S. This is nice for `redshift -v`.

lasers:
Sometimes we can get same output repeatedly with certain command or script. This diff prevents updating if the output is same as the last one... and will also stop cluttering up the log.
```diff
diff --git a/py3status/modules/async_script.py b/py3status/modules/async_script.py
index d6fdb8965..4cbca4b05 100644
--- a/py3status/modules/async_script.py
+++ b/py3status/modules/async_script.py
@@ -59,6 +59,7 @@ class Py3status:
         self.command_output = None
         self.command_color = None
         self.command_error = None  # cannot throw self.py3.error from thread
+        self.last_command = None
 
         if not self.script_path:
             self.py3.error("script_path is mandatory")
@@ -93,8 +94,10 @@ class Py3status:
                 if re.search(r'^#[0-9a-fA-F]{6}$', output) and not self.force_nocolor:
                     self.command_color = output
                 else:
-                    self.command_output = output
-                    self.py3.update()
+                    if self.last_command != output:
+                        self.command_output = output
+                        self.py3.update()
+                    self.last_command = output
         except Exception as e:
             self.command_error = str(e)
             self.py3.update()
```

ultrabug:
Thanks for your patience @girst !

Thanks for the patch @lasers I'm gonna add it too :)

Tested good on my part

girst:
no worries; it did say "low prio" directly in the title ;-)

PR #1267, opened 2018-01-28
py3: command_output call command with LC_ALL=C

vvoland:
Many modules parse command's output by searching for keywords which
are mostly English words. This fails when command is able to provide
localized output based on user's locale

This PR adds to `localized` parameter (False by default) to py3.command_output 
to control whether it should set environment variable LC_ALL to C which
forces application to use default (mostly English) locale.

Having this, command output will be independent from user locale.

To consider:
- localized = False by default means that existing modules have to be modified if they
rely on output being localized, but since most (all?) modules scrap data from command's
output instead of displaying the output directly, it might not be a problem.

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -105,6 +105,8 @@ class Py3:
     def __init__(self, module=None):
         self._audio = None
         self._config_setting = {}
+        self._english_env = dict(os.environ)
+        self._english_env['LC_ALL'] = 'C'
         self._format_placeholders = {}
         self._format_placeholders_cache = {}
         self._is_python_2 = sys.version_info < (3, 0)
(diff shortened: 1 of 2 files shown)


py3status/py3.py
```diff
@@ -879,14 +881,15 @@ def command_run(self, command):
...
-    def command_output(self, command, shell=False, capture_stderr=False):
+    def command_output(self, command, shell=False, capture_stderr=False, localized=False):
         """
         Run a command and return its output as unicode.
         The command can either be supplied as a sequence or string.
 
         :param command: command to run can be a str or list
         :param shell: if `True` then command is run through the shell
         :param capture_stderr: if `True` then STDERR is piped to STDOUT
+        :param localized: if `FALSE` then command is forced to use its default (English) locale
```
tobes: `False` would be better

vvoland:
@tobes Done. I also made the localized output in `external_script` optional through **localize** parameter (True by default). Not sure if it should jammed into this PR though. Should I create another one?

tobes:
@vvoland another nice bit of code thanks

PR #1275, opened 2018-01-31
Added emerge status module

AnwariasEu:
Added a module that displays current emerge information.
Multiple format options are available (documented in the module file).

User needs to have read access to the log file in /var/log/emerge.log. This can be acomplished by the Gentoo 'portage' group.

### DIFF
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -25,7 +25,7 @@ Format examples:
     '{category} - {pkg}'
     Result: "x11-misc - py3status"
 
-@author: AnwariasEu
+@author AnwariasEu
 """
 
 import os
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -104,8 +104,6 @@ class Py3status:
                         ret['category'] = res.group('category')
                         ret['pkg'] = res.group('pkg')
                         ret['action'] = res.group('action')
-                    else:
-                        raise Exception('Regex did not match a line')
                     break
         return ret
 
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-
 """
 Emerge_status provides a short information on the current
 emerge status if there is an emerge process currently running
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -44,7 +44,7 @@ class Py3status:
         Returns true if at least one instance of emerge is running.
         """
         try:
-            self.py3.command_output(['pidof', 'emerge'])
+            self.py3.command_output(['pgrep', 'emerge'])
             return True
         except:
             return False
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -13,6 +13,7 @@ Format placeholders:
     {total} Total number of packages that will be emerged
     {category} Category of the currently emerged package
     {pkg} Name of the currently emerged packaged
+    {action} Current emerge action
 
 Format examples:
      {current} / {total}'
@@ -63,6 +64,7 @@ class Py3status:
         ret['total'] = None
         ret['category'] = None
         ret['pkg'] = None
+        ret['action'] = None
 
         input_data = []
 
@@ -92,6 +94,7 @@ class Py3status:
                 if match:
                     status_re = (
                         "\((?P<current>[0-9]+) of (?P<total>[0-9]+)\) "
+                        "\((?P<action>[a-zA-Z0-9\/]+( [a-zA-Z0-9]+)?) "
                         "(?P<category>[a-zA-Z0-9\-]+)\/(?P<pkg>[a-zA-Z0-9\.]+)"
                     )
                     res = re.search(status_re, match.group())
@@ -100,6 +103,7 @@ class Py3status:
                         ret['total'] = res.group('total')
                         ret['category'] = res.group('category')
                         ret['pkg'] = res.group('pkg')
+                        ret['action'] = res.group('action')
                     else:
                         raise Exception('Regex did not match a line')
                     break
@@ -123,6 +127,7 @@ class Py3status:
                 self.ret['total'] = "0"
                 self.ret['category'] = ""
                 self.ret['pkg'] = ""
+                self.ret['action'] = "None"
                 response['full_text'] = self.py3.safe_format(
                     self.format, self.ret)
 
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -29,6 +29,8 @@ Format examples:
 
 import re
 
+STRING_NOT_INSTALLED = "emerge not installed"
+
 
 class Py3status:
     """
@@ -48,6 +50,10 @@ class Py3status:
             return True
         except:
             return False
+
+    def post_config_hook(self):
+        if not self.py3.check_commands('emerge'):
+            raise Exception(STRING_NOT_INSTALLED)
 
     def _getProgress(self):
         """
--- a/py3status/modules/emerge_status.py
+++ b/py3status/modules/emerge_status.py
@@ -5,8 +5,16 @@ emerge status if there is an emerge process currently running
 
 Configuration parameters:
     cache_timeout: how often we refresh this module in second (default 0)
-    format: Display format to user (default 'EMRG: {current} / {total}')
-    hide_if_zero: Don't show in bar if there is no emerge running (default False)
+    format: *(default '[[\\?if=!hide_if_stopped {prefix}]|'
+                     '[\\?if=is_running {prefix}]]'
+                     '[[\\?if=is_running [\\?if=!total=0 {current}/{total}'
+                     '[\\?if=show_action {action} ]'
+                     '[\\?if=show_pkg {category}/{pkg}]]]|'
+                     '[\\?if=!hide_if_stopped 0/0]]')*
+    hide_if_stopped: Hide all information if no emerge is running (default False)
+    prefix: prefix in statusbar (default "EMRG: ")
+    show_action: En/Disable showing action (default True)
+    show_pkg: Dis-/Enable showing category and pkg name (default True)
 
 Format placeholders:
     {current} Number of package that is currently emerged
@@ -15,26 +23,52 @@ Format placeholders:
     {pkg} Name of the currently emerged packaged
     {action} Current emerge action
 
-Format examples:
-     {current} / {total}'
-    Result: "3 / 14"
-    '{current} / {total} = {pkg}'
-    Result: "3 / 14 = py3status"
-    '{category} - {pkg}'
-    Result: "x11-misc - py3status"
+Examples:
+```
+emerge_status {
+    hide_if_stopped = True
+    prefix = emerge
+    show_action = False
+    show_pkg = False
+}
+```
+
 
 @author AnwariasEu
 """
 
 import re
+import copy
+
+STRING_NOT_INSTALLED = 'not installed'
 
 
 class Py3status:
     """
     """
     cache_timeout = 0
-    format = u'EMRG: {current} / {total}'
-    hide_if_zero = False
+    format = (
+        # show prefix if either show_stopped or if emerge is running
+        '[[\?if=!hide_if_stopped {prefix}]|[\?if=is_running {prefix}]]'
+        # if emerge is running and total != 0 show
+        '[[\?if=is_running [\?if=!total=0 {current}/{total}'
+        # show action and cat/pkg only if user wants it
+        '[\?if=show_action {action} ][\?if=show_pkg {category}/{pkg}]]]|'
+        # if no emerge is running but show_stopped is set show 0/0
+        '[\?if=!hide_if_stopped 0/0]]')
+    hide_if_stopped = False
+    prefix = "EMRG: "
+    show_action = True
+    show_pkg = True
+
+    def __init__(self):
+        self.ret_default = {
+            'current': 0,
+            'total': 0,
+            'category': "",
+            'pkg': "",
+            'action': "",
+            'is_running': False}
 
     def _emergeRunning(self):
         """
@@ -44,7 +78,7 @@ class Py3status:
         try:
             self.py3.command_output(['pgrep', 'emerge'])
             return True
-        except:
+        except Exception:
             return False
 
     def post_config_hook(self):
@@ -60,11 +94,6 @@ class Py3status:
         emerge_log_file = '/var/log/emerge.log'
 
         ret = {}
-        ret['current'] = None
-        ret['total'] = None
-        ret['category'] = None
-        ret['pkg'] = None
-        ret['action'] = None
 
         input_data = []
 
@@ -85,24 +114,22 @@ class Py3status:
 
         for line in input_data:
             if "*** terminating." in line:
-                ret['current'] = "0"
-                ret['total'] = "0"
+                # Copy content of ret_default no only the references
+                ret = copy.deepcopt(self.ret_default)
                 break
             else:
-                match = re.search(">>> emerge.*", line)
-                if match:
-                    status_re = (
-                        "\((?P<current>[0-9]+) of (?P<total>[0-9]+)\) "
-                        "\((?P<action>[a-zA-Z0-9\/]+( [a-zA-Z0-9]+)?) "
-                        "(?P<category>[a-zA-Z0-9\-]+)\/(?P<pkg>[a-zA-Z0-9\.]+)"
-                    )
-                    res = re.search(status_re, match.group())
-                    if res:
-                        ret['current'] = res.group('current')
-                        ret['total'] = res.group('total')
-                        ret['category'] = res.group('category')
-                        ret['pkg'] = res.group('pkg')
-                        ret['action'] = res.group('action')
+                status_re = re.compile(
+                    "\((?P<cu>[\d]+) of (?P<t>[\d]+)\) "
+                    "(?P<a>[a-zA-Z\/]+( [a-zA-Z]+)?) "
+                    "\((?P<ca>[\w\-]+)\/(?P<p>[\w\.]+)"
+                )
+                res = status_re.search(line)
+                if res is not None:
+                    ret['current'] = res.group('cu')
+                    ret['total'] = res.group('t')
+                    ret['category'] = res.group('ca')
+                    ret['pkg'] = res.group('p')
+                    ret['action'] = res.group('a')
                     break
         return ret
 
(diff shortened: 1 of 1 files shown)


py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,144 @@
...
+
+        """
+        Try to open file, if unable throw error to be catched later
+        """
+        try:
+            with open(self.emerge_log_file, 'r') as fp:
+                for line in fp:
+                    input_data.append(line)
+        except IOError as err:
+            ret['err'] = True
```
lasers: This could be better off with `self.py3.error()` instead. I don't know how often this can happen so if it's not something very often, then we could skip `Try/Except` instead. The exception would be shown on the bar.

We also could check for binary `emerge` in `post_config_hook` so this can't happen on other distributions.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,144 @@
...
+
+        returns a dict containing current and total value.
+        """
+        ret = {}
+        ret['current'] = None
+        ret['total'] = None
+        ret['err'] = False
+        ret['err_text'] = ""
+        ret['category'] = None
+        ret['pkg'] = None
```
lasers: See if we can simplify things enough to start with just `ret = {}`.

py3status/modules/emerge_status.py
AnwariasEu: This can happen if the file does not exist e.g. the module is used on system where there is no portage and therefore no emerge.log.
But it can also happen the user has no read access to the file, which happens in case the user is not part of the 'portage' group.

Error handling can definitely be improved.

AnwariasEu:
First of all thanks for the feedback.
I'm pretty new to contributing to other work. 
I made some fixes in more commits, hopefully i will submit them correctly.

> * (Imo) I think it It would be nice to add prefix in the format too like DPMS, XSCR, UPD, etc, eg EMG {current}/{total} or whatever you think is better.

You mean in den default format?
I chose no prefix by personal preference, but there can definitely a prefix in the default.

> * Can we get some example emerge.log too?

Short except, want me to pastebin more?
```1484779948: Started emerge on: Jan 18, 2017 22:52:27
1484779948:  *** emerge --quiet --ask --tree --verbose sys-kernel/gentoo-sources
1484779959:  >>> emerge (1 of 2) sys-devel/bc-1.06.95-r1 to /
1484779962:  === (1 of 2) Cleaning (sys-devel/bc-1.06.95-r1::/usr/portage/sys-devel/bc/bc-1.06.95-r1.ebuild)
1484779962:  === (1 of 2) Compiling/Merging (sys-devel/bc-1.06.95-r1::/usr/portage/sys-devel/bc/bc-1.06.95-r1.ebuild)
1484779973:  === (1 of 2) Merging (sys-devel/bc-1.06.95-r1::/usr/portage/sys-devel/bc/bc-1.06.95-r1.ebuild)
1484779974:  >>> AUTOCLEAN: sys-devel/bc:0
1484779975:  === (1 of 2) Post-Build Cleaning (sys-devel/bc-1.06.95-r1::/usr/portage/sys-devel/bc/bc-1.06.95-r1.ebuild)
1484779975:  ::: completed emerge (1 of 2) sys-devel/bc-1.06.95-r1 to /
1484779975:  >>> emerge (2 of 2) sys-kernel/gentoo-sources-4.4.39 to /
1484781033:  *** terminating.
1484818385: Started emerge on: Jan 19, 2017 09:33:04
1484818385:  *** emerge --ask sys-kernel/gentoo-sources
1484818412:  >>> emerge (1 of 1) sys-kernel/gentoo-sources-4.4.39 to /
1484818412:  === (1 of 1) Cleaning (sys-kernel/gentoo-sources-4.4.39::/usr/portage/sys-kernel/gentoo-sources/gentoo-sources-4.4.39.ebuild)
1484818413:  *** Finished. Cleaning up...
1484818413:  *** exiting unsuccessfully with status '1'.
1484818413:  *** terminating.
```

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,146 @@
...
+
+        input_data = []
+
+        """
+        Try to open file, if unable throw error to be catched later
+        """
+        try:
+            with open(self.emerge_log_file, 'r') as fp:
+                for line in fp:
+                    input_data.append(line)
```
lasers: Here's an idea. This might be better to use `lines = open(self.log_file).readlines()` and maybe do `[-1]` too for last line only and fix up if/else to work with the last line. Based on the logs, we could make a `status` out of it too... like this: `Cleaning 1/2`, `Emerging 1/2`, `Compiling 1/2`, `Merging 1/2`, `Post-Build Cleaning 1/2`.... if users want that over static `EMG 1/2`... It's a bit more fun. :-)

py3status/modules/emerge_status.py
lasers: We could make a `STRING_NOT_INSTALLED` (see `moc.py`) so this would only run on machines with installed `emerge` == portage (?). I wonder if people who uses `emerge` regularly may already add themselves to `portage` group? If not, we could make a small note about `portage` group. Then we might not have to worry about error handling anymore. And if there was one, users can read the exception on the bar instead of custom error string.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,144 @@
...
+
+import os
+import re
+
+
+class Py3status:
+    """
+    """
+    cache_timeout = 1
+    emerge_log_file = '/var/log/emerge.log'
```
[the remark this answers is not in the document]
AnwariasEu: The location does not change, i mostly used this option for debugging.
Addressed in 8057c2b.

py3status/modules/emerge_status.py
AnwariasEu: emerge is the main binary of the portage package manager.
Checking for emerge binary implemented in 42e811f.
Error handling addressed in  e9d9ab8.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,146 @@
...
+
+        """
+        Traverse emerge.log from bottom up to get latest information
+        """
+
+        input_data.reverse()
+
+        for line in input_data:
+            terminated = re.search("\*\*\* terminating\.", line)
+            if terminated:
```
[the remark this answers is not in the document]
AnwariasEu: Addressed in 53afb9b.

py3status/modules/emerge_status.py
AnwariasEu: Readlines implemented in 5fc185a.
Status implemented as action in 1d78b05.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,144 @@
...
+            self.ret = self._getProgress()
+            if self.ret['err']:
+                response['full_text'] = self.py3.safe_format(
+                    u"emerge_status error: {err_text}", self.ret['err_text'])
+            else:
+                response['full_text'] = self.py3.safe_format(
+                    self.format, self.ret)
+        else:
+            if self.hide_if_zero:
+                response['full_text'] = ""
```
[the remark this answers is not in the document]
AnwariasEu: I personally think this a rather confusing formatting style and more likely to discourage users from using a custom format. Even more since the function is not, yet, documented.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,144 @@
...
+    def _emergeRunning(self):
+        """
+        Use pgrep to check if emerge is running.
+
+        Returns true if at least one instance of emerge is running.
+        """
+        tmp = os.popen("pgrep emerge").read()
+        pcount = tmp.count("\n")
+        if pcount > 0:
+            return True
```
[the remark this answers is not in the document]
AnwariasEu: Addressed in 590ed83. And fixed in 91d9e07.

py3status/modules/emerge_status.py
AnwariasEu: I guess this change only makes sense if the undocumented modifiers like `\?if`  are used?

py3status/modules/emerge_status.py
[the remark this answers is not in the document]
lasers: I think @tobes may have a branch to deal with this stuff. What is more is that the formatters can do more than just hiding zeroes... IIRC it can do `\?if=max_length 1`, `\?if=min_length 1`, `\?not_zero`, `\?if=placeholder`, `{placeholder:d}` (decimal), `{placeholder:.0f}` (rounding a float), and can grow more later. See `./tests/test_formatter.py` for more possibilities. 

Maybe we will get it documented for `3.8` or `3.9`, but will be easier to fix this module in the `format` now rather than to try and fix it it later / deprecate.

Also, you can add `Examples: ` with `# hide if zero` comment too.

py3status/modules/emerge_status.py
lasers: This is to kill lines. The format can use `[{example}]` to make non-existent  placeholders and `None` placeholders go away. It looks like some values would be filled one way or other in the loop.

Or you could leave `ret = {}` and move the others down to `*** terminating line`. Use `''` instead of `None` (to make it look nice).

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,146 @@
...
+        return ret
+
+    def emerge_status(self):
+        """
+        Emerge Status main routine
+        """
+        response = {}
+        response['cached_until'] = self.py3.time_in(self.cache_timeout)
+        if self._emergeRunning():
+            self.ret = self._getProgress()
```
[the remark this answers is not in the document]
lasers: With `cache_timeout = 0`, we can just read the log instead.
* Skips running `i3-msg` every second.
* Skips running `pgrep` every second.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        try:
+            self.py3.command_output(['pgrep', 'emerge'])
+            return True
+        except:
+            return False
+
+    def post_config_hook(self):
+        if not self.py3.check_commands('emerge'):
+            raise Exception("emerge not installed")
```
lasers: Look at `moc.py`. Make constant `STRING_NOT_INSTALLED` with string `not_installed`. The prefix name will be `emerge_status` which is a name of this module unless you want to change it to something else like... `emerge`. :-)

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,139 @@
...
+Format placeholders:
+    {current} Number of package that is currently emerged
+    {total} Total number of packages that will be emerged
+    {category} Category of the currently emerged package
+    {pkg} Name of the currently emerged packaged
+    {action} Current emerge action
+
+Format examples:
+     {current} / {total}'
+    Result: "3 / 14"
```
lasers: Wrap this part in `Examples: """ (docstrings) """`. See other modules for some examples. (ha)

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Traverse emerge.log from bottom up to get latest information
+        """
+
+        input_data.reverse()
+
+        for line in input_data:
+            if "*** terminating." in line:
+                ret['current'] = "0"
+                ret['total'] = "0"
```
lasers: You can use real `0` and `None` everywhere instead of `"0"` and `"None"`.

py3status/modules/emerge_status.py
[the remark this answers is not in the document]
lasers: Agreed. Check the log's last line instead of running `sh`+`i3-msg`+`pgrep` all times. More nice on your system resources.

AnwariasEu:
I have not been using a Gentoo setup lately.
I have fixed the string formatting (using the new modfiers) and will test it for a while/distribute it for testing.
Will probably update the branch in a week.

lasers:
_/me gives a friendly nudge_

AnwariasEu:
`Agreed. Check the log's last line instead of running sh+i3-msg+pgrep all times. More nice on your system resources.`  
Sadly this is not possible. We need to check by process id to see if emerge is running since the logfile is not deterministic if the user stops a running emerge process e.g. using ctrl+c. In that case the module would assume there is still a running emerge job (this behavior is even restart save).

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+                    break
+        return ret
+
+    def emerge_status(self):
+        """
+        Emerge Status main routine
+        """
+        response = {}
+        response['cached_until'] = self.py3.time_in(self.cache_timeout)
+        response['is_running'] = self._emergeRunning()
```
lasers: Putting unrelated things in `response` is misleading. Use `is_running = self._emergeRunning()` instead.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+        ret = copy.deepcopy(self.ret_default)
```
lasers: Is this too much? Does `ret = self.ret_default` works fine?

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+        input_data = []
+
+        """
+        Try to open file, if unable throw error to be catched later
+        """
+        try:
+            with open(emerge_log_file, 'r') as fp:
+                input_data = fp.readlines()
+        except IOError as err:
+            raise Exception(err)
```
lasers: If you want to show error, you can remove `try/except` and py3status will catch the error. Otherwise, I think this wouldn't work correctly without the log, so you probably want to remove `try/except`. It's either that or keep going with nothing.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+    show_pkg = True
+
+    def __init__(self):
+        self.ret_default = {
+            'current': 0,
+            'total': 0,
+            'category': "",
+            'pkg': "",
+            'action': "",
+            'is_running': False}
```
lasers: I'm not sure if this makes sense, but at same time, it makes sense too. If emerge is not running, you're going to start passing in fake data. `total = 0` is not really `0` for next several days until we finally run `emerge`. Ideal behavior could be that we don't pass anything and the placeholders would display `{total}` instead... and we come up with something in the `format` to deal with that.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Get current progress of emerge.
+
+        returns a dict containing current and total value.
+        """
+        emerge_log_file = '/var/log/emerge.log'
+
+        ret = {}
+
+        input_data = []
```
lasers: You might not need this as you already have one below.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+            return True
+        except Exception:
+            return False
+
+    def post_config_hook(self):
+        if not self.py3.check_commands('emerge'):
+            raise Exception("emerge not installed")
+
+    def _getProgress(self):
+        """
```
lasers: Python naming conventions for functions [[pep link]](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), ie `_get_progress`. :+1:

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+        Traverse emerge.log from bottom up to get latest information
+        """
+
+        input_data.reverse()
+
+        for line in input_data:
+            if "*** terminating." in line:
+                # Copy content of ret_default no only the references
+                ret = copy.deepcopt(self.ret_default)
```
lasers: Typo.

py3status/modules/emerge_status.py
AnwariasEu: Since ret_default is a dictionary filled with references the assignment `ret = self.ret_default` will copy all references and not their actual values. As a result when setting e.g. ret['current'] we override the default value as well. I wanted to use ** here but this has only been introduced in python 3.5 and travis for 3.4 did not like it.

AnwariasEu:
> Here's what does not make sense.
>   hide_if_stopped does nothing.
>   prefix does nothing.
>    show_action does nothing.
>    show_pkg does nothing.

> If users want or don't want something, they can tweak everything in the format instead of limited customization boolean configs. We probably need to work on your format too.

I personally don't think format strings of that size are user friendly and easy to edit.
I want to offer the user a simple set of variables he can tweak to change the output of the module.

lasers:
Your current `format` is maybe the largest default `format` in py3status right now. It also looks unfriendly because of your set of variables. You're putting everything in the `format` then have users toggle on/off a set of variables instead of having users change things in the `format` (like many other modules).

We absolutely want users to change things in the `format`. Otherwise, we can keep getting pull requests that this user wanted to hide this... This user want to show this only if... This users want to start changing colors only if this is running... or to hide everything if a certain boolean is True.

We also want to offer users set of `format` examples instead where they can copy and paste what they like into their config... then if necessary, make small changes from there. Some of my examples can be found in `thunderbird_todos`, `coin_market`, `mail`, `keyboard_locks`, `loadavg`, `lm_sensors`, etc.

When it comes to customization, putting everything (if possible) in the `format` is easily more powerful than a set of static configs. Your set of static configs would become useless in an instant the users added the `format` to their config. For this to work, you require users to keep all those set of static configs in the `format` too... We have other thing for that instead... Placeholders.

If users can add colors, then they can add `{action}` too instead of toggling `show_action`.

(offtopic below) We already can do this with `percent = 1234.56789`.
* `{percent}` --> `1234.56789`
* `{percent:.3f}` --> `1234.567`
* `{percent:.2f}` --> `1234.57`
* `{percent:.0f}` --> `1235`
* `{percent:d}` --> `1234` <-- Maybe or `1235`.


There are other things we could add in the future too... such as allowing users to print both weather temperatures ie `format = '{weather_temp}°C and {weather_temp:c2f}°F'` assuming it's already in Celsius and `c2f` would convert Celsius to Fahrenheit. This is far more flexible than a boolean `Fahrenheit` config which would only can print one weather temperature in `format`.

py3status/modules/emerge_status.py
AnwariasEu: Thanks for the explanation! Totally makes sense to keep things consistent.

py3status/modules/emerge_status.py
AnwariasEu: Maybe i can leave out definition of the `input_data` but `ret` needs to be be initialized somewhere as a dict. If i do so at beginning of the function or inside the if (right before the assignments of the values e.g. `ret['current']`) does not change that much.

py3status/modules/emerge_status.py
```diff
@@ -0,0 +1,160 @@
...
+        # if no emerge is running but show_stopped is set show 0/0
+        '[\?if=!hide_if_stopped 0/0]]')
+    hide_if_stopped = False
+    prefix = "EMRG: "
+    show_action = True
+    show_pkg = True
+
+    def __init__(self):
+        self.ret_default = {
+            'current': 0,
```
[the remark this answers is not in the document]
AnwariasEu: Why put the initialization there?
Are there performance implications that favor using the `post_config_hook` instead of the builtin `__init__` ?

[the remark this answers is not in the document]
lasers: For more information, see `Module init() method #399` PR.

https://github.com/ultrabug/py3status/blob/d395f2e57b95b1307eeed1c44d0bbc50825e7c3b/py3status/module.py#L130-L137

Tested just now... If we try to use `self.py3` helpers inside `__init__()`, we get this.
```bash
$ python static_string.py
  File "static_string.py", line 39, in <module>
    module_test(Py3status)
  File "py3status/py3status/module_test.py", line 67, in module_test
    module = module_class()
  File "static_string.py", line 22, in __init__
    self.py3.log('__init__: {}'.format(self.format))
AttributeError: 'Py3status' object has no attribute 'py3'
```
It's preferred too. All modules already had been updated to use `post_config_hook`.

AnwariasEu:
- Fixed the error message.
- Simplified the format string. I thinks its still not easy to read/write, but it matches the new formatting.
Perhaps it can be further simplified.
- Removed try/execpt when trying to open the file.
- Cache timeout is still missing.

ultrabug:
Thanks a lot @AnwariasEu 

I'm merging this now since you've been working on this for a while. I'm allowing myself to change some logic and QA on the module to be more gentle with disk I/O 

I've tested it, it's great!

PR #1292, opened 2018-02-27
add new module: gitlab_todo

sebPomme:
Hello,

I came with a new module, gitlab_todo
It looks like that
okay
I will glad to know what do you think about it

Pomme

### DIFF
--- a/py3status/modules/README.md
+++ b/py3status/modules/README.md
@@ -48,6 +48,8 @@
 
 **[github](#github)** — Display Github notifications and issue/pull requests for a repo.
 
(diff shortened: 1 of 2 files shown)


sebPomme:
I will fix the code to match with the "builder"

ultrabug:
@sebPomme thanks for your contribution

As you can see in #1214 there was a pending gitlab module contribution already. Could you test it and check if it works for your use case please?

Thanks

ultrabug:
@sebPomme I merged the more complete `gitlab` module as I trust it handled todos as well. Thanks again for your contribution!

PR #1309, opened 2018-03-12
Add wwan network manager module

cyrinux:
Hi !
I would like to used wwan_status but it was not working for me.
I this lot of laptop installation use networkmanger and modemmanager for wwan now.
This module based on wwan_status used NetworkManager, ModemManager and dbus services for get status.
This module doesn't report the ip address because i think it is useless. I can re-add it if you think this can be usefull, say me.
Also, i use this api doc https://developer.gnome.org/NetworkManager/stable/nm-dbus-types.html#NMDeviceModemCapabilities and I'm not sure if I find by the correct way the "netgen".
 
This is my first pull request for this great project, I hope my module will help.

For me the module do the job for the moment,
I listen you, I will be happy if the module is adopted, I can improve it, say me.
ps: I'm python newbie.


cyrinux:
i clean it

PR #1328, opened 2018-04-02
Reintroduce PR #1077

maximbaz:
Because #1077 [is not in master and not in 3.8](https://github.com/ultrabug/py3status/pull/1077#issuecomment-377995471) 😞 

/cc @ultrabug


ultrabug:
Dunno what happened, I must have done a mistake somewhere along the road I'm sorry

maximbaz:
No worries, I'm using 3.8 with this patch on top and it works, glad that this will be part of 3.9 and I won't have to manually patch anything again 😉

PR #1330, opened 2018-04-03
add py3-cmd to Control section of README

kmoe:
Update README with a reference to the new `py3-cmd` util so users are aware they don't have to send `USR1` for bar updates.


ultrabug:
Thanks a lot @kmoe ! I love documentation contributions :+1:

PR #1343, opened 2018-05-06
bitcoin_price: Changed request url from http to https

zack-baker:
Updated the request url to use HTTPS as opposed to just HTTP, since the api supports HTTPS requests.


ultrabug:
Sensible, thanks @zack-baker !

PR #1353, opened 2018-05-26
Module simplify

tobes:
This PR clears out some legacy cruft.

* remove support for multiple methods for a module - this add complexity and is no longer used by any modules for some time.

* remove legacy calling for modules

This is a first push on a fresh round of code cleaning.  It removes complexity from the modules which is no longer needed.  As part of this cleanup I have added a `while True/break` to reduce the diff for this PR.  If accepted then I will make another PR just to remove that.

There is still much room for further cleanups\simplification here but this feels like a good first step

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -521,7 +521,7 @@ class Py3statusWrapper:
             try:
                 my_m = Module(module, user_modules, self)
                 # only handle modules with available methods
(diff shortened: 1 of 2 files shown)


ultrabug:
Oh yes, this is due to die for so long, thank you for this!

PR #1356, opened 2018-05-30
update documentation to include the pacman-contrib requirement.

markrileybot:
Seems that the command line utility 'checkupdates' was moved from the pacman package to pacman-contrib.  I've updated the documentation to reflect the new dependency.


ultrabug:
Doc contribs are my favourites, thanks a lot @markrileybot !

PR #1357, opened 2018-05-30
Revert "Module simplify"

ultrabug:
Reverts ultrabug/py3status#1353

This breaks click events


tobes:
hmm what was the bug with the click events?  I've just tried to have a look but can't cause the issue.  Do you know how you triggered it?  Was it a click in a group or something like that?

PR #1396, opened 2018-07-18
air_quality: display nothing instead of error on api error

cyrinux:
There is actually a 502 bad gateway on the api.
A quick fix to display nothing instead of a big error when api error.


cyrinux:
Thanks @lasers, hope the api will be back, yet die, strange

ultrabug:
Thanks!

PR #1435, opened 2018-08-19
New module: speedtest

cyrinux:
Hi family,

Here a new module for do speedtest.
It needs speedtest-cli to be working.

image

Some point:
* This module if not refreshed at py3status, but only next on click / py3-cmd refresh.
I think I will personally refresh it with a networkmanager dispatcher hook.

* I have first done a "hack" which is the `{quality}` placeholder. Done this because there is yet a bug for int/float comparaison with the formatter. Even when this will be fix i think we can keep it, but, what is a "good" or "bad" connection? Must I add params for report quality based or user preference?

* To be quick, i do upload or download only if placeholders are present. Need at least `{download}` or `{upload}`

Waiting for review :)

(I will have lot of travis fix to do first I think)

### DIFF
--- a/py3status/modules/speedtest.py
+++ b/py3status/modules/speedtest.py
@@ -6,7 +6,7 @@ Configuration parameters:
     button_refresh: button to run the check (default 2)
     button_share: button to open the result link (default None)
     format: display format for this module, {download} and/or {upload} required
-        (default '● [{ping} ms] [↑ {upload} {upload_unit}] [↓ {download} {download_unit}]')
+        (default 'Speedtest [Up {upload} {upload_unit}] [Down {download} {download_unit}]')
     server_id: speedtest server to use, `speedtest-cli --list` to get id (default None)
     si_units: use SI units (default False)
     sleep_timeout: when speedtest-cli is allready running, this interval will be used
@@ -25,42 +25,42 @@ Configuration parameters:
     unit_size: unit for bytes_received/bytes_sent (default 'MB')
 
 Format placeholders:
-    {bytes_sent} bytes sent during test, human readable
-    {bytes_sent_unit} unit used for bytes_sent, eg 'MB'
-    {bytes_sent_raw} bytes sent during test, for format comparation
-    {bytes_received} bytes received during test, human readable
-    {bytes_received_raw} bytes received during test, for format comparation
-    {bytes_received_unit} unit used for bytes_received, eg 'MB'
-    {client_country} client country code, eg 'FR'
-    {client_ip} client ip, eg '78.194.13.7'
-    {client_isp} client isp, eg 'Free SAS'
-    {client_ispdlavg} client isp download average, eg '0'
-    {client_isprating} client isp rating, eg 3.7
-    {client_ispulavg} client isp upload average, eg '0'
-    {client_lat} client latitude, eg '48.8534'
-    {client_loggedin} client logged in, eg '0'
-    {client_lon} client longitude, eg '2.3487999999999998'
-    {client_rating} client rating, eg '0'
-    {download} download rate from speedtest server, human readable
-    {download_raw} download rate from speedtest server, for format comparation
-    {download_unit} unit used as , eg 'MB/s'
-    {ping} ping time in ms to speedtest server
-    {quality} quality of the connection, eg ok, bad, faster, slower
-    {timestamp} timestamp of the run
-    {server_cc} server country code, eg 'FR'
-    {server_country} server country, eg 'France'
-    {server_d} server distance, eg '2.316599376968091'
-    {server_host} server host, eg 'speedtest.telecom-paristech.fr:8080'
-    {server_id} server id, eg '11977'
-    {server_lat} server latitude, eg '48.8742'
-    {server_latency} server latency, eg 8.265
-    {server_lon} server longitude, eg 2.3470
-    {server_name} server name, eg 'Paris'
-    {server_sponsor} server sponsor, eg 'Télécom ParisTech'
-    {server_url} server url, eg 'http://speedtest.telecom-paristech.fr/upload.php'
-    {upload} upload rate to speedtest server, human readable
-    {upload_raw} upload rate to speedtest server, for format comparation
-    {upload_unit} unit used for upload, eg 'MB/s'
+    {bytes_sent}            bytes sent during test, human readable
+    {bytes_sent_unit}       unit used for bytes_sent, eg 'MB'
+    {bytes_sent_raw}        bytes sent during test, for format comparation
+    {bytes_received}        bytes received during test, human readable
+    {bytes_received_raw}    bytes received during test, for format comparation, eg
+    {bytes_received_unit}   unit used for bytes_received, eg 'MB'
+    {client_country}        client country code, eg 'FR'
+    {client_ip}             client ip, eg '78.194.13.7'
+    {client_isp}            client isp, eg 'Free SAS'
+    {client_ispdlavg}       client isp download average, eg '0'
+    {client_isprating}      client isp rating, eg 3.7
+    {client_ispulavg}       client isp upload average, eg '0'
+    {client_lat}            client latitude, eg '48.8534'
+    {client_loggedin}       client logged in, eg '0'
+    {client_lon}            client longitude, eg '2.3487999999999998'
+    {client_rating}         client rating, eg '0'
+    {download}              download rate from speedtest server, human readable
+    {download_raw}          download rate from speedtest server, for format comparation
+    {download_unit}         unit used as , eg 'MB/s'
+    {ping}                  ping time in ms to speedtest server
+    {quality}               quality of the connection, eg ok, bad, faster, slower
+    {timestamp}             timestamp of the run, eg 'TODO'
+    {server_cc}             server country code, eg 'FR'
+    {server_country}        server country, eg 'France'
+    {server_d}              server distance, eg '2.316599376968091'
+    {server_host}           server host, eg 'speedtest.telecom-paristech.fr:8080'
+    {server_id}             server id, eg '11977'
+    {server_lat}            server latitude, eg '48.8742'
+    {server_latency}        server latency, eg 8.265
+    {server_lon}            server longitude, eg 2.3470
+    {server_name}           server name, eg 'Paris'
+    {server_sponsor}        server sponsor, eg 'Télécom ParisTech'
+    {server_url}            server url, eg 'http://speedtest.telecom-paristech.fr/upload.php'
+    {upload}                upload rate to speedtest server, human readable
+    {upload_raw}            upload rate to speedtest server, for format comparation
+    {upload_unit}           unit used for upload, eg 'MB/s'
 
 Color thresholds:
     format:
@@ -73,6 +73,8 @@ This can be usefull to compare two runs.
 
 Requires:
     speedtest-cli: Command line interface for testing internet bandwidth using speedtest.net
+
+@author Cyril Levis (@cyrinux)
 
 Examples:
 ```
@@ -90,13 +92,14 @@ speedtest {
         ],
         "upload": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
     }
-    format = '[\?color=ping {ping} ms] [\?color=upload ↑ {upload} {upload_unit}]'
-    format += ' [\?color=download ↓ {download} {download_unit}]'
+    format = '[\?color=ping {ping} ms] [\?color=upload Up {upload} {upload_unit}]'
+    format += ' [\?color=download Down {download} {download_unit}]'
 }
 
 # colored based on comparation between two last runs
 speedtest {
-    format = ' [[\?if=download_raw<previous_download_raw&color=degraded ↓]|[\?color=ok ↑]]'
+    format = '[[\?if=download_raw<previous_download_raw&color=degraded Faster]'
+    format += '|[\?color=ok Slower]]'
 }
 
 # compare only download
@@ -105,14 +108,12 @@ speedtest {
     format += '[\?if=quality=faster&color=download_raw ↑ {download} {download_unit}]'
     format += '[\?if=quality=ok&color=download_raw -> {download} {download_unit}]'
     format += '[\?if=quality=bad&color=download_raw x {download} {download_unit}]]'
-    format += '|'
+    format += '|speedtest'
 }
 ```
 
-@author Cyril Levis (@cyrinux)
-
 SAMPLE OUTPUT
-{'full_text': u'\u25cf 208.18 ms \u2191 0.247 MB/s \u2193 0.453 MB/s'}
+{'full_text': u'Speedtest Down 0.247 MB/s Up 0.453 MB/s'}
 """
 
 from json import loads
@@ -130,7 +131,7 @@ class Py3status:
     button_refresh = 2
     button_share = None
     format = (
-        u"\u25cf [{ping} ms] [↑ {upload} {upload_unit}] [↓ {download} {download_unit}]"
+        u"Speedtest [Up {upload} {upload_unit}] [Down {download} {download_unit}]"
     )
     server_id = None
     si_units = False
@@ -159,33 +160,29 @@ class Py3status:
 
         self.placeholders = self.py3.get_placeholders_list(self.format)
 
+        self.command = ["speedtest-cli --json --secure"]
+
         # if download* or upload* missing, run complete test
-        no_upload = ""
-        no_download = ""
         if any(
             key.startswith(x)
             for x in ["download", "upload"]
             for key in self.placeholders
         ):
             if not any(key.startswith("upload") for key in self.placeholders):
-                no_upload = "--no-upload"
+                self.command += ["--no-upload"]
 
             if not any(key.startswith("download") for key in self.placeholders):
-                no_download = " --no-download"
+                self.command += ["--no-download"]
 
-        server = ""  # if not specified, nearest server is use
         if self.server_id and int(self.server_id):
-            server = "--server {}".format(self.server_id)
+            self.command += ["--server {}".format(self.server_id)]
 
-        share = ""
         if self.button_share and all(
             x in self.placeholders for x in ["download", "upload"]
         ):
-            share = "--share"
+            self.command += ["--share"]
 
-        self.command = "speedtest-cli --json --secure --timeout {} {} {} {} {}".format(
-            self.timeout, no_upload, no_download, server, share
-        )
+        self.command += ["--timeout {}".format(self.timeout)]
 
     def _is_running(self):
         try:
@@ -195,7 +192,8 @@ class Py3status:
             return False
 
     def _get_speedtest_data(self):
-        return loads(self.py3.command_output(self.command)) or None
+        command = ' '.join(self.command)
+        return loads(self.py3.command_output(command)) or None
 
     def speedtest(self):
         speedtest_data = {}
@@ -253,8 +251,7 @@ class Py3status:
             self.py3.storage_set("speedtest_data", current_data)
 
             # get speedtest result url
-            if current_data and "share" in current_data:
-                self.url = current_data["share"]
+            self.url = current_data.get("share")
 
             # create placeholders
             speedtest_data.update(current_data)
@@ -286,7 +283,7 @@ class Py3status:
         button = event["button"]
         if button == self.button_share and self.url:
             self.py3.command_run("xdg-open %s" % self.url)
-        elif button != self.button_refresh:
+        if button != self.button_refresh:
             self.py3.prevent_refresh()
 
 
--- a/py3status/modules/speedtest.py
+++ b/py3status/modules/speedtest.py
@@ -6,7 +6,9 @@ Configuration parameters:
     button_refresh: button to run the check (default 2)
     button_share: button to open the result link (default None)
     format: display format for this module, {download} and/or {upload} required
-        (default 'Speedtest [Up {upload} {upload_unit}] [Down {download} {download_unit}]')
+        *(default 'Speedtest [\?color=darkgray ping [\?color=ping {ping} ms] '
+                    'up [\?color=upload {upload} {upload_unit}] '
+                    'down [\?color=download {download} {download_unit}]]')*
     server_id: speedtest server to use, `speedtest-cli --list` to get id (default None)
     si_units: use SI units (default False)
     sleep_timeout: when speedtest-cli is allready running, this interval will be used
@@ -113,7 +115,15 @@ speedtest {
 ```
 
 SAMPLE OUTPUT
-{'full_text': u'Speedtest Down 0.247 MB/s Up 0.453 MB/s'}
+[
+    {"full_text": "Speedtest " },
+    {"full_text": "ping ", "color": "#a9a9a9"},
+    {"full_text": "9.498 ms", "color": "#ebdbb2"},
+    {"full_text": " up ", "color": "#a9a9a9"},
+    {"full_text": "78.42 MB/s", "color": "#fb4934"},
+    {"full_text": " down ", "color": "#a9a9a9"},
+    {"full_text": "39.58 MB/s", "color": "#fb4934"}
+]
 """
 
 from json import loads
@@ -131,7 +141,9 @@ class Py3status:
     button_refresh = 2
     button_share = None
     format = (
-        u"Speedtest [Up {upload} {upload_unit}] [Down {download} {download_unit}]"
+        u"Speedtest [\?color=darkgray ping [\?color=ping {ping} ms] "
+        "up [\?color=upload {upload} {upload_unit}] "
+        "down [\?color=download {download} {download_unit}]]"
     )
     server_id = None
     si_units = False
@@ -160,7 +172,7 @@ class Py3status:
         self.first_run = True
         self.placeholders = self.py3.get_placeholders_list(self.format)
         self.thresholds_init = self.py3.get_color_names_list(self.format)
-        self.command = ["speedtest-cli --json --secure"]
+        self.command = "speedtest-cli --json --secure"
 
         self.quality = {
             -1: 'failed',
@@ -178,20 +190,20 @@ class Py3status:
             for key in self.placeholders
         ):
             if not any(key.startswith("upload") for key in self.placeholders):
-                self.command += ["--no-upload"]
+                self.command += " --no-upload"
 
             if not any(key.startswith("download") for key in self.placeholders):
-                self.command += ["--no-download"]
+                self.command += " --no-download"
 
         if self.server_id and int(self.server_id):
-            self.command += ["--server {}".format(self.server_id)]
+            self.command += " --server " + str(self.server_id)
 
         if self.button_share and all(
             x in self.placeholders for x in ["download", "upload"]
         ):
-            self.command += ["--share"]
+            self.command += " --share"
 
-        self.command += ["--timeout {}".format(self.timeout)]
+        self.command += " --timeout " + str(self.timeout)
 
     def _is_running(self):
         try:
--- a/py3status/modules/speedtest.py
+++ b/py3status/modules/speedtest.py
@@ -3,7 +3,6 @@
 Do a bandwidth test with speedtest-cli.
 
 Configuration parameters:
-    button_refresh: button to run the check (default 2)
     button_share: button to open the result link (default None)
     format: display format for this module, {download} and/or {upload} required
         *(default 'Speedtest [\?color=darkgray ping [\?color=ping {ping} ms] '
@@ -11,14 +10,9 @@ Configuration parameters:
                     'down [\?color=download {download} {download_unit}]]')*
     server_id: speedtest server to use, `speedtest-cli --list` to get id (default None)
     si_units: use SI units (default False)
-    sleep_timeout: when speedtest-cli is allready running, this interval will be used
-        to allow faster retry refreshes (default 5)
     thresholds: specify color thresholds to use *(default {
                 'download': [(0, 'bad'), (1024, 'degraded'), (1024 * 1024, 'good')],
                 'ping': [(200, 'bad'), (150, 'orange'), (100, 'degraded'), (10, 'good')],
-                'quality': [(-1, "bad"), (0, "darkgrey"), (1, "degraded"),
-                    (2, "good"), (4, "degraded"), (3, "good"),
-                ],
                 'upload': [(0, 'bad'), (1024, 'degraded'), (1024 * 1024, 'good')]})*
     timeout: timeout when communicating with speedtest.net servers (default 10)
     unit_bitrate: unit for download/upload rate (default 'MB/s')
@@ -45,8 +39,6 @@ Format placeholders:
     {download_raw}          download rate from speedtest server, for format comparation
     {download_unit}         unit used as , eg 'MB/s'
     {ping}                  ping time in ms to speedtest server
-    {quality}               quality code of the connection, eg 0, 1, 2, 3, 4
-    {quality_name}          quality name of the connection, eg failed, bad, good, slower, faster
     {timestamp}             timestamp of the run, eg '2018-08-30T16:27:25.318212Z'
     {server_cc}             server country code, eg 'FR'
     {server_country}        server country, eg 'France'
@@ -64,7 +56,7 @@ Format placeholders:
     {upload_unit}           unit used for upload, eg 'MB/s'
 
 Color thresholds:
-    xxx: print a color based on the value of `xxx` placeholder
+    field_name: print a color based on the value of `field_name` placeholder
 
 The module will be triggered on clic only. Not at start.
 
@@ -83,14 +75,6 @@ speedtest {
     thresholds = {
         "download": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
         "ping": [(200, "bad"), (150, "orange"), (100, "degraded"), (10, "good")],
-        "quality": [
-            (-1, "bad"),
-            (0, "darkgrey"),
-            (1, "degraded"),
-            (2, "good"),
-            (4, "degraded"),
-            (3, "good"),
-        ],
         "upload": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
     }
     format = '[\?color=ping {ping} ms] [\?color=upload Up {upload} {upload_unit}]'
@@ -105,10 +89,9 @@ speedtest {
 
 # compare only download
 speedtest {
-    format = '[[\?if=quality_name=slower&color=download_raw ↓ {download} {download_unit}]'
-    format += '[\?if=quality_name=faster&color=download_raw ↑ {download} {download_unit}]'
-    format += '[\?if=quality_name=good&color=download_raw -> {download} {download_unit}]'
-    format += '[\?if=quality_name=bad&color=download_raw x {download} {download_unit}]]'
+    format = '[[\?if=download_raw<previous_download_raw ↓ {download} {download_unit}]'
+    format += '[\?if=download_raw>previous_download_raw ↑ {download} {download_unit}]'
+    format += '[\?if=download_raw=previous_download_raw -> {download} {download_unit}]'
     format += '|speedtest'
 }
 ```
@@ -137,7 +120,6 @@ class Py3status:
     """
 
     # available configuration parameters
-    button_refresh = 2
     button_share = None
     format = (
         u"Speedtest [\?color=darkgray ping [\?color=ping {ping} ms] "
@@ -146,18 +128,9 @@ class Py3status:
     )
     server_id = None
     si_units = False
-    sleep_timeout = 5
     thresholds = {
         "download": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
         "ping": [(200, "bad"), (150, "orange"), (100, "degraded"), (10, "good")],
-        "quality": [
-            (-1, "bad"),
-            (0, "darkgrey"),
-            (1, "degraded"),
-            (2, "good"),
-            (4, "degraded"),
-            (3, "good"),
-        ],
         "upload": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
     }
     timeout = 10
@@ -168,21 +141,14 @@ class Py3status:
         if not self.py3.check_commands("speedtest-cli"):
             raise Exception(STRING_NOT_INSTALLED)
 
+        self.button_refresh = 2
+        self.sleep_timeout = 2.5
         self.speedtest_data = {}
         self.cached_until = self.py3.CACHE_FOREVER
         self.placeholders = self.py3.get_placeholders_list(self.format)
         self.thresholds_init = self.py3.get_color_names_list(self.format)
         self.command = "speedtest-cli --json --secure"
         self.url = None
-
-        self.quality = {
-            -1: "failed",
-            0: "unknow",
-            1: "bad",
-            2: "good",
-            3: "slower",
-            4: "faster",
-        }
 
         # if download* or upload* missing, run complete test
         if any(
@@ -213,18 +179,18 @@ class Py3status:
         except self.py3.CommandError:
             return False
 
-    def _cast_number(self, data):
+    def _cast_number(self, value):
         try:
-            data = int(data)
+            value = int(value)
         except ValueError:
             try:
-                data = float(data)
+                value = float(value)
             except ValueError:
                 pass
-        return data
+        return value
 
     def _get_speedtest_data(self):
-        return loads(self.py3.command_output(self.command)) or None
+        return loads(self.py3.command_output(self.command))
 
     def _start(self):
         self.cached_until = self.py3.time_in(self.sleep_timeout)
@@ -241,20 +207,6 @@ class Py3status:
                     current_data.get("upload", 0)
                 )
 
-                # create "quality" #maybe bad name
-                if "total" in previous_data:
-                    if current_data["total"] >= previous_data["total"]:
-                        quality_key = 4
-                    else:
-                        quality_key = 3
-                else:
-                    if current_data["total"] <= 0:
-                        quality_key = 1
-                    else:
-                        quality_key = 2
-                current_data["quality"] = quality_key
-                current_data["quality_name"] = self.quality[quality_key]
-
         # zero-ing if not fetched, raw version and units convertion
         for x in ["download", "upload", "bytes_received", "bytes_sent"]:
             unit = self.unit_size if "bytes" in x else self.unit_bitrate
@@ -265,9 +217,9 @@ class Py3status:
             )
 
         # extra data, not sure we want to expose
-        current_data.update(self.py3.flatten_dict(current_data, delimiter='_'))
-        for x in ['client', 'server']:
-            del current_data[x]
+        current_data = self.py3.flatten_dict(
+            current_data, delimiter='_', intermediates=False
+        )
 
         # store last data fetched
         self.py3.storage_set("speedtest_data", current_data)
@@ -277,17 +229,17 @@ class Py3status:
 
         # create placeholders
         self.speedtest_data.update(current_data)
-        if previous_data:
-            self.speedtest_data.update(
-                {"previous_" + k: v for (k, v) in previous_data.items()}
-            )
+        self.speedtest_data.update(
+            {"previous_" + k: v for (k, v) in previous_data.items()}
+        )
 
         # cast number
         self.speedtest_data.update(
             {
-                k: self._cast_number(v)
-                for (k, v) in self.speedtest_data.items()
-                if isinstance(v, Number)
+                k: self._cast_number(self.speedtest_data[k])
+                for k in self.placeholders if isinstance(
+                    self.speedtest_data[k], Number
+                )
             }
         )
 
@@ -304,12 +256,12 @@ class Py3status:
 
     def on_click(self, event):
         button = event["button"]
-        if button == self.button_refresh:
-            self._start()
         if button == self.button_share and self.url:
             self.py3.command_run("xdg-open %s" % self.url)
         if button != self.button_refresh:
             self.py3.prevent_refresh()
+        if button == self.button_refresh:
+            self._start()
 
 
 if __name__ == "__main__":
--- a/py3status/modules/speedtest.py
+++ b/py3status/modules/speedtest.py
@@ -12,9 +12,9 @@ Configuration parameters:
     server_id: speedtest server to use, `speedtest-cli --list` to get id (default None)
     si_units: use SI units (default False)
     thresholds: specify color thresholds to use *(default {
-                'download': [(0, 'bad'), (1024, 'degraded'), (1024 * 1024, 'good')],
-                'ping': [(200, 'bad'), (150, 'orange'), (100, 'degraded'), (10, 'good')],
-                'upload': [(0, 'bad'), (1024, 'degraded'), (1024 * 1024, 'good')]})*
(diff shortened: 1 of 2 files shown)


lasers:
You should turn off `fatal: Cannot setup tracking information; starting point 'cyrinux/addSpeedtestCliModule' is not a branch.`. It prevent people from checking out your branch and often make a big mess putting files all over in current branch every time we try to pull in changes.

cyrinux:
@lasers I will try :/ It is a running joke of my git config.

cyrinux:
Erf ... :/

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,302 @@
+# -*- coding: utf-8 -*-
+"""
+Do a bandwidth test with speedtest-cli.
+
+Configuration parameters:
+    button_refresh: button to run the check (default 1)
```
lasers: All `button_refresh` should be `2`.... Consistency across all modules.

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,302 @@
...
+    button_share: button to open the result link (default None)
+    format: display format for this module, {download} and/or {upload} required
```
lasers: >{download} and/or {upload} required

Why? Bad code? Users should be able to skip this if they want to. :laughing:

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,294 @@
...
+    timeout = 10
+    unit_bitrate = "MB/s"
+    unit_size = "MB"
+
+    def post_config_hook(self):
+        if not self.py3.check_commands("speedtest-cli"):
+            raise Exception(STRING_NOT_INSTALLED)
+
+        self.first_run = True
+        self.placeholders = list(set(self.py3.get_placeholders_list(self.format)))
```
lasers: I think it already does this. You can test this in terminal.
```diff
diff --git a/py3status/modules/static_string.py b/py3status/modules/static_string.py
index dbcec8c6..c9d6f8bc 100644
--- a/py3status/modules/static_string.py
+++ b/py3status/modules/static_string.py
@@ -17,8 +17,13 @@ class Py3status:
     """
     # available configuration parameters
     format = 'Hello, world!'
+    format += '{a} {d}'
+    format += '{b} {c}'
+    format += '{c} {b}'
+    format += '{d} {a}'

     def static_string(self):
+        self.py3.log(self.py3.get_placeholders_list(self.format))
         return {
             'cached_until': self.py3.CACHE_FOREVER,
             'full_text': self.py3.safe_format(self.format),
```

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,294 @@
...
+        ):
+            share = '--share'
+
+        self.command = "speedtest-cli --json --secure --timeout {} {} {} {} {}".format(
+            self.timeout, dont_upload, dont_download, server, share
+        )
+
+        # fail if download or upload missing
+        if not any(x in ['download', 'upload'] for x in self.placeholders):
+            raise Exception('%s' % (MISSING_PLACEHOLDER))
```
lasers: Throw this exception first thing first so you don't have to do some unnecessary work.

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,294 @@
...
+            return True
+        except:
+            return False
+
+    def _get_speedtest_data(self):
+        data = {}
+        try:
+            data = loads(self.py3.command_output(self.command))
+        except self.py3.CommandError as ce:
+            raise Exception('%s') % ce.output
```
lasers: Seems like you maybe could return this without try/except.

py3status/modules/speedtest_cli.py
cyrinux: Maybe I missunderstand your remark. 
I have done a system which autoswitch speedtest-cli based on params. 
But yes, I can run basic speedtest-cli with download AND upload if no download or upload placeholders used.

py3status/modules/speedtest_cli.py
cyrinux: Ok

py3status/modules/speedtest_cli.py
cyrinux: Find this in another module, not test but i feel your are true.

py3status/modules/speedtest_cli.py
cyrinux: True

cyrinux:
@lasers could you help me with travis? Don't understand what is wrong with thresholds in docstring. Thx 😗

py3status/modules/speedtest_cli.py
cyrinux: Fixed

py3status/modules/speedtest_cli.py
cyrinux: I changed the behavior

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+        "quality": [
+            ("ok", "good"),
+            ("bad", "degraded"),
+            ("faster", "good"),
+            ("slower", "degraded"),
+            ("faster", "good"),
+        ],
+        "upload": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
+    }
+    timeout = 10
```
lasers: `--timeout TIMEOUT     HTTP timeout in seconds. Default 10`.

(Opinion) It's good that we try to make things possible. However, I think we should just kill this option.

cyrinux:
Thanks lasers will check all this.

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+
+                if current_data and len(current_data) > 1:
+                    # create a "total" for know if cnx is better or not
+                    # between two run
+                    current_data["total"] = int(current_data.get("download", 0)) + int(
+                        current_data.get("upload", 0)
+                    )
+
+                    # create "quality" #maybe bad name
+                    if "total" in previous_data:
```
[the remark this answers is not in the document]
cyrinux: Fixed

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+            server = "--server {}".format(self.server_id)
+
+        share = ""
+        if self.button_share and all(
+            x in self.placeholders for x in ["download", "upload"]
+        ):
+            share = "--share"
+
+        self.command = "speedtest-cli --json --secure --timeout {} {} {} {} {}".format(
+            self.timeout, dont_upload, dont_download, server, share
```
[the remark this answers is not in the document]
cyrinux: ok. For extra space i have recheck for me it is good like this.

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+            # store last data fetched
+            self.py3.storage_set("speedtest_data", current_data)
+
+            # get speedtest result url
+            if "share" in current_data:
+                self.url = current_data["share"]
+
+            # create placeholders
+            for x in current_data:
+                speedtest_data[x] = current_data[x]
```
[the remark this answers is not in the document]
cyrinux: seems ok

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+            for x in previous_data:
+                speedtest_data["previous_" + x] = previous_data[x]
```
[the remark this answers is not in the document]
cyrinux: done

py3status/modules/speedtest_cli.py
cyrinux: I have recently use case with poor connection I want to test and where setting timeout help me ...

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+        return {
+            "cached_until": cached_until,
+            "full_text": self.py3.safe_format(self.format, speedtest_data),
+        }
+
+    def on_click(self, event):
+        button = event["button"]
+        if button == self.button_share and self.url:
+            self.py3.command_run("xdg-open %s" % self.url)
+        elif button != self.button_refresh:
```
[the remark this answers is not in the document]
cyrinux: another refresh?

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
cyrinux: done

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+            for x in previous_data:
+                speedtest_data["previous_" + x] = previous_data[x]
+
+            # # cast
+            for x in speedtest_data:
+                try:
+                    value = float(speedtest_data[x])
+                except:
+                    value = speedtest_data[x]
+                speedtest_data[x] = value
```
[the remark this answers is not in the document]
cyrinux: fixed

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+                    for y in current_data[x]:
+                        current_data[x + "_" + y] = current_data[x][y]
+                    del current_data[x]
+
+            # store last data fetched
+            self.py3.storage_set("speedtest_data", current_data)
+
+            # get speedtest result url
+            if "share" in current_data:
+                self.url = current_data["share"]
```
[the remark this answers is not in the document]
cyrinux: which behavior do you want?
I clean the url now on each refresh.

py3status/modules/speedtest.py
```diff
@@ -191,7 +191,7 @@ def _is_running(self):
         try:
             self.py3.command_output(["pgrep", "speedtest-cli"])
             return True
-        except:
+        except self.py3.CommandError:
```
lasers: :+1: You can try `pre-push` hook now. See `.git-hooks/pre-push`.

py3status/modules/speedtest_cli.py
```diff
@@ -0,0 +1,301 @@
...
+                current_data[x + "_raw"] = current_data[x]
+                current_data[x], current_data[x + "_unit"] = self.py3.format_units(
+                    current_data[x], unit=self.unit_bitrate, si=self.si_units
+                )
+
+            for x in ["bytes_received", "bytes_sent"]:
+                current_data[x] = current_data.get(x, 0)
+                current_data[x + "_raw"] = current_data[x]
+                current_data[x], current_data[x + "_unit"] = self.py3.format_units(
+                    current_data[x], unit=self.unit_size, si=self.si_units
```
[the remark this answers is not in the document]
cyrinux: ok I understand now

py3status/modules/speedtest.py
cyrinux: Activated

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,299 @@
...
+        return {
+            "cached_until": cached_until,
+            "full_text": self.py3.safe_format(self.format, speedtest_data),
+        }
+
+    def on_click(self, event):
+        button = event["button"]
+        if button == self.button_share and self.url:
+            self.py3.command_run("xdg-open %s" % self.url)
+        elif button != self.button_refresh:
```
lasers: It will refresh after sharing. Change `elif` to `if`.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,299 @@
...
+        if self.first_run:
+            self.first_run = False
+            cached_until = self.py3.CACHE_FOREVER
+        else:
+            current_data = {}
+            previous_data = {}
+            cached_until = self.py3.time_in(self.sleep_timeout)
+
+            if not self._is_running():
+                previous_data = self.py3.storage_get("speedtest_data")
```
lasers: I assume you want `or` at end with something else to reduce additional checkings below.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,299 @@
...
+                    for y in current_data[x]:
+                        current_data[x + "_" + y] = current_data[x][y]
+                    del current_data[x]
+
+            # store last data fetched
+            self.py3.storage_set("speedtest_data", current_data)
+
+            # get speedtest result url
+            if current_data and "share" in current_data:
+                self.url = current_data["share"]
```
lasers: You can use `current_data.get()` instead.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,299 @@
...
+This can be usefull to compare two runs.
+
+Requires:
+    speedtest-cli: Command line interface for testing internet bandwidth using speedtest.net
+
+Examples:
+```
+# colored based on thresholds
+speedtest {
+    thresholds = {
```
lasers: This example is broken in the config. Possibly hidden at startup too.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,299 @@
...
+            self.py3.prevent_refresh()
+
+
+if __name__ == "__main__":
+    """
+    Run module in test mode.
+    """
+    from py3status.module_test import module_test
+
+    module_test(Py3status)
```
lasers: There is a Python API too. https://github.com/sivel/speedtest-cli/wiki
Whether it's better or not, I dunno. I just wanted to mention this.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    button_refresh = 2
+    button_share = None
+    format = (
+        u"Speedtest [Up {upload} {upload_unit}] [Down {download} {download_unit}]"
+    )
```
lasers: Your `format` prints extra spaces on right side on start. Test all your examples in real config, not inside `speedtest.py`. In thresholds... `1024 * 1024` throws an error.

Put colors in this format too except the words. We'll play with colors. Find what works well here. I'm aiming to have modules with a simple, clean theme so users can just add what they want and it would still look good on the bar.. Follow nocolor default, secondary darkgray, then colorized information.

For your module, I'm visualizing default `Speedtest` + lowercase darkgray `ping`, `up`, `down` prefixes with colorized stats... like `loadavg`s detailed format or `lm_sensors` default format.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+            cached_until = self.py3.CACHE_FOREVER
+        else:
+            current_data = {}
+            previous_data = {}
+            cached_until = self.py3.time_in(self.sleep_timeout)
+
+            if not self._is_running():
+                previous_data = self.py3.storage_get("speedtest_data")
+                current_data = self._get_speedtest_data()
+                cached_until = self.py3.CACHE_FOREVER
```
lasers: We don't need three `cached_until`. Move this to top.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+    server_id = None
+    si_units = False
+    sleep_timeout = 5
+    thresholds = {
+        "download": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
+        "ping": [(200, "bad"), (150, "orange"), (100, "degraded"), (10, "good")],
+        "quality": [
+            ("ok", "good"),
+            ("bad", "degraded"),
+            ("faster", "good"),
```
lasers: Not implemented, not supported. Looking at your code, you can use state code instead just like `wwan`. :smiley_cat:

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+
+        self.placeholders = self.py3.get_placeholders_list(self.format)
+
+        self.command = ["speedtest-cli --json --secure"]
+
+        # if download* or upload* missing, run complete test
+        if any(
+            key.startswith(x)
+            for x in ["download", "upload"]
+            for key in self.placeholders
```
lasers: We should remove this over-complicated if-statement. `post_config_hook` is run only once.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+                    k: float(v)
+                    for (k, v) in speedtest_data.items()
+                    if isinstance(v, Number)
+                }
+            )
+
+            # thresholds
+            for x in self.thresholds:
+                if x in speedtest_data:
+                    self.py3.threshold_get_color(speedtest_data[x], x)
```
lasers: This looks wrong. Look at `loadavg` to compare examples.

py3status/modules/speedtest.py
lasers: Do you need all three `ifs` or just two `ifs`?

py3status/modules/speedtest.py
cyrinux: seen

py3status/modules/speedtest.py
cyrinux: Idk, for the moment i see how get what i want with 3 ... I will reread...

py3status/modules/speedtest.py
cyrinux: good idea

py3status/modules/speedtest.py
cyrinux: ok will try

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+    {server_name}           server name, eg 'Paris'
+    {server_sponsor}        server sponsor, eg 'Télécom ParisTech'
+    {server_url}            server url, eg 'http://speedtest.telecom-paristech.fr/upload.php'
+    {upload}                upload rate to speedtest server, human readable
+    {upload_raw}            upload rate to speedtest server, for format comparation
+    {upload_unit}           unit used for upload, eg 'MB/s'
+
+Color thresholds:
+    format:
+        xxx: print a color based on the value of `xxx` placeholder
```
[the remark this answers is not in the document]
cyrinux: done

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,296 @@
...
+
+        if self.server_id and int(self.server_id):
+            self.command += ["--server {}".format(self.server_id)]
+
+        if self.button_share and all(
+            x in self.placeholders for x in ["download", "upload"]
+        ):
+            self.command += ["--share"]
+
+        self.command += ["--timeout {}".format(self.timeout)]
```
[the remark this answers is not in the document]
cyrinux: done

py3status/modules/speedtest.py
cyrinux: 3

cyrinux:
I'm fixing a wrong behaviour with refresh (which auto-refresh).

tobes:
> This module if not refreshed at py3status,

You probably want the main function to not really do anything and then have the speed test function only ever called via `on_click()` that will help make life much simpler and can potentially lead to a much cooler user experience

cyrinux:
Yes, true, this must be simple! Will try this. Thx @tobes

py3status/modules/speedtest.py
```diff
@@ -25,11 +25,11 @@
...
 
 Format placeholders:
-    {bytes_sent}            bytes sent during test, human readable, eg 'TODO'
+    {bytes_sent}            bytes sent during test, human readable, eg '52.45'
     {bytes_sent_unit}       unit used for bytes_sent, eg 'MB'
-    {bytes_sent_raw}        bytes sent during test, for format comparation, eg 'TODO'
-    {bytes_received}        bytes received during test, human readable, eg 'TODO'
-    {bytes_received_raw}    bytes received during test, for format comparation, eg 'TODO'
+    {bytes_sent_raw}        bytes sent during test, for format comparation, eg '52445184.0'
+    {bytes_received}        bytes received during test, human readable, eg '70.23'
```
tobes: 70.23 what?  70.23kb/s 70.23Tb/s

tobes: You know I do sometimes a the code ;p

My point is that the example is not valid. Maybe this is not clear from my comment

lasers: But the example is valid... is what I was trying to say.

If a CPU placeholder prints a percent number, then the example shouldn't show `%` too. And if it does, then the `eg` example is wrong.

If it actually print `%` too.. then the code is wrong. Sorry if I was not clear too.

tobes: OK makes sense the documentation could be clearer

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,313 @@
...
+            current_data[x] = current_data.get(x, 0)
+            current_data[x + "_raw"] = current_data[x]
+            current_data[x], current_data[x + "_unit"] = self.py3.format_units(
+                current_data[x], unit=unit, si=self.si_units
+            )
+
+        # extra data, not sure we want to expose
+        for x in ["server", "client"]:
+            if x in current_data:
+                for y in current_data[x]:
```
lasers: You can use `self.py3.flatten_dict()` instead.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,313 @@
...
+            self.speedtest_data.update(
+                {"previous_" + k: v for (k, v) in previous_data.items()}
+            )
+
+        # # cast
+        self.speedtest_data.update(
+            {
+                k: float(v)
+                for (k, v) in self.speedtest_data.items()
+                if isinstance(v, Number)
```
lasers: Some are `int`. Users won't use this much, but we shouldn't see `0.0` or `2321.0` on the bar though.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,313 @@
...
+                        quality_key = 1
+                    else:
+                        quality_key = 2
+                current_data["quality"] = quality_key
+                current_data["quality_name"] = self.quality[quality_key]
+
+        # zero-ing if not fetched, raw version and units convertion
+        for x in ["download", "upload", "bytes_received", "bytes_sent"]:
+            unit = self.unit_size if "bytes" in x else self.unit_bitrate
+            current_data[x] = current_data.get(x, 0)
```
lasers: We don't have to run `self.py3.format_units()` and other things all times on all four placeholders... Only what we need from `format`. It might be good to skip the rest of work if we get `None`.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,313 @@
...
+    def _start(self):
+        self.cached_until = self.py3.time_in(self.sleep_timeout)
+
+        if not self._is_running():
+            self.cached_until = self.py3.CACHE_FOREVER
+            previous_data = self.py3.storage_get("speedtest_data") or {}
+            current_data = self._get_speedtest_data() or {}
+
+            if current_data and len(current_data) > 1:
+                # create a "total" for know if cnx is better or not
```
lasers: You used `or {}` above. You shouldn't need `if current_data` anymore.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,313 @@
...
+                # between two run
+                current_data["total"] = int(current_data.get("download", 0)) + int(
+                    current_data.get("upload", 0)
+                )
+
+                # create "quality" #maybe bad name
+                if previous_data and "total" in previous_data:
+                    if current_data["total"] >= previous_data["total"]:
```
lasers: You used `or {}` above. You shouldn't need `if previous_data:` anymore.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,313 @@
...
+        # store last data fetched
+        self.py3.storage_set("speedtest_data", current_data)
+
+        # get speedtest result url
+        self.url = current_data.get("share")
+
+        # create placeholders
+        self.speedtest_data.update(current_data)
+        if previous_data:
+            self.speedtest_data.update(
```
lasers: You used `or {}` above. You shouldn't need `if previous_data:` anymore.

py3status/modules/speedtest.py
cyrinux: Must I keep default placeholders `bytes_send` and `bytes_received` and create `*_human` one for `self.py3.safe_format_units`?

py3status/modules/speedtest.py
tobes: soon this will not be needed  :)

cyrinux: I made quality due to impossibilty to compare previous and current "raw" data.
I can remove it because you and tobes fix this if i understand.

cyrinux: @tobes why `self.py3.flatten_dict()` will not be needed?

py3status/modules/speedtest.py
cyrinux: so I int all Number?

lasers: Which placeholders specifically do you require to have `int/float`?

`speedtest-cli --json` gave me json with several floating numbers. There are some string ints/floats (eg two different `lat` + `lon`, `server_id`, etc). Are you casting this because you knew about https://github.com/ultrabug/py3status/issues/1391?

The only thing that can require casting is... `isprating`, `rating`, `ispdlavg`, `ispulavg`, `loggedin`. I'm not sure if users would use this to compare something with.

py3status/modules/speedtest.py
lasers: You should still don't need `if previous data:`. It would be updated with an empty dict instead.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,321 @@
...
+            current_data[x] = current_data.get(x, 0)
+            current_data[x + "_raw"] = current_data[x]
+            current_data[x], current_data[x + "_unit"] = self.py3.format_units(
+                current_data[x], unit=unit, si=self.si_units
+            )
+
+        # extra data, not sure we want to expose
+        current_data.update(self.py3.flatten_dict(current_data, delimiter='_'))
+        for x in ['client', 'server']:
+            del current_data[x]
```
lasers: I think you might be creating `flatten_dict(current_data, delimiter='_', intermediates=True)`, then you delete the unwanted dicts. Does `current_data = self.py3.flatten...` works the same here?

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,321 @@
...
+        self.speedtest_data = {}
+        self.cached_until = self.py3.CACHE_FOREVER
+        self.placeholders = self.py3.get_placeholders_list(self.format)
+        self.thresholds_init = self.py3.get_color_names_list(self.format)
+        self.command = "speedtest-cli --json --secure"
+        self.url = None
+
+        self.quality = {
+            -1: "failed",
+            0: "unknow",
```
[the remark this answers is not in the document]
cyrinux: shortcut, will remove quality

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,321 @@
+# -*- coding: utf-8 -*-
+"""
+Do a bandwidth test with speedtest-cli.
+
+Configuration parameters:
+    button_refresh: button to run the check (default 2)
```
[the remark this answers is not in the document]
cyrinux: ok!

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,321 @@
...
+    button_share: button to open the result link (default None)
+    format: display format for this module, {download} and/or {upload} required
+        *(default 'Speedtest [\?color=darkgray ping [\?color=ping {ping} ms] '
+                    'up [\?color=upload {upload} {upload_unit}] '
+                    'down [\?color=download {download} {download_unit}]]')*
+    server_id: speedtest server to use, `speedtest-cli --list` to get id (default None)
+    si_units: use SI units (default False)
+    sleep_timeout: when speedtest-cli is allready running, this interval will be used
+        to allow faster retry refreshes (default 5)
```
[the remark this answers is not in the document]
cyrinux: ok, hardcoded to 2.5

py3status/modules/speedtest.py
cyrinux: not need intermediates ;)

cyrinux:
We progress, i have error when test fail at middle because my wifi works bad. Must also test without internet.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+        "upload": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
+    }
+    format = '[\?color=ping {ping} ms] [\?color=upload Up {upload} {upload_unit}]'
+    format += ' [\?color=download Down {download} {download_unit}]'
+}
+
+# colored based on comparation between two last runs
+speedtest {
+    format = '[[\?if=download_raw<previous_download_raw&color=degraded Faster]'
+    format += '|[\?color=ok Slower]]'
```
lasers: `color=ok` does not exist anywhere here.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+        self.py3.storage_set("speedtest_data", current_data)
+
+        # get speedtest result url
+        self.url = current_data.get("share")
+
+        # create placeholders
+        self.speedtest_data.update(current_data)
+        self.speedtest_data.update(
+            {"previous_" + k: v for (k, v) in previous_data.items()}
+        )
```
lasers: (Opinion) Are you going to use `_raw` and `previous_` yourself? If not, then I think it would be better not to add them in first place. Make it more about current speed test, not previous or raw. Users should be able to change `unit_bitrate` and `unit_size` to get same (raw) results too.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+        if not self._is_running():
+            self.cached_until = self.py3.CACHE_FOREVER
+            previous_data = self.py3.storage_get("speedtest_data") or {}
+            current_data = self._get_speedtest_data() or {}
+
+            if len(current_data) > 1:
+                # create a "total" for know if cnx is better or not
+                # between two run
+                current_data["total"] = int(current_data.get("download", 0)) + int(
+                    current_data.get("upload", 0)
```
lasers: (Q) What is this for? It's the only `total` words in the code. Is this meant to be removed too (quality)?

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
+# -*- coding: utf-8 -*-
+"""
+Do a bandwidth test with speedtest-cli.
+
+Configuration parameters:
+    button_share: button to open the result link (default None)
```
lasers: (Conflicts) I'm trying to figure out if it's better to keep this `button_share` config or to use undocumented `{share}` placeholder instead + `index` so when users click `share`, it would open up an url... The latter requires users to touch `format`. I want to see if we can default this without `button_share` config too.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+    format: display format for this module, {download} and/or {upload} required
+        *(default 'Speedtest [\?color=darkgray ping [\?color=ping {ping} ms] '
+                    'up [\?color=upload {upload} {upload_unit}] '
+                    'down [\?color=download {download} {download_unit}]]')*
+    server_id: speedtest server to use, `speedtest-cli --list` to get id (default None)
+    si_units: use SI units (default False)
+    thresholds: specify color thresholds to use *(default {
+                'download': [(0, 'bad'), (1024, 'degraded'), (1024 * 1024, 'good')],
+                'ping': [(200, 'bad'), (150, 'orange'), (100, 'degraded'), (10, 'good')],
+                'upload': [(0, 'bad'), (1024, 'degraded'), (1024 * 1024, 'good')]})*
```
lasers: (Qeustion) You have sources for this numbers? How do you know what's good, degraded, bad, etc?

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+    {"full_text": " down ", "color": "#a9a9a9"},
+    {"full_text": "39.58 MB/s", "color": "#fb4934"}
+]
+"""
+
+from json import loads
+from numbers import Number
+
+STRING_NOT_INSTALLED = "not installed"
+MISSING_PLACEHOLDER = "{download} or {upload} placeholder required"
```
lasers: Dead string.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+    def _get_speedtest_data(self):
+        return loads(self.py3.command_output(self.command))
+
+    def _start(self):
+        self.cached_until = self.py3.time_in(self.sleep_timeout)
+
+        if not self._is_running():
+            self.cached_until = self.py3.CACHE_FOREVER
+            previous_data = self.py3.storage_get("speedtest_data") or {}
+            current_data = self._get_speedtest_data() or {}
```
lasers: This part might not make sense. If you don't have data, you're still running everything below. I think you need to change so switch your `if statements` in a way that if it's still running, don't run thresholds, looping through, making fake zero values, etc... repeatedly. Just skip them all until you get the data first.

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,273 @@
...
+    {client_ispulavg}       client isp upload average, eg '0'
+    {client_lat}            client latitude, eg '48.8534'
+    {client_loggedin}       client logged in, eg '0'
+    {client_lon}            client longitude, eg '2.3487999999999998'
+    {client_rating}         client rating, eg '0'
+    {download}              download rate from speedtest server, human readable
+    {download_raw}          download rate from speedtest server, for format comparation
+    {download_unit}         unit used as , eg 'MB/s'
+    {ping}                  ping time in ms to speedtest server
+    {timestamp}             timestamp of the run, eg '2018-08-30T16:27:25.318212Z'
```
lasers: One annoying thing about `Speedtest` is that if you refresh, you can't tell when it's finished. I think you should try make `{query_seconds}` or `{time_seconds}`. Maybe it can be part of default too next to `Speedtest`... so when we click it, we see the `[\?color=deepskyblue {seconds}]` going up... until it stopped there. That's when we know it's done. Lack of progress feedback sucks big time.  Just a thought. I guess `sleep_timeout = 0`.

cyrinux: good idea, i thought on this too, will try

py3status/modules/speedtest.py
cyrinux: raw was for do comparation, if now with patch and casting we can compare without raw, i can remove them.

cyrinux:
About, casting, on low connection, float is important...
Should must be better to keep float and use `{placeholder:0f}` no?

On no connection (if speedtest-cli raise error), must i raise an error to say there is no connection? Or keep my zeroing value is enought?

cyrinux:
I think about cast download and upload or yes ping for example. I must be able to choose if I want 29 ms or  28.56 ms.
For the moment on error in zero `ping` , `*download`, `*upload`.

py3status/modules/speedtest.py
cyrinux: Before I go further, I have done some try and I understand I need Threads to get working what we want. Before running command is a sort of long time polling command which block the rest of script.
If I run command in a Threads this should works.
We forget the idea or I go with Threads?

lasers: Your code, your decision. We can forget this idea if you want to. 

Alternatively, see how it works when we empty the data on every `start()`. It should help.

cyrinux:
peek 2018-09-04 23-09

py3status/modules/speedtest.py
lasers: Maybe... Users have different speeds and needs. I don't want to keep painting things red for some users. Alternatively, we could paint them all with a color that is similar to speedtest logo/theme... kinda like arbitrary `loadavg` blue... Maybe try `#6AFFF3`... Not sure how it looks (color is from speedtest.net's stars). Hope it's not too bright like the stars. :)

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+        "down [\?color=download {download} {download_unit}]]"
+    )
+    server_id = None
+    si_units = False
+    thresholds = {
+        "download": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
+        "ping": [(200, "bad"), (150, "orange"), (100, "degraded"), (10, "good")],
+        "upload": [(0, "bad"), (1024, "degraded"), (1024 * 1024, "good")],
+    }
+    timeout = 10
```
lasers: `timeout` should be `None` so it does not get added to the command too in case speedtest-cli changes its default timeout in the future. (imho)

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+            for key in self.placeholders
+        ):
+            if not any(key.startswith("upload") for key in self.placeholders):
+                self.command += " --no-upload"
+
+            if not any(key.startswith("download") for key in self.placeholders):
+                self.command += " --no-download"
+
+        if self.server_id and int(self.server_id):
+            self.command += " --server " + str(self.server_id)
```
lasers: We probably don't need this `and int()` check and `"... {}".format(self.server.id)` looks nice (imho).

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+    def kill(self):
+        self.killed.set()
+
+    def on_click(self, event):
+        button = event["button"]
+        if button == self.button_share and self.url:
+            self.py3.command_run("xdg-open %s" % self.url)
+        if button != self.button_refresh:
+            self.py3.prevent_refresh()
+        if button == self.button_refresh:
```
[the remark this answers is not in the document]
cyrinux: ok maybe more readable

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+
+            # thresholds
+            for x in self.thresholds_init:
+                if x in self.speedtest_data:
+                    self.py3.threshold_get_color(self.speedtest_data[x], x)
+
+    def speedtest(self):
+        # calculate elapsed time since start
+        if self.start_time:
+            self.elapsed_time = int(time.time() - self.start_time)
```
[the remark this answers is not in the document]
cyrinux: ok will try this 👍

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+
+        self.button_refresh = 2
+        self.cached_until = self.py3.CACHE_FOREVER
+        self.command = "speedtest-cli --json --secure --timeout {}".format(
+            self.timeout
+        )
+        self.elapsed_time = None
+        self.killed = threading.Event()
+        self.placeholders = self.py3.get_placeholders_list(self.format)
+        self.sleep_timeout = 0
```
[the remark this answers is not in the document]
cyrinux: i'm lost between my several code editor 🥇

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+
+            # create placeholders
+            self.speedtest_data.update(current_data)
+            self.speedtest_data.update(
+                {"previous_" + k: v for (k, v) in previous_data.items()}
+            )
+
+            # stop timer / stop refreshing
+            self.start_time = None
+            self.elapsed_time = None
```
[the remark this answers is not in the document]
cyrinux: yep

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+        if not self.py3.check_commands("speedtest-cli"):
+            raise Exception(STRING_NOT_INSTALLED)
+
+        self.button_refresh = 2
+        self.cached_until = self.py3.CACHE_FOREVER
+        self.command = "speedtest-cli --json --secure --timeout {}".format(
+            self.timeout
+        )
+        self.elapsed_time = None
+        self.killed = threading.Event()
```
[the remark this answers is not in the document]
cyrinux: I use yubikey one as model as I have done for usbguard ;)
Will try to find the comment you mention...

py3status/modules/speedtest.py
```diff
@@ -0,0 +1,298 @@
...
+        if any(
+            key.startswith(x)
+            for x in ["download", "upload"]
+            for key in self.placeholders
+        ):
+            if not any(key.startswith("upload") for key in self.placeholders):
+                self.command += " --no-upload"
+
+            if not any(key.startswith("download") for key in self.placeholders):
+                self.command += " --no-download"
```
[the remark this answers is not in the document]
cyrinux: understand, i removed _raw so i can test with "in" and "not" startwith

cyrinux:
Hello @lasers I think I'm done for this last review, can't find info about daemonize or not, but works well like this. What is the problem?

lasers:
@tobes Can you explain the differences between `Event()` vs `daemon`?

py3status/modules/speedtest.py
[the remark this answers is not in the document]
lasers:
```python
tests = ["download", "upload"]
for x in tests:
    if x not in self.placeholders:
        self.command += " --no-{}".format(x)
```

py3status/modules/speedtest.py
```diff
@@ -108,8 +108,9 @@
 """
 
 from json import loads
-import time
 import threading
+import time
```
lasers: Use `from time import time`.

py3status/modules/speedtest.py
```diff
@@ -264,7 +277,8 @@ def speedtest(self):
         # calculate elapsed time since start
         elapsed_time = None
         if self.start_time:
-            elapsed_time = int(time.time() - self.start_time)
+            elapsed_time = self._cast_number(time.time() - self.start_time)
```
lasers: I think you don't need `_cast_number` if you used `class Meta`.

cyrinux:
Good morning, Ok will look this.

tobes:
> @tobes Can you explain the differences between Event() vs daemon?

`daemon` is a Thread type and is about when the Thread is killed

`Event()` is about waiting on stuff

They are like apples and oranges - comparison is not possible

lasers:
To clarify... We should want to stay with `Event()` here? You suggested daemon in `usbguard` so it would make sense to have that keep running all times. Here, we wait no more than 30 seconds for `speedtest-cli` process to come back with a JSON data, then we don't need to run anything until users refreshes the `speedtest` module.

I think `yubikey` may need `daemon` too because it seems to be running all times too.

cyrinux:
I don't understand all yet sorry ... my goal is to have the module not stuck when the test is running to be able to refresh it. Thread make the job. Must i use other thing?

py3status/modules/speedtest.py
```diff
@@ -265,14 +255,6 @@ def _start(self):
         self.start_time = None
         self.cached_until = self.py3.CACHE_FOREVER
 
-        # cast number
-        self.speedtest_data.update(
-            {
-                k: self._cast_number(self.speedtest_data[k])
```
lasers: Is this correct? I was talking about casting `elapsed_time` alone.  Just a heads up.

cyrinux: My try was good with default format, but i have not try other placeholders after this commit. Maybe only placeholder update in Meta class are good.

py3status/modules/speedtest.py
cyrinux:
```
tests = ["download", "upload"]
for x in tests:
    if x not in self.placeholders:
        self.command += " --no-{}".format(x)
```
I can't have no-upload and no-download together, this is invalid params combo.

tobes:
> I think yubikey may need daemon too

@lasers - threads are complicated and it is easy to fuck things up - so don't play with them because you can.

Sometimes it is better to wait to see if there is a problem than to try to fix a perceived one. 

Change because we can is almost always a bad thing

tobes:
> Here the Event version.  Which one is better?


I'm 100% sure that you do not want events.

I've not looked at this code and realistically won't until this PR is closed and you make a PR version 2 one because the conversation here is so long - I'm not even going to start reading through it.

I like the idea of the module and am sure I'll want lots of changes if I do review it :)

again not looked at the code but I would

have a thread that runs the speed test and logs data as it goes and then terminates.  The module would update monitoring the data from the thread.  when the thread is ended the module stops updating.  Thread is triggered by click event.

Maybe you need an Event() what is your code doing?  anyhow have fun coding this

cyrinux:
I have done a Thread like usbguard/yubikey and one with Event() with do_no_disturb as model.
If I don't say wrong, that you describe is what I done in my first version. @lasers you confirm?

lasers:
@cyrinux Add a log inside your thread (or use `--debug` on py3status). You'll see `speedtest` is refreshing too much instead of every `0.1` or `1.0` second.

cyrinux:
Thanks for hint ;) Fixed (and other thing)

cyrinux:
I works this morning, on a SIMPLE version, so was trying to reuse only one thread spawner per run (like my previous version so but little cleaner). Was thinking of take a lock for prevent running 2 threads but seems this works like this without.
Also, should I readd check to check if a speedtest, started manually for example, is allready running  and in this case do nothing? This is to prevent wrong result...btw if people is downloading holidays movie by torrent result will be not ok also...

lasers:
Try https://github.com/lasers/py3status/tree/othermodule-speedtest.

Too much changes and I don't want to upset/touch your code. ;-)

cyrinux:
Hi lasers, seems works well with default config. Display reset is cool improvement.

lasers:
@cyrinux Summary...

* Daemon. :fireworks: 
* Storage statistics between restarts so we don't lose the bar data.
* One button to start/reset.
* Share placeholder instead of button config. If users want it, add `{share}`.
* Untested with no connection or such. Something you might want to test.
* I don't like that I killed data when we start speedtest. I think it would be better if we check elapsed boolean instead of data.
* Copied three color hex from speedtest website.
* I think there are too many trivial settings relating to units and speedtest that users won't care about, so I hide them all.

If you have questions/concerns... or see issues, please share them.

cyrinux:
Nice. I would prefer `.0f` for counter, and on start, be able to reset previous saved result before run.

lasers:
Pushed another commit. 

>I would prefer .0f for counter

When starting counter, there is a _tiniest_ delay as we latch onto the nearest second, but accurate. The initial nearest second value will always be different based on restarts or such. I gave it `.2f` to be consistent with other placeholders. Your module, your call. I'm just toying. Same for other placeholders.

>be able to reset previous saved result before run

This part kills a formatting functionality and it also does not make sense to reset data for no reason and we're trying to storage this too. I added `elapsed` bool (and an example) so you can see the difference between hiding data (instead of resetting) versus keeping data.  This gives users both behaviors.

cyrinux:
Well, I understand. So it is great like this. Tried last version, it's ok.

lasers:
@cyrinux Do you approve my `othermodule-speedtest` work on top of your module? I then can add my commits directly to this branch. Otherwise, you can keep working on this or is it finished already?

As always, you are free to do what you want. My work contains too many changes and it is possible that you don't agree with some of them. Let me know so we know where this module stands today.

cyrinux:
@lasers you can commit over this one with yours change. Thanks

lasers:
I have a lingering question... Will users change anything in `thresholds` here?
```
        *(default {"upload": [(0, "violet")], "ping": [(0, "#fff381")],
        "download": [(0, "cyan")], "elapsed_time": [(0, "#1cbfff")]})*
```

We're not using `thresholds` correctly in a sense that we're just putting in default `zero` colors... and I question if users will use/change it to something else or if they are content with default colors.

If users are content with default colors, then we could change it in `format` instead. On the other hand, we have ` xxx: print a color based on the value of `xxx` placeholder` thing going on here so it might be good to leave this alone just in case.

Do you see yourself setting anything custom in `thresholds` for this module? Which variables?

cyrinux:
Hi @lasers, no finally i use default thresholds, because i never use same connection. the day we will have profile feature maybe? :)


image

ultrabug:
Tested ok here

@cyrinux I will allow myself some modifications on default output once merged, hope you won't mind them

Thanks

PR #1439, opened 2018-08-20
external_script: show notification with full output on click, refresh only on button_refresh

maximbaz:
Replaces #1431.

Show notification with full `output` on click.
Refresh only on `button_refresh` click.
Still needs #1432 for most scripts that I use.

### DIFF
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -13,11 +13,11 @@ Configuration parameters:
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
-    format_notification: see placeholders below (default None)
     localize: should script output be localized (if available)
         (default True)
-    notification_state: specify notification state to use: current, click
-        (default 'current')
+    notifications: specify a nested dict to send a notification if matched
+        against the keys, see ``Notifications`` section for more information
+        (default {})
     script_path: script you want to show output of (compulsory)
         (default None)
     strip_output: shall we strip leading and trailing spaces from output
@@ -28,10 +28,23 @@ Format placeholders:
     {output} first line of the output of script given by "script_path"
     {output_full} full output of script given by "script_path"
 
-Format notification placeholders:
-    {line} number of lines in the output
-    {output} first line of the output of script given by "script_path"
-    {output_full} full output of script given by "script_path"
+Notifications:
+    Specify a nested dictionary of notification states and options to use.
+
+    Notification states:
+        'changed': display a notification only if it is changed
+        'click': display a notification of last output on click
+        'current': display a current notification regardless
+
+    Notification options:
+        'title': notification title
+        'msg': notification message
+        'level': must be 'info', 'error' or 'warning'.
+        'rate_limit': time period in seconds not to be repeated
+        'icon': must be an icon path or icon name.
+
+        You can add `format` placeholders in `msg` and `title`.
+        The `msg` and `title` options will also be formatted.
 
 Examples:
 ```
@@ -41,23 +54,28 @@ external_script {
     script_path = "/usr/bin/whoami"
 }
 
-# always display a notification, no output means no notification
+# display changed notifications, same output means no notification
 external_script {
-    format_notification = '{output}
-    script_path = "~/my_python_script.py"
+    notifications = {'changed': {'msg': '{output}'}}
+    script_path = "~/my_script.py"
 }
 
-# display a notification only if output have more than 20 lines
+# display current notifications, no output means no notification
 external_script {
-    format_notification = '\?if=line>20 {output}'
-    script_path = "~/my_python_script.py"
+    notifications = {'current': {'msg': '{output}'}}
+    script_path = "~/my_script.py"
+}
+
+# display current notification only if output have more than 20 lines
+external_script {
+    notifications = {'current': {'msg': '\?if=line>20 {output}'}}
+    script_path = "~/my_script.py"
 }
 
 # display a notification of last full output on click
 external_script {
-    format_notification = '{output_full}'
-    notification_state = 'click'
-    script_path = "~/my_python_script.py"
+    notifications = {'click': {'msg': '{output_full}'}}
+    script_path = "~/my_script.py"
 }
 ```
 
@@ -72,7 +90,6 @@ example
 
 import re
 STRING_ERROR = 'missing script_path'
-STRING_STATE_ERROR = 'invalid notification state `{}`'
 
 
 class Py3status:
@@ -81,33 +98,46 @@ class Py3status:
     # available configuration parameters
     cache_timeout = 15
     format = '{output}'
-    format_notification = None
     localize = True
-    notification_state = 'current'
+    notifications = {}
     script_path = None
     strip_output = False
 
     def post_config_hook(self):
         if not self.script_path:
             raise Exception(STRING_ERROR)
-        elif self.notification_state not in ['click', 'current']:
-            raise Exception(STRING_STATE_ERROR.format(self.notification_state))
 
         self.button_refresh = 2
-        notification_boolean = bool(self.format_notification)
-        self.init = {
-            'notify_click': (
-                notification_boolean and self.notification_state == 'click'
-            ),
-            'notify_current': (
-                notification_boolean and self.notification_state == 'current'
-            ),
-        }
+        self.notification = {'normal': []}
+        if self.notifications:
+            for x in ['changed', 'click', 'current']:
+                self.notification[x] = self.notifications.get(x, False)
+                if self.notification[x]:
+                    if x in ['changed', 'current']:
+                        self.notification['normal'].append(x)
+            self.last_changed = self.py3.storage_get('changed')
 
-    def _notify_user(self):
-        self.py3.notify_user(
-            self.py3.safe_format(self.format_notification, self.script_data)
-        )
+    def _get_notification(self, state):
+        temporary = self.notification[state].copy()
+        for x in ['title', 'msg']:
+            if x in temporary:
+                temporary[x] = self.py3.get_composite_string(
+                    self.py3.safe_format(temporary[x], self.script_data)
+                )
+        return temporary
+
+    def _notify_user(self, state=None):
+        if state is None:
+            if self.notification['changed']:
+                changed = self._get_notification('changed')
+                if changed != self.last_changed:
+                    self.last_changed = changed
+                    self.py3.storage_set('changed', changed)
+                    self.py3.notify_user(**changed)
+            if self.notification['current']:
+                self.py3.notify_user(**self._get_notification('current'))
+        elif state == 'click':
+            self.py3.notify_user(**self._get_notification('click'))
 
     def external_script(self):
         output_lines = None
@@ -156,9 +186,9 @@ class Py3status:
         response['full_text'] = self.py3.safe_format(
             self.format, self.script_data
         )
-        self.script_data['output_full'] = output_full
 
-        if self.init['notify_current']:
+        self.script_data['output_full'] = output_full
+        if self.notification['normal']:
             self._notify_user()
 
         return response
@@ -166,8 +196,8 @@ class Py3status:
     def on_click(self, event):
         button = event["button"]
         if button != self.button_refresh:
-            if self.init['notify_click']:
-                self._notify_user()
+            if self.notification['click']:
+                self._notify_user('click')
             self.py3.prevent_refresh()
 
 
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -10,11 +10,10 @@ code such as #FF0000 for red).
 The script should not have any parameters, but it could work.
 
 Configuration parameters:
-    button_refresh: button to refresh the module (default 2)
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
-    format_notification: see placeholders below (default '{output}')
+    format_notification: see placeholders below (default None)
     localize: should script output be localized (if available)
         (default True)
     script_path: script you want to show output of (compulsory)
@@ -27,6 +26,7 @@ Format placeholders:
     {output} first line of the output of script given by "script_path"
 
 Format notification placeholders:
+    {line} number of lines in the output
     {output} full output of script given by "script_path"
 
 i3status.conf example:
@@ -55,15 +55,15 @@ class Py3status:
     """
     """
     # available configuration parameters
-    button_refresh = 2
     cache_timeout = 15
     format = '{output}'
-    format_notification = '{output}'
+    format_notification = None
     localize = True
     script_path = None
     strip_output = False
 
     def post_config_hook(self):
+        self.button_refresh = 2
         if not self.script_path:
             raise Exception(STRING_ERROR)
 
@@ -72,9 +72,10 @@ class Py3status:
         response = {}
         response['cached_until'] = self.py3.time_in(self.cache_timeout)
         try:
-            self.output = self.py3.command_output(
-                self.script_path, shell=True, localized=self.localize)
-            output_lines = self.output.splitlines()
+            output = self.py3.command_output(
+                self.script_path, shell=True, localized=self.localize
+            )
+            output_lines = output.splitlines()
             if len(output_lines) > 1:
                 output_color = output_lines[1]
                 if re.search(r'^#[0-9a-fA-F]{6}$', output_color):
@@ -105,8 +106,15 @@ class Py3status:
         else:
             output = ''
 
+        self.script_data = {
+            'line': len(output_lines),
+            'output': output,
+            'output_full': ' '.join(output.splitlines())
+        }
         response['full_text'] = self.py3.safe_format(
-            self.format, {'output': output, 'line': len(output_lines)})
+            self.format, self.script_data
+        )
+        self.script_data['output_full'] = output
         return response
 
     def on_click(self, event):
@@ -114,7 +122,7 @@ class Py3status:
         if button != self.button_refresh:
             if self.format_notification:
                 self.py3.notify_user(self.py3.safe_format(
-                    self.format_notification, {'output': self.output})
+                    self.format_notification, self.script_data)
                 )
             self.py3.prevent_refresh()
 
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -72,7 +72,8 @@ class Py3status:
         response = {}
         response['cached_until'] = self.py3.time_in(self.cache_timeout)
         try:
-            self.output = self.py3.command_output(self.script_path, shell=True, localized=self.localize)
+            self.output = self.py3.command_output(
+                self.script_path, shell=True, localized=self.localize)
             output_lines = self.output.splitlines()
             if len(output_lines) > 1:
                 output_color = output_lines[1]
@@ -116,6 +117,7 @@ class Py3status:
                     self.format_notification, {'output': self.output})
                 )
             self.py3.prevent_refresh()
+
 
 if __name__ == "__main__":
     """
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -16,6 +16,8 @@ Configuration parameters:
     format_notification: see placeholders below (default None)
     localize: should script output be localized (if available)
         (default True)
+    notification_state: specify notification state to use: current, click
+        (default 'current')
     script_path: script you want to show output of (compulsory)
         (default None)
     strip_output: shall we strip leading and trailing spaces from output
@@ -24,17 +26,38 @@ Configuration parameters:
 Format placeholders:
     {line} number of lines in the output
     {output} first line of the output of script given by "script_path"
+    {output_full} full output of script given by "script_path"
 
 Format notification placeholders:
     {line} number of lines in the output
     {output} first line of the output of script given by "script_path"
+    {output_full} full output of script given by "script_path"
 
-i3status.conf example:
-
+Examples:
 ```
+# add a script
 external_script {
     format = "my name is {output}"
     script_path = "/usr/bin/whoami"
+}
+
+# always display a notification, no output means no notification
+external_script {
+    format_notification = '{output}
+    script_path = "~/my_python_script.py"
+}
+
+# display a notification only if output have more than 20 lines
+external_script {
+    format_notification = '\?if=line>20 {output}'
+    script_path = "~/my_python_script.py"
+}
+
+# display a notification of last full output on click
+external_script {
+    format_notification = '{output_full}'
+    notification_state = 'click'
+    script_path = "~/my_python_script.py"
 }
 ```
 
@@ -49,6 +72,7 @@ example
 
 import re
 STRING_ERROR = 'missing script_path'
+STRING_STATE_ERROR = 'invalid notification state `{}`'
 
 
 class Py3status:
@@ -59,23 +83,41 @@ class Py3status:
     format = '{output}'
     format_notification = None
     localize = True
+    notification_state = 'current'
     script_path = None
     strip_output = False
 
     def post_config_hook(self):
-        self.button_refresh = 2
         if not self.script_path:
             raise Exception(STRING_ERROR)
+        elif self.notification_state not in ['click', 'current']:
+            raise Exception(STRING_STATE_ERROR.format(self.notification_state))
+
+        self.button_refresh = 2
+        notification_boolean = bool(self.format_notification)
+        self.init = {
+            'notify_click': (
+                notification_boolean and self.notification_state == 'click'
+            ),
+            'notify_current': (
+                notification_boolean and self.notification_state == 'current'
+            ),
+        }
+
+    def _notify_user(self):
+        self.py3.notify_user(
+            self.py3.safe_format(self.format_notification, self.script_data)
+        )
 
     def external_script(self):
         output_lines = None
         response = {}
         response['cached_until'] = self.py3.time_in(self.cache_timeout)
         try:
-            output = self.py3.command_output(
+            output_full = self.py3.command_output(
                 self.script_path, shell=True, localized=self.localize
             )
-            output_lines = output.splitlines()
+            output_lines = output_full.splitlines()
             if len(output_lines) > 1:
                 output_color = output_lines[1]
                 if re.search(r'^#[0-9a-fA-F]{6}$', output_color):
@@ -109,21 +151,23 @@ class Py3status:
         self.script_data = {
             'line': len(output_lines),
             'output': output,
-            'output_full': ' '.join(output.splitlines())
+            'output_full': ' '.join(output_full.splitlines())
         }
         response['full_text'] = self.py3.safe_format(
             self.format, self.script_data
         )
-        self.script_data['output_full'] = output
+        self.script_data['output_full'] = output_full
+
+        if self.init['notify_current']:
+            self._notify_user()
+
         return response
 
     def on_click(self, event):
         button = event["button"]
         if button != self.button_refresh:
-            if self.format_notification:
-                self.py3.notify_user(self.py3.safe_format(
-                    self.format_notification, self.script_data)
-                )
+            if self.init['notify_click']:
+                self._notify_user()
             self.py3.prevent_refresh()
 
 
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -27,7 +27,7 @@ Format placeholders:
 
 Format notification placeholders:
     {line} number of lines in the output
-    {output} full output of script given by "script_path"
+    {output} first line of the output of script given by "script_path"
 
 i3status.conf example:
 
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -14,6 +14,7 @@ Configuration parameters:
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
+    format_notification: see placeholders below (default '{output}')
     localize: should script output be localized (if available)
         (default True)
     script_path: script you want to show output of (compulsory)
@@ -23,7 +24,10 @@ Configuration parameters:
 
 Format placeholders:
     {line} number of lines in the output
-    {output} output of script given by "script_path"
+    {output} first line of the output of script given by "script_path"
+
+Format notification placeholders:
+    {output} full output of script given by "script_path"
 
 i3status.conf example:
 
@@ -54,6 +58,7 @@ class Py3status:
     button_refresh = 2
     cache_timeout = 15
     format = '{output}'
+    format_notification = '{output}'
     localize = True
     script_path = None
     strip_output = False
@@ -106,7 +111,10 @@ class Py3status:
     def on_click(self, event):
         button = event["button"]
         if button != self.button_refresh:
-            self.py3.notify_user(self.output)
+            if self.format_notification:
+                self.py3.notify_user(self.py3.safe_format(
+                    self.format_notification, {'output': self.output})
+                )
             self.py3.prevent_refresh()
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


lasers:
Why am I getting `fatal: Cannot setup tracking information; starting point 'maximbaz/external_script_notification' is not a branch.` from just you and @cyrinux?

maximbaz:
Don't know, what `git` commands do you use? Are you sure you fetched my remote before checking out the branch?

lasers:
```
[chris:~/src/py3status] master(+168/-1)+ ± git fetch maximbaz
[chris:~/src/py3status] master(+168/-1)+ ± git checkout https://github.com/ultrabug/py3status/pull/1439
fatal: Cannot setup tracking information; starting point 'maximbaz/external_script_notification' is not a branch.
```

py3status/modules/external_script.py
```diff
@@ -22,6 +23,7 @@
 
 Format placeholders:
     {output} output of script given by "script_path"
+    {count_lines} number of lines in the output
```
lasers: I want to avoid `count_` prefix here. Use `{line}` just like `{path}`.

If one day, somebody want specific lines, then we can do `{line_3}` too without messing up the placeholders. Also, it is much shorter and clean.
* `[\?if=line&color=line {line} lines]`
* `[\?if=count_lines&color=count_lines {count_lines} lines]`.

py3status/modules/external_script.py
```diff
@@ -83,9 +86,14 @@ def external_script(self):
...
 
         response['full_text'] = self.py3.safe_format(
-            self.format, {'output': output_text})
+            self.format, {'output': output_text, 'count_lines': len(output_lines)})
         return response
 
+    def on_click(self, event):
+        button = event["button"]
+        if button != self.button_refresh:
+            self.py3.notify_user(self.output)
```
lasers: This hardcode the notification with whole output. We want `self.format_notification` for this so users can specify if they want whole output, first line output, or number of lines in an output... just like `format`. It looks like you should make another placeholder to display whole output.

Based on current code, `{output}` is technically `{line_1}` right now because if we try `script_path = 'cal'`, we will get first line only.. instead of whole `cal` in `{output}`.

py3status/modules/external_script.py
maximbaz: Why? In my mind that's more confusing, why `line_1` is a string and `line` is suddenly a number.

`num_lines` maybe?

maximbaz: how about `line_count` then? no plurals, will stay next to `line_?` and yet explains well that this is a count.

lasers: Similar things with percents... eg `{brightness}` instead of `{brightness_percent}`. Sometimes, we add `_percent` because of sysdata's `{mem_used}`.  `{mem_used}` is same as `{line}` since we don't suffix `{mem_used}` with anything.

We have to consider `format` too. If we prefix this with `_count`, then everything gets expanded much longer. Placeholder, if statements, color thresholds, etc. This approach where we don't be explicit with counts seems to work well for several modules I worked on... such as `mail` where I don't do `count_mail`. The users can use `[\?if=mail {mail} Mail]` instead of `[\?if=count_mail {count_mail} Mail]`.

py3status/modules/external_script.py
lasers: The way I see it... We have a choice. Implement this PR where it will do things wrong and will do things wrong for long time or implement a PR that does things right in first place and not having to worry about it for long time.

For instance, if an user drops in... asking us how to turn off the notification because he simply want to refresh his script without dealing with notifications, can it be turned off right now?

If not, then we have to fix it eventually. That's why I want to fix this right in first place for the notifications. The ideal behavior is to leave notifications off by default too.

py3status/modules/external_script.py
maximbaz: To be honest I don't see an issue with verbose configuration, at least it's readable without referring to docs. To a new user, it will not be obvious that `format = '{line}'` will actually print a number.

But, I am ready to give up on this and change to `line` if you really want so 🙂

lasers: Please change. We need to get away with long names if we can... since most of the modules are small or simple enough not to warrant spelling out `_count`. Those simple module often are being used with threshold and other things too.

sysdata, pomodoro, taskwarrior, etc are not simple modules, but it looks like we can avoid them too... so I want to take advantage of that situation and extend this to all modules. It's a matter of verbosity level... I chalk it down a bit.

py3status/modules/external_script.py
maximbaz: I don't think not being able to format notification message is necessarily wrong, current behavior shows first line of the output on the status bar, this change allows you to additionally see the rest of the lines on click. 

As for disabling notifications, that's why initially I implemented it with two buttons, button_refresh and button_show_notification. Should I return to that approach?

lasers: Ask @cyrinux to help you with `format_notification`. It won't be a problem... The problem is that `{output}` placeholder is printing one line... so I assume we have to find a different name that explains `{output}`, but not `{output}`... eg `{full_output}`... but `{output}` is a good name to use.

lasers: @maximbaz Okay, the solution is make `format_notification = None`, but we won't document `{output_full}` placeholder so only you and few others will know about it... The rest can use `{output}` (aka 1line) or `{line}` (aka count_lines).

If/when we break this module someday, we could do it at that time.

So you just have to implement `format_notification`. Pass `self.py3.safe_form...` object to `self.py3.notifiy_user()`` instead of `self.output` string.

maximbaz: Done, now it is possible to set `self.format_notification = None` to disable notification. I intentionally implemented as least as possible functionality to keep this PR focused on its original goal.

maximbaz: It is named `output` because it is a proper name, and a short one as you seem to prefer - the fact that in another place `output` currently only prints the first line is the problem of that code.

py3status/modules/external_script.py
```diff
@@ -10,9 +10,11 @@
 The script should not have any parameters, but it could work.
 
 Configuration parameters:
+    button_refresh: button to refresh the module (default 2)
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
+    format_notification: see placeholders below (default '{output}')
```
lasers: This config need to be `None` by default so we don't annoy other users.

py3status/modules/external_script.py
```diff
@@ -21,7 +23,11 @@
         (default False)
 
 Format placeholders:
-    {output} output of script given by "script_path"
+    {line} number of lines in the output
+    {output} first line of the output of script given by "script_path"
+
+Format notification placeholders:
+    {output} full output of script given by "script_path"
```
lasers: Missing `{line}` too. See below.

py3status/modules/external_script.py
```diff
@@ -64,8 +72,9 @@ def external_script(self):
         response = {}
         response['cached_until'] = self.py3.time_in(self.cache_timeout)
         try:
-            output = self.py3.command_output(self.script_path, shell=True, localized=self.localize)
-            output_lines = output.splitlines()
+            self.output = self.py3.command_output(
+                self.script_path, shell=True, localized=self.localize)
+            output_lines = self.output.splitlines()
```
lasers: Leave this alone as-is. Put `output` inside the dict instead... and assign `self.py3.safe_format()` to `self.composite`. Feed composite to `self.py3.notify_user`.. then we can haz all same placeholders.

Don't document `{output_full}`. Fewer behavior changes/breakages by skipping out on making new configs/placeholders only to be deprecated later, eg `button_show_notification` and hardcoded `self.output`.

You're almost there.

maximbaz: I don't want to tackle formatting in this PR, I don't like forever-hanging huge PRs with lots of unrelated changes. This PR is solving one specific problem - showing full output in a notification without refreshing module on click. Formatting needs to be reviewed here anyway: `output` needs to not be truncated, `line_?` needs to be introduced, color parsing needs to be improved, etc - all of this deserves to be done separately.

py3status/modules/external_script.py
maximbaz: Why would they be annoyed? Middle click (default button for refresh) still works as it was working before, the rest of the buttons had no action assigned to them, now they do something useful - show full output of the script.

lasers: Everybody have different goals with their `external_script` and we don't know what their goals are. It can be something trivial like to clicking immediately to update a number of mails or updates... and not wanting to see a notification about last number of mails or updates... or to simply execute something periodically with static text... only to be notified about the same good ol' static text.

I assume 99% of time, it will be be same as the one they have on their bar. You want _New!_ full output in your notification. That's fine and we can add it... but not with default `notification = '{output}'` or `notification = '{output_full}'`.

 With `None` by default, we can quietly introduce this new feature without _literally_ notifying people about it.

py3status/modules/external_script.py
lasers: We're not tackling formatting. We're just adding `output_full` to a dict... doing same `self.output =` thing, but in different place... and with composite instead of string, and passing the composite to notify_user.. so we don't have to run same thing twice on every click.

py3status/modules/external_script.py
maximbaz: Clicking to refresh is what `button_refresh` is for, it doesn't show notifications and is there exactly for these cases. You said so yourself, it is good to keep `button_refresh` to `2` by default so that everyone knows: middle click is refresh with no extra actions. Other buttons do extra actions.

I don't see this as a feature "only for me", which I should hide from other users, I genuinely think it makes sense to be on by default. Same in battery_level by the way, I don't see why this is an opt-in feature, in my mind button `1` should always show notification and button `2` should be for silent refreshes.

py3status/modules/external_script.py
maximbaz: I just don't see the benefit of doing this, at this moment it only complicates the code. When there are more formatter placeholders, of course it makes sense to pre-compute them all and store in a dict, and do some micro-optimizations like wrapping in a composite - but not right now when there are so many unresolved issues and formatting itself needs to be redesigned.

lasers: Just pushed out modifications. I think this does what you want... and it looks like we don't really need `button_refresh`... so we can hide it... and I was slightly off... I don't want you to pass a composite. I wanted you to make and use `self.dict` instead... so all format have same placeholders.

maximbaz: I think we do need `button_refresh`, I want the middle click (button `2`) to be a silent refresh without any notification (because the notification it will show will contain outdated info anyway).

lasers: Sensible. I pushed it out. You can use `format_notification = '{output_full}'`... and the `format` support hidden `{output_full}` too. Things look okay on your side? Squash this PR?

py3status/modules/external_script.py
```diff
@@ -51,11 +57,13 @@ class Py3status:
     # available configuration parameters
     cache_timeout = 15
     format = '{output}'
+    format_notification = None
     localize = True
     script_path = None
     strip_output = False
 
     def post_config_hook(self):
+        self.button_refresh = 2
```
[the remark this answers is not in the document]
lasers: Okay because without `on_click`, the module still does not refresh on any other click events than `2` and we don't have any other button, so we don't need to mention it... and the behavior still will be same as before, but with optional notifications on click events (except `2`) this time.

`button_refresh` is not needed and `{output_full}` is not documented, but this is implemented to do what you want... and `having format_notification` implemented correctly is what I want.

py3status/modules/external_script.py
```diff
@@ -51,11 +57,13 @@ class Py3status:
     # available configuration parameters
     cache_timeout = 15
     format = '{output}'
+    format_notification = None
```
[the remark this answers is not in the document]
lasers: `button_refresh` is used because some modules connects to Internet. We don't want people to scroll and start hitting up Internet.

It's in `external_script`, `whatismyip`, `github`, `google_calendar`.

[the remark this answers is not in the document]
lasers: Talking specifically about mouse scroll wheel. We can get 50+ events per second if we scroll hard.... I don't think it's good idea to force notifications on existing `external_script` users because we don't know their cases and most of them can not want it... so they have to fix thier config to turn it off... Having it off (and always off by default) is better.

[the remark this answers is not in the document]
maximbaz: Well not true, it's just hiding the issue - those who enabled notifications will suffer from this issue. Lets introduce button_show_notification then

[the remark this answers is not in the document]
lasers: Why do we need `button_show_notification` if it works okay on several buttons?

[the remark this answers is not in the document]
maximbaz: Because like you said, I can accidentally generate 50+ events with mouse scroll - your argument "I don't want such accidents to happen, therefore disable notifications by default", my argument "I _also_ don't want this to happen, but disabling notifications is not an option for me, so let's find a better way to fix this issue".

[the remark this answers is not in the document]
lasers: I did not disable notifications because of 50+ events. When we add `on_click` to the module, the automagically `2` is turned off. For that reason alone, we sometimes need to add it to deal with Internet modules. For other modules, we just let it refresh.

In this case, `button_refresh` is not added because there simply are no other button. Your use case is that you wanted the notification on last output and at same not, not triggering anything.

I disable notifications because it is simply a bad behavior to force notifications on other users.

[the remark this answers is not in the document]
lasers: Obviously we don't agree on same things.... :-) 

Pinging @ultrabug for his opinion on this matter.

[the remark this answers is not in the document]
maximbaz: I guess I simply don't understand your use-case, with what you are proposing most buttons will not be doing anything (not even module refresh, absolutely nothing). Why not put some useful action on button click if these buttons are unused? 

By analogy we can say that Chromium is forcing an action "open new window" when you press "Ctrl+N". This shortcut is unused, so they put some action on it. Same here, button is unused and I have a useful action that I can put on this button.

I'll update the code in this PR to have it show my intention for this, and yeah let's have a third opinion on it 🙂

py3status/modules/external_script.py
```diff
@@ -82,10 +92,26 @@ def external_script(self):
...
+        self.script_data = {
+            'line': len(output_lines),
+            'output': output_text,
+            'output_full': ' '.join(output.splitlines())
+        }
         response['full_text'] = self.py3.safe_format(
-            self.format, {'output': output_text})
+            self.format, self.script_data
+        )
+        self.script_data['output_full'] = output
```
maximbaz: @lasers why is this line needed? it already contains full output since line 98, isn't?

py3status/modules/external_script.py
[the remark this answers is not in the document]
lasers: Because this is `external_script`. It does not really have a specific purpose and your use case is a bit different in that you only wanted a notification on last output... External scripts often have some intervals in them... and `format_notification` might confuse users in thinking that they'll be getting a notification on every interval... (say I put in 7200, then I expect a notification every two hours). This is not the case... so I'm not sure if we should even add `format_notification` here too... It's too early to tell.

What you wanted is minimally implemented.  Minimal is the keyword here.  If users come by wanting notification on every two hours, then yes, it would be easier to implement that minimally too... 

If we add buttons, then it's hard to undo them so I'm trying to avoid not adding unnecessary configs because if an user stop by with a PR containing something really useful for `external_script` and it requires a button, then by all means, we should let him do that.

`external_script` is hard in that it does not really have specific purposes... It might be better to enhance https://py3status.readthedocs.io/en/latest/configuration.html#special-on-click-commands to accept `{placeholders}` too, not just `$OUTPUT`...

And obviously, to enhance `on_click` too with new options to better address #1264.

py3status/modules/external_script.py
lasers: To remove newlines, put them all on one same line for the bar.
And to preserve newlines for the notification.

Try `cal`.

py3status/modules/external_script.py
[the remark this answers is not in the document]
maximbaz: > if an user stop by with a PR containing something really useful for external_script and it requires a button, then by all means, we should let him do that.

ah, my feelings are hurt 😢, my PR is not considered cool/useful enough to make a case for reserving a button... 😛 

Although I find it strange that we prefer to let notifications be shown on mouse scroll to creating a dedicated `button_show_notification`, I'm fine with conceding this to you — in reality there is a throttle on notifications, so even if you scroll extensively, you will get at most one notification every 5 seconds. Accidents will not be too painful.

However, I still believe that by default showing a notification on click is better than doing absolutely nothing. If you are annoyed by notification, no configuration change is needed — simply don't click on the module, click is not doing anything except showing a notification.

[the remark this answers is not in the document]
lasers: >However, I still believe that by default showing a notification on click is better than doing absolutely nothing.

Doing absolutely nothing is the best way to ensure that we don't break user configs. Instead of trying to put something in the `format_notification` by default... only to later remove it., we break one thing. We can defer the `format_notification` to users instead.

This is `external_script` though. Users uses this to execute their scripts. We don't know their scripts. Their scripts could already contain some notifications too... so now they get dual notifications... One for clicking the module (and not caring about last output)... and one for when  when their script is done doing something.

I believe there can be some grief to have notifications enabled by default versus none.

_Still waiting for third opinion... No rush... Toy with the current module._

maximbaz:
@lasers I've just found something out... if I define `on_click 1 = ...` in `.i3status.conf`, this does not override `on_click` in the module itself — do you know if this is intentional? I assume it is, but I didn't know this.

Because of this behavior, of course it doesn't make any sense to have notifications be enabled by default, it is very confusing to define a custom `on_click` and still see notifications popping up. I think now the code now matches what you and I want, we don't need to wait for the third opinion anymore 🙂

lasers:
>do you know if this is intentional? I assume it is, but I didn't know this.

I don't know if it is intentional. @tobes is much more qualified to answer this than I am. Also, maybe it had something to do with https://github.com/ultrabug/py3status/issues/1270's PR. I can try to look around maybe later.

Also, I recall @tobes was working on something to enhance button functionality so we can specify functions too... for more customization and/or richer experience. Overriding functions could be one of them. Also, it's possible that they both have same name and had nothing to do with each others.

maximbaz:
It most probably is intentional, because `on_click()` function is called for all button clicks, while `on_click 1 = ...` in `.i3status.conf` is there to define command only for one specific button, it would have been bad if that completely shadowed the entire `on_click()` function possibly disabling other buttons.

So, I'm happy with merging it as it is, have another look on your side if it looks like you want as well and let's move forward 😉

py3status/modules/external_script.py
```diff
@@ -21,7 +22,12 @@
         (default False)
 
 Format placeholders:
-    {output} output of script given by "script_path"
+    {line} number of lines in the output
+    {output} first line of the output of script given by "script_path"
+
+Format notification placeholders:
+    {line} number of lines in the output
+    {output} first line of the output of script given by "script_path"
```
maximbaz: Fixed documentation here, `{output_full}` still remains undocumented as per your request.

cyrinux: Why don't document this?

maximbaz:
A friendly ping @lasers, anything else you want to see implemented before this can get merged? I think it currently does as you wanted it to work.

lasers:
I hold off on this mainly because I think we should try `external_script: new module from scratch` to address #1265 too and that `format_notification` only work with last notification... and it only works when we clicks too... instead of normal notifications, ie `battery_level`, `pomodoro`, `wwan`.

We need to make another config `notification` (state) to better support `current` use cases in `external_script` even if it's only one placeholder right now... number of lines...

This would be similar to the new `state` config in `do_not_disturb` which is implemented nicely and not hardcoding `format_notification` to something like `do_not_disturb`'s last state.

---


Yes. `format_notification` is ideally implemented, but I'm going to wait until we get `notification` config too with the following options because it's better to wait until we support this too... instead of pushing merge button and forget about it immediately. :slightly_smiling_face: 

* `current`: Display notifications based on formatted placeholders (and not duplicate notifications, but only if something changed in the output between last output and current output... and also respecting formatted line, eg `\?if=line>750 hi the compilation may be almost done`, send a notification. This would behave same as `wwan`.

* `last`: Simply show the notification of last output and/or formatted placeholders without refreshing.

This PR deals with your `last` usecase. I want this to work with `current` usecase too.

maximbaz:
OK if you want to play with rewriting `external_script`, I would be happy to test, please just don't forget about it, I don't want to use my own fork for the next half a year 😉

py3status/modules/external_script.py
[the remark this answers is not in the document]
tobes: why is this feature hidden?  It should either be documented or removed imho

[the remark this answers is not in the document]
lasers: New `{output_full}` is the true `{output}`.
... and `{output}` is the true `{line_1}` since it does `output[0]` in the code.

We keep `{output_full}` output as-is... for notifications, but will remove newlines for the bar (although undocumented).

And we should address #1265 PR too because I assume `async_script` could want `{output_full}` too... so maybe we should address all of this in one swift PR... or something like that.

Most users are using `{output}` so we can't swap that. If we have to break something, I prefer we do it once instead of twice, thrice.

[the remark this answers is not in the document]
maximbaz: For this one shift PR, you already have a prototype for rewritten external_script, right?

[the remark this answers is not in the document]
lasers: I did start something particularly to deal with explicit lines and to parse py3status's format in the output lines too instead of external_script's second line hidden color... I don't have `Popen` in there, and I pretty much stopped working on that already. See what @tobes want to do here.

@tobes He need `{output_full}` for his notifications and `format_notifications` are made to work with `last` (data), not `current` data like most modules... so I wanted to add `notification` (state) where users can specify what notification state to use (current, last) (default: current).

lasers:
I implemented `notification_state`. This option will accept `current` or `click`. Is `notification_state` a good config name to use? Alternative name could be just `state` or `notification`. Anyway, this `state` part is done... Test this with your script and hopefully we can get this PR merged soon enough next.

py3status/modules/external_script.py
lasers: Documented. :sob:

tobes:
> I implemented notification_state. 

This just makes the PR too complicated imho, also there is no commentary about why you add this.

can we just do the 'one thing == one PR'

lasers:
@tobes `format_notification` is (often) used for current refreshes. @maximbaz have a different goal of printing the notification about last data without triggering a refresh. This can confuse users about `format_notification`'s purpose.

I introduce `notification_state = 'current'` to be consistent with `format_notification`'s behavior in that it only works with current data... and still be able to do what @maximbaz want to do here. 

If I take it out, then it will only work with last data without refreshing. If we need to do 'one thing == one PR', then we need @maximbaz to forgo his goal first and implement this with current data.

maximbaz:
Having `notification_state = 'click'` is fine for me, @lasers do you still plan to re-write this module or you are OK with this PR?

tobes:
So for me I have a few thoughts

a) this seems over-engineered for some thing that should be fairly simple

b) I think that the  on_click is being abused

c) the code is becoming complicated with no explanatory comments `self.init` is a `dict`?

d) it is overly complicated for the user

e) this changes the behaviour in a way that could upset some users.  eg I have a script I refresh by middle clicking, suddenly I have notifications appearing when I refresh - wtf

As I understand we have two issues:

1) we want to show the output to the user

2) we do not want to re-run long running scripts (or maybe we want to specifically see the given output without re-running for some other reason)

For the notification (1)

`button_notify` for notifying the output

this feels clean simple and the user can decide

for (2) this is more complicated we have options

i. `button_notify_refresh` and `button_notify_no_refresh` instead of `button_notify` the names are bad but hopefully you get the idea.  easy for the user and gives them options

ii. `refresh_on_notify` option.  Less flexible but maybe simpler to comprehend

Overall I think this reveals issues with our on_click which currently is limited and I think could be changed to make this sort of problem much more trivial to implement in a module whilst being simple from a users perspective.  

Given the current situation I think the best approach is a `button_notify`, `refresh_on_notify` combination and I'd just notify the output (trimmed to some length to stop too much crazy)

maximbaz:
> e) this changes the behaviour in a way that could upset some users. eg I have a script I refresh by middle clicking, suddenly I have notifications appearing when I refresh - wtf

This doesn't happen, see line 168 — I also refresh modules with middle click. The point of not introducing `button_refresh` is that right now it works great out of the box and doesn't conflict with any existing scenario, so no need to introduce a useless option.

`refresh_on_notify` is also useless because you can just do middle click for refresh.

py3status/modules/external_script.py
```diff
@@ -13,23 +13,52 @@
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
+    format_notification: see placeholders below (default None)
     localize: should script output be localized (if available)
         (default True)
+    notification_state: specify notification state to use: current, click
+        (default 'current')
```
maximbaz: @lasers I would suggest to make `click` the default here, right now notifications are appearing after each refresh (by default every 15 seconds) and I don't think this is a good default. 

Also, right now it's not possible to have notifications appearing _both_ after refresh and on click. I would suggest to replace `notification_state` with two new boolean options, `notify_after_refresh` (default `False`) and `notify_after_click` (default `True`)

lasers: I agree `current` wasn't a good default. I made a new state `changed` that will print notifications only if it is changed. This would be a better default to use too. However, I added+consolidated more notification-related things so it is going to be an empty (nested) dictionary for default.

With this new nested dictionary, users can specify what notification should happen on `current`, `changed`, and `click` states along with (all) options: Levels, icons, message body or titles.

Although it may be over-engineered, I feel this approach is more clean than adding boolean + button configs that might not convey information well. Please check it out.

If you want to use all three states, you can... Just add them all in the dictionary, eg.
```python
    notifications = {
        'click': {'title': 'Click', 'msg': 'Clicking... {output_full}'},
        'current': {'title': 'Current', 'msg': 'Current... {output_full}'},
        'changed': {'title': 'Changed', 'msg': 'Changed... {output_full}'},
    }
```
If there are any other states that we can add, we should be able to add that too... rather than creating more new boolean + button configs.

maximbaz: it's awesome now 👍

maximbaz:
@lasers just want to confirm with you, since you were driving implementation of `notifications` in this PR, are you interested in taking on the work of moving this to core? As far as I understood, the idea is that `notifications` object will be supported by all modules simultaneously, and all that will remain in this PR is the addition of `{output_full}` format placeholder.

py3status/modules/external_script.py
```diff
@@ -96,10 +178,28 @@ def external_script(self):
...
+        self.script_data['output_full'] = output_full
+        if self.notification['normal']:
+            self._notify_user()
+
         return response
 
+    def on_click(self, event):
+        button = event["button"]
+        if button != self.button_refresh:
+            if self.notification['click']:
```
maximbaz: Just a reminder for future self: this throws exception if `notification` object doesn't contain `click` key

lasers: What configs? I'm not able to reproduce. The `notification` dict should already have all three.

https://github.com/ultrabug/py3status/blob/c6a1c2146294ed1c90ba7153df3c9fb41d130fe9/py3status/modules/external_script.py#L111-L114

maximbaz: This is my exact config:

```
order += 'external_script date'

external_script date {
  cache_timeout = 60
  script_path = 'date +"%a, %d %b"'
  format = 'Calendar: {output}'
  on_click 1 = 'exec gsimplecal'
}
```

Add this, then restart i3 (important), then click on the module.

I get an error, and in `journalctl` I see

```
py3status[12014]: on_click event in `external_script date` failed (KeyError) external_script.py line 199.
py3status[12014]: on_click event in `external_script date` failed (KeyError) external_script.py line 199. Please try to fix this and reload i3wm (Mod+Shift+R)
```

lasers:
@maximbaz I'm not sure what to do here. It might be more fun for you to try it first. I have my py3status things that I want to do first since they're all finished or nearly finished.

I think it would not be supported by all modules simultaneously and would behave much similar to `thresholds` where we have to add this to each modules that wants it. Wild guess, we need two functions... First to validate what we need (for `self.notifications`) in `post_config_hook`, then second to ... put them in good places... similar to `threshold_get_color`. I can give it a shot eventually.

py3status/modules/external_script.py
```diff
@@ -52,20 +99,55 @@ class Py3status:
...
     script_path = None
     strip_output = False
 
     def post_config_hook(self):
         if not self.script_path:
             raise Exception(STRING_ERROR)
 
+        self.button_refresh = 2
+        self.notification = {'normal': []}
+        if self.notifications:
```
maximbaz: @lasers could you describe what is happening in this `if` block?

lasers: `post_config_hook` checking to find out which notifications to use... so we use a dict/list of booelans instead of constantly testing the conditionals on each interval.

maximbaz: I'm looking at this right now and I think it's a bit over-engineered, both external_script and battery_level (and potentially updates) only need notifications on click (battery has extra special case of showing notification on certain threshold), is there any module that currently shows notifications always or when changed?

lasers: IIRC I think battery's implementation of showing notification on clicks is a bit silly. We likely could make it show up on the bar instead. Maybe with a toggle to show/hide what we want.

Current and changed is similar. I'm not even sure if it makes sense to make a `changed` state... because we technically could make `changed` placeholder so you do `\?if=changed {test}` in current state. Maybe better not to. Fleshing out a nice notification solution is hard though.

* `wwan` on changed.
* `github` on always (always failing). Could be in `post_config_hook` instead.
* `battery_level` is both click and changed.
`pomodoro` is changed when timer reached zero. Not sure how we would add into this as it's not `changed`.

Things to take from this is that we are only making notification states to allow different kind of notifications to happen in a module. I guess we could use `changed` here.

* `graphite` seems to be always.

maximbaz: * `wwan` on changed - 👍 
* `github` is not always, it's a weird way of showing error, completely custom use of notifications and should not be generalized
* `battery_level` is click and on certain thresholds, this is again custom and is not "changed" (i.e. it will not show notification when changing from 81% to 80%).
* `pomodoro` yes, it's again not changed but some custom logic.

I'm not sure how much we will be able to generalize if every module has some special use of notifications...

py3status/modules/external_script.py
```diff
@@ -52,20 +99,55 @@ class Py3status:
...
             raise Exception(STRING_ERROR)
 
+        self.button_refresh = 2
+        self.notification = {'normal': [], 'click': False}
+        if self.notifications:
+            for x in ['changed', 'click', 'current']:
+                self.notification[x] = self.notifications.get(x, False)
+                if self.notification[x]:
+                    if x in ['changed', 'current']:
+                        self.notification['normal'].append(x)
```
maximbaz: Still cannot grasp this logic, if notification is added for "changed", we add it to "normal"? What is normal?

lasers: With only two, it would be just `if current or changed:`.
With only 10 or 100, it would be just `if current or changed or .. or .. or .. or .. False:`

I add `"True"` to the list and we check the list itself to see if it's empty or not... instead of going through the list to find a first `"True"`. When we go in, we now have (already optimized) list.

It's a bit different here because of notifications so we check both `if` statements. In other scenarios, we might use that list to loop through something.

Normal is both `current+changed`, but I don't want to say `current`. Idk.

maximbaz:
This thread became too complicated and it's hard to get feedback on it. Let's start again, with a simple patch that just solves my need: #1585

PR #1453, opened 2018-08-23
docs: update formater and configuration documentation

L0ric0:
Added a formater section in the Configuration manual as users should not have to search the py3-modul documentation to format their status bar. Also updated the save_format doc string to include the conditional syntax of the formater.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -220,6 +220,79 @@ Some modules may allow more than one threshold to be defined.  If all the thresh
         }
     }
 
+Formater
+--------
+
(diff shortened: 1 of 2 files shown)


doc/configuration.rst
```diff
@@ -220,6 +220,79 @@ Some modules may allow more than one threshold to be defined.  If all the thresh
         }
     }
 
+Formater
```
ultrabug: I believe this is spelled `Formatter`

ultrabug:
Thanks a lot for your first contribution @L0ric0 !!

PR #1455, opened 2018-08-26
allow us to get params from formats and also update them

tobes:
So this is an improvement on #1333 that actually cares about the format rather than randomly guessing what might be important.

I've added code to allow the `update_placeholders()` functionality to also update used params as that makes sense but is most of the code.

I added some testing for these functions too but they are fairly basic but good to have


ultrabug:
two :heart: from lasers, I guess my only option is to merge :+1:

PR #1458, opened 2018-08-26
New module: MEGA sync

maximbaz:
Because Dropbox decided to abandon Linux users in November, it's time to migrate somewhere. I chose MEGA 😛 

MEGA allows configuring sync for multiple folders, here's the output of `mega-sync`:

```
❯ mega-sync 
ID LOCALPATH      REMOTEPATH     ActState   SyncState     SIZE  FILES   DIRS
 0 ~/MEGA         /              Active     Synced     4.45 GB   3275    678 
```

What this module does is that it parses every line and creates a bunch of format placeholders for every line, such as `LOCALPATH_0`, `SyncState_0`, etc.

I decided not to do what Dropbox module does in terms of `status_*` configs, instead just parse the fields and let people use formatter. Personally, I want to hide the module when everything is synced, so I included the commit that adds this functionality to formatter.

```
format = "[\?if=SyncState_0!~Synced&color=degraded MEGA: {SyncState_0}]"
```

### DIFF
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-Display status of MEGA sync.
+Display status of MEGAcmd.
 
 Configuration parameters:
     cache_timeout: refresh interval for this module (default 10)
@@ -41,7 +41,7 @@ class Py3status:
         if not self.py3.check_commands("mega-sync"):
             raise Exception(STRING_NOT_INSTALLED)
 
-    def megasync(self):
+    def mega(self):
         output = self.py3.command_output("mega-sync").splitlines()
         columns = output[0].split()
 
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -45,7 +45,7 @@ class Py3status:
     def mega(self):
         output = self.py3.command_output("mega-sync").splitlines()
 
-        format_sync = ""
+        format_sync = None
         if len(output) > 0:
             columns = output[0].lower().split()
             megasync_data = []
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -5,7 +5,7 @@ Display status of MEGAcmd.
 Configuration parameters:
     cache_timeout: refresh interval for this module (default 10)
     format: display format for the module (default "MEGA {format_sync}|No MEGA")
-    format_sync: display format for every sync (default "{SyncState}")
+    format_sync: display format for every sync (default "{syncstate}")
     format_sync_separator: show separator if more than one sync (default " ")
 
 Format placeholders:
@@ -35,7 +35,7 @@ class Py3status:
     # available configuration parameters
     cache_timeout = 10
     format = "MEGA {format_sync}|No MEGA"
-    format_sync = "{SyncState}"
+    format_sync = "{syncstate}"
     format_sync_separator = " "
 
     def post_config_hook(self):
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -25,6 +25,7 @@ SAMPLE OUTPUT
 """
 
 STRING_NOT_INSTALLED = "MEGAcmd is not installed"
+STRING_NOT_CONFIGURED = "no sync is configured yet"
 
 
 class Py3status:
@@ -43,8 +44,10 @@ class Py3status:
 
     def mega(self):
         output = self.py3.command_output("mega-sync").splitlines()
-        columns = output[0].split()
+        if len(output) == 0:
+            self.py3.error(STRING_NOT_CONFIGURED)
 
+        columns = output[0].split()
         megasync_data = []
         for line in output[1:]:
             cells = dict(zip(columns, line.split()))
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -48,7 +48,7 @@ class Py3status:
         if len(output) == 0:
             self.py3.error(STRING_NOT_CONFIGURED)
 
-        columns = [col.lower() for col in output[0].split()]
+        columns = output[0].lower().split()
         megasync_data = []
         for line in output[1:]:
             cells = dict(zip(columns, line.split()))
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -12,7 +12,8 @@ Format placeholders:
     {format_sync} Format for every sync returned by 'mega-sync' command.
 
 format_sync placeholders:
-    Any column returned by 'mega-sync' command, e.g.: ID, SyncState, LOCALPATH
+    Any column returned by 'mega-sync' command - in lower case!
+    For example: id, syncstate, localpath
 
 Requires:
     MEGAcmd: command-line interface for MEGA
@@ -47,7 +48,7 @@ class Py3status:
         if len(output) == 0:
             self.py3.error(STRING_NOT_CONFIGURED)
 
-        columns = output[0].split()
+        columns = [col.lower() for col in output[0].split()]
         megasync_data = []
         for line in output[1:]:
             cells = dict(zip(columns, line.split()))
--- a/py3status/modules/mega.py
+++ b/py3status/modules/mega.py
@@ -4,7 +4,7 @@ Display status of MEGAcmd.
 
 Configuration parameters:
     cache_timeout: refresh interval for this module (default 10)
-    format: display format for this module (default "MEGA: {format_sync}")
+    format: display format for the module (default "MEGA {format_sync}|No MEGA")
     format_sync: display format for every sync (default "{SyncState}")
     format_sync_separator: show separator if more than one sync (default " ")
 
@@ -22,11 +22,10 @@ Requires:
 @license BSD
 
 SAMPLE OUTPUT
-{'full_text': 'MEGA: Synced'}
+{'full_text': 'MEGA Synced'}
 """
 
 STRING_NOT_INSTALLED = "MEGAcmd is not installed"
-STRING_NOT_CONFIGURED = "no sync is configured yet"
 
 
 class Py3status:
@@ -35,7 +34,7 @@ class Py3status:
 
     # available configuration parameters
     cache_timeout = 10
-    format = "MEGA: {format_sync}"
+    format = "MEGA {format_sync}|No MEGA"
     format_sync = "{SyncState}"
     format_sync_separator = " "
 
@@ -45,17 +44,17 @@ class Py3status:
 
     def mega(self):
         output = self.py3.command_output("mega-sync").splitlines()
-        if len(output) == 0:
-            self.py3.error(STRING_NOT_CONFIGURED)
 
-        columns = output[0].lower().split()
-        megasync_data = []
-        for line in output[1:]:
-            cells = dict(zip(columns, line.split()))
-            megasync_data.append(self.py3.safe_format(self.format_sync, cells))
+        format_sync = ""
+        if len(output) > 0:
+            columns = output[0].lower().split()
+            megasync_data = []
+            for line in output[1:]:
+                cells = dict(zip(columns, line.split()))
+                megasync_data.append(self.py3.safe_format(self.format_sync, cells))
 
-        format_sync_separator = self.py3.safe_format(self.format_sync_separator)
-        format_sync = self.py3.composite_join(format_sync_separator, megasync_data)
+            format_sync_separator = self.py3.safe_format(self.format_sync_separator)
+            format_sync = self.py3.composite_join(format_sync_separator, megasync_data)
 
         return {
             "cached_until": self.py3.time_in(self.cache_timeout),
--- a/py3status/modules/mega_sync.py
+++ b/py3status/modules/mega_sync.py
@@ -7,7 +7,8 @@ Configuration parameters:
     format: display format for this module (default "MEGA: {SyncState_0}")
 
 Format placeholders:
-    <column>_<ID> where <column> and <ID> are returned by 'mega-sync' command.
+    <column>_<ID> where <column> is the names returned by 'mega-sync' command.
+    For example: SyncState_0, LOCALPATH_3, ActState_7
 
 Requires:
     MEGAcmd: command-line interface for MEGA
@@ -16,7 +17,7 @@ Requires:
 @license BSD
 
 SAMPLE OUTPUT
-{'color': '#00FF00', 'full_text': 'MEGA: Synced'}
+{'full_text': 'MEGA: Synced'}
 """
 
 
@@ -33,24 +34,18 @@ class Py3status:
             raise Exception("MEGAcmd is not installed")
 
     def mega_sync(self):
-        try:
-            output = self.py3.command_output("mega-sync").splitlines()
-        except self.py3.CommandError as e:
-            error = e.output or e.error
-            self.py3.error("error: " + error)
-
-        format_data = {}
+        output = self.py3.command_output("mega-sync").splitlines()
         columns = output[0].split()
+        megasync_data = {}
         for line in output[1:]:
             cells = dict(zip(columns, line.split()))
-            suffixed_cells = {
-                key + "_" + cells["ID"]: value for key, value in cells.items()
-            }
-            format_data.update(suffixed_cells)
+            megasync_data.update(
+                {key + "_" + cells["ID"]: value for key, value in cells.items()}
+            )
 
         return {
             "cached_until": self.py3.time_in(self.cache_timeout),
-            "full_text": self.py3.safe_format(self.format, format_data),
+            "full_text": self.py3.safe_format(self.format, megasync_data),
         }
 
 
--- a/py3status/modules/mega_sync.py
+++ b/py3status/modules/mega_sync.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+"""
+Display status of MEGA sync.
+
+Configuration parameters:
+    cache_timeout: refresh interval for this module (default 10)
+    format: display format for this module (default "MEGA: {SyncState_0}")
+
+Format placeholders:
+    <column>_<ID> where <column> and <ID> are returned by 'mega-sync' command.
+
+Requires:
+    MEGAcmd: command-line interface for MEGA
+
+@author Maxim Baz (https://github.com/maximbaz)
+@license BSD
+
+SAMPLE OUTPUT
+{'color': '#00FF00', 'full_text': 'MEGA: Synced'}
+"""
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    cache_timeout = 10
+    format = "MEGA: {SyncState_0}"
+
+    def post_config_hook(self):
+        if not self.py3.check_commands("mega-sync"):
+            raise Exception("MEGAcmd is not installed")
+
+    def mega_sync(self):
+        try:
+            output = self.py3.command_output("mega-sync").splitlines()
+        except self.py3.CommandError as e:
+            error = e.output or e.error
+            self.py3.error("error: " + error)
+
+        format_data = {}
+        columns = output[0].split()
+        for line in output[1:]:
+            cells = dict(zip(columns, line.split()))
+            suffixed_cells = {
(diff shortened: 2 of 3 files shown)


py3status/formatter.py
```diff
@@ -339,6 +339,14 @@ def __init__(self, info):
             self.variable, self.value = info.split('<')
             self.condition = '<'
             self.check_valid = self._check_valid_condition
+        elif '!~' in info:
+            self.variable, self.value = info.split('!~')
+            self.condition = '!~'
+            self.check_valid = self._check_valid_condition
```
lasers: `!~` is what I want to add too, but I don't know if that'll be accepted.

Without this, we have to use `'\?if=!long_str~long Oh my... What a long string.'` instead.

I sometimes used this by accident... `\?if!=number_of_updates` instead of `\?if=!number_of_updates`.

py3status/modules/mega_sync.py
```diff
@@ -0,0 +1,63 @@
...
+        for line in output[1:]:
+            cells = dict(zip(columns, line.split()))
+            suffixed_cells = {
+                key + "_" + cells["ID"]: value for key, value in cells.items()
+            }
+            format_data.update(suffixed_cells)
+
+        return {
+            "cached_until": self.py3.time_in(self.cache_timeout),
+            "full_text": self.py3.safe_format(self.format, format_data),
```
lasers: Can we rename `format_data` to `mega_data`? Nothing to do with `format`.

py3status/modules/mega_sync.py
```diff
@@ -0,0 +1,63 @@
...
+            self.py3.error("error: " + error)
+
+        format_data = {}
+        columns = output[0].split()
+        for line in output[1:]:
+            cells = dict(zip(columns, line.split()))
+            suffixed_cells = {
+                key + "_" + cells["ID"]: value for key, value in cells.items()
+            }
+            format_data.update(suffixed_cells)
```
lasers: You can consolidate this a bit more `... data.update({` above to kill a line.

py3status/modules/mega_sync.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    cache_timeout = 10
+    format = "MEGA: {SyncState_0}"
```
lasers: Your `SAMPLE OUTPUT` is different from this format.

Can we change format to something like this...
* `'MEGA [\?color=??? {SyncState_0}]'`
* `'Mega [\?color=??? {SyncState_0}]'`

I'm attempting to create a consistent coloring scheme in modules. Start off with default nocolor title, secondary darkgray, then colorized data. It looks like no secondary darkgray is needed here. `lm_sensors` need it.

Can you use `thresholds` to colorize this module or do you need string thresholds too?

maximbaz: I don't think any thresholds can be used here...? 

Fine if I fix `SAMPLE_OUTPUT` to be colorless? We could make a default format that bases color on the value of `SyncState_0`, but it has higher chances to be broken in the future, e.g. what if `mega_sync` starts to localize output?

py3status/modules/mega_sync.py
```diff
@@ -0,0 +1,63 @@
...
+    def post_config_hook(self):
+        if not self.py3.check_commands("mega-sync"):
+            raise Exception("MEGAcmd is not installed")
+
+    def mega_sync(self):
+        try:
+            output = self.py3.command_output("mega-sync").splitlines()
+        except self.py3.CommandError as e:
+            error = e.output or e.error
+            self.py3.error("error: " + error)
```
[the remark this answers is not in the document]
maximbaz: Apparently not, removed 👍

py3status/modules/mega_sync.py
maximbaz: Sure 🙂

py3status/modules/mega_sync.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    """
+
+    # available configuration parameters
+    cache_timeout = 10
+    format = "MEGA: {SyncState_0}"
+
+    def post_config_hook(self):
+        if not self.py3.check_commands("mega-sync"):
+            raise Exception("MEGAcmd is not installed")
```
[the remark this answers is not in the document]
maximbaz: I was thinking about naming it `mega`, but then there are a bunch of other tools in MEGAcmd (`mega-backup`, `mega-speedlimit`, `mega-webdav`, etc.), I can easily envision people creating more py3status modules for those specific actions. And I don't think one huge module would be a good fit for all these tools, what they do is very very different. So let's call it `megasync` then.

py3status/modules/mega_sync.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+"""
+Display status of MEGA sync.
+
+Configuration parameters:
+    cache_timeout: refresh interval for this module (default 10)
+    format: display format for this module (default "MEGA: {SyncState_0}")
+
+Format placeholders:
+    <column>_<ID> where <column> and <ID> are returned by 'mega-sync' command.
```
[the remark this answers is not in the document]
maximbaz: Added, is that what you wanted?

py3status/formatter.py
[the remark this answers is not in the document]
lasers: It can be accepted... It's still a basic formatter. We need to decide what is good and what is not good. :slightly_smiling_face:

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
lasers: I'm with `mega_sync`, `megasync`, `mega`. Just wanted to point this out. :+1:

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
lasers: Yes. Still unclear to us who don't use `megasync`. I'll have to install this to understand what exactly the available placeholders are.... Thanks for adding. :+1:

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
maximbaz: OK, extracted the constant

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
maximbaz: Yeah, I guess this is the case, you will run `mega-sync`, see the output including the values specific to your own installation of megasync and then you will see which values you want to see on py3status.

Here's mine, by the way, I have one row with ID=0, so everything is suffixed with `_0`:

```
❯ mega-sync 
ID LOCALPATH      REMOTEPATH     ActState   SyncState     SIZE  FILES   DIRS
 0 ~/MEGA         /              Active     Synced     4.45 GB   3275    678 
```

py3status/modules/mega_sync.py
lasers: We have `command_output(..... localized=False)` so it should be forced to use English. :wink: 

I'm trying to have colors in default modules so they all don't come out plain boring. See `timewarrior` (new module). It defaults with `degraded` color. I think it's much easier for users to remove colors than to add colors.

Side note: With `timewarrior`, you can do `timew start py3status` to see tag (in secondary darkgray color). If you're going to use this for real, you need to specify filter too, or everything gets shown eventually.

maximbaz: Yeah but I don't think people would want me to make their mega-sync py3status forcefully use English, since we run command once and the output of it is what will be shown on the panel 😉 

I would be curious how many people actually use default `format`, maybe what I do is unusual but I _always_ want to change something 😄 

Do you like this for a default?

```
format = "MEGA: [\?if=SyncState_0~Syncedd&color=good {SyncState_0}|[\?color=degraded {SyncState_0}]]"
```

lasers: You really don't need to use `~` here? Does `=Synced` work fine here? I don't have `mega-sync`, but I'm seeing a different issue. Will type it out soon. I'd kill colon off `MEGA`.

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
lasers: Can you add more paths to sync? I think this calls for looping with `format_sync` or `format_path` instead of having you populating everything in one `format`. That explains why you used numbered prefixes. See `nvidia_smi`. It does basically what you want to do here.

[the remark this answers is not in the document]
maximbaz: Exactly, yeah that's what I wanted, thanks

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
lasers: >but then there are a bunch of other tools in MEGAcmd (mega-backup, mega-speedlimit, mega-webdav, etc.), I can easily envision people creating more py3status modules for those specific actions. And I don't think one huge module would be a good fit for all these tools,

If this is true... then we could name it `mega` because if you're making `format_sync` for your `mega-sync`, the other tools can be `format_speedlimit`, `format_backup`, `format_webdev`, etc... just the same.

Users would have to specify what they want to use in the `format` just same as what you're working right now... to put `{format_sync}` in `format`  instead of numbered sync placeholders.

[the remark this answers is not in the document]
maximbaz: Now it makes sense, renamed again

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
maximbaz: Done now. I have one issues though, with this negative equals format doesn't seem to work...

What am I missing @lasers?

```
format_sync = "[\?!if=SyncState=Synced {SyncState}]"
```

I tried this:

```
format_sync = "[\?if=ID=0 {SyncState}]"
```

And it correctly shows me one of two entries, but if I put `!` before `if`, then I see two entries. Why?

```
format_sync = "[\?!if=ID=0 {SyncState}]"
```

py3status/formatter.py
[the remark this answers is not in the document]
tobes: `~` is a bit operator so I'm not sure that abusing it is a good thing.  Also it isn't needed directly by the module so maybe should be a separate PR

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
lasers: I'm not sure if it's good to keep keys as-is. Maybe lower the line before you make column keys.
```
ID LOCALPATH      REMOTEPATH     ActState   SyncState     SIZE  FILES   DIRS

... to... 

id, localpath, remotepath, actstate, syncstate, size, files, dirs
```
We don't have `[\?!if=`. We have `[\?if=!`. Try that.

py3status/formatter.py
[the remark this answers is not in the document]
maximbaz: Yup, thanks, already reverted as soon as I learned about `=` 😉

py3status/modules/mega_sync.py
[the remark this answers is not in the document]
maximbaz: Keeping keys as-is is easy for copy-paste, requiring users to do post-processing will only lead for typos and wasted time... Let's keep them as is, we can always complain on MEGAcmd github that their headers are ugly and inconsistent if it really bothers you 😉 

> We don't have [\?!if=. We have [\?if=!. Try that.

Yeah, worked, thanks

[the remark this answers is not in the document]
lasers: It does not bother me. What bother me is that we can break formats if we don't try and lowercase this... @tobes might disagree with keeping this as-is too.

We try to make our placeholders lowercase if possible. Otherwise, it's stupid to use `${REMOTEPATH}` because of their output... What if they change it later to `${RemotePath}` then we break lot of people's format just the same... so it's generally safe/better/more consistent if we keep or make all things lowercase.

Do you want to pop over there and ask them why their headers aren't inconsistent? :wink:

py3status/modules/mega.py
```diff
@@ -47,7 +48,7 @@ def mega(self):
         if len(output) == 0:
             self.py3.error(STRING_NOT_CONFIGURED)
 
-        columns = output[0].split()
+        columns = [col.lower() for col in output[0].split()]
```
lasers: `columns = output[0].lower().split()` is better.

maximbaz: Indeed

py3status/modules/mega.py
```diff
@@ -0,0 +1,74 @@
...
+    format_sync_separator = " "
+
+    def post_config_hook(self):
+        if not self.py3.check_commands("mega-sync"):
+            raise Exception(STRING_NOT_INSTALLED)
+
+    def mega(self):
+        output = self.py3.command_output("mega-sync").splitlines()
+        if len(output) == 0:
+            self.py3.error(STRING_NOT_CONFIGURED)
```
lasers: I think it's wrong to report this  if `mega-sync` is not throwing an error. At this point, `mega-sync` know users are logged in with nothing configured... so it's not really an error. It's your custom error.  We can fix this module a little differently to be more gentle with users and not complaining about this part.

* Users are not signed in. Errors are thrown.
* Users are signed in, but had nothing configured. Be quiet. Show `No MEGA`.
* Users are signed in, with something configured. Show `MEGA {format_output}`

lasers: See `timewarrior`. Example `format = '[MEGA {format_time}]|No MEGA'`

py3status/modules/mega.py
```diff
@@ -0,0 +1,74 @@
...
+    format: display format for this module (default "MEGA: {format_sync}")
+    format_sync: display format for every sync (default "{SyncState}")
+    format_sync_separator: show separator if more than one sync (default " ")
+
+Format placeholders:
+    {format_sync} Format for every sync returned by 'mega-sync' command.
+
+format_sync placeholders:
+    Any column returned by 'mega-sync' command - in lower case!
+    For example: id, syncstate, localpath
```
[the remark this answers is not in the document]
maximbaz: I purposefully don't want to list placeholders because this module doesn't rely on them or control them, people shouldn't rely on this description because it will quickly be outdated. If a new column is added in upstream, or an existing column is updated, our own documentation will get out of sync and people will be confused and will be filing issues. 

It is not that hard to execute `mega-sync` in terminal and get up-to-date list of columns, and it is documented so it's not like people don't know where to get the available placeholders 😉 

If other modules list placeholders that they don't really rely on and control, they should be fixed too and explain to users how to get up to date info.

py3status/modules/mega.py
maximbaz: Hmm, but why a string "No MEGA" in successful color is better than a string "you don't have sync, configure one" in an error color? The latter one draws your attention and specifically describes what you need to do to get the module working, "No MEGA" would have been a very confusing output for me.

lasers: What successful color? Default color? You view this as something important. I view this as a simple number of syncs, file paths, or updates. I think it is fine to have none.

We will throw error if users don't have this installed. After that, `mega-sync` will throw errors if users are not logged in... or when things does not go its way. That's when we can deter responsibility to `mega-sync` and not having to make custom errors... And also because we don't know how `mega-sync` work.

To reiterate...
* Not installed? We throw an error.
* Users not signed in? `mega-sync` throw an error. We deter everything here.
* Users signed in + no number of syncs... `mega-sync` does not throw an error.
* Users signed in, but no output? `mega-sync` does not throw an error, but you want to...
* Users does not have Internet, `mega-sync` does not throw an error, but your code might...

We also don't throw errors on other things such as...

* No Internet connection on `whatismyip`.
* Audio players not configured or running (iirc mpd throws error, grrr.)
* No new mails, timew intervals, tasks, updates, etc.
* et cetera

Generally, we want to go for fewer configs, fewer errors, fewer issues, etc.

Users seeing `No MEGA` can figure out their issues fast because they're smart cookies and they already run linux, i3wm, and py3status too... Also because `mega-sync` does not consider this a fatal error in first place so basically, both `mega-sync` and we are saying that users can set this up later when they want to... instead of trying to make an error on possibly something trivial.

maximbaz: OK, makes sense.

py3status/modules/mega.py
```diff
@@ -0,0 +1,73 @@
...
+STRING_NOT_INSTALLED = "MEGAcmd is not installed"
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    cache_timeout = 10
+    format = "MEGA {format_sync}|No MEGA"
```
lasers: Ha. Did you test this? I think this will come out as `'MEGA '`.

`'[MEGA {format_sync}|No MEGA]'` should be what makes the condition work.

Speaking of `'MEGA '`, I think `format_sync = ""` is a bad approach because it will (generally) hide things
from users like this example I'd rather we either pass `None` or we don't pass anything at all to the dict... so we see stray `{format_sync}` instead.

* `'MEGA '`
* `'MEGA None'`
* `'MEGA {format_sync}'`  <--- my preference, more clean.

maximbaz: I did test, it works as expected. Changed to None instead of empty string, behaves in the same way, but if you prefer it this way, let it be so.

maximbaz:
Friendly ping 😉

maximbaz:
Another friendly ping 😉

py3status/modules/mega_sync.py
lasers: > I don't think any thresholds can be used here...?
> 
> Fine if I fix `SAMPLE_OUTPUT` to be colorless? We could make a default format that bases color on the value of `SyncState_0`, but it has higher chances to be broken in the future, e.g. what if `mega_sync` starts to localize output?

We have string thresholds now. Do you want to update this module? :smiley_cat:

Side note: Defaulting `Mega Synced Synced` could look horrible (if multiple ids).

I don't use `mega`*, but it looks like `format_sync` could use new threshold (and still print something else than `Synced`... Maybe `format_sync = '\?color=syncstate {size}'` is nice than always printing `Synced`.

py3status/modules/mega.py
lasers: >I did test, it works as expected.

It did not work before (twas tested), but now it works... because of recent revert 5fdb2b3d.

py3status/modules/mega_sync.py
maximbaz: Looks like I'm the only user of the module so far, and I don't need thresholds (my module is hidden when State == "Synced"), mind if we postpone this to the time when there is someone who actually needs thresholds?

Shall I change the default value of `format_sync` to be `{size}`?

py3status/modules/mega.py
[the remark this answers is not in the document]
lasers: It's same with `nvidia_smi` for me, but I make a list of (some) placeholders users might like...

>It is not that hard to execute mega-sync in terminal and get up-to-date list of columns

Also, I had to scroll up and down this issue to find a placeholder because I was looking for that one placeholder you mentioned somewhere... It was `{size}`.

It's nice for users not using `mega` module to be able to find a list of placeholders without installing and running `mega` command. We might have to make an account too. *cough*

Also, I want to fix this up so it'll be ready to go in. It's fine if we miss some. I didn't include other 20+ placeholders from `nvidia_smi` as it depends on what gpu users have.

>because it will quickly be outdated.

Is it still up to date? Anything changed? Headers?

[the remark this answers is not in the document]
maximbaz: I don't see why users would be interested in available placeholders of a mega module if they don't even have a mega account. You as a reviewer of the initial PR are an exception 😉 

megacmd got a new major version since the time I submitted the PR, I didn't track if the headers change or not, but as a maintainer of this module I'm happy that I don't even need to worry about it.

py3status/modules/mega_sync.py
lasers: It's 100% your module. Sure on postpone... I wanted to make (better) default formats so users might like them more (and use them more) and not having to change `format` much. I'm aware you want to keep things hidden which is fine (but could be a bad default behavior).

>Shall I change the default value of format_sync to be {size}?

That's the thing. What do you think would make a better default format? Always having Synced or some numbers on the bar? Some numbers, but with synced colors (good, degraded, bad) threshold? What about prefixing darkgray `{id}` in case there are more than one? Et cetera.

PR #1460, opened 2018-08-27
weather_owm: work without timezoneapi but also if you set city or location

tobes:
#1433 reported an issue when https://timezoneapi.io/ is blocked/unavailable.

Whilst looking at this I found that there is a lack of flexibility.  I often am interested in weather in several locations and it is nice to be able to choose one, We need this to solve the #1433 issue

we can now set either `location` or `city`

Added examples as well as some general re-factoring needed for this functionality.


tobes:
issue reporter confirms fix https://github.com/ultrabug/py3status/issues/1433#issuecomment-416432267

ultrabug:
thanks!

PR #1466, opened 2018-08-28
clock: use format parsing to find placeholder

tobes:
clock did it's own parsing which is bad so use formatter so we do it correctly


ultrabug:
that formatter thing

PR #1478, opened 2018-09-01
DO NOT MERGE: config allow base64()

tobes:
So I was looking for issues to close and found  #1211

So it is actually a nice idea so I thought I'd implement it. basically it allow us to use `base64(...)` in the config.  It works like the old base64 stuff in that it tries to keep it secret (not that it really does but it keeps things out of the logs etc).

It does allow functionality lacking before like using inside a dict in the config.

So if this is acceptable ~~I'll knock up some documentation - but I'll wait till there is a request.~~  But closing that would get us to the magic single page of issues

Also looking at the `env()` and `shell()` they could do with a clean up too, as well as unifying the code but again I'd like to get this in first.   Also depreciate the old method as tis fits in with the others

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -131,17 +131,31 @@ Store per config cache in different directories.
 
 Configuration obfuscation
 -------------------------
-
-.. note::
-    New in version 3.1
-
 Py3status allows you to hide individual configuration parameters so that they
 do not leak into log files, user notifications or to the i3bar. Additionally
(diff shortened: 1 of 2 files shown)


tobes:
I'll leave this here for now but I'd like to wait till the config parser cleanup are done and use my branch based on those #1489 #1484

ultrabug:
@tobes resolve conflicts, rename PR and I merge?

tobes:
replaced  by  #1498

PR #1482, opened 2018-09-02
config parser: add explicit exceptions

tobes:
This add exceptions to the config parser and removes `# noqa`

 `# noqa` flake8 error suppression should be used extremely sparingly and never to just hide errors.  The whole point of flake8 is to help improve code quality so ignoring it's suggestions undermines the point of using it.

I am confident that `ValueError` is the only exception thrown in the casting as I do not believe we can ever have a situation where `TypeError` is raised.

For module import a variety of exceptions can occur and so catching any feels appropriate in this case.


ultrabug:
merci

PR #1484, opened 2018-09-02
config parser: unify quote removal

tobes:
So we do quote removal in two places in the config parser.  This changes things to just do it in one place

1) rename `make_name()` to the more readable `remove_quotes()`

2) `make_value()` now uses this code rather than having it's own version

3) improve the robustness of the code we check that the first and last character are both the same quote type.


ultrabug:
pro PR

PR #1491, opened 2018-09-02
py3status lexer updates for docs

tobes:
So I was looking at the docs and things are getting a mess with the config syntax highlighting.
Things have got more complicated and the highlighting was always fairly basic.

I think the `shell(..)` and `env(..)` where the final straw.
https://py3status.readthedocs.io/en/latest/configuration.html#environment-variables

Anyhow this gives us a ~~spring~~autumn clean and makes everything super shiny once again.  Anyhow it's all fairly ugly regex stuff and seeing as I'm the only one who works on this bit I'm not going to go into much detail.  But there are some new comments (mainly for me when I revisit this in a couple of years or so)

if you are testing make sure to `make clean html` or else you won't get a clean build

No screenshots to preserve the mystery

### DIFF
--- a/py3status/autodoc.py
+++ b/py3status/autodoc.py
@@ -7,7 +7,7 @@ import re
 
 from docutils import nodes
 from docutils.parsers.rst import Directive
-from pygments.lexer import RegexLexer
+from pygments.lexer import RegexLexer, bygroups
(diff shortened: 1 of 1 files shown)


ultrabug:
Hell...

PR #1515, opened 2018-09-12
New module: Velib metropole

cyrinux:
Hi guys,

A new module for shared velib bike in Paris.

Working version.

PR following https://github.com/ultrabug/py3status/issues/1512

Waiting for comment.

ps: When pybikes will be python3 compliant (https://github.com/eskerda/pybikes/pull/284) I will do generic shared bikes modules.

😗

### DIFF
--- a/py3status/modules/velib_metropole.py
+++ b/py3status/modules/velib_metropole.py
@@ -130,9 +130,12 @@ class Py3status:
         self.optimal_area = False
         self.button_refresh = 2
 
+        self.thresholds_init = self.py3.get_color_names_list(self.format)
+
         # get placeholders
         self.placeholders = ["station_gps_latitude", "station_gps_longitude"]
         for x in [self.format, self.format_station]:
+            self.thresholds_init += self.py3.get_color_names_list(x)
             self.placeholders += self.py3.get_placeholders_list(x)
 
     def _camel_to_snake_case(self, data):
@@ -252,13 +255,6 @@ class Py3status:
             if self.station_index == 0:
                 self.station_index = 1
 
-            # thresholds TODO: FIX ME
-            for x in self.thresholds:
-                if x in self.stations_data[self.station_index]:
-                    self.py3.threshold_get_color(
-                        self.stations_data[self.station_index][x], x
-                    )
-
             # forge data output
             velib_data = {
                 "stations": self.number_of_stations,
@@ -267,6 +263,12 @@ class Py3status:
                 ),
                 "index": self.station_index,
             }
+
+        for x in self.thresholds_init:
+            if x in self.stations_data[self.station_index]:
+                self.py3.threshold_get_color(
+                    self.stations_data[self.station_index][x], x
+                )
 
         return {
             "cached_until": self.py3.time_in(cached_until),
--- a/py3status/modules/velib_metropole.py
+++ b/py3status/modules/velib_metropole.py
@@ -60,6 +60,7 @@ from re import sub
 from time import time
 
 STRING_MISSING_STATIONS = "No velib stations set"
+VELIB_ENDPOINT = "https://www.velib-metropole.fr/webapi/map/details"
 
 
 class Py3status:
@@ -117,12 +118,13 @@ class Py3status:
         # Area is then shrink to only get
         # specified stations
         self.gps_dict = {
-            "top_latitude": 49.0,
-            "top_longitude": 2.5,
-            "bottom_latitude": 48.6,
-            "bottom_longitude": 2.2,
-            "zoom_level": 15,
+            "gpsTopLatitude": 49.0,
+            "gpsTopLongitude": 2.5,
+            "gpsBotLatitude": 48.6,
+            "gpsBotLongitude": 2.2,
+            "zoomLevel": 15,
         }
+        self.toggled = False
         self.idle_time = 0
         self.request_timeout = 10
         self.station_states = {"Operative": 0, "Close": 1}
@@ -130,6 +132,7 @@ class Py3status:
         self.stations = {}
         self.optimal_area = False
         self.first_start = True
+        self.button_refresh = 2
 
         # get placeholders
         self.placeholders = []
@@ -161,30 +164,25 @@ class Py3status:
 
         latitudes = []
         longitudes = []
+
         for x in data:
             latitudes.append(float(data[x]["station_gps_latitude"]))
             longitudes.append(float(data[x]["station_gps_longitude"]))
+
         self.gps_dict.update(
             {
-                "top_latitude": max(latitudes),
-                "top_longitude": max(longitudes),
-                "bottom_latitude": min(latitudes),
-                "bottom_longitude": min(longitudes),
+                "gpsTopLatitude": max(latitudes),
+                "gpsTopLongitude": max(longitudes),
+                "gpsBotLatitude": min(latitudes),
+                "gpsBotLongitude": min(longitudes),
             }
         )
         self.optimal_area = True
 
     def _get_velib_data(self):
-        url = """https://www.velib-metropole.fr/webapi/map/details? \
-            gpsTopLatitude={top_latitude} \
-            &gpsTopLongitude={top_longitude} \
-            &gpsBotLatitude={bottom_latitude} \
-            &gpsBotLongitude={bottom_longitude} \
-            &zoomLevel={zoom_level}"""
-
         try:
             return self.py3.request(
-                url.format(**self.gps_dict), timeout=self.request_timeout
+                VELIB_ENDPOINT, params=self.gps_dict, timeout=self.request_timeout
             ).json()
         except self.py3.RequestException:
             return None
@@ -209,18 +207,14 @@ class Py3status:
                     new_station[key] = self._cast_number(value)
 
             # station_due_date_s: station due date in dateime iso format
-            if all(
-                "station_due_date" in list for list in [new_station, self.placeholders]
-            ):
+            if all("station_due_date" in list for list in [new_station, self.placeholders]):
                 station_due_date = datetime.fromtimestamp(
                     new_station["station_due_date"]
                 )
                 new_station.update({"station_due_date_s": station_due_date.isoformat()})
 
             # station_state_code: station code for thresholds
-            if all(
-                "station_state" in list for list in [new_station, self.placeholders]
-            ):
+            if all("station_state" in list for list in [new_station, self.placeholders]):
                 new_station.update(
                     {
                         "station_state_code": int(
@@ -249,14 +243,19 @@ class Py3status:
         else:
             cached_until = self.idle_time - current_time
 
-        if not refresh or self.first_start:
-            self.velib_metropole_data = self._get_velib_data()
-            if self.velib_metropole_data:
-                self.stations = self._manipulate(self.velib_metropole_data)
-                # not need to try/except i think, first run
-                # will be "slower" like first implem but cleaner?
-                if not self.optimal_area:
-                    self._set_optimal_area(self.stations)
+        if not self.toggled and not refresh:
+            self.toggled = False
+            data = self.velib_metropole_data
+        else:
+            data = self.velib_metropole_data = self._get_velib_data()
+
+        if self.first_start:
+            data = self.velib_metropole_data = self._get_velib_data()
+
+        if data:
+            self.stations = self._manipulate(data)
+            if not self.optimal_area:
+                self._set_optimal_area(self.stations)
 
         self.number_of_stations = len(self.stations) or 0
 
@@ -292,14 +291,15 @@ class Py3status:
 
     def on_click(self, event):
         button = event["button"]
-        if self.stations:
-            if button == self.button_next:
-                self.station_index += 1
-                self.station_index %= self.number_of_stations + 1
-            elif button == self.button_previous:
-                self.station_index -= 1
-                self.station_index %= self.number_of_stations + 1
-        elif button == 2:
+        self.toggled = False
+        if button == self.button_next:
+            self.station_index += 1
+            self.station_index %= self.number_of_stations + 1
+        elif button == self.button_previous:
+            self.station_index -= 1
+            self.station_index %= self.number_of_stations + 1
+        elif button == self.button_refresh:
+            self.toggled = True
             self.idle_time = 0
         else:
             self.py3.prevent_refresh()
--- a/py3status/modules/velib_metropole.py
+++ b/py3status/modules/velib_metropole.py
@@ -272,12 +272,11 @@ class Py3status:
     def on_click(self, event):
         button = event["button"]
         if self.stations:
+            self.refresh = False
             if button == self.button_next:
-                self.refresh = False
                 self.station_index += 1
                 self.station_index %= self.number_of_stations + 1
             elif button == self.button_previous:
-                self.refresh = False
                 self.station_index -= 1
                 self.station_index %= self.number_of_stations + 1
         elif button == self.button_refresh:
--- a/py3status/modules/velib_metropole.py
+++ b/py3status/modules/velib_metropole.py
@@ -1,10 +1,7 @@
 # -*- coding: utf-8 -*-
 """
 Display Velib shared bike avaibility on our favorite stations.
-Use https://www.velib-metropole.fr/map data.
-
-You only need to set id of stations you want to monitor in station_codes param.
-Then scroll to cycle stations, and middle clic to force a refresh.
+Use api from https://www.velib-metropole.fr/map.
 
 Configuration parameters:
     button_next: Display next station (default 4)
@@ -18,7 +15,7 @@ Configuration parameters:
         *(default '{station_name}: [\?color=station_state_code {station_state}]'
         '[\?soft  ][\?color=greenyellow {nb_bike}/{nb_free_e_dock}]'
         '[\?soft  ][\?color=deepskyblue {nb_ebike}/{nb_free_e_dock}]')*
-    station_codes: List of velib stations to monitor.
+    stations: List of velib stations to monitor.
         You can get stations id on map here https://www.velib-metropole.fr/map
         (default [20043, 11014, 20012, 20014, 10042])
     thresholds: Configure colors of format station.
@@ -27,7 +24,7 @@ Configuration parameters:
 Format placeholders:
     {format_station}            format for station details
     {index}                     current index of displayed station, eg 1
-    {stations}                  count of stations find in station_codes, eg 12
+    {stations}                  count of stations find in stations, eg 12
 
 format_station placeholders:
     {credit_card}               station take credit card?, eg 'no'
@@ -77,7 +74,7 @@ class Py3status:
         "[\?soft  ][\?color=greenyellow {nb_bike}/{nb_free_e_dock}]"
         "[\?soft  ][\?color=deepskyblue {nb_ebike}/{nb_free_e_dock}]"
     )
-    station_codes = [20043, 11014, 20012, 20014, 10042]
+    stations = [20043, 11014, 20012, 20014, 10042]
     thresholds = [(0, "good"), (1, "bad")]
 
     class Meta:
@@ -105,12 +102,12 @@ class Py3status:
         }
 
     def post_config_hook(self):
-        if not self.station_codes:
+        if not self.stations:
             raise Exception(STRING_MISSING_STATIONS)
 
         # string to list if necessary
-        if not isinstance(self.station_codes, list):
-            self.station_codes = [self.station_codes]
+        if not isinstance(self.stations, list):
+            self.stations = [self.stations]
 
         # take the whole map for first run
         # default values take all stations
@@ -124,14 +121,13 @@ class Py3status:
             "gpsBotLongitude": 2.2,
             "zoomLevel": 15,
         }
-        self.toggled = False
+        self.scrolling = False
         self.idle_time = 0
         self.request_timeout = 10
         self.station_states = {"Operative": 0, "Close": 1}
         self.station_index = 1
-        self.stations = {}
+        self.stations_data = {}
         self.optimal_area = False
-        self.first_start = True
         self.button_refresh = 2
 
         # get placeholders
@@ -161,14 +157,8 @@ class Py3status:
         """
         reduce the zone to reduce the size of fetched data on refresh
         """
-
-        latitudes = []
-        longitudes = []
-
-        for x in data:
-            latitudes.append(float(data[x]["station_gps_latitude"]))
-            longitudes.append(float(data[x]["station_gps_longitude"]))
-
+        latitudes = [float(data[x]["station_gps_latitude"]) for x in data]
+        longitudes = [float(data[x]["station_gps_longitude"]) for x in data]
         self.gps_dict.update(
             {
                 "gpsTopLatitude": max(latitudes),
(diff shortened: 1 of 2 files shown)


py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+        (default 'Velib Metropole: {index}/{stations} {format_station}')
+    format_station: How to display the velib station data.
+        *(default '{station_name}: [\?color=station_state_code {station_state}]'
+            '[\?soft  ][\?color=greenyellow {nb_bike}/{nb_free_e_dock}]'
+            '[\?soft  ][\?color=deepskyblue {nb_ebike}/{nb_free_e_dock}]')*
+    station_codes: List of velib stations to monitor.
+        You can get stations id on map here https://www.velib-metropole.fr/map
+        (default [20043, 11014, 20012, 20014, 10042])
+    thresholds: Configure colors of format station.
+        (default {'station_state_code': [(0, 'good'), (1, 'bad')]})
```
lasers: Default with `[(0, 'good'), (1, 'bad')]` instead. If this module does not work with that option, then your threshold code is probably a little off.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+                        "station_due_date": ".0f",
+                    },
+                    "format_strings": ["format_station"],
+                }
+            ]
+        }
+
+    def post_config_hook(self):
+        if not self.station_codes:
+            raise Exception("%s" % (STRING_MISSING_STATIONS))
```
lasers: No need for `%s` here.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+    button_refresh: refresh stations (default 2)
+    cache_timeout: The time between API polling in seconds
+        It is recommended to keep this at a higher value to avoid rate
+        limiting with the API's. (default 60)
+    format: How to display the velib data.
+        (default 'Velib Metropole: {index}/{stations} {format_station}')
+    format_station: How to display the velib station data.
+        *(default '{station_name}: [\?color=station_state_code {station_state}]'
+            '[\?soft  ][\?color=greenyellow {nb_bike}/{nb_free_e_dock}]'
+            '[\?soft  ][\?color=deepskyblue {nb_ebike}/{nb_free_e_dock}]')*
```
lasers: You can unindent this a bit more.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+        # take the whole map for first run
+        # default values take all stations
+        # around Paris
+        # Area is then shrink to only get
+        # specified stations
+        self.gps_top_right_latitude = 49.0
+        self.gps_top_right_longitude = 2.5
+        self.gps_bot_left_latitude = 48.6
+        self.gps_bot_left_longitude = 2.2
+        self.gps_zoom_level = 15
```
lasers: (Opinion) Might be more nice to have one thing... a `gps` dict instead.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+
+        self.station_states = {"Operative": 0, "Close": 1}
+
+        json = self._get_velib_data()
+        if json:
```
lasers: `velib_metropole_data` or something else. Not `json`. (plz).

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+            self.number_of_stations, self.stations = self._manipulate(json)
+            self._get_usefull_area(self.stations)
+
+        # get placeholders
+        self.placeholders = []
+        for x in ["format", "format_station"]:
+            self.placeholders.append(self.py3.get_placeholders_list(x))
```
lasers: You're not using `self.placeholders` anywhere.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+                "&zoomLevel=%s",
+            ]
+        )
+        url = url % (
+            self.gps_top_right_latitude,
+            self.gps_top_right_longitude,
+            self.gps_bot_left_latitude,
+            self.gps_bot_left_longitude,
+            self.gps_zoom_level,
+        )
```
lasers: This part looks like something that could be messed up easily. Come up with something more clean. You don't need to use `"".join()` and use `{example_name}` instead. You can do `format(**self.gps_dict)` too.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+                {
+                    "station_due_date_s": station_due_date.isoformat(),
+                    "station_state_code": int(
+                        self.station_states[new_station["station_state"]]
+                    ),
+                }
+            )
+
+            stations[index] = new_station
+            index += 1
```
lasers: Use `enumerate()` instead of homemade `index`.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+
+        return len(stations), stations
+
+    def velib_metropole(self):
+        if self.refresh is True:
+            # get number of stations and station list
+            json = self._get_velib_data()
+            if json:
+                self.number_of_stations, self.stations = self._manipulate(json)
```
lasers: (Opinion) You can do `len(stations)` here instead. More clean since you're not doing special with `stations` inside `manipulate()` to get that number of stations.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+        if not self.stations:
+            full_text = None
+        else:
+            # reset station_index counter
+            if self.station_index == 0:
+                self.station_index = 1
+
+            # thresholds
+            for x in self.thresholds:
+                if x in self.stations[self.station_index]:
```
lasers: Your thresholds code looks off.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+
+            # search station
+            station = next(
+                (item for item in data if int(item["station"]["code"]) == int(code)),
+                None,
+            )
+
+            # flat, camel to snake and cast...
+            for k, v in self.py3.flatten_dict(station, delimiter="_").items():
+                new_station[self._camel_to_snake_case(k)] = self._cast_number(v)
```
lasers: You're running this all times. This and `datetime.fromtimestamp` below too. Chance is that users don't always use them so you should see if you can be more efficient by only working with placeholders first instead of running this on all variables in every station.... for unrelated variables on only one station.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+            # get number of stations and station list
+            json = self._get_velib_data()
+            if json:
+                self.number_of_stations, self.stations = self._manipulate(json)
+
+        # reset refresh
+        self.refresh = True
+
+        if not self.stations:
+            full_text = None
```
lasers: (Tip) Never return `full_text = None`.

We can return safeformatted composite with empty velib (dict) data instead of `full_text = None` because if we don't, this will mess with `format = '{station}|No Velib` where we can't get `No Velib` to work.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+        self.gps_bot_left_latitude = 48.6
+        self.gps_bot_left_longitude = 2.2
+        self.gps_zoom_level = 15
+
+        self.station_states = {"Operative": 0, "Close": 1}
+
+        json = self._get_velib_data()
+        if json:
+            self.number_of_stations, self.stations = self._manipulate(json)
+            self._get_usefull_area(self.stations)
```
lasers: (Opinion) I'm pretty sure if you used `try/except` + `self.gps` dict to get useful area in code, it would look more clean than reproducing this code again in the `post_config_hook`. Remember `try/except` in `hddtemp`? Do something like that.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+        except ValueError:
+            try:
+                value = int(value)
+            except ValueError:
+                pass
+        return value
+
+    def _get_usefull_area(self, data):
+        """
+        reduce the zone to reduce the size of fetched data on refresh
```
lasers: (Opinion) You're not returning anything here. Rename `_get_` to something else like `_set_`... and maybe more descriptive naming than `get_usefull_area`. Also, typo.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+        if self.stations:
+            self.refresh = False
+            if button == self.button_next:
+                self.station_index += 1
+                self.station_index %= self.number_of_stations + 1
+            elif button == self.button_previous:
+                self.station_index -= 1
+                self.station_index %= self.number_of_stations + 1
+        elif button == self.button_refresh:
+            self.refresh = True
```
lasers: (Opinion) This `self.refresh` code seems a little icky... and could lead to errors somewhere... See `whatismyip` for the nice example as it would track your `self.cache_timeout` and  allow you to scroll away without refreshes.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+"""
+Display Velib shared bike avaibility on our favorite stations.
+Use https://www.velib-metropole.fr/map data.
+
+You only need to set id of stations you want to monitor in station_codes param.
+Then scroll to cycle stations, and middle clic to force a refresh.
+
+Configuration parameters:
+    button_next: Display next station (default 5)
+    button_previous: Display previous station (default 4)
```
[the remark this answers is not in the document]
cyrinux: I would prefer to keep the possibitly to set them

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,292 @@
...
+    button_refresh: refresh stations (default 2)
```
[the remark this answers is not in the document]
cyrinux: ok i hide refresh button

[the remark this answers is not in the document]
tobes: >  I think it'd be good if we start hiding this option everywhere. 

why?  this is the most useful option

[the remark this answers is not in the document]
lasers: Four modules have this option right now. 
```python
./whatismyip.py:175:        elif button == self.button_refresh:
./github.py:259:        elif button != self.button_refresh:
./gitlab.py:165:        if button != self.button_refresh:
./google_calendar.py:496:            elif button == self.button_refresh:
```
Option 1) We can add this option for API modules everywhere. 
Option 2) We can `getattr` those 4 modules into nonexistent.

If we go with Option 2, then users can refresh modules everywhere with button `2`. This is already enforced for long time in many modules thanks to your code. Users can pick up on this soon enough or they will learn about this with (new) documentation. 

I think this could be akin to `request_timeout` in that it's good to expose those options, but maybe users won't care (and that it will refresh again in a minute) and maybe we should try and hide this config to simplify code and/or to enforce consistency refresh behavior for all modules.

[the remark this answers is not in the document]
tobes: @lasers The thing is not to change __anything__ but to create an issue to discuss a change before implementing

Also just to let you know I'll be merging #1504 soon as I have sign-off on it - I'll look at fixing lm_sensors too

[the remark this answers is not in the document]
cyrinux: lovers, so, I hide / hardcode refresh or not? 🐌

py3status/modules/velib_metropole.py
cyrinux: ok will try to put this at another place

cyrinux: done without try/except, hope this is well

py3status/modules/velib_metropole.py
```diff
@@ -165,38 +164,34 @@ def _get_usefull_area(self, data):
...
-            ]
-        )
-        url = url % (
-            self.gps_top_right_latitude,
-            self.gps_top_right_longitude,
-            self.gps_bot_left_latitude,
-            self.gps_bot_left_longitude,
-            self.gps_zoom_level,
-        )
+        url = """https://www.velib-metropole.fr/webapi/map/details? \
```
lasers: Alternatively, you can do `url = (`  + `)` without requiring `\`.

py3status/modules/velib_metropole.py
lasers: Look at `loadavg`. You're missing `self.py3.get_color_names_list` as it's more useful than checking inside `thresholds` and allow things like `thresholds = [(0, 'bad'), (1, 'good')]` to work too.

py3status/modules/velib_metropole.py
```diff
@@ -262,25 +282,27 @@ def velib_metropole(self):
...
-            self.refresh = False
             if button == self.button_next:
                 self.station_index += 1
                 self.station_index %= self.number_of_stations + 1
             elif button == self.button_previous:
                 self.station_index -= 1
                 self.station_index %= self.number_of_stations + 1
-        elif button == self.button_refresh:
-            self.refresh = True
+        elif button == 2:
```
lasers: Cool beans. Use `self.button_refresh` and keep `2` in `post_config_hook` with others.

py3status/modules/velib_metropole.py
```diff
@@ -208,46 +203,71 @@ def _manipulate(self, data):
...
+        if refresh:
+            self.idle_time = current_time + self.cache_timeout
+            cached_until = self.cache_timeout
+        else:
+            cached_until = self.idle_time - current_time
 
-        # reset refresh
-        self.refresh = True
+        if not refresh or self.first_start:
+            self.velib_metropole_data = self._get_velib_data()
```
lasers: Good try. This is refreshing on every scroll now. You're missing `self.toggled` code from `whatismyip` too... to prevent data refreshes on scrolling.

py3status/modules/velib_metropole.py
```diff
@@ -208,46 +203,71 @@ def _manipulate(self, data):
...
+            if all(
+                "station_state" in list for list in [new_station, self.placeholders]
+            ):
+                new_station.update(
+                    {
+                        "station_state_code": int(
+                            self.station_states[new_station["station_state"]]
+                        )
+                    }
+                )
```
lasers: Show a magic trick. Turn 7 lines here into 2 lines.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,314 @@
...
+        )
+        self.optimal_area = True
+
+    def _get_velib_data(self):
+        url = """https://www.velib-metropole.fr/webapi/map/details? \
+            gpsTopLatitude={top_latitude} \
+            &gpsTopLongitude={top_longitude} \
+            &gpsBotLatitude={bottom_latitude} \
+            &gpsBotLongitude={bottom_longitude} \
+            &zoomLevel={zoom_level}"""
```
tobes: these can be passed to the request directly as a params dict
```
params = {
    'a': 1,
    'b': 'hello',
}

self.py3.request(
    url,
    params=params,
)
```

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,314 @@
...
+        self.stations = {}
+        self.optimal_area = False
+        self.first_start = True
+        self.button_refresh = 2
+
+        # get placeholders
+        self.placeholders = []
+        for x in [self.format, self.format_station]:
+            self.placeholders += self.py3.get_placeholders_list(x)
+        self.placeholders += ["station_gps_latitude", "station_gps_longitude"]
```
lasers: (Nit) You can init the list with this.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        reduce the zone to reduce the size of fetched data on refresh
+        """
+
+        latitudes = []
+        longitudes = []
+
+        for x in data:
+            latitudes.append(float(data[x]["station_gps_latitude"]))
+            longitudes.append(float(data[x]["station_gps_longitude"]))
```
lasers: (Suggestion) You probably can reduce 7 lines here to 2 lines by using list comprehension or maybe map.

py3status/modules/velib_metropole.py
lasers:
```python
code = int(self.station_states[new_station["station_state"]])
new_station["station_state_code"] = code
```

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,314 @@
...
+
+        # time
+        if refresh:
+            self.idle_time = current_time + self.cache_timeout
+            cached_until = self.cache_timeout
+        else:
+            cached_until = self.idle_time - current_time
+
+        if not self.toggled and not refresh:
+            self.toggled = False
```
lasers: Your toggle is a bit off. This should be `if self.toggled and not refresh`. Maybe just rename it to `self.scrolling` then fix `on_click` code so scrolling is true here instead of opposite.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,314 @@
...
+            data = self.velib_metropole_data
+        else:
+            data = self.velib_metropole_data = self._get_velib_data()
+
+        if self.first_start:
+            data = self.velib_metropole_data = self._get_velib_data()
```
lasers: We fetch data twice on startup now... because of this and code above. Initialize `idle_timeout = 0` and `self.scrolling = False` should be enough... then we shouldn't need `first_start` anymore.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,314 @@
...
+
+        if data:
+            self.stations = self._manipulate(data)
+            if not self.optimal_area:
+                self._set_optimal_area(self.stations)
+
+        self.number_of_stations = len(self.stations) or 0
```
lasers: Is this `0 or 0`?

py3status/modules/velib_metropole.py
cyrinux: done 2 lines from 6 with simple list comprehension

lasers:
Minor fix for scrolling. 
```diff
diff --git a/py3status/modules/velib_metropole.py b/py3status/modules/velib_metropole.py
index 2c100645..27eeb0f0 100644
--- a/py3status/modules/velib_metropole.py
+++ b/py3status/modules/velib_metropole.py
@@ -236,7 +236,7 @@ class Py3status:
             cached_until = self.idle_time - current_time
 
         if self.scrolling and not refresh:
-            self.scrolling = True
+            self.scrolling = False
             data = self.velib_metropole_data
         else:
             data = self.velib_metropole_data = self._get_velib_data()
@@ -277,15 +277,14 @@ class Py3status:
 
     def on_click(self, event):
         button = event["button"]
-        self.scrolling = True
-        if button == self.button_next:
-            self.station_index += 1
-            self.station_index %= self.number_of_stations + 1
-        elif button == self.button_previous:
-            self.station_index -= 1
+        if button in [self.button_next, self.button_previous]:
+            self.scrolling = True
+            if button == self.button_next:
+                self.station_index += 1
+            elif button == self.button_previous:
+                self.station_index -= 1
             self.station_index %= self.number_of_stations + 1
         elif button == self.button_refresh:
-            self.scrolling = False
             self.idle_time = 0
         else:
             self.py3.prevent_refresh()

```

lasers:
Do `git merge upstream/string-thresholds`. I suppose you can remove `station_state_code` in favor of string thresholds. Sometime I see your Operative/Close with wrong colors. Possibly nice to rename `index` with `station` -- `station index, eg 1`. I'll look at your thresholds next.

lasers:
Minor fix for thresholds.
```diff
diff --git a/py3status/modules/velib_metropole.py b/py3status/modules/velib_metropole.py
index 2c100645..02a42b3f 100644
--- a/py3status/modules/velib_metropole.py
+++ b/py3status/modules/velib_metropole.py
@@ -105,15 +105,12 @@ class Py3status:
         if not self.stations:
             raise Exception(STRING_MISSING_STATIONS)
 
-        # string to list if necessary
+        # int to list if necessary
         if not isinstance(self.stations, list):
             self.stations = [self.stations]
 
-        # take the whole map for first run
-        # default values take all stations
-        # around Paris
-        # Area is then shrink to only get
-        # specified stations
+        # take the whole map for first run default values take all stations
+        # around Paris Area is then shrink to only get specified stations
         self.gps_dict = {
             "gpsTopLatitude": 49.0,
             "gpsTopLongitude": 2.5,
@@ -130,14 +127,18 @@ class Py3status:
         self.optimal_area = False
         self.button_refresh = 2
 
-        self.thresholds_init = self.py3.get_color_names_list(self.format)
-
-        # get placeholders
+        # placeholders
         self.placeholders = ["station_gps_latitude", "station_gps_longitude"]
         for x in [self.format, self.format_station]:
-            self.thresholds_init += self.py3.get_color_names_list(x)
             self.placeholders += self.py3.get_placeholders_list(x)
 
+        # thresholds
+        self.thresholds_init = {}
+        for name in ['format', 'format_station']:
+            self.thresholds_init[name] = self.py3.get_color_names_list(
+                getattr(self, name, '')
+            )
+
     def _camel_to_snake_case(self, data):
         if not isinstance(data, (int, float)):
             s1 = sub("(.)([A-Z][a-z]+)", r"\1_\2", data)
@@ -215,6 +216,10 @@ class Py3status:
                     }
                 )
 
+            for x in self.thresholds_init['format_station']:
+                if x in new_station:
+                    self.py3.threshold_get_color(new_station[x], x)
+
             stations_data.append(new_station)
 
         # forge return
@@ -264,7 +269,7 @@ class Py3status:
                 "index": self.station_index,
             }
 
-        for x in self.thresholds_init:
+        for x in self.thresholds_init['format']:
             if x in self.stations_data[self.station_index]:
                 self.py3.threshold_get_color(
                     self.stations_data[self.station_index][x], x

```

cyrinux:
@lasers Thanks ;)

cyrinux:
Hmm my try (after push...) show me not the good color for Close state. Will check tomorrow.

py3status/modules/velib_metropole.py
```diff
@@ -277,15 +282,13 @@ def velib_metropole(self):
 
     def on_click(self, event):
         button = event["button"]
-        self.scrolling = True
-        if button == self.button_next:
-            self.station_index += 1
-            self.station_index %= self.number_of_stations + 1
-        elif button == self.button_previous:
-            self.station_index -= 1
-            self.station_index %= self.number_of_stations + 1
```
lasers: You forget this.

lasers:
(Opinion) With list of composite, you're going to expose everything on the bar. You probably want to cache a dict of composite (instead of a dict of data) so thresholds would work correctly there and you wouldn't have to do any work between scrolling since it would be already done on each refresh.

py3status/modules/velib_metropole.py
```diff
@@ -0,0 +1,293 @@
...
+                self._set_optimal_area(self.stations_data)
+
+        self.number_of_stations = len(self.stations_data)
+
+        if not self.stations_data:
+            velib_data = {"index": 1, "stations": 0, "format_station": {}}
+        else:
+            # reset station_index counter
+            if self.station_index == 0:
+                self.station_index = 1
```
lasers: If you have to do this, then your `index` code is probably off. Your index should start at `0`. You should only have to do `+ 1` thing on velib_data's `"index": self.station_index` (and not everywhere)

cyrinux:
Ok, will try...will understand one day

ultrabug:
Impressive module @cyrinux I just tried it. I think there's a typo on the default format since you always show the number of edocks even for the mechanical bikes.

I'll merge this and correct that

Good job and thank you guys!

PR #1525, opened 2018-09-19
Some commands in some environments does require LANGUAGE=C to print output in english.

wojtex:
Add `LANGUAGE="C"` to "English" environment so that commands like `hamster` will print output in English (esp. on systems where `LANGUAGE` is set to something different).

Useful link about locale-related environment variables:
[https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables)

### DIFF
--- a/py3status/py3.py
+++ b/py3status/py3.py
@@ -109,6 +109,7 @@ class Py3:
         self._config_setting = {}
         self._english_env = dict(os.environ)
         self._english_env['LC_ALL'] = 'C'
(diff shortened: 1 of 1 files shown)


ultrabug:
Good catch, thanks a lot for your first contribution to this project @wojtex !

PR #1547, opened 2018-10-11
WIP: dbus helper for modules

tobes:
#1546 adds cool udev support

this adds dbus support for modules - the code is still in crude proof of concept form.

The example module is battery_level and it just updates on a dbus power change.  Due to how the module currently works it triggers better when we unplug the power than when we plug it in.  This is more about the module.

It is very crude and needs more work but I wanted to show the code as relates to #1546 so I think it is good to consider the needs of both

The idea of this is

* we can subscribe to dbus events
* get dbus properies (missing)
* trigger dbus methods (missing)
* stop modules doing this and having multiple gobject loops etc as currently

but the real issue for now is how modules interact with this and how it fits in with udev

### DIFF
--- a/py3status/dbus_helper.py
+++ b/py3status/dbus_helper.py
@@ -1,0 +1,53 @@
+from threading import Thread
+
+
+class DBus:
+
+    def __init__(self, py3_wrapper):
+        try:
+            from gi.repository import GObject
+            from pydbus import SystemBus, SessionBus
+            self.GObject = GObject
+            self.SessionBus = SessionBus
+            self.SystemBus = SystemBus
+            self.initialized = True
+        except ImportError:
+            # FIXME logging
+            self.initialized = False
+            pass
+        self.py3_wrapper = py3_wrapper
+        self.started = False
+        self.bus_system = None
+        self.bus_session = None
+
+    def get_bus(self, bus):
+        if bus == "system":
+            if self.bus_system is None:
+                self.bus_system = self.SystemBus()
+            return self.bus_system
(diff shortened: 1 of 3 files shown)


ultrabug:
Ok then here, can we go for the same `on_dbus_` logic with reused constants for triggered actions please?

tobes:
I'm not sure that works because dbus is quite specific and has a whole interface eg calling methods, getting properties, as well as subscribing to events. think - `mpis` so controlling your music player(s) and showing current song etc so it is a bit more complicated.  Maybe we also can do a on dbus path/event but even this gets tricky - but we can see.

ultrabug:
hmm on a second thought indeed this sounds more like the first implementation I made of udev...

boucman:
@tobes do you intend to work on this some more ?

I might have time to work on it if it's fine with you...

ultrabug:
@boucman by all means, go ahead mate :+1:

ultrabug:
Timeout, closing, reopen if needed mates; Thanks!

boucman:
Well, I had a look and honestly it looked good as is... i'm not sure what needs to be added...

boucman:
though I should have said so..

tex:
can you merge this? dbus support would be useful for my plugin...

ultrabug:
@tex no I'm sorry, some modules like `mpris` do implement dbus related logic you could get inspiration from

tex:
@ultrabug Thanks! That is exactly what I was looking for!

PR #1549, opened 2018-10-16
adds ~/.config/i3 as a default config directory, closes #1548

### LINKED ISSUE #1548: default config location
I have installed ubuntu 18.04 and it makes sense to put my `i3status.conf` in `~/.config/i3/` but this is not found by default.  I think we should add this path.

ultrabug:
I consider it a bug on our part


tobes:
Thanks for the quick fix

Issue #1561, opened 2018-10-29
battery_level: IndexError for battery_level.py line 231 when docked


eddie-dunn:
The error `Instance 'battery_level', user method 'battery_level' failed (IndexError) battery_level.py line 231.` is reported when I dock my laptop.

```
$ acpi -bi
Battery 0: Full, 100%
Battery 0: design capacity 6842 mAh, last full capacity 6002 mAh = 87%
Battery 1: Discharging, 0%, rate information unavailable
Battery 2: Discharging, 0%, rate information unavailable
```

```python
# battery_level:230-231
battery["capacity"] = int(findall("(?<= )(\d+)(?= mAh)",
                                  acpi_battery_lines[1])[1])
```
It appears the string 'rate information unavailable' is not handled gracefully when trying to extract the capacity from `acpi -bi`.

Unless someone objects (or beats me to it), I can submit a patch that resolves this issue.

Also, going through the code I think that it needs some general refactoring love. Are you guys open to such endeavours?

tobes:
> Unless someone objects (or beats me to it), I can submit a patch that resolves this issue.

patches welcome

> Also, going through the code I think that it needs some general refactoring love. Are you guys open to such endeavours?

Yeah this module has been around for quite a while and had bits added over time, so it is probably due some love.

eddie-dunn:
After looking into this a bit more, the reason we get this issue is that I was running `battery_level` in its default ACPI mode, with two battery driven devices connected (Battery 1 and 2 in the `acpi -bi` output quoted above). In my case it was my wireless mouse and keyboard.

With the setting `measurement_mode = "sys"` enabled, `battery_level` no longer crashes.

Also, I did some digging in the source code and the way acpi mode is implemented is super fragile, and relies on regexes and that the user gets their output in English which may not always be the case. As such, I will not work on solving this bug, since acpi mode is not worth the effort.

This leads me to the following questions:

1. Why isn't 'sys' the default mode if `acpi` is installed?
1. Can it be set to always be the default?
1. Can we deprecate/remove 'acpi' mode?
1. Checking `/sys/class/power_supply` is actually quite powerful; we can get battery information also for e.g., HID devices. I can look into expanding the use case for `battery_level` to report information for other things found there as well. Would that be of interest to you?

maximbaz:
So `acpi` method is default because `sys` was implemented later. Since `sys` seem to be more stable, I see no reason not to make it default and deprecate `acpi` mode.

[...]

PR #1562, opened 2018-11-01
battery_level: Default to using sys instead of acpi

eddie-dunn:
Solves #1561


eddie-dunn:
I've tested it locally and for me, at least, `sys` is used if `measurement_mode` is not given.

ultrabug:
I'm making an exception into 3.14 rules to congratulate and thank @eddie-dunn for his first-time contribution to the project!

PR #1565, opened 2018-11-02
Print module information to sdtout instead of stderr

### LINKED ISSUE #1564: Module information printed to stderr
Using `py3status module details` or `py3status module list` sends the output to stderr instead of stdout.

I was trying to pipe details into a file to read it with my text editor and got surprised with an empty file. Is there any reason for this, or can I submit a PR?

RoberTnf:
Fix #1564


ultrabug:
Thank you @RoberTnf !

PR #1573, opened 2018-11-10
battery_level: refresh with udev, increase default cache_timeout

maximbaz:
cc @ultrabug for another great example of using new udev feature.

With this change, `battery_level` module will refresh instantly as soon as you plug or unplug charging cable. Because of this, there's no need to refresh the module as often as every 30 seconds, I set it to 60 seconds for now.

Guess what, until today I had `cache_timeout` set to `1` because I wanted to instantly know when I plug in my charger. NO MORE.


ultrabug:
Awesome, thanks!

Issue #1582, opened 2018-11-25
[RFC] Zen of py3status


ultrabug:
Following up on #1566 I'd like to propose a short yet good enough zen to allow us to create a frame for our decisions. 

If I'm missing something important, feel free to propose it.

This is meant to be added to the py3status philosophy!

# efficient and simple defaults

remember the philosophy:
> no added configuration file, use the standard i3status.conf

we like py3status because it's a drop-in replacement of i3status
i3 users don't expect fancy and magical things, they use i3 because it's simple and efficient
keep configuration options and default format as simple as possible

# it's not because you can that you should

on modules, expose things that you WILL use, not things that you COULD use
on core, make features and options as seamless as possible (lazy loading, sane default, no strong requirements)

# good enough is good enough

strive for and accept "good enough" features / proposals and keep discussions on this topic only
you can always iterate further after it's merged IF it's worth it

# one feature / idea at a time

trust and foster iteration with your peers by refraining from digressions
keep discussions focused and easy to get into

# modules are responsible for user information and interactions

that is what's written in the bar and its behavior on clicks etc

# core is responsible for user experience 

core features and configuration options should focus on users overall experience
things that are	related to the actual overall output of the bar are handled by core
smart things overlaying	modules	themselves also end up in the core

# rely on i3status color scheme

no fancy colors by default, only i3status good/degraded/bad
if we want to provide enhanced coloring, this should be	a core feature

# rely on the i3bar protocol

what's possible with it, we should offer, no less no more

maximbaz:
Agree on all points, this is a good summary of #1566. Could you please also comment on the topic of breaking changes, consolidating modules (I'm thinking about examples like the new `update` module which replaces all of the existing `*_updates` modules), rewriting modules from scratch and cleaning codebase of deprecated options when major version changes?

ultrabug:
@maximbaz I think most of those answers are contained in my version 4.0 point of view #1584

ultrabug:
https://py3status.readthedocs.io/en/latest/contributing.html#zen-of-py3status

For further modifications, we welcome PRs ;)

PR #1585, opened 2018-11-25
external_script: notifications

maximbaz:
In light of new Zen #1582, I'm submitting a simple implementation for the pattern I use with `external_script`. In my mind is good enough until we come up with some sophisticated support for notifications in core. This replaces #1439.

-----

Before:

* external_script shows only first line of output, no way to see entire output
* refreshes on any button click

After:

* external_script still shows first line of output, but on click you get a notification with full output
* refreshes only on `button_refresh`, i.e. left click shows notification and doesn't trigger refresh
* also added `line` format placeholder to show the total number of lines in the output (could rename if you want, `line` was proposed in #1439).

Why:

I use `external_script` with the same pattern for different purposes. Don't expect you to understand what exactly they are doing, just showing you a repeating pattern for which I'm implementing this PR.

<details>
<summary>Examples of why I need this for <code>external_script</code></summary>

* Check updates for official packages:
    * I have a script that outputs updates, one line per package
    * In status bar I have `official updates: 3` (this is why I need `line` placeholder)
    * On click I show notification that shows names of those 3 packages
    * I don't want left click to refresh the module, because script is slow to execute, thus need for `button_refresh`, but I want to refresh on middle click

* Check updates for unofficial packages:
    * I have a script that outputs updates, one line per package
    * In status bar I have `unofficial updates: 3`
    * On click I show notification that shows names of those 3 packages
    * I want left click to only show notification, and middle click to refresh

* Check updates for VCS packages:
    * I have a script that outputs updates, one line per package
    * In status bar I have `VCS updates: 3`
    * On click I show notification that shows names of those 3 packages
    * I want left click to only show notification, and middle click to refresh

* Check which packages need to be rebuild because of updated dependencies:
    * I have a script that outputs them, one line per package
    * In status bar I have `packages to rebuild: 3`
    * On click I show notification that shows names of those 3 packages
    * I want left click to only show notification, and middle click to refresh

* Check which configs need to be updated because of updated packages:
    * I have a script that outputs them, one line per config
    * In status bar I have `configs to update: 3`
    * On click I show notification that shows names of those 3 configs
    * I want left click to only show notification, and middle click to refresh

</details>

### DIFF
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -10,7 +10,8 @@ code such as #FF0000 for red).
 The script should not have any parameters, but it could work.
 
 Configuration parameters:
-    button_refresh: button to refresh the module (default 2)
+    button_show_notification: button to show notification with full output
+        (default None)
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
@@ -51,7 +52,7 @@ class Py3status:
     """
     """
     # available configuration parameters
-    button_refresh = 2
+    button_show_notification = None
     cache_timeout = 15
     format = '{output}'
     localize = True
@@ -105,7 +106,7 @@ class Py3status:
 
     def on_click(self, event):
         button = event["button"]
-        if button != self.button_refresh:
+        if button == self.button_show_notification:
             self.py3.notify_user(self.output)
             self.py3.prevent_refresh()
 
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -10,6 +10,7 @@ code such as #FF0000 for red).
 The script should not have any parameters, but it could work.
 
 Configuration parameters:
+    button_refresh: button to refresh the module (default 2)
     cache_timeout: how often we refresh this module in seconds
         (default 15)
     format: see placeholders below (default '{output}')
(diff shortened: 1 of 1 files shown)


py3status/modules/external_script.py
```diff
@@ -97,9 +100,15 @@ def external_script(self):
...
         response['full_text'] = self.py3.safe_format(
-            self.format, {'output': output})
+            self.format, {'output': output, 'line': len(output_lines)})
         return response
 
+    def on_click(self, event):
+        button = event["button"]
+        if button != self.button_refresh:
+            self.py3.notify_user(self.output)
+            self.py3.prevent_refresh()
```
maximbaz: One problem I see with this simple approach is that if a user added a custom `on_click 1` handler, this will now execute both their handler and show notification. Maybe worth adding a flag `show_click_notifications`, or instead of `button_refresh` implement `button_show_notification` (so condition changes to `if button == self.button_show_notification`). Thoughts?

maximbaz: I already changed to `button_show_notification`, it's disabled by default so won't suddenly conflict with custom `on_click 1` handlers. I think we are good here.

py3status/modules/external_script.py
```diff
@@ -21,6 +23,7 @@
         (default False)
 
 Format placeholders:
+    {line} number of lines in the output
```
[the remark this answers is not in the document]
maximbaz: Well, to me too, I was following earlier comment 🙂 Changed back to `lines` now.

ultrabug:
Good enough and respects default behavior. This is in!

Thanks

PR #1607, opened 2018-12-29
backlight: Added "hide_when_unavailable" option

m45t3r:
Use case: user share the same configuration between laptops/desktops and don't want to have a "backlight: no unavailable devices" error when using backlight mode in desktop.


ultrabug:
Simple & clear, thanks @m45t3r !

And thanks for your interesting recipes around py3status + nixOS on Twitter ;)

PR #1614, opened 2019-01-05
Sysdata temperatures

JackDoan:
I adjusted how sysdata gets temperature data. Now, in addition to the way it uses lm_sensors, you can also just point it at a hwmon path. This allows you to get temperature data for things that aren't CPUs, and it made it actually work on my Ryzen machine.


py3status/modules/sysdata.py
```diff
@@ -15,7 +15,9 @@
         (default '°C')
     thresholds: thresholds to use for color changes
         (default [(0, "good"), (40, "degraded"), (75, "bad")])
-    zone: thermal zone to use. If None try to guess CPU temperature
+    zone: thermal zone to use. If None try to use zone_path
```
maximbaz: we can add "or guess CPU temperature"

py3status/modules/sysdata.py
```diff
@@ -267,13 +267,37 @@ def _get_cputemp(self, zone, unit):
...
+                with open(zone_path) as f:
+                    cpu_temp = f.readline()
+                cpu_temp = float(cpu_temp) / 1000  # convert from mdegC to degC
+            except IOError:
+                cpu_temp = "zone_path does not exist"
+            except ValueError:
+                cpu_temp = "?"
+
+        else:
+            cpu_temp = self._get_cputemp_with_lmsensors(zone)
```
maximbaz: In the comment in the top you said that "zone" has priority over "zone_path" (if "zone" is None, try to use "zone_path"), but the logic doesn't seem to match that description, "zone_path" will be used even if "zone" is not None. Can you make them consistent?

JackDoan:
I've adjusted this so that if zone is a file, it tries to read it as a hwmon sensor. If that doesn't work, it falls back on the old behavior.

py3status/modules/sysdata.py
```diff
@@ -267,13 +264,36 @@ def _get_cputemp(self, zone, unit):
...
-            if unit == "K":
-                cpu_temp += 273.15
         else:
             cpu_temp = "?"
 
         return cpu_temp
 
+    def _get_cputemp(self, zone, unit):
+        # if unit not in [u"°F", u"°C", "K"]:
+        #     return "unknown unit"
```
maximbaz: uncomment 😛

py3status/modules/sysdata.py
```diff
@@ -267,13 +264,36 @@ def _get_cputemp(self, zone, unit):
...
+                # ValueError can be thrown if zone was a file that didn't have a float
+                # if zone was not a file, it might be a sensor!
+                cpu_temp = self._get_cputemp_with_lmsensors(zone=zone)
+
+        else:
+            cpu_temp = self._get_cputemp_with_lmsensors()
+
+        if cpu_temp is float:
+            if unit == u"°F":
+                cpu_temp = cpu_temp * (9 / 5) + 32
```
maximbaz: be careful with this, on python2 `9 / 5` returns `1`... Either just write `1.8`, or use explicit floats `9.0 / 5.0`

py3status/modules/sysdata.py
```diff
@@ -15,7 +15,8 @@
         (default '°C')
     thresholds: thresholds to use for color changes
         (default [(0, "good"), (40, "degraded"), (75, "bad")])
-    zone: thermal zone to use. If None try to guess CPU temperature
+    zone: Either a path in sysfs to CPU temperature sensor, or an lm_sensors thermal zone to use.
```
maximbaz: It looks neat, but I'm only a bit worried that when people know they want to specify lm_sensors zone, trying to read a non-existent file and raising&catching exceptions every few seconds could be expensive operation, what do you think?

By the way, is specifying sysfs path superior and people better do this instead of lm_sensors zone anyway?

py3status/modules/sysdata.py
JackDoan: Ah. I wasn't sure about this block, because the unit is also checked in `post_config_hook()`. Do we want to fail this way?

py3status/modules/sysdata.py
JackDoan: Changed to explicit floats

py3status/modules/sysdata.py
maximbaz: Hmm you are right. I'm a bit surprized that we aren't raising an error in this case (not in `post_config_hook` either), but instead making the module return some string... Could you test what happens with this block commented when you use unknown unit, what do you see as a user? If it's clear for user that unit is wrong, I'm ok removing this commented block 👍

py3status/modules/sysdata.py
JackDoan: I'm looking into this. I found [this](https://stackoverflow.com/questions/82831/how-do-i-check-whether-a-file-exists-without-exceptions) but it doesn't mention performance

JackDoan: I'm of the opinion that the sysfs way is better. It pulls the same data that lm_sensors does. The catch is that you need to know where to look. :/

maximbaz: In my experiment there's no perf difference, so I'm fine keeping everything as it is now.

py3status/modules/sysdata.py
JackDoan: Here's the relevant section from my config:

```
		sysdata tdie {
  		    zone = "/sys/class/hwmon/hwmon1/temp1_input"
			cache_timeout = 3
			thresholds = [(0, ""), (55, "degraded"), (65, "bad")]
			format =  '[\?color=temp '
			format += 'T_die: {cpu_temp:.0f}{temp_unit}'
			format += ']'
                        temp_unit = 'foo'
		}
```

With the block uncommented: `T_die unknown unitunknown unit`
With the block commented out: `T_die 30unknown unit`

maximbaz: Awesome, let's remove this then 👍

ultrabug:
@JackDoan please resolve conflicts mate

maximbaz:
Ping @JackDoan 😛

JackDoan:
Sorry y'all, I really only have time to work on this stuff on Saturdays. I don't mean to leave you hanging. Expect something in the next couple hours.

JackDoan:
Oh oops there was only one.

maximbaz:
Awesome :) Is it ready for merge from your perspective @JackDoan, or you want to do something else here?

JackDoan:
I think it's good to go!

PR #1659, opened 2019-01-17
net_rate: prevent log flooding on weird network interface state

cyrinux:
When i'm offline, after resuming my laptop, I can get lot of garbage from net_rate.
A quick working fix is done by handle KeyError too. 
ps: Error was on line 163.

### DIFF
--- a/py3status/modules/net_rate.py
+++ b/py3status/modules/net_rate.py
@@ -162,7 +162,7 @@ class Py3status:
             # get the deltas into variable
             delta = deltas[interface] if interface else None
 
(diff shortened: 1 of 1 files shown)


py3status/modules/net_rate.py
```diff
@@ -162,7 +162,7 @@ def net_rate(self):
             # get the deltas into variable
             delta = deltas[interface] if interface else None
```
maximbaz: we chatted offline, for context: KeyError was happening on this line

PR #1660, opened 2019-01-18
[RFC] Remove Python 2 compatibility

m45t3r:
This is mostly a PR to RFC a removal of Python 2 compatibility, as discussed in issue https://github.com/ultrabug/py3status/issues/1584.

What I basically did:
- Run `2to3` tool to help detect places where useless compatibility code was being used: `u""`, imports with old name, `__future__` imports, etc.
- Grep the code for "python 2" and "python2", remove obvious code that was Python 2 specific.

Probably this should be missing things and also may introduce new bugs. However this PR serves as a benchmark on how much code can be simplified.


m45t3r:
Some of our modules use `MySQLdb`. `MySQLdb` never had a Python 3 release, so this modules wouldn't work in a Python 3 world only.

The modules:
- `py3status/modules/rt.py`
- `py3status/modules/sql.py`
- `py3status/modules/glpi.py`

The only one of those that offers an alternative that works in Python 3 is `py3status/modules/rt.py`, that can also works with `pymysql`. `pymysql` seems to offer a compatible API, so maybe we could drop support to `MySQLdb` and substitute the usage of `MySQLdb` to `pymysql`. Any opnions?

m45t3r:
Well, I think that is it for now. I am using this branch in my desktop to look for any regressions, however as far I tested for my configuration it is working very well.

I don't really know what you @ultrabug (and the other maintainers) will want to do with this PR. It was most for experiment, however it can be merged without problems.

ultrabug:
Thanks for this amazing work @m45t3r !

So far, 4.0 dev has not started so we will not be dropping python2 support until we work on the 4 branch.

Your preliminary work will help a lot tho and could very well be the start of the 4.0 branch when we feel ready to move on this!

Cereal2nd:
50218cf breaks the people that already use py3status with mysqldb, they will need to change there config

m45t3r:
@Cereal2nd if this PR is merged it would break anyway since AFAIK there is no port for `MySQLdb` for Python 3.

Unless I am wrong.

Cereal2nd:
I'm talking about the sql module,
I applied the PR locally and everything works, except we place the 'MySQLdb' in the config, and we have to replace this with 'pymysql', so this will be a backward incompatibility

m45t3r:
@Cereal2nd Huh, yeah, I understand that, however this incompatibility is unavoidable.

Probably we should document it somewhere (in release notes or something).

m45t3r:
Rebased and add a preliminary changelog.

ultrabug:
@thiagokokada hi mate

as you may have noticed, we have worked on removing support for PY2 now thx to @hugovk 

does that sound okay to close this PR so you can maybe add your own new ideas on it if you wish to?

PR #1674, opened 2019-01-28
arch_updates: add auracle support

Compunctus:
.. since cower was deprecated by author and replaced with auracle.
This commit also prioritizes auracle over cower if both are installed.


ultrabug:
Thanks a lot for your first contribution to the project @Compunctus !

PR #1700, opened 2019-02-07
graphite: use func mapping instead of eval('func')

lasers:
Untested.


ultrabug:
Elegant indeed, cheers

PR #1704, opened 2019-02-08
Use python's native way of detecting location of i3status bin

maximbaz:
`which` is not guaranteed to be installed on the system, so let's not depend on it.

### DIFF
--- a/py3status/cli.py
+++ b/py3status/cli.py
@@ -1,6 +1,6 @@
 import argparse
 import os
-import subprocess
+import shutil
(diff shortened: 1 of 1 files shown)


maximbaz:
meh, okay :(

maximbaz:
I think I'm going to abandon it, because I just found that `check_commands()` in `py3.py` also uses `which`, and there we can't just ignore the absence of `which` on the system. 

Damn you, python2 😞

PR #1795, opened 2019-04-29
keyboard_layout: xkblayout-state should prefer variant if available

### LINKED ISSUE #1794: xkblayout-state should prefer variant if available
Hi,

I am using xkblayout-state and when I did an update, xkblayout-state is now my preferred command. I switch between us and dvorak often so in order to get it working with my setup I need to have 

"xkblayout-state print "%E"

when its querying the keymaps before setting them.

This issue is a follow-up to #1780

ecks:
Fixes #1794 

**Changes:**

- changes "xkblayout-state print %S" to "xkblayout-state print %E". This way the keyboard variant is returned if there is one set.

### DIFF
--- a/py3status/modules/keyboard_layout.py
+++ b/py3status/modules/keyboard_layout.py
@@ -167,7 +167,7 @@ class Py3status:
     def _set_xkblayout(self):
         layout = self._layouts[self._active]
         layout_pos = (
-            self.py3.command_output(["xkblayout-state", "print", "%S"])
+            self.py3.command_output(["xkblayout-state", "print", "%E"])
             .split()
             .index(layout)
         )


ultrabug:
@saengowp ping please, can you test & report it's okay for you too?

Thanks @ecks !

saengowp:
@ultrabug It's working for me. Sorry for introducing bug @ecks.

1. This PR definitely fixes problem for user who have multiple keyboard variant.
2. The only way this PR could regress is when user has multiple layout with different layout but same variant eg. `us(dvorak)` and `de(dvorak)`.

I don't know which bug would affect the most user.

ultrabug:
Thanks guys!

PR #1809, opened 2019-05-22
arch_updates module: auracle throws exit code on zero updates

nwg-piotr:
The bug manifests in the arch_updates module when no AUR updates found. The `auracle sync` command in such case returns exit status 1:

```
Command `auracle sync` returned non-zero exit status 1
```
Instead of returning 0 as the result of `len(updates.splitlines())`, the exception code is being executed, that results in strange display like "UPD: 0/None" for default format.

image

As a simple workaround, I propose returning 0 instead of None.

### DIFF
--- a/py3status/modules/arch_updates.py
+++ b/py3status/modules/arch_updates.py
@@ -84,8 +84,8 @@ class Py3status:
         try:
             updates = self.py3.command_output(["auracle", "sync"])
             return len(updates.splitlines())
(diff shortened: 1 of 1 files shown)


nwg-piotr:
You're right, as always. :) I should've taken a closer look to the CommandError class.

By the way: it seems [cower](https://github.com/falconindy/cower) no longer exists. Shouldn't it be removed from the module?

PR #1812, opened 2019-06-03
volume_status: add blocks

dpeukert:
This adds the blocks feature from other modules to volume_status. The last character gets used for values over 100 as per the implementation in battery_level.


ultrabug:
Hi @dpeukert thanks for your first contribution!

It's clean and simple :+1:

PR #1813, opened 2019-06-04
arch_updates module: add trizen support

dpeukert:


dpeukert:
Now that I am thinking about it, the correct long term solution is probably to let the user provide their own command so that we don't have to add suport for every AUR helper under the sun manually.

maximbaz:
I agree, we even had a prototype already, while I don't agree with that particular implementation, I think the general idea is very interesting and worth revisiting at some point. 

https://github.com/ultrabug/py3status/pull/1552

ultrabug:
I let the Arch contributors decide by themselves then :)

py3status/modules/arch_updates.py
```diff
@@ -94,6 +96,13 @@ def _get_cower_updates(self):
         except self.py3.CommandError as ce:
             return len(ce.output.splitlines())
 
+    def _get_trizen_updates(self):
+        try:
+            updates = self.py3.command_output(["trizen", "-Sua"])
```
lasers: Please check if we want `Sua` or `Qua` optionally with `q` (quiet). 
1) https://github.com/trizen/trizen/issues/160
2) Executing `trizen -Qua` gave me an interactive prompt.
3) For 1 update, `Qua` gave me 1. `Sua` gave 4 (due to prompt lines).

py3status/modules/arch_updates.py
```diff
@@ -14,8 +14,10 @@
 
 Requires:
     pacman-contrib: contributed scripts and tools for pacman systems
-    auracle: a flexible command line client for arch linux's user repository
-    yay: yet another yogurt. pacman wrapper and aur helper written in go
+    one of the following AUR helpers:
```
lasers: We'll see if this works okay in readthedocs. Ty for cleaning up trailing whitespace.

py3status/modules/arch_updates.py
dpeukert: The prompt is only displayed in interactive shells (https://github.com/trizen/trizen/blob/master/trizen#L2570-L2580) so that shouldn't be a problem. From a quick look at the code, it actually looks like both commands call the same function, but `-Q` (query) is probably a more logical choice (I used `-Sua` because of trizen/trizen#33). `-q` disables version number output which we don't need so that should fine to use.

dpeukert:
```
 _______
< -Quaq >
 -------
        \
         \
          🦆
```
`-Quaq` is also much more fun than `-Sua`.

py3status/modules/arch_updates.py
dpeukert: image
I'll change it back to the way it was.

py3status/modules/arch_updates.py
lasers: >From a quick look at the code, it actually looks like both commands call the same function

Sorry. I think 4 came from me forgetting to override aur with `trizen` when switching branches.

>but -Q (query) is probably a more logical choice (I used -Sua as recommended by trizen/trizen#33). -q disables version number output which we don't need so that should fine to use.

Agreed. `-Quaq` looks nice and is recent than `-Sua` (2017 vs 2018... by 5~ months or so).

>-Quaq is also much more fun than -Sua.

:+1:

dpeukert:
Turns out there is a difference between the two commands, `-Sua` exits with 0 if no updates are available and `-Qua` exits with 1. The problem is that it also exits with 1 if the command wasn't successful and therefore offers no way differentiate between the two. Fix in #1815.

ultrabug:
Thanks mate

PR #1814, opened 2019-06-04
Fix logo url in README

dpeukert:
Changed the logo URL in the README to point to the main repo and not Tobaloidee's fork, in case the fork gets deleted in the future.


ultrabug:
Very good catch @dpeukert !

PR #1823, opened 2019-07-15
Introduce entry point based discovery of packaged custom modules

### LINKED ISSUE #1822: Add possibility to register modules via entry points
As an additional way of adding modules to py3status we can use entry points to find modules that are installed in the same host interpreter like documented [here](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins).

Advantages of this approach: 
* a py3status module can be a real package installable via pip and bring its own dependencies. 
* the modules does not need to be contained in one file and does not need to be copied into one of the configuration directories and be free of dependencies
* modules can be developed independently of py3status with a different life cycle from py3status itself

obestwalter:
fixes #1822


doc/writing_modules.rst
```diff
@@ -27,6 +27,34 @@ which if you are used to XDG_CONFIG paths relates to:
...
+package from your module and publish it on PyPI.
+
+py3status can discover custom modules if they are installed in the same host interpreter
+and if an entry_point in the package ``setup.py`` is defined::
+
+    setup(
+        entry_points={"py3status": ["module = path.to.py3status_module"]},
+    )
+
+Although it is possible to install packages from PYPi in the distributed Python
```
ultrabug: `PyPI` typo

doc/writing_modules.rst
```diff
@@ -27,6 +27,34 @@ which if you are used to XDG_CONFIG paths relates to:
...
+interpreter it is recommended to install py3status into a virtualenv and install
+the custom packages into the same virtuelenv if you want to use this approach.
```
ultrabug: `virtualenv`

doc/writing_modules.rst
ultrabug: also, by `distributed` you mean `system` right?

doc/writing_modules.rst
```diff
@@ -27,6 +27,34 @@ which if you are used to XDG_CONFIG paths relates to:
...
+
+py3status can discover custom modules if they are installed in the same host interpreter
+and if an entry_point in the package ``setup.py`` is defined::
+
+    setup(
+        entry_points={"py3status": ["module = path.to.py3status_module"]},
+    )
+
+Although it is possible to install packages from PYPi in the distributed Python
+interpreter it is recommended to install py3status into a virtualenv and install
```
ultrabug: I'm not so sure about advocating people to install py3status in a virtualenv. This targets mostly "advanced" users because people expect the `py3status` command to be available on their system.

I don't want to start dealing with ppl complaining to not find their py3status "bin" or i3wm startup problems due to missing PATHs.

doc/writing_modules.rst
```diff
@@ -27,6 +27,34 @@ which if you are used to XDG_CONFIG paths relates to:
...
+
+If you have installed py3status in a virtualenv (maybe because your custom module
+has dependencies that need to be available) you can also create an installable
+package from your module and publish it on PyPI.
+
+py3status can discover custom modules if they are installed in the same host interpreter
+and if an entry_point in the package ``setup.py`` is defined::
+
+    setup(
+        entry_points={"py3status": ["module = path.to.py3status_module"]},
```
ultrabug: Would it be possible to provide a concrete example? I'm not sure I even understand what `path.to.py3status_module` should be I'm sorry

tests/test_user_modules.py
```diff
@@ -0,0 +1,54 @@
...
+    def return_fake_entry_points(*_):
+        class FakePy3status(object):
+            Py3status = "I am a fake class"
+
+            def __init__(self, name):
+                self.name = name
+                self.module_name = "module_name_" + self.name
+
+            @staticmethod
+            def load():
```
ultrabug: not sure why on python3.6 this breaks travis : https://travis-ci.org/ultrabug/py3status/jobs/558894076

doc/writing_modules.rst
obestwalter: True. That's not what I was trying to say anyway. I would say that creating an installable py3status package that is registered via entry point is already an advanced usage and in that advanced usage scenario it then makes sense to have py3status running in a virtualenv to avoid cluttering your host interpreter with depdendencies. I'll see if I can rewrite that with a warning to make this clearer. Will be very busy though during the next two weeks, but after that I can pick things up here and finish them.

tests/test_user_modules.py
obestwalter: ah - linting, should be fixed with last commit

doc/writing_modules.rst
obestwalter: Concrete eample is always good, we can use pew3wm module as example. In a nutshell it should point to the object that you want access to when iterating over the enty points (in our case the module where the Py3status class lives)

`entry_points={"py3status": ["module = pew3wm.pewpew"]},` - the pewpew module contains the `Py3status` class.

ultrabug: ok then I think having `path.to.py3status_module` is misleading

maybe change to `pypi_package_name.module_name`?

doc/writing_modules.rst
ultrabug: Maybe gimme write access to your fork, so I can try working on the phrasing mate?

obestwalter: Sure - just added you as a collaborator there, but I actually thought that allowing changes from maintainers would take care of that but seems not to be the case.

ultrabug: > I actually thought that allowing changes from maintainers would take care of that but seems not to be the case.

I'm sorry I'm not sure I understand?

obestwalter: I meant this: https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

ultrabug:
Just added logs and safety measures to entry_point loading!

Will continue on docs now, I love the mechanism :)

doc/writing_modules.rst
ultrabug: oohh first time I hear about that so I'm apologize I didn't even try before asking mate!

ultrabug:
@obestwalter I allowed myself to change the documentation, hope it feels right to you?

obestwalter:
Great! That's shaping up quite nicely. :tada:

ultrabug:
@obestwalter since you tagged this PR as WIP, do you see anything else that should be done before merging?

Thanks

obestwalter:
Regarding the use of entry points, I think all is fine as that is a quite standard way of doing it. As you now spruced up the docs, added logging and proper error handling, I guess we are good to go (that's basically why I had marked it WIP :)).

The only hypothetical caveat I would add is that accessing pkg_resources is slowing down startup time [by about a second](https://github.com/pypa/setuptools/issues/510), but working from my fork of py3status I don't notice any negative. You had already added an interesting hack to speed up entry point enumeration by adding [fatentrypoints](https://pypi.org/project/fastentrypoints/), so I guess we're good to go :+1:

objarni:
Bump, any reason why this isn't merged / published yet...?? Want it :+1:

ultrabug:
I'm in vacation, will merge and release new version when I come back in a bit more than a week!On Jul 26, 2019 19:08, Olof Bjarnason <notifications@github.com> wrote:Bump, any reason why this isn't merged / published yet...?? Want it 👍

ultrabug:
Here we are! Thanks!

PR #1832, opened 2019-08-19
Mpris module: Fix AttributeError: 'MockPy3statusWrapper' object has no attribute 'is_gevent'

valdur55:
Mpris module uses is_gevent but module_test doesn''t have such attribute.

Addresses #1827


ultrabug:
Thanks !

PR #1834, opened 2019-08-25
[wifi] device by MAC address as alternative to name

ownaginatious:
This is a pretty minor change that allows for selecting a WiFi device by its MAC address as an alternative to its name in the `wifi` module. Reason for wanting this is because there is currently an issue/dispute/race condition between `iwd` and `systemd-udevd` over who should control WiFi interface names.

Given that it's causing network interface names to be unpredictable, it seems like people are generally resorting to selecting a network interface by its MAC address instead now (e.g. with `systemd-networkd` config files). This simply adds the same to `py3status`.

### DIFF
--- a/py3status/modules/wifi.py
+++ b/py3status/modules/wifi.py
@@ -8,7 +8,8 @@ Configuration parameters:
     blocks: a string, where each character represents quality level
         (default "_▁▂▃▄▅▆▇█")
     cache_timeout: Update interval in seconds (default 10)
-    device: specify device name to use, otherwise auto (default None)
+    device: specify name or MAC address of device to use, otherwise auto
+        (default None)
     down_color: Output color when disconnected, possible values:
         "good", "degraded", "bad" (default "bad")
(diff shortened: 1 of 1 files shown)


X-dark:
This breaks automatic device detection. I have to explicitly set the device name, where it was working without any device setting before.

X-dark:
```
phy#0
	Unnamed/non-netdev interface
		wdev 0x4
		addr e4:xx:xx
		type P2P-device
		txpower 0.00 dBm
	Interface wlan0
		ifindex 3
		wdev 0x1
		addr e4:xx:xx
		ssid xxxx
		type managed
		channel 48 (5240 MHz), width: 20 MHz, center1: 5240 MHz
		txpower 17.00 dBm
		multicast TXQ:
			qsz-byt	qsz-pkt	flows	drops	marks	overlmt	hashcol	tx-bytes	tx-packets
			0	0	0	0	0	0	0	0		0
```

PR #1868, opened 2019-11-16
[module] salaah_time

azzamsa:
Hi.

I hope this module could be useful. :)

For more info: https://github.com/azzamsa/py3s-salaah-time

### DIFF
--- a/py3status/modules/salaah_time.py
+++ b/py3status/modules/salaah_time.py
@@ -1,0 +1,216 @@
+# -*- coding: utf-8 -*-
+"""
+Salaah Time.
+
+Configuration parameters:
+    longitude: longitude of your city
+    latitude: latitude of your city
+    timezone: timezone of your country (GMT+n)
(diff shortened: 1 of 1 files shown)


azzamsa:
This module is related to religion/believe, you can reject if this is not suitable for central repo. I can host it in my own repo with installation info.

ultrabug:
Hi @azzamsa 

First of all, thank you for your contributions mate!

Yes indeed, we try to not add too much modules to py3status and I believe that you would be interested to learn about how to publish one or a set of py3status modules on PyPi so that users could just install your package and use your modules without any other setup !

Here it is : https://py3status.readthedocs.io/en/latest/writing_modules.html#publishing-custom-modules-on-pypi

azzamsa:
> First of all, thank you for your contributions mate!

:)

> ... PyPi so that users could just install your package and use your modules without any other setup !

I'll go with this, awesome.

PR #1870, opened 2019-12-08
brightness : use logind as a fallback

boucman:
since v243 logind provides a dbus method to set the brightness.
This neither needs X nor root privilege.

Use logind as a fallback when xbacklight is not found

I considered having systemd as a third tool next to light and xbacklight, but I think 
having it as a fallback works better since it's not a dedicated tool but something everybody has.

I considered doing a direct dbus call instead of calling busctl, but there is no common way to 
deal with dbus currently, so I went for the easy way.

For reference
https://github.com/systemd/systemd/commit/b3be8c46af32b1135b0e118415eb2060bd69ae1a
https://github.com/systemd/systemd/pull/12413
https://github.com/systemd/systemd/pull/12424

Also note https://github.com/systemd/systemd/issues/14288 that might be a reason to delay this PR
(but still worth reviewing the current decisions)

### DIFF
--- a/py3status/modules/backlight.py
+++ b/py3status/modules/backlight.py
@@ -16,7 +16,8 @@ Configuration parameters:
     cache_timeout: How often we refresh this module in seconds (default 10)
     command: The program to use to change the backlight.
         Currently xbacklight and light are supported. The program needs
-        to be installed and on your path.
+        to be installed and on your path. If no program is installed, this
+        module will attempt to use logind support instead
         (default 'xbacklight')
     device: Device name or full path to use, eg, acpi_video0 or
         /sys/class/backlight/acpi_video0, otherwise automatic
@@ -32,11 +33,14 @@ Configuration parameters:
 Format placeholders:
     {level} brightness
 
-Requires:
+Requires: one of
     xbacklight: need for changing brightness, not detection
     light: program to easily change brightness on backlight-controllers
+    pydbus + logind v243:
+        python library for dbus access
+        logind allows to change brightness without X
 
-@author Tjaart van der Walt (github:tjaartvdwalt)
+@author Tjaart van der Walt (github:tjaartvdwalt), Jérémy Rosen (github:boucman)
 @license BSD
 
 SAMPLE OUTPUT
@@ -45,6 +49,11 @@ SAMPLE OUTPUT
 
 from __future__ import division
 import os
+
+try:
+    from pydbus import SystemBus
+except ImportError:
+    pass
 
 STRING_NOT_AVAILABLE = "no available device"
 
@@ -99,6 +108,14 @@ class Py3status:
         }
 
     def post_config_hook(self):
+        try:
+            self._logind_proxy = SystemBus().get(
+                bus_name="org.freedesktop.login1",
+                object_path="/org/freedesktop/login1/session/self",
+            )
+        except NameError:
+            self._logind_proxy = None
+
         if not self.device:
             self.device = get_device()
         elif "/" not in self.device:
@@ -124,13 +141,10 @@ class Py3status:
         except self.py3.CommandError:
             pass
 
-        if self.command_available and self.brightness_initial:
+        if self.brightness_initial:
             self._set_backlight_level(self.brightness_initial)
 
     def on_click(self, event):
-        if not self.command_available:
-            return None
-
         level = self._get_backlight_level()
         button = event["button"]
         if button == self.button_up:
@@ -147,7 +161,16 @@ class Py3status:
             self._set_backlight_level(level)
 
     def _set_backlight_level(self, level):
-        self.py3.command_run(self._command_set(level))
+        if self.command_available:
+            self.py3.command_run(self._command_set(level))
+            return
+        if self._logind_proxy:
+            for brightness_max_line in open("%s/max_brightness" % self.device, "rb"):
+                brightness_max = int(brightness_max_line)
+            brightness = brightness_max * level / 100
+            self._logind_proxy.SetBrightness(
+                "backlight", os.path.basename(os.path.normpath(self.device)), brightness
+            )
 
     def _get_backlight_level(self):
         if self.command_available:


maximbaz:
Wow, this is very nice!

> Use logind as a fallback when xbacklight is not found

I think we should consider this being the first option, not the fallback, what do you think? Less dependencies by default. Can we also retrieve the brightness value, as to get rid of dependencies altogether?

> I considered doing a direct dbus call instead of calling busctl, but there is no common way to deal with dbus currently, so I went for the easy way.

We have quite a few modules using `pydbus`, so I think we have a common way to deal with dbus at least in terms of this project, let's try to use it in this case as well? 🙂

boucman:
> > Use logind as a fallback when xbacklight is not found
> 
> I think we should consider this being the first option, not the fallback, what do you think? Less dependencies by default. 

well, the default value is xbacklight and the fallback happens when xbacklight is not here...
so with the current setup, it should work out of the box, whether xbacklight is here or not

>Can we also retrieve the brightness value, as to get rid of dependencies altogether?
>

It can't retrieve the value. The only thing it does is provide proxy to write to the `brightness` file.
Everything else we have to do by reading the /sys files ourselves, but that code was already in
place. 

So yes, reading without xbacklight works. It just doesn't use logind, it reads the value directly
 
> > I considered doing a direct dbus call instead of calling busctl, but there is no common way to deal with dbus currently, so I went for the easy way.
> 
> We have quite a few modules using `pydbus`, so I think we have a common way to deal with dbus at least in terms of this project, let's try to use it in this case as well?

yeah... but each module has to implement its own glib thread, which is kinda meh...

I'll reimplement it that way, yes, but I think i'll look into resurecting https://github.com/ultrabug/py3status/pull/1547

maximbaz:
Gotcha, makes sense 👍

boucman:
Note to self: also mention the change in the module documentation

boucman:
ok, so doing it properly was trivial, PTAL

py3status/modules/backlight.py
```diff
@@ -147,7 +152,15 @@ def on_click(self, event):
             self._set_backlight_level(level)
 
     def _set_backlight_level(self, level):
-        self.py3.command_run(self._command_set(level))
+        if self.command_available:
+            self.py3.command_run(self._command_set(level))
+            return
+        for brightness_max_line in open("%s/max_brightness" % self.device, "rb"):
+            brightness_max = int(brightness_max_line)
```
maximbaz: I know you copied this code from below, but what kind of weird syntax is this, we loop over lines and overwrite previous values? 🤔 I wonder if we should make it more readable somehow...

boucman: well... using a for loop to read all lines of a file is a classic. In this case the file is guaranteed to only contain one line (kernel API) so the cod is correct..

Alternative would  be
```python
with open(...) as f: brightness_max = int(f.read()) 
```
that might be consider cleaner but it's a matter of taste at that point

maximbaz: In 22 days, when python2 reaches EOL, this can simply become something like `int(Path(...).read_text())` 😄 

But ok, at least I see how the current code works, if we were to refactor here, let's do it separately anyways.

boucman:
after anayzis, the problem is "kinda" a busctl...

logind has to as dbus-daemon to chekc the identitty of the caller before writing to /sys (to check you have proper permissions) but when logind does the check, busctl is already gone, so there is no PID left to check.

in our case, we should ideally call SetBrightness with the "no_repy" flag, but I couldn't find how to set that flag in the pydbus code... I'm not sure it's implemented.

ultrabug:
Comment from IRC : @boucman: you're making pydbus + logind mandatory instead of using them as a fallback if no command is available

I'd prefer respecting current users by not adding a new dependency unless it becomes a core feature of py3status (which it is not yet)

So please use logind + its deps as a fallback of unavailable commands

maximbaz:
This code only gets executed if no other commands are available though, the `set` method starts with this:

image

lasers:
Try/except `from pydbus import SystemBus` to see if we can `logind`. Otherwise, mandatory dep.

maximbaz:
Aaaaah that's what you mean - yes sure, makes sense 👍

ultrabug:
@lasers was faster to reply than me :)

(btw, gonna release 3.22 now)

py3status/modules/backlight.py
```diff
@@ -128,9 +143,6 @@ def post_config_hook(self):
             self._set_backlight_level(self.brightness_initial)
```
maximbaz: the condition that leads to this line needs to be adjusted too I think, removing `self.command_available`

py3status/modules/backlight.py
```diff
@@ -45,6 +49,10 @@
 
 from __future__ import division
 import os
+try:
+    from pydbus import SystemBus
+except ModuleNotFoundError:
+    None
```
maximbaz: nit: `pass` would probably be more natural in this case

boucman: right... i'm not very good at python. Still learning the coding style...

boucman:
should be good now

py3status/modules/backlight.py
```diff
@@ -32,9 +33,12 @@
...
 
-Requires:
+Optional: one of
     xbacklight: need for changing brightness, not detection
     light: program to easily change brightness on backlight-controllers
+    pydbus + logind v243:
+        python library for dbus access
+        logind allows to change brightness without X
 
 @author Tjaart van der Walt (github:tjaartvdwalt)
```
ultrabug: You deserve to add yourself to the liste mate

py3status/modules/backlight.py
```diff
@@ -32,9 +33,12 @@
 Format placeholders:
     {level} brightness
 
-Requires:
+Optional: one of
```
ultrabug: Please replace "optional" by "requires"

`Requires: one of`

ultrabug:
Thanks a lot @boucman

boucman:
Thx to you... now moving on to dbus integration :)

PR #1871, opened 2019-12-08
Add support for the Strawberry music player & Add playback_status format placeholder to Clementine module

NJBS:
- Adds a new module for the [Strawberry](https://www.strawberrymusicplayer.org/) music player
- Adds a `playback_status` format placeholder for the Clementine & Strawberry modules

The modules are very similar so I'm not sure if it would be preferable to combine them somehow.

### DIFF
--- a/py3status/modules/clementine.py
+++ b/py3status/modules/clementine.py
@@ -8,6 +8,7 @@ Configuration parameters:
 
 Format placeholders:
     {current} currently playing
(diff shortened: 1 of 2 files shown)


ultrabug:
Yes please make it happen on `mpris`

NJBS:
Oh, I didn't even realize `mpris` was a module. Indeed it already does exactly what I was going for here already. Thanks :).

PR #1874, opened 2019-12-13
implemented real force_on_change #1873

Andrwe:


root360-AndreasUlm:
> Are you by any chance using py3-cmd refresh --all or similiar?

I'm not using this command.
I'm just starting py3status using the following i3-config line:
`status_command /home/user/pyenv/py3status/bin/py3status -c /home/user/.config/i3status/config_$(hostname) -l /home/user/py3status.log`
The command is running within a python virtual env to allow running latest version.

> What about python-pyudev? Do you have that installed too?

As the module has no dependency for python-pyudev in pip this module is not installed:
```
$ /home/a_ulm/pyenv/py3status/bin/pip show py3status | grep Requi
Requires: 
Required-by:
$ /home/user/pyenv/py3status/bin/pip list | grep udev
$
```

Unfortunately I cannot install pyudev on the system as it's a company laptop without root privileges.

ultrabug:
@root360-AndreasUlm I think because your setup does not support pyudev, periodic refreshes happen and indeed resets your click chosen layout.

When pyudev is in place, xrandr does only update on change so the problem you mention is not present.

Please note that py3status itself does not specify any dependency used by modules, that's why you can't see it using pip.

I will try this without udev and merge if it works good; thanks !

ultrabug:
Finally tested; sorry it took so long, works perfect for people without udev support!

PR #1879, opened 2020-01-07
IMAP: add OAuth2 support

justyn:
Hi,

I've added OAuth support to the IMAP module.

Starting in June, Google is [apparently going to stop](https://gsuiteupdates.googleblog.com/2019/12/less-secure-apps-oauth-google-username-password-incorrect.html) general apps from using password authentication on G Suite accounts. They'd like people to use OAuth instead.

This patch is quite similar to the Google Calendar module (and the same client secret should be able to be used) but it uses google-auth and google-auth-oauthlib instead of oauth2client because apparently [oauth2client is deprecated](https://oauth2client.readthedocs.io/en/latest/).

Let me know if you need any changes made.

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -4,7 +4,13 @@ Display number of unread messages from IMAP account.
 
 Configuration parameters:
     allow_urgent: display urgency on unread messages (default False)
+    auth_scope: scope to use with OAuth2 (default 'https://mail.google.com/')
+    auth_token: path to where the pickled access/refresh token will be saved
+        after successful credential authorization.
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks a lot for your contribution @justyn !

As you can see, the first step would be to make sure that your proposal pass the CI tests please?

justyn:
Hi @ultrabug 
Thanks, do you have any ideas on why the tests would pass on Pythons 2.7, 3.4, 3.5, 3.7 and 3.8 but fail on 3.6?

ultrabug:
Awesome contribution thank you so much @justyn !

PR #1880, opened 2020-01-07
Fix tests for python 3.8

maximbaz:
Finally had time to figure out why unit tests have been failing for me on Arch since the last release, but Travis was green.

https://docs.python.org/3/library/ast.html

> Deprecated since version 3.8: Class ast.Constant is now used for all constants.
Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases.

Adding python 3.8 to testing matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+  - "3.7"
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


ultrabug:
Ooooh, thanks @maximbaz !

PR #1881, opened 2020-01-11
imap module: implement retry logic (works around badly configured mail servers)

girst:
This implements some changes I've had running for over a year now, but never had the time to clean up and submit here (partly, because I've finally completed the move to a better mail host).

In essence, some mail providers (cough, cough, yahoo, cough) are just plain awful at keeping connections (automatic logout after a few minutes of not issuing a command,  randomly just refusing to log in, ...). This implements a fail-safe retry logic (retry two times, then error out).

I've opted to not expose the retry count and delay to the user, since configuring this doesn't make much of a difference, but I'm open to changes.

while I was at it, I also opted to refactor the long log messages to use String.format() instead of string concatenation. It makes sense to look at each commit on its own, since the last one adds a lot of noise due to the reformatting. Feel free to squash the commits on merging.

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -258,7 +258,9 @@ class Py3status:
             except socket_error:
                 raise imaplib.IMAP4.abort("Server didn't respond to 'IDLE' in time")
             if not response.lower().startswith("+ idling"):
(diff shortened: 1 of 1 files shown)


ultrabug:
Hey @girst thanks for another cool contribution mate!

That sounds perfect

PR #1882, opened 2020-01-17
imap module: guard against possible race condition in IDLE thread

justyn:
The IDLE thread indicates that a change has occurred (and so new mail
should be checked for) by refreshing the module and ending itself.

Because on refresh the module checks for the existence of the IDLE
thread to see if mail should be checked, there is a chance that the
module is already refreshing before the thread has completely finished,
and therefore the mail would not be checked.

To guard against this an attribute idle_thread_ending is set before
refreshing the module with the IDLE thread.

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -120,6 +120,7 @@ class Py3status:
             0  # IMAPcommands are tagged, so responses can be matched up to requests
         )
         self.idle_thread = Thread()
+        self.idle_thread_ending = False
 
         if self.client_secret:
             self.client_secret = os.path.expanduser(self.client_secret)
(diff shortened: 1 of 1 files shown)


justyn:
I can update this after PR #1881 is merged.

girst:
> Because on refresh the module checks for the existence of the IDLE
thread to see if mail should be checked,

No, it doesn't: the thread is started when the module is loaded and will keep
 running in the background (notice the `while True` around the whole of `_get_mail_count()`). If the thread dies (due to an exception being thrown), we fall back to polling for 1 cycle (we then restart the thread). This was an inte
ntional decision, so that we back off a little on errors.

Have you encountered any problems due to this in practice?

justyn:
Ah yes, I see now.

Apologies, I'm still getting to grips with the code (and py3status).

I had experienced a situation where the imap module appeared to have frozen (as in, it was no longer updating) and it wasn't clear why. I haven't been able to reproduce this though. I was looking through the code to see if there were any contenders, and thought I spotted this.

Thanks for the input, I'll close this PR.

girst:
On Fri, Jan 17, 2020 at 10:32:18AM -0800, Justyn Butler wrote:
>Ah yes, I see now.
>
>Apologies, I'm still getting to grips with the code (and py3status).

no worries! ;-)

>I had experienced a situation where the imap module appeared to have 
>frozen (as in, it was no longer updating) and it wasn't clear why. I 
>haven't been able to reproduce this though. 

it's very likely that the retry-logic pull request will fix it for you 
(I've been running a hacked-up version of it for well over a year and 
never had problems--my apologies for keeping it for myself)

PR #1884, opened 2020-01-17
mpd_status using idle (major cpu savings)

girst:
Issuing commands every two seconds, even if nothing changes causes high
CPU load with Mopidy-MPD, and possibly other MPD servers. With polling
every 2 seconds (the default), mopidy (while paused) was consuming
between 10% and 15%, with occational spikes even up to 20% cpu; using
idling, this number dropped to 0.0%.

This patch is similar to IDLE in imap.py, where we listen for changes on
subsystems we're interested in (99% that's just 'player', but we'll
default to 'mixer' (changes in volume) and 'options' (e.g. repeat mode)
too, just in case a user has those states displayed in her status bar).

### DIFF
--- a/py3status/modules/mpd_status.py
+++ b/py3status/modules/mpd_status.py
@@ -202,74 +202,74 @@ class Py3status:
         return response
 
     def _get_status(self):
-      while True:
-        try:
-            status = self._get_mpd().status()
-            song = int(status.get("song", 0))
-            next_song = int(status.get("nextsong", 0))
+        while True:
+            try:
+                status = self._get_mpd().status()
+                song = int(status.get("song", 0))
+                next_song = int(status.get("nextsong", 0))
 
-            state = status.get("state")
+                state = status.get("state")
 
-            if (state == "pause" and self.hide_when_paused) or (
-                state == "stop" and self.hide_when_stopped
-            ):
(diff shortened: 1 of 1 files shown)


girst:
rebased on top of master. (not that it makes too much of a difference)

py3status/modules/mpd_status.py
```diff
@@ -214,6 +201,76 @@ def attr_getter(attr):
...
+                    changed = self._get_mpd().idle()
+                    while not any([c in self.idle_subsystems for c in changed]):
+                        changed = self._get_mpd().idle()
+                else:
+                    return
+
+            except (ValueError, socket.error, ConnectionError, CommandError) as e:
+                # ValueError can happen when status.get(...) returns None; e.g.
+                # during reversal of playlist
+                if type(e) is ValueError:
```
ultrabug: Can we use `isinstance` instead?

girst: done

ultrabug:
Thank you!

girst:
On Sat, Jan 18, 2020 at 01:33:34PM -0800, Ultrabug wrote:
>Thank you!

thanks for your maintenance service, too!

PR #1885, opened 2020-01-19
RFC: implement degraded_on_stale

girst:
see discussion in #1883

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -67,7 +67,7 @@ import os
 from threading import Thread
 from time import sleep
 from ssl import create_default_context
-from socket import error as socket_error
+from socket import setdefaulttimeout, error as socket_error
 
 STRING_UNAVAILABLE = "N/A"
 NO_DATA_YET = -1
@@ -209,10 +209,12 @@ class Py3status:
         if self.client_secret:
             # Use OAUTH
             self._get_creds()
+        setdefaulttimeout(self.read_timeout)
         connection = imaplib.IMAP4_SSL(self.server, int(self.port))
         return connection
 
     def _connection_starttls(self):
+        setdefaulttimeout(self.read_timeout)
         connection = imaplib.IMAP4(self.server, int(self.port))
         connection.starttls(create_default_context())
         return connection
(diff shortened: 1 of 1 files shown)


girst:
I've tacked on a second change (properly setting a timeout when opening the socket the first time)

ultrabug:
I must admit that I got lost in all your work with @justyn on #1886 so I do not know if you reached an optimal yet or if I should wait for your signal guys

girst:
On Sun, Jan 26, 2020 at 05:01:08AM -0800, Ultrabug wrote:
>I must admit that I got lost in all your work with @justyn on #1886 so I do not know if you reached an optimal yet or if I should wait for your signal guys

Yeah, that's a lengthy brainstorm. Essentially, we discovered a stall 
when connecting to captive portal networks, which I'm fairly confident 
(*knocks on wood*) 27b1991 fixes.

I'd like to see some confirmation about this from justyn, but otherwise 
I believe this has reached an optimum and is good to go.

justyn:
> Yeah, that's a lengthy brainstorm. Essentially, we discovered a stall when connecting to captive portal networks, which I'm fairly confident (*knocks on wood*) [27b1991](https://github.com/ultrabug/py3status/commit/27b1991dd6d946c453cabd2b23bde90f97109a7f) fixes. I'd like to see some confirmation about this from justyn, but otherwise I believe this has reached an optimum and is good to go.

Sorry I couldn't reply properly before.

TLDR: yes this latest change seems to fix the problem! Great stuff @girst!

I did wonder if the two calls to setdefaulttimeout could be combined?

girst:
On Sun, Jan 26, 2020 at 05:34:11AM -0800, Justyn Butler wrote:
>Sorry I couldn't reply properly before.

no worries!

>TLDR: yes this latest change seems to fix the problem! Great stuff @girst!

great, this gets the green light from me, then.

>I did wonder if the two calls to setdefaulttimeout could be combined?

I'd rather have them as close to the call to imaplib.IMAP{,_SSL} to make 
it clear that this sets the socket timeouts inside imaplib.

ultrabug:
Hi guys, did you reach a consensus on what to do to move on plz?

girst:
On Mon, Feb 03, 2020 at 11:42:22AM -0800, Ultrabug wrote:
>Hi guys, did you reach a consensus on what to do to move on plz?

that's good to go!

ultrabug:
Thanks @girst

PR #1896, opened 2020-01-25
Drop support for EOL Python 2.6-3.4

hugovk:
* Add `python_requires` to help pip
* Upgrade Python syntax with https://github.com/asottile/pyupgrade/ `--py3-plus`

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 import re
 
 from py3status.autodoc import (
--- a/doc/example_module.py
+++ b/doc/example_module.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 # This is an example module to be used as a template.
 # See https://github.com/ultrabug/py3status/wiki/Write-your-own-modules
 # for more details.
--- a/fastentrypoints.py
+++ b/fastentrypoints.py
@@ -74,8 +74,7 @@ def get_args(cls, dist, header=None):  # noqa: D205,D400
             )
             # pylint: disable=E1101
             args = cls._get_script_args(type_, name, header, script_text)
-            for res in args:
-                yield res
+            yield from args
 
 
 # pylint: disable=E1101
--- a/py3status/__init__.py
+++ b/py3status/__init__.py
@@ -8,13 +8,6 @@ try:
 except ImportError:
     pass
 
-try:
-    # python3
-    IOPipeError = BrokenPipeError
-except NameError:
-    # python2
-    IOPipeError = IOError
-
 
 def main():
     from py3status.argparsers import parse_cli_args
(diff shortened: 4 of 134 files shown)


tirkarthi:
I am generally against large changes like this since it reduces the usefulness of git blame. I will leave it to the maintainers on the same. Thanks.

ultrabug:
while fuzzing a bit with git blame doesn't shock me too much I think this move deserves two steps

1- remove support on tox + setup.py + **add a startup notification to users running 2.7 that next version will not work for them**

release

2- do all the rest you propose here

would that sound okay?

hugovk:
Sounds good, will make a new PR!

hugovk:
> 1- remove support on tox + setup.py + **add a startup notification to users running 2.7 that next version will not work for them**

A couple of questions:

This PR removes support for 2.6, 2.7, 3.2, 3.3 and 3.4.

1. Should the new PR only remove classifiers for Python 2, or all of those?

2. Similarly, should the notification only be for users running Python 2, or all of those?

hugovk:
Actually, if the next release still supports a given version, those classifiers should still be present to indicate support.

And likewise those versions should also be tested, to know that it really does support them.

Another question:

Would you like the notification in setup.py (install time) or \_\_init__.py (run time)?

One thing to point out: if the last release warns users that it is the last release to support a given version, people "stuck" on that version will always get that warning. So perhaps install time is best?

(Many projects don't worry about warning in advance, and let `python_requires` do the heavy lifting: it means those with modern pip will get the most recent release that supports their running Python version.)

ultrabug:
I apologize @hugovk I'm short in time lately, please bear with me

hugovk:
Sure, there's no rush, take your time!

ultrabug:
@hugovk

Overall I would prefer we keep this PR short on dropping support only, not all those code changes please?

> Should the new PR only remove classifiers for Python 2, or all of those?

On documentation + setup.py, let's drop everything <3.5

> Actually, if the next release still supports a given version, those classifiers should still be present to indicate support.
> And likewise those versions should also be tested, to know that it really does support them.

I think the code could still support it for the next version but should drop testing etc. A bit like "we do not officially support that anymore".

> Similarly, should the notification only be for users running Python 2, or all of those?
> Would you like the notification in setup.py (install time) or __init__.py (run time)?
> (Many projects don't worry about warning in advance, and let python_requires do the heavy lifting: it means those with modern pip will get the most recent release that supports their running Python version.)

That sounds reasonable to not warn then okay.

Does that sound okay?

hugovk:
Sounds fine to me. 

I'll make a new PR with only https://github.com/ultrabug/py3status/pull/1896/commits/8df834da2111da28c50472bf792d6c88c47656e0 + https://github.com/ultrabug/py3status/pull/1896/commits/368a631c6ed8ca1bb1a65ca8cc3d0322baefde6b + just the RST change of https://github.com/ultrabug/py3status/pull/1896/commits/5b10a85bffe8bf0c5299549158d0b5418fa09249.

ultrabug:
First batch merged & 3.26 released, thanks

ultrabug:
@hugovk can you rebase this plz?

hugovk:
Rebased!

ultrabug:
Thanks a lot @hugovk and sorry for the delay!

hugovk:
You're welcome, no problem, and thank you for maintaining!

PR #1899, opened 2020-01-30
Fix imap idle for courier

nylocx:
As stated in RFC 2177 (https://tools.ietf.org/html/rfc2177)  
> The server requests a response to the IDLE command using the continuation ("+") [...]

For example curier responds with "+ entering idle mode" instead of "+ idling".

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -257,6 +257,8 @@ class Py3status:
                 response = socket.read(4096).decode("ascii")
             except socket_error:
                 raise imaplib.IMAP4.abort("Server didn't respond to 'IDLE' in time")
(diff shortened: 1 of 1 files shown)


girst:
lgtm, thanks!

girst:
One thing: can you please add a comment like the following above the changed line, so future developers know what kind of message is expected to be transmitted?

    # "+ idling", "+ entering idle mode", ...

ultrabug:
Thanks a lot for your first contribution @nylocx and @girst for the review!

Would you care to add @girst comment please?

nylocx:
Oh, sorry completely missed that comment about the comment, I will fix it tomorrow morning.

PR #1900, opened 2020-01-31
frame module: expose output of submodules

MikaYuoadas:
Hello,

It's a simple proposition to expose the submodules output that can then be used in conditional formats (see docstring for one example of such use case).


ultrabug:
Bonjour @MikaYuoadas !

Thanks a lot for your proposal, please bear with me while I get the time to review it properly...

ultrabug:
Thank you @MikaYuoadas , nicely done

PR #1904, opened 2020-02-09
Drop support for EOL Python 2.6-3.4

hugovk:
A shorter version of https://github.com/ultrabug/py3status/pull/1896, as discussed there.


doc/contributing.rst
```diff
@@ -128,8 +127,6 @@ after a reload.
 Python versions
 ---------------
 
-py3status code, including modules, should run under both python 2 and python 3.
-
 Starting with version 4.0, py3status will only run using python 3.
```
ultrabug: Can you say 3.26 instead?

setup.py
```diff
@@ -49,20 +49,17 @@ def read(fname):
...
         ]
     },
+    python_requires=">=3.5",
     classifiers=[
         "License :: OSI Approved :: BSD License",
         "Operating System :: POSIX :: Linux",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 2.6",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
```
ultrabug: I guess we can remove this, no?

hugovk: It can, but it's very commonly used to mean _any_ 3.x is supported:

https://pypi.org/search/?c=Programming+Language+%3A%3A+Python+%3A%3A+3

There's another classifier for _exactly_ 3.0:

https://pypi.org/search/?c=Programming+Language+%3A%3A+Python+%3A%3A+3.0

doc/contributing.rst
hugovk: Done in https://github.com/ultrabug/py3status/pull/1904/commits/81a935c52d9fb575dbdf96586ae246e80b4e64ee!

setup.py
ultrabug: When this first iteration of cleanup will have taken place, we will drop support for 3.5 and thus will need to remove this

But for now, that's good yep

ultrabug:
Thanks a lot!

Now I will release a new version for this only

PR #1910, opened 2020-03-03
Adding "locale" config parameter on clock module.

inemajo:
Time format use locale for displaying some options (like %a, %A ...), this
parameter is used to override the system locale.

Example:

clock master {
    cycle = 30
    locale = "fr_FR"
    format = ["{Europe/Paris}"]
    format_time = "%a %d/%m/%Y {icon} %H:%M"
}

Override the system locale to "fr_FR" locale.


ultrabug:
Hello @inemajo 

Thanks for your first & cool contribution to the project. I apologize for the reply delay!

This indeed sounds like a simple yet efficient feature :+1:

PR #1921, opened 2020-03-16
NetworkManager: allow using the currently active AP in formats

bendardenne:
Currently, you can only configure your format with access points numbers, e.g.  `ap1_ssid`, `ap2_bars`, ...

This works fine most of the time because `nmcli` only returns one AP, the currently active one. So most of the time the `ap1_xxx` placeholders make sense. However, occasionally, `nmcli` may return additional APs (for instance when a scan has been performed). In that case, AP1 may not always be the currently active AP, and if your format contains `ap1_ssid` py3status will show you as being connected to an incorrect SSID. 

This change allows to use `ap_ssid`, `ap_rate` instead of `ap1_ssid` etc. so that py3status always shows the information related to the currently used AP, and not just a specific AP number.

### DIFF
--- a/py3status/modules/networkmanager.py
+++ b/py3status/modules/networkmanager.py
@@ -165,7 +165,6 @@ class Py3status:
                         current_ap[key.replace(used_ap + "_", "ap_")] = device[key]
                 device.update(current_ap)
 
-
             for x in self.thresholds_init:
                 if x in device:
                     self.py3.threshold_get_color(device[x], x)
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @bendardenne thanks for your first contribution!

Could you please rebase it on master?

Thanks!

ultrabug:
Tested ok, backward compatible so shouldn't break current users special config

Merci @bendardenne

PR #1923, opened 2020-03-30
add the filter ability of the mail module to the imap module

### LINKED ISSUE #1902: imap: add filtering like in mail plugin
it would be nice to have mailbox filtering in the imap plugin like in the mail plugin, so that it is possible to filter for example for all mailboxes that don't contain spam

L0ric0:
adds a new parameter `filter` which can be a list of regular expressions. If the `mailbox` parameter is not set to an empty string the old behavior will be used.

fix #1902

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -383,17 +383,13 @@ class Py3status:
                                 STRING_INVALID_FILTER.format(filters),
                                 self.py3.CACHE_FOREVER,
                             )
-                        for line in lines:
-                            self.py3.log(line)
+                        if self.debug:
+                            for line in lines:
+                                self.py3.log(line)
 
                 for directory in directories:
-                    if (
-                        self.connection.select(directory, readonly=True)[0]
-                        == "OK"
-                    ):
-                        unseen_response = self.connection.search(
-                            None, self.criterion
-                        )
+                    if self.connection.select(directory, readonly=True)[0] == "OK":
+                        unseen_response = self.connection.search(None, self.criterion)
                         mails = unseen_response[1][0].split()
                         tmp_mail_count += len(mails)
                     else:
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -14,6 +14,23 @@ Configuration parameters:
     criterion: status of emails to check for (default 'UNSEEN')
     debug: log warnings (default False)
     degraded_when_stale: color as degraded when updating failed (default True)
+    filter: You can specify a list of filters to decide which folders to search.
+        By default, we search only the INBOX folder (ie: `['^INBOX$']`). We
+        can use regular expressions, so if you use more than one, it would be
+        joined by a logical operator `or`.
+
+        `'.*'` will match all folders.
+        `'pattern'` will match folders containing `pattern`.
+        `'^pattern'` will match folders beginning with `pattern`.
+        `'pattern$'` will match folders ending with `pattern`.
+        `'^((?![Ss][Pp][Aa][Mm]).)*$'` will match all folders
+        except for every possible case of `spam` folders.
+
+        For more documentation, see https://docs.python.org/3/library/re.html
+        and/or any regex builder on the web. Don't forget to escape characters.
+
+        The filters will only be used if the mailbox configuration value is
+        set to an empty value.
     format: display format for this module (default 'Mail: {unseen}')
     hide_if_zero: hide this module when no new mail (default False)
     mailbox: name of the mailbox to check (default 'INBOX')
@@ -64,12 +81,15 @@ SAMPLE OUTPUT
 """
 import imaplib
 import os
+import re
+from csv import reader
 from threading import Thread
 from time import sleep
 from ssl import create_default_context
 from socket import setdefaulttimeout, error as socket_error
 
 STRING_UNAVAILABLE = "N/A"
+STRING_INVALID_FILTER = "invalid imap filters `{}`"
 NO_DATA_YET = -1
 
 
@@ -86,6 +106,7 @@ class Py3status:
     criterion = "UNSEEN"
     debug = False
     degraded_when_stale = True
+    filter = ["^INBOX$"]
     format = "Mail: {unseen}"
     hide_if_zero = False
     mailbox = "INBOX"
@@ -130,6 +151,8 @@ class Py3status:
 
         if self.security not in ["ssl", "starttls"]:
             raise ValueError("Unknown security protocol")
+        if not isinstance(self.filter, list):
+            raise Exception(STRING_INVALID_FILTER.format(filter))
 
     def imap(self):
         # I -- acquire mail_count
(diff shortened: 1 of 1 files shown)


py3status/modules/imap.py
```diff
@@ -325,19 +360,58 @@ def _get_mail_count(self):
...
                 tmp_mail_count = 0
-                directories = self.mailbox.split(",")
+                if self.mailbox:
+                    directories = self.mailbox.split(",")
+                else:
+                    try:
+                        directories
+                    except:
+                        directories = []
+                        filters = "|".join(self.filter)
```
girst: why not just pass a single regex in filter?

py3status/modules/imap.py
```diff
@@ -325,19 +360,58 @@ def _get_mail_count(self):
...
+                        objs = [x.decode() for x in self.connection.list()[1]]
+                        folders = [x[-1] for x in reader(objs, delimiter=" ")]
```
girst: note: connection.list() returns the unwieldy
`[b'(\\HasChildren) "." INBOX', b'(\\HasNoChildren \\Sent) "." INBOX.Sent']`. i'm not thrilled about importing `csv`, but i concede that i don't have a better solution off hand.

py3status/modules/imap.py
```diff
@@ -325,19 +360,58 @@ def _get_mail_count(self):
...
+                                directories = []
+                                break
+                            lines.append("[{}] {}".format(subscribed, name))
+                        if not directories:
+                            self.py3.error(
+                                STRING_INVALID_FILTER.format(filters),
+                                self.py3.CACHE_FOREVER,
+                            )
+                        for line in lines:
+                            self.py3.log(line)
```
girst: that should be removed, otherwise the log gets spammed constantly. same goes for `lines = ["===== IMAP ====="]`, and everything that's connected to it (e.g. '`subscribed`)

py3status/modules/imap.py
```diff
@@ -325,19 +360,58 @@ def _get_mail_count(self):
...
+                        objs = [x.decode() for x in self.connection.list()[1]]
+                        folders = [x[-1] for x in reader(objs, delimiter=" ")]
+                        lines = ["===== IMAP ====="]
+                        for name in folders:
+                            subscribed = " "
+                            try:
+                                if re.search(filters, name):
+                                    subscribed = "x"
+                                    folder = name.replace("\\", "\\\\")
+                                    folder = folder.replace('"', '\\"')
```
girst: you can use `r''`-strings to prevent backslash escapes

py3status/modules/imap.py
girst: actually, i'm pretty sure your approach breaks on the second item in my inbox (there's a non-quoted space there).

py3status/modules/imap.py
L0ric0: that would certainly be possible but i think that 
```
filter = ['pattern']
filter += ['pattern2']
...
```
is a lot more readable than 
```
filter = 'pattern|pattern2'
```

py3status/modules/imap.py
L0ric0: the code here is the same as the one in the `mail` module. I don't know how much it was tested there. 
Can you get the result of the `list()` operation for the item from the server, so that i can try to fix it?

py3status/modules/imap.py
L0ric0: it should only be printed once as in line 374,375 the try checks if the variable directorys is defined and if it is not runs the regexes. This is also important as the regex can be quite heavy on cpu. It would be possible to put it behind a debug flag but i would leave it in there to be able to debug the regex.

L0ric0:
i am sorry about the git history but i don't know how to cut it down to one commit.
The formating I didn't notice as my editor runs black on saving the file.

L0ric0:
the usecase i have is that i create mailboxes with
```
elsif allof(address :contains :domain "from" "github.com",
    header :regex "subject" "^(Re: )?\\[([^/]*)\\/([^]]*)\\]") {
        set "user" "${2}";
        set "repository" "${3}";
        fileinto :create "GitHub/${user}/${repository}";
}

girst:
On Mon, Mar 30, 2020 at 11:35:03AM -0700, L0ric0 wrote:
>the usecase i have is that i create mailboxes with

can you please post a list of mailbox names this generates?

girst:
while i'm still not convinced that this feature should be merged, i have 
created an experiment on my repository at 
https://github.com/fork-graveyard/py3status/ branch imapwildcard

you can use the IMAP wildcards % and * in the mailbox config.

please try if that works for you and report back.

L0ric0:
the mailbox names are `GitHub/<username>/<repository>` (i.e. `GitHub/ultrabug/py3status`)

L0ric0:
> while i'm still not convinced that this feature should be merged, i have created an experiment on my repository at https://github.com/fork-graveyard/py3status/ branch imapwildcard you can use the IMAP wildcards % and * in the mailbox config. please try if that works for you and report back.


with some modifications so that it runs this works for my use case. The problem I see is that we introduce more divergence between the imap and the mail module, as the mail module supports regex matching but no idle and the imap module supports wildcard matching and idle. maybe we could make the mail module import and run the imap module?

The other point that i worry about is that inverse matching is not possible with wildcards. So you can't filter out spam directories or single ones that you don't care about. 

Would you want to merge the wildcard support or the regex support both work for me. If you want to merge the wildcard support would it be ok for me to make some edits? For example that the wildcard matching only takes place in the beginning.

girst:
tl;dr: i believe this is out of scope for imap.py, but i'm open to 
discuss wildcards.

you are asking for features that i'm guessing only you will use 
(selecting hierarchial directory structures using regular expressions, 
"wildcards only in the beginning"). in such a case there is nothing 
wrong with keeping patches on top of py3status. i'm doing the same on my 
machine.

i have opened a new issue regarding wildcards in imap.py, and want to 
use this to get some kind of interest measurement.
https://github.com/ultrabug/py3status/issues/1925

On Tue, Mar 31, 2020 at 01:15:38AM -0700, L0ric0 wrote:
>with some modifications so that it runs this works for my use case. The 

send your changes as a pull request to my experimental branch please.  
i'll take a look.

>problem I see is that we introduce more divergence between the imap and 
>the mail module, as the mail module supports regex matching but no idle 
>and the imap module supports wildcard matching and idle. maybe we could 
>make the mail module import and run the imap module?

the divergence is on mail.py, since they are the second imap 
implementation in py3status.
idle support for mail.py was considered, but not done in the end, 
because it would require rewriting most of its code.

>Would you want to merge the wildcard support or the regex support both 
>work for me.

regular expressions: hard no from me. wrong tool for the job.
wildcards: maybe, iff we can find a way to implement it cleanly and 
reliably, and there is enough demand for it (it's not just about writing 
the code, but also about maintaining it).

>For example that the wildcard matching only takes place in the 
>beginning.

those wildcard operators are an IMAP thing, not something I cooked up.  
if we want to support it, we can either do it to spec [1] or not at 
all--everything else is confusing.

[1]: https://tools.ietf.org/html/rfc3501#section-6.3.8

L0ric0:
> those wildcard operators are an IMAP thing, not something I cooked up.
if we want to support it, we can either do it to spec [1] or not at
all--everything else is confusing.

yes i have seen this in your code. I mean that we only query the server once on startup for the directory names and not every time we check for mails, as i think that this wouldn't be needed. But I also don't know how heavy the list request is.

Ok as regex support is to much i will close this request

PR #1924, opened 2020-03-30
[MPD module] use currentsong command if possible

girst:
when playling a live stream with embedded title information (e.g.
through mopidy), the currentsong command can provide the embedded track
metadata like the currenlty playing title, instead of just the filename.

**Edit**: looks like travis isn't talking to github right now. [tests passed](https://travis-ci.org/github/ultrabug/py3status/builds/668850084).


ultrabug:
Simple enough, thanks mate!

PR #1926, opened 2020-04-02
imap.py: mailbox wildcard support

### LINKED ISSUE #1902: imap: add filtering like in mail plugin

### LINKED ISSUE #1925: RFC: supporting wildcards in imap.py
This ticket tries to estimate demand for selecting mailboxes using [wildcards as specified by the imap rfc](https://tools.ietf.org/html/rfc3501#section-6.3.8). Please comment on your use case(s).

girst:
@L0ric0 and I worked out some code that we feel is solid.

Fixes #1902, fixes #1925, refrerences #1923.

### DIFF
--- a/py3status/modules/imap.py
+++ b/py3status/modules/imap.py
@@ -16,7 +16,8 @@ Configuration parameters:
     degraded_when_stale: color as degraded when updating failed (default True)
     format: display format for this module (default 'Mail: {unseen}')
     hide_if_zero: hide this module when no new mail (default False)
-    mailbox: name of the mailbox to check (default 'INBOX')
+    mailbox: name of the mailbox to check. can handle `%` and `*` wildcards
+        (default 'INBOX')
     password: login password (default None)
     port: number to use (default '993')
     read_timeout: timeout for read(2) syscalls (default 5)
@@ -246,7 +247,7 @@ class Py3status:
         finally:
             self.connection = None
 
-    def _idle(self):
+    def _idle(self, directories=["INBOX"]):
         """
         since imaplib doesn't support IMAP4r1 IDLE, we'll do it by hand
         """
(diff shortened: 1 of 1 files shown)


ultrabug:
Hello @girst 

To my own very surprise I've finally tested this and it does not affect the default behavior. I sincerely apologize for the delay.

But I can't get this to work on my IMAP server and there is no added documentation on how to do so...

This is what I get for example:

 ```
LIST command error: BAD [b"parse error: wrong character; expected ' ' but got '%'"] (error) module.py line 940.
```

Do you still have interest in this?

girst:
On Fri, Oct 30, 2020 at 12:39:33PM -0700, Ultrabug wrote:
>To my own very surprise I've finally tested this and it does not affect 
>the default behavior. I sincerely apologize for the delay.

yes, this is an opt-in patch. and there is no reason to apologize!

>
>But I can't get this to work on my IMAP server and there is no added 
>documentation on how to do so...

all it should take is modifying the mailbox= config option. the imap rfc 
lists some examples at <https://tools.ietf.org/html/rfc3501#page-42>.  

>
>This is what I get for example:
>
> ```
>LIST command error: BAD [b"parse error: wrong character; expected ' ' 
>but got '%'"] (error) module.py line 940.
>```

interesting™. can you share the mailbox name you configured and which 
imap server you're connecting to? afair, i tested this on dovecot and it 
worked. a stack trace of when the error occured would be useful as well.

>
>Do you still have interest in this?

to be frank, i've only written this patch as a saner alternative to 
other proposals floating around at the time and never had any interest 
in it myself. i can fix the patch up if we can diagnose the error you've 
run into, but i'm equally ok with closing this as wontfix and not 
accepting other wildcard-style patches for the imap module. by this 
point i believe i am the person most familiar with the imap.py code 
(having (re-)written most of it), and would rather not see it 
haphazardly extended without my input.

ultrabug:
Thanks @girst ; then with your permission I'm closing this one since we have no strong need of it.

Cheers

Issue #1915, opened 2020-03-06
Improve documentation of `volume_status` module


jose1711:
There are multiple occurrences of `backend` in https://py3status.readthedocs.io/en/latest/modules.html#volume-status and yet it does not say that it's defined by `command` parameter. User has no way to find out other than inspecting the source code.

Since in reality there are just two actual backends (`alsa` represented by `amixer` or `pulseaudio` represented by `pamixer` and `pactl`) a few hints on what to use for `device` (e. g. anything returned by `pactl list sinks | grep Name` command) would be nice.

PR #1932, opened 2020-04-14
volume_status: document commands to print possible devices

lasers:
This document commands to print possible devices. /cc @jose1711 and others. Will this do? 

Closes https://github.com/ultrabug/py3status/issues/1915.


ultrabug:
Thanks!

PR #1942, opened 2020-07-02
Remove deprectated function in modules.py

### LINKED ISSUE #1941: Handle return-type annotations
The current version seems to be unable to deal with code that has type annotations for return values.

```
↪ py3status --version
py3status version 3.28 (python 3.8.3) on i3
```

[This custome module](https://github.com/CryptoCopter/py3status-watson/blob/060184e6464ed90766330dc686ce045f47b46380/py3status-watson/watson.py) I have written will result in the following error whey trying to execute it:

```
↪ python3 watson.py
Module `test_module` could not be loaded
Function has keyword-only parameters or annotations, use inspect.signature() API which can support them
```

however, when I remove the return type annotations, it works fine.

The variable annotations don't seem create any issues.

CryptoCopter:
- `inspect.getargspec` was deprected in favor of
`inspect.getfullargspec`, which also supports return type annotations.
Fixes #1941

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -567,7 +567,7 @@ class Module:
         # on_click method has extra events parameter
         if method_name == "on_click":
             arg_count = 2
-        args, vargs, kw, defaults = inspect.getargspec(method)
+        args, vargs, kw, defaults, _, _, _ = inspect.getfullargspec(method)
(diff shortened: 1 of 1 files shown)


ultrabug:
Very good catch @CryptoCopter thanks a lot for your first contribution to the project!

PR #1943, opened 2020-07-07
Broken chromium mpris interface workaround

Jacotsu:
This pull request is basically a workaround that bypasses chromium broken mpris interface thus allowing py3status to show which media is currently being played in the browser.

**Bugs**:
- Control buttons don't work as chromium doesn't expose any interface in dbus, maybe can be fixed with some IPC
- If a player with higher priority was running the title of the old media will remain until you switch to a tab with a new media
- If playback is started before py3status the status bar won't be synchronized until a new media is played

At the moment i'm not planning on fixing those bugs because the most sensible way to do it would be to fix chromium's MPRIS.

### DIFF
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -137,21 +137,22 @@ class BrokenDBusMpris:
             def combined_function(*args):
                 callback(*self.filter_messages(*args))
 
-            self._subscription = self._dbus\
-                .subscribe(signal_fired=combined_function)
+            self._subscription = self._dbus.subscribe(signal_fired=combined_function)
 
         def disconnect(self):
             self._subscription.disconnect()
 
         # For some reason the dbus subscribe filtering doesn't work
         def filter_messages(self, *args):
-            dbus_params = ['/org/mpris/MediaPlayer2',
-                           'org.freedesktop.DBus.Properties',
-                           'PropertiesChanged']
+            dbus_params = [
+                "/org/mpris/MediaPlayer2",
+                "org.freedesktop.DBus.Properties",
+                "PropertiesChanged",
+            ]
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks for your first contribution to py3status @JackDoan !

It seems like your IDE has pushed quite a bunch of other changes unrelated to your proposal. Some of them are even breaking the CI.

Could you fix that first please?

JackDoan:
Hey, I think you tagged the wrong person. 

@Jacotsu

ultrabug:
apologizes @JackDoan :(

Jacotsu:
> Thanks for your first contribution to py3status @JackDoan !
> 
> It seems like your IDE has pushed quite a bunch of other changes unrelated to your proposal. Some of them are even breaking the CI.
> 
> Could you fix that first please?

Oh, I didn't know that you used black for formatting; the latest commit should be formatted correctly now. I've made some other changes like `if metadata.get("xesam:artist") is not None:` -> `if metadata.get("xesam:artist"):` to make the code more pythonic, but i can also revert those if needed

ultrabug:
Thank you @Jacotsu ; I wonder about the `BrokenDBusMpris` name : do you expect this code to work with other things than Chromium? Else why not call it `ChromiumDBusMpris` maybe?

Jacotsu:
> Thank you @Jacotsu ; I wonder about the `BrokenDBusMpris` name : do you expect this code to work with other things than Chromium? Else why not call it `ChromiumDBusMpris` maybe?

Yes, it does work with other things as long as the application publishes the data correctly and you specify the correct identity during initialization.

ultrabug:
Thanks!

PR #1944, opened 2020-07-16
add new module: watson

CryptoCopter:
Add new module to display the status of [watson](https://github.com/TailorDev/Watson), a cli time tracker.

Two tests are still failing, but I am not entirely sure why:

```
`ormat placeholders:` is not in module `watson.py`
```
I don't think(?) I have any format placeholders, and there seem to be other modules which also don't declare any in the docstring, so is there a way to have the tests ignore that?

```
watson
cache_time: (default 10) not in module docstring
watson_state_file: (default 'UNKNOWN JoinedStr') not in module docstring
``` 

I'm pretty sure, that I do have them in the docstring... so not sure what's going on here...


ultrabug:
Thanks for your contribution @CryptoCopter and sorry for the ridiculous time it took me to respond

I'll make comments directly in the review but what you're looking for is something like

```
Configuration parameters:
    cache_timeout: Number of seconds before state is re-read
        (default 10)
    watson_state_file: Path to the file which watson uses to track its own state
        (default '~/.config/watson/state')
```

py3status/modules/watson.py
```diff
@@ -0,0 +1,87 @@
+"""
+Display the current status of the watson time-tracking tool.
+
+Configuration Parameters:
+    cache_time: Number of seconds before state is re-read (default: 10)
```
ultrabug, line 5:
```suggestion
    cache_timeout: Number of seconds before state is re-read
        (default 10)
```

cache_timeout is a standard to be respected on modules, please rename

when writing down default values, you have to get to a new line, indent, then not use ":" to stipulate the default

py3status/modules/watson.py
```diff
@@ -0,0 +1,87 @@
...
+    watson_state_file: Path to the file which watson uses to track its own state (default: '~/.config/watson/state')
```
ultrabug, line 6: same for default docstring

ultrabug: we could also argue that this parameter is simply named `state_file` since we're in the watson module

py3status/modules/watson.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Dict, List, Union
+
+
+class Py3status:
+    """
+    """
+
+    # available configuration parameters
+    cache_time = 10
+    watson_state_file = f"{Path.home()}/.config/watson/state"
```
ultrabug, line 31: py3status still supports python3.5 so we cannot use fstrings yet

ultrabug: also, tests are failing because the default written in the docstring `'~/.config/watson/state'` is not the same as here and it should be please

If you want to expand user home path, do it later on a `post_config_hook` method preferably

CryptoCopter: Okay, then variable annotations need to go as well, these were only introduced in 3.6.
FYI: 3.5 reaches end-of-life on 2020-09-13 - so about a month from now.
Shall I make those changes and then submit another PR next month to undo them, or will you be supporting 3.5 past eol?

CryptoCopter: also, what would be an acceptable default value for `state_file`?

ultrabug: > Shall I make those changes and then submit another PR next month to undo them, or will you be supporting 3.5 past eol?

Yes please

ultrabug: > also, what would be an acceptable default value for `state_file`?

check for instance the google_calendar module source code, the one in the doctstring sounds reasonable if that's the default watson is waiting for, else you could go for `~/.config/py3status/watson_state`

CryptoCopter: okay, will do

CryptoCopter:
Okay, so while I do believe that I applied your change requests correctly, the tests still complain that the configuration parameters are not documented...

ultrabug:
@CryptoCopter I'll check again, but in my comment above I showed you how the test passed for me

ultrabug:
@CryptoCopter ahah, it's just that your have `Configuration Parameters` with a capital P at parameters while it should be lowercase!

Almost there :+1:

CryptoCopter:
oops, thanks for the insight.

should be fixed now

PR #1946, opened 2020-08-13
Support more battery in battery_level module

dkm:
Previous code only matches with "BAT*" names in the /sys/class/power_supply.
There does not seem to be any naming convention in the documentation:
https://www.kernel.org/doc/html/latest/power/power_supply_class.html
For example, the battery on the GPD Pocket is exposed as "max170xx_battery".

This patch allows for easier addition of glob patterns to use to match for
battery nodes and adds "*bat*" in the list.

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -83,6 +83,7 @@ discharging
 from re import findall
 from glob import iglob
 
+import itertools
 import math
 import os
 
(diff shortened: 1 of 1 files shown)


ultrabug:
Hello @dkm thanks for your first contribution to py3status and please accept my apologizes for the time it took me to respond...

I tested it OK, let's get this in!

dkm:
no worries and thanks for all the work on py3status !

PR #1949, opened 2020-09-10
Skip batteries with missing info

Urban82:
Skip batteries with missing info that leads to exceptions.

In my system, I have a Logitech Keyboard and a Logitech Mouse that appears in `/sys/class/power_supply` but they raise exceptions when parsed because of missing info:

`/sys/class/power_supply/hidpp_battery_0/uevent`
```
POWER_SUPPLY_NAME=hidpp_battery_0
POWER_SUPPLY_ONLINE=1
POWER_SUPPLY_STATUS=Discharging
POWER_SUPPLY_SCOPE=Device
POWER_SUPPLY_MODEL_NAME=Wireless Keyboard K270
POWER_SUPPLY_MANUFACTURER=Logitech
POWER_SUPPLY_SERIAL_NUMBER=OMISSIS
POWER_SUPPLY_CAPACITY_LEVEL=Full
```

PR #1951, opened 2020-09-13
battery_level: fix filter in /sys

### LINKED ISSUE #1950: Commit 9cf96a91 Broke battery_level module
Hello,

From my investigation commit 9cf96a91 broke the `battery_level` module. 

I'm running on Arch Linux and the latest update to py3status seemed to have pulled this in. 

I pulled the source from this github repo and installed py3status on my machine, with 9cf96a91 reverted, and the `battery_level` module seems to be working fine again.

Can we revert this commit until a working solution is refined?

dkm:
fix #1950 by really checking the dir in /sys contains the needed file (uevent).

### DIFF
--- a/py3status/modules/battery_level.py
+++ b/py3status/modules/battery_level.py
@@ -302,6 +302,9 @@ class Py3status:
         for path in path_its:
             r = _parse_battery_info(path)
 
+            if not r:
+                continue
+
             capacity = r.get(
                 "POWER_SUPPLY_ENERGY_FULL", r.get("POWER_SUPPLY_CHARGE_FULL")
             )
@@ -314,6 +317,16 @@ class Py3status:
             )
             current_now = r.get("POWER_SUPPLY_CURRENT_NOW", 0)
             voltage_now = r.get("POWER_SUPPLY_VOLTAGE_NOW", 0)
+
+            # missing values may indicate this is not a battery and should be skipped
+            if not (
+                capacity
+                and present_rate
+                and remaining_energy
+                and current_now
+                and voltage_now
+            ):
+                continue
 
             battery = {}
             battery["capacity"] = capacity


py3status/modules/battery_level.py
```diff
@@ -315,6 +318,16 @@ def _parse_battery_info(sys_path):
             current_now = r.get("POWER_SUPPLY_CURRENT_NOW", 0)
             voltage_now = r.get("POWER_SUPPLY_VOLTAGE_NOW", 0)
 
+            # missing values may indicate this is not a battery and should be skipped
+            if not (
+                capacity
+                and present_rate
```
Urban82, line 324: I think this check should not be included here: it avoids to include fully charged batteries that will enter in the `except ZeroDivisionError` on R343, instead.

dkm: True... Maybe instead of relying on a division error to detect fully charged, we could check for the existence of the corresponding key/value. That would be easier to read. What do you think ?

Urban82: it can be done with a pre-check, before the division: if all other values are presents but only `present_rate` is missing, the battery is present and fully charge (or, at least, not discharging and not charging)

Urban82: I added a proposal in my last commit from PR #1949 (048c26c89fea7ab51ed1130366af0d997d326d13)

```
            # missing values may indicate this is not a battery and should be skipped
            print(path, capacity, remaining_energy, present_rate, current_now, voltage_now)
            if (
                capacity == None
                or present_rate == None
                or remaining_energy == None
                or current_now == None
                or voltage_now == None
            ):
                continue

            battery = {}
            battery["capacity"] = capacity
            battery["charging"] = "Charging" in r["POWER_SUPPLY_STATUS"]
            battery["percent_charged"] = int(
                math.floor(remaining_energy / capacity * 100)
            )
            if present_rate == 0:
                # Battery is either full charged or is not discharging
                battery["time_remaining"] = FULLY_CHARGED
            else:
                if battery["charging"]:
                    time_in_secs = (capacity - remaining_energy) / present_rate * 3600
                else:
                    time_in_secs = remaining_energy / present_rate * 3600
                battery["time_remaining"] = self._seconds_to_hms(time_in_secs)

            battery["power"] = current_now * voltage_now / 1e12
```

PR #1954, opened 2020-09-26
Don't join the artist fields in BrokenDBusMpris

ritze:
The artist fields are already joined later in line 558 / 555 .

### DIFF
--- a/py3status/modules/mpris.py
+++ b/py3status/modules/mpris.py
@@ -162,9 +162,6 @@ class BrokenDBusMpris:
                         self._parent.PlaybackStatus = msg["PlaybackStatus"]
                     if "Metadata" in msg:
                         self._parent.Metadata = msg["Metadata"]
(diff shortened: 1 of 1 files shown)


ultrabug:
what's the consequence of this @ritze plz?

ritze:
dbus returns an array of strings. In line 558/555 we just take the first array entry.

If BrokenDBusMpris doesn't return an array of strings too, but joins the strings already, as it was the case before, we will just take the first char of the array in line 558/555 instead of taking the first string within the array:

```python
self._data["artist"] = metadata.get("xesam:artist")[0]
```

Or in other words: Instead of showing the song "Song 2" from "Blur" listening via YouTube with using BrokenDBusMpris, the song would not displayed as expected as "Blur - Song 2", but as "B - Song 2".

ultrabug:
Thanks @ritze !

PR #1963, opened 2020-10-28
 Add poweron architecture ppc64le to travis build

asellappen:
This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing,We typically build applications for customers and ISVs, and while we don't use this package directly, 
we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or third level dependency and couldn't tell you explicitly which features we use or our usage model,For more info tag @gerrith3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: python
+arch:
+  - amd64
+  - ppc64le
 python:
   - "3.5"
   - "3.6"


ultrabug:
Hello @asellappen 

I feel like adding this on travis would mean that our project officially supports this architecture which we do not since I don't have any access to such a machine

So I'm unsure about the necessity of this I'm sorry

gerrith3:
@ultrabug if you want, https://osuosl.org/services/powerdev/ provides free access to power VMs if you want.  As far as support, this is already in debian/ubuntu ppc64le so even if you don't support it it is out there and used.  You could also state that it is either not supported, or community supported in your README.

ultrabug:
I see @gerrith3 thanks for explaining further.

Since Travis passes, let's get this in. I hope this helps you guys.

PR #1967, opened 2020-11-08
hueshift module: fix initialization incrementing cached value by delta

eugenenelou:
Fix hueshift being initialized with the cached value+delta.


ultrabug:
Merci Eugène

Good catch, I'm not using this module but reading the code makes me understand your point!

PR #1974, opened 2020-11-16
Add spotifyd support

evrardjp:
Without this patch, spotify module only supports spotify client,
but doesn't support spotifyd. Spotifyd users need to use mpris
or other, while the spotify module is just fine.

This is problematic as it means users will write new plugins or
new code for almost no reason.

This fixes it by making the dbus commands more flexible, to be
able to use org.mpris.MediaPlayer2.spotifyd for spotifyd users
or org.mpris.MediaPlayer2.spotify for regular users. If a new
player out there supports a similar mpris interface with a new
client name, the users can now simply override the dbus_client
variable.

### DIFF
--- a/py3status/modules/spotify.py
+++ b/py3status/modules/spotify.py
@@ -6,6 +6,10 @@ Configuration parameters:
     button_play_pause: button to toggle play/pause (default None)
     button_previous: button to switch to previous song (default None)
     cache_timeout: how often to update the bar (default 5)
+    dbus_client: Used to override which app is used as a client for
+        spotify. If you use spotifyd as a client, set this to
+        'org.mpris.MediaPlayer2.spotifyd'
+        (default 'org.mpris.MediaPlayer2.spotify')
     format: see placeholders below (default '{artist} : {title}')
     format_down: define output if spotify is not running
         (default 'Spotify not running')
@@ -62,7 +66,7 @@ from time import sleep
 
 import dbus
 
-SPOTIFY_CMD = """dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify
+SPOTIFY_CMD = """dbus-send --print-reply --dest={dbus_client}
                  /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player.{cmd}"""
 
 
@@ -75,6 +79,7 @@ class Py3status:
     button_play_pause = None
     button_previous = None
     cache_timeout = 5
+    dbus_client = "org.mpris.MediaPlayer2.spotify"
     format = "{artist} : {title}"
     format_down = "Spotify not running"
     format_stopped = "Spotify stopped"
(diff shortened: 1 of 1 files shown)


py3status/modules/spotify.py
```diff
@@ -16,6 +16,10 @@
     sanitize_words: which meta data to remove
         *(default ['bonus', 'demo', 'edit', 'explicit', 'extended',
             'feat', 'mono', 'remaster', 'stereo', 'version'])*
+    dbus_client: Used to override which app is used as a client for
```
ultrabug, line 19: please add this in alphabetical order related to the other options in the docstring

py3status/modules/spotify.py
```diff
@@ -91,9 +95,10 @@ class Py3status:
         "stereo",
         "version",
     ]
+    dbus_client = "org.mpris.MediaPlayer2.spotify"
```
ultrabug, line 98: please add this in the alphabetical order within the available config params above

py3status/modules/spotify.py
```diff
@@ -91,9 +95,10 @@ class Py3status:
...
 
     def _spotify_cmd(self, action):
-        return SPOTIFY_CMD.format(cmd=action)
+        return SPOTIFY_CMD.format(dbus_client=self.dbus_client,cmd=action)
```
ultrabug, line 101: python lint will require a space after the comma please

ultrabug (changes requested):
Thanks a lot for your first contribution to py3status @evrardjp !

This is well thought and implemented, some small glitches are in the way and I'll merge this

py3status/modules/spotify.py
evrardjp: Oh wow, that's a wrong push, sorry for that.

py3status/modules/spotify.py
evrardjp: done.

py3status/modules/spotify.py
evrardjp: done

evrardjp (commented):
Done

evrardjp:
Rebased.

evrardjp:
Haha thanks! Will fix in a bit. Thanks for those swift reviews!

py3status/modules/spotify.py
```diff
@@ -75,6 +79,7 @@ class Py3status:
     button_play_pause = None
     button_previous = None
     cache_timeout = 5
+    dbus_client = "org.mpris.MediaPlayer2.spotifyd"
```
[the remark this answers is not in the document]
evrardjp: oh yeah good catch! Sorry for that. Regression on my updated patch.

py3status/modules/spotify.py
```diff
@@ -6,6 +6,10 @@
     button_play_pause: button to toggle play/pause (default None)
     button_previous: button to switch to previous song (default None)
     cache_timeout: how often to update the bar (default 5)
+    dbus_client: Used to override which app is used as a client for
+        spotify. If you use spotifyd as a client, set this to
+        org.mpris.MediaPlayer2.spotifyd
+        (default org.mpris.MediaPlayer2.spotify)
```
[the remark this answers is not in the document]
evrardjp: Done.

ultrabug:
Awesome, merci Jean-Philippe !

evrardjp:
Thanks for your patience :)

PR #1978, opened 2020-11-28
Use pathlib to manipulate paths and files

eumiro:
The project uses different ways to manipulate paths and files. This is an attempt to replace all `os.path.*`, `glob.*`, and `open` calls with `pathlib.Path` objects and methods that make the directory tree structure clearer and that allow better manipulation of disk-related operations.

### DIFF
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -1,8 +1,8 @@
-import os
 import inspect
 
 from collections import OrderedDict
 from importlib.machinery import SourceFileLoader
+from pathlib import Path
 from time import time
 from random import randint
 
@@ -110,9 +110,8 @@ class Module:
         Return user-written class object from given path.
         """
         class_inst = None
-        module_name, file_ext = os.path.splitext(os.path.split(filepath)[-1])
-        if file_ext.lower() == ".py":
-            py_mod = SourceFileLoader(module_name, filepath).load_module()
+        if filepath.suffix == ".py":
+            py_mod = SourceFileLoader(filepath.stem, filepath).load_module()
             if hasattr(py_mod, cls.EXPECTED_CLASS):
                 class_inst = py_mod.Py3status()
         return class_inst
@@ -588,7 +587,7 @@ class Module:
             # user provided modules take precedence over py3status provided modules
             if self.module_name in user_modules:
                 include_path, f_name = user_modules[self.module_name]
-                module_path = os.path.join(include_path, f_name)
+                module_path = Path(include_path) / f_name
                 self._py3_wrapper.log(f'loading module "{module}" from {module_path}')
                 self.module_class = self.load_from_file(module_path)
             # load from py3status provided modules
--- a/py3status/modules/file_status.py
+++ b/py3status/modules/file_status.py
@@ -52,8 +52,8 @@ missing
 {'color': '#FF0000', 'full_text': u'\u25a0'}
 """
 
-from glob import glob
-from os.path import basename, expanduser
+from pathlib import Path
+
 
 STRING_NO_PATHS = "missing paths"
 
@@ -107,7 +107,7 @@ class Py3status:
         # convert str to list + expand path
         if not isinstance(self.paths, list):
             self.paths = [self.paths]
-        self.paths = list(map(expanduser, self.paths))
+        self.paths = [Path(path).expanduser() for path in self.paths]
 
         self.init = {"format_path": []}
         if self.py3.format_contains(self.format, "format_path"):
@@ -117,7 +117,9 @@ class Py3status:
 
     def file_status(self):
         # init data
-        paths = sorted([files for path in self.paths for files in glob(path)])
+        paths = sorted(
+            files for path in self.paths for files in path.parent.glob(path.name)
+        )
         count_path = len(paths)
         format_path = None
 
@@ -130,7 +132,7 @@ class Py3status:
                 path = {}
                 for key in self.init["format_path"]:
                     if key == "basename":
-                        value = basename(pathname)
+                        value = pathname.name
                     elif key == "pathname":
                         value = pathname
                     else:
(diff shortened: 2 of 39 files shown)


ultrabug:
Hey @eumiro, I hope you're doing good, I missed seeing you at EuroPython this year :/

Thanks a lot for this big contribution which makes a lot of sense, I shall review it properly and will get back to you :+1:

eumiro:
Glad to hear that, @ultrabug .
Travis says there are problems in Python 3.5, but that is caused by the f-strings, which I completely ignored. I can replace them by `"".format` in my PR. Or are you considering dropping the Python 3.5 support in the near™ future?

ultrabug:
@eumiro python3.5 reached EOL more than two months ago, I guess we could start getting rid of it

I think we should get a proper PR to update the setup.py to remove python3.5 and then we can proceed with yours?

We'll then need another PR with the run of pyupgrade to transform everything else while we're at it

Sounds like a plan?

eumiro:
Great. See #1979

ultrabug:
Cool :+1:

py3status/modules/file_status.py
```diff
@@ -107,7 +104,7 @@ def post_config_hook(self):
         # convert str to list + expand path
         if not isinstance(self.paths, list):
             self.paths = [self.paths]
-        self.paths = list(map(expanduser, self.paths))
+        self.paths = [path.expanduser() for path in self.paths]
```
ultrabug, line 107: I may have read through this too quick but how are the paths in self.path magically converted to Path so you can use expanduser() like this?

eumiro: That's what reviews are for :+1: Thank you for checking it.

Fixed to

```
        self.paths = [Path(path).expanduser() for path in self.paths]
```

Is there actually no test coverage run in tox?

ultrabug: I think this mistake is present on some other files, I'll get through them again then

> Is there actually no test coverage run in tox?

not those ones no; we've always been willing to but..

py3status/module.py
```diff
@@ -588,7 +586,7 @@ def load_methods(self, module, user_modules):
             # user provided modules take precedence over py3status provided modules
             if self.module_name in user_modules:
                 include_path, f_name = user_modules[self.module_name]
```
ultrabug, line 588: this returns Path objects right?

py3status/modules/google_calendar.py
```diff
@@ -195,8 +194,8 @@ def post_config_hook(self):
         self.events = None
         self.no_update = False
 
-        self.client_secret = os.path.expanduser(self.client_secret)
-        self.auth_token = os.path.expanduser(self.auth_token)
+        self.client_secret = self.client_secret.expanduser()
```
ultrabug, line 197: here this module is missing Path

py3status/modules/imap.py
```diff
@@ -123,8 +122,8 @@ def post_config_hook(self):
         self.idle_thread = Thread()
 
         if self.client_secret:
-            self.client_secret = os.path.expanduser(self.client_secret)
-        self.auth_token = os.path.expanduser(self.auth_token)
+            self.client_secret = self.client_secret.expanduser()
```
ultrabug, line 125: this modules needs Path

py3status/py3.py
```diff
@@ -1103,7 +1103,7 @@ def play_sound(self, sound_file):
             if cmd:
                 if cmd == "ffplay":
                     cmd = "ffplay -autoexit -nodisp -loglevel 0"
-                sound_file = os.path.expanduser(sound_file)
+                sound_file = sound_file.expanduser()
```
ultrabug, line 1106: missing Path?

eumiro:
Thanks @ultrabug for the review, I have fixed and pushed all your suggestions. I think the strings should be converted to `Path` as soon as possible, even in the user modules, which should be then clearly defined in the interface.

fastentrypoints.py
```diff
@@ -82,21 +82,21 @@ def get_args(cls, dist, header=None):  # noqa: D205,D400
 
 
 def main():
-    import os
     import re
     import shutil
     import sys
+    from pathlib import Path
```
lasers, line 88: Suggestion: Change this upstream. We can get it later along with any other changes.

ultrabug (changes requested):
That's the last comment I have then we're good to go!

Sorry for the time it took me to get back to you mate

py3status/modules/file_status.py
```diff
@@ -117,7 +117,7 @@ def post_config_hook(self):
 
     def file_status(self):
         # init data
-        paths = sorted([files for path in self.paths for files in glob(path)])
+        paths = sorted(files for path in self.paths for files in path.iterdir())
```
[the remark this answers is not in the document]
eumiro: You're right, thank you for the review. I have now changed it to `path.parent.glob(path.name)` which should work for wildcards like `/a/b/*.d`. Do you think that would solve the problem?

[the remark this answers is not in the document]
ultrabug: Yep I guess it should now, cheers

ultrabug:
@eumiro merged and tested, it's broken... my review wasn't enough obviously !

ultrabug:
Ok I believe I tackled most of them and master is now working as intended

eumiro:
Thank you for the quick bugfixes and sorry for introducing those bugs. Can these cases be added to tests?

ultrabug:
The whole test suite is to be reworked so I guess we should not waste our time in the v3 branch

I'll make sure some proper testing methodology is done on the version 4 I'm working on tho

X-dark:
Seems there is still some issues:

```
   File "/build/py3status-git/src/py3status/setup.py", line 14, in <module>
     sys.path.remove(module_path)
 ValueError: list.remove(x): x not in list
```

when trying to build with `python setup.py install --root="$pkgdir/" --optimize=1`

ultrabug:
Thanks for reporting @X-dark

PR #1979, opened 2020-11-29
Drop Python 3.5 support

eumiro:
Python 3.5 is after EOL. This PR removes the Python 3.5 support from `setup.py` and `.travis.yml` and then updates all `*.py` files using `pyupgrade --py36-plus` automatically (mostly introducing `f-strings`).


ultrabug:
Thanks!

PR #1980, opened 2020-12-13
Add GitHub Actions CI

eumiro:
Since the `master` still does not use this, the results won't be immediately visible in this PR, so you may have to test it in an extra branch on the main repo and only when it works (or needs bugfixing), you can merge it into `master`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,4 +1,5 @@
 [tox]
+envlist = py36, py37, py38, py39, mypy
 skip_missing_interpreters = True
 
 [testenv]
@@ -34,3 +35,11 @@ flake8-ignore =
 flake8-select = F
 flake8-max-complexity = 18
 flake8-max-line-length = 88
+
+[gh-actions]
+python =
+    3.6: py36
+    3.7: py37
+    3.8: py38, mypy
+    3.9: py39
+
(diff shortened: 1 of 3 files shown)


eumiro:
It looks like the [run](https://github.com/eumiro/py3status/actions/runs/418559544) was successful.

ultrabug:
Very interesting thanks @eumiro 

I guess we could drop travis in that case right?

tox.ini
```diff
@@ -34,3 +35,11 @@ flake8-ignore =
 flake8-select = F
 flake8-max-complexity = 18
 flake8-max-line-length = 88
+
+[gh-actions]
+python =
+    3.6: py36
+    3.7: py37
+    3.8: py38, mypy
```
ultrabug, line 43: can you explain mypy wrt 3.8 only plz?

eumiro: This comes from the [`tox-gh-actions`](https://github.com/ymyzk/tox-gh-actions) page. I believe it's enough to put it along a single version.

eumiro:
> Very interesting thanks @eumiro
> 
> I guess we could drop travis in that case right?

Yes, but just let run both next to each other for a few more PRs before you drop travis.

ultrabug:
Thanks!

PR #1981, opened 2020-12-14
Use f-strings

eumiro:
Since `py3status` is Python 3.6+, we can use f-strings for immediate string formatting. Some templates will be still formatted later using `.format`.

### DIFF
--- a/py3status/autodoc.py
+++ b/py3status/autodoc.py
@@ -101,7 +101,7 @@ def markdown_2_rst(lines):
             code = not code
             space = " " * (len(line.rstrip()) - 3)
             if code:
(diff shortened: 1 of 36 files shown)


ultrabug:
I wonder why pyupgrade didn't catch most of them?

ultrabug:
plz rebase mate?

eumiro:
rebased, pushed.

I tried to reconstruct the `pyupgrade` (both with `py3-plus` and `py36-plus`) call with the older commit, but did not find anything. Anyway, now we're py3.6+.

ultrabug:
Thanks mate

PR #1982, opened 2020-12-14
Use list/generator/dict comprehensions where appropriate

eumiro:
Remove some redundant code, such as square brackets, as well as `map/lambda` calls and use list/generator/dict comprehensions where appropriate.


ultrabug:
So many improvements, thanks mate!

PR #1983, opened 2020-12-15
Fix sys.path manipulation in setup.py

eumiro:
Fixes bug introduced in 37603bf3d08d08ebf8fe38ef5d419e51f03821ec and reported by @X-dark within #1978

`sys.path` contains `str` only and so should be all `Path`s converted to `str`.


ultrabug:
Thanks!

PR #1989, opened 2020-12-30
Fixes mpris for Chrome, too.

treibholz:
Chrome is exactly as broken as Chromium, but has the player_id "chrome",
instead of "chromium".

This adds a rather unelegant but simple workaround, to "fix" it in the same way.


ultrabug:
Thank you @treibholz

PR #1991, opened 2021-01-08
Remove obsolete numeric relics from Python 2

eumiro:
Remove unnecessary int/float conversions in Python3 code that were needed in Python2.

Another idea would be to use `ratio` (0.0–1.0) instead of `percents` (0–100) and only convert them to string at the end (`f'{0.1234:.1%}'` converts nicely to `12.3%`), which would save a lot of multiplication/division by 100 in the code. What do you think?

### DIFF
--- a/py3status/modules/xsel.py
+++ b/py3status/modules/xsel.py
@@ -66,7 +66,7 @@ class Py3status:
         selection = " ".join(selection.split())
         if len(selection) >= self.max_size:
             if self.symmetric is True:
-                split = int(self.max_size / 2) - 1
+                split = self.max_size // 2 - 1
                 selection = selection[:split] + ".." + selection[-split:]
             else:
                 selection = selection[: self.max_size]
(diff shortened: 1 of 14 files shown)


py3status/modules/xsel.py
```diff
@@ -66,7 +66,7 @@ def xsel(self):
         selection = " ".join(selection.split())
         if len(selection) >= self.max_size:
             if self.symmetric is True:
-                split = int(self.max_size / 2) - 1
+                split = (self.max_size // 2) - 1
```
ultrabug, line 69: I guess we can loose the parenthesis too?

ultrabug:
> Another idea would be to use ratio (0.0–1.0) instead of percents (0–100)

This could break existing user configurations I guess, so I would not do that

Thanks

py3status/modules/xsel.py
eumiro: Oh yes, thanks, fixed, repushed.

ultrabug:
Thanks!

PR #1993, opened 2021-01-08
Modernize Python3 code

eumiro:
A few independent commits, please feel free to discuss them individually:

1. access the keys of a dictionary implicitly when iterating or checking
2. get all necessary variables in the header of a loop
3. remove redundant parentheses (leave those in returning tuples though)
4. chain conditions
5. unescape characters in regexes that do not have to be escaped

### DIFF
--- a/py3status/modules/net_iplist.py
+++ b/py3status/modules/net_iplist.py
@@ -77,7 +77,7 @@ class Py3status:
 
     def post_config_hook(self):
         self.iface_re = re.compile(r"\d+: (?P<iface>[\w\-]+):")
-        self.ip_re = re.compile(r"\s+inet (?P<ip4>[\d\.]+)(?:/| )")
+        self.ip_re = re.compile(r"\s+inet (?P<ip4>[\d.]+)(?:/| )")
         self.ip6_re = re.compile(r"\s+inet6 (?P<ip6>[\da-f:]+)(?:/| )")
 
     def net_iplist(self):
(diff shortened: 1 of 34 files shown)


eumiro:
Rebased to the current `master`.

py3status/modules/net_iplist.py
```diff
@@ -77,7 +77,7 @@ class Py3status:
 
     def post_config_hook(self):
         self.iface_re = re.compile(r"\d+: (?P<iface>[\w\-]+):")
-        self.ip_re = re.compile(r"\s+inet (?P<ip4>[\d\.]+)(?:/| )")
+        self.ip_re = re.compile(r"\s+inet (?P<ip4>[\d.]+)(?:/| )")
```
ultrabug, line 80: ain't that catch "any caracter" instead of the intended "dot" ?

ultrabug (commented):
Thanks a lot for this amazing code QA mate

py3status/modules/net_iplist.py
eumiro: No.
https://docs.python.org/3/library/re.html#regular-expression-syntax

    Special characters lose their special meaning inside sets. For example, `[(+*)]` will match any of the literal characters `'('`, `'+'`, `'*'`, or `')'`.

eumiro: No.
https://docs.python.org/3/library/re.html#regular-expression-syntax

    Special characters lose their special meaning inside sets. For example, `[(+*)]` will match any of the literal characters `'('`, `'+'`, `'*'`, or `')'`.

ultrabug: wow TIL something new again <3

ultrabug:
Thank you mate!

PR #1996, opened 2021-01-12
Use time.perf_counter to measure short time periods

eumiro:
When using `time.time()` twice to get a short timedelta, such as refresh rate, timeout, or elapsed time, where the zero (unix epoch) is not important, this may return false (even negative) results if the system clock gets updated during that time.

To measure such short timedeltas, `time.perf_counter()` is better, because at each call it returns a positive float that is higher than the previous value.

The negative point is that this float is process-dependent and cannot be shared between processes. Therefore I have avoided modification of the code that stores this value on disk for the next instance to load it.

There are several name misuses of `time` in the code. They should be renamed to something more meaningful.

### DIFF
--- a/py3status/modules/uptime.py
+++ b/py3status/modules/uptime.py
@@ -60,7 +60,7 @@ SAMPLE OUTPUT
 {'full_text': 'up 1 days 18 hours 20 minutes'}
 """
 
-from time import time
+import time
 from datetime import datetime
 from collections import OrderedDict
 from pathlib import Path
@@ -94,7 +94,7 @@ class Py3status:
     def uptime(self):
         with Path("/proc/uptime").open() as f:
             up = int(float(f.readline().split()[0]))
-            offset = time() - up
+            offset = time.time() - up
 
         uptime = {}
         for unit, interval in self.time_periods.items():
--- a/py3status/storage.py
+++ b/py3status/storage.py
@@ -1,9 +1,9 @@
 import os
+import time
 
 from pathlib import Path
 from pickle import dump, load
 from tempfile import NamedTemporaryFile
-from time import time
 
 
 class Storage:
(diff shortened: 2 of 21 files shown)


py3status/modules/uptime.py
```diff
@@ -94,7 +94,7 @@ def post_config_hook(self):
     def uptime(self):
         with Path("/proc/uptime").open() as f:
             up = int(float(f.readline().split()[0]))
-            offset = time() - up
+            offset = time.time() - up
```
ultrabug, line 97:
```suggestion
            offset = time.perf_counter() - up
```

py3status/storage.py
```diff
@@ -90,7 +90,7 @@ def storage_set(self, module_name, key, value):
         if module_name not in self.data:
             self.data[module_name] = {}
         self.data[module_name][key] = value
-        ts = time()
+        ts = time.time()
```
ultrabug, line 93:
```suggestion
        ts = time.perf_counter()
```

ultrabug (changes requested):
Thanks mate, TIL something new :)

py3status/modules/uptime.py
eumiro: Not in this case. The `up` is a unix timestamp and we want to calculate the difference to it, so we have to ask for `time.time()`.

py3status/storage.py
eumiro: And here it goes to `storage`, so probably it will be stored on the disk and loaded by a new process. `time.perf_counter()` returns a float that is based on *some* time in the past, apparently the start of the process. So you cannot copy it between processes.

ultrabug:
Oh now I get what you mean by process dependant @eumiro !.. hmm

ultrabug:
:+1: thanks

ya-isakov:
@eumiro perf_counter doesn't work for cache_until, because perf_counter could start at any second of real time, so cache will be always shifted against wall clock. I've created a ticket https://github.com/ultrabug/py3status/issues/2028, where this is demonstrated via python cli.

PR #2009, opened 2021-02-10
volume_status: run with either pulseaudio or pipewire installed

Compunctus:
pipewire is now able to completely replace pulseaudio, so there may be
no "pulseaudio" command.


ultrabug:
TIL about pipewire :scream: thanks @Compunctus !

PR #2011, opened 2021-02-12
net_iplist show only non deprecated ipv6 addresses

cstruck:
net_iplist shows all ipv6 addresses on a interface which can be alot. This patch removes all old deprecated from the list.


ultrabug:
tested ok, thanks for your first contribution @cstruck !

PR #2013, opened 2021-02-21
[Module : Google Calendar] Additional options

ben75020:
This pull requests adds two options to the Google Calendar plugin :
1. The call to the browser can be customized if required (you might not
want use the browser default while opening an event).

2. It is now possible to open the Meet Hangout VC link (associated from the caledar details)
instead of the google calendar event.
We'll always fallback to the htmlLink (event itself) in the case the
hangout VC does not exists.

Those two parameters are optional.

Example of usage :
```
google_calendar {
    browser_invocation = "i3-msg '[title=\".*?Meet.*?\" workspace=\"0\"] kill; workspace 0; exec /usr/bin/google-chrome --app={}'"
    preferred_event_link = "hangoutLink"
}
```
This way, I ask my window manager i3 to close any previous opened
google meet before reopening one in a dedicated workspace.

These changes have been tested on my local computer.

### DIFF
--- a/py3status/modules/google_calendar.py
+++ b/py3status/modules/google_calendar.py
@@ -53,6 +53,8 @@ Configuration parameters:
         (default 3)
     response: Only display events for which the response status is
         on the list. (default ['accepted'])
+    browser_invocation: Command to run to open browser. Curly braces stands for URL opened.
+    (default "xdg-open {}")
     thresholds: Thresholds for events. The first entry is the color for event 1,
         the second for event 2, and so on.
         (default [])
@@ -60,6 +62,10 @@ Configuration parameters:
         string; e.g. if time_to_max is 60, `{format_timer}` will only be
         displayed for events starting in 60 minutes or less.
         (default 180)
+    preferred_event_link: link to open in the browser.
+        accepted values : hangoutLink (open the VC room associated with the event),
+           (default)    : htmlLink (open the event's details in Google Calendar).
+        fallback to htmlLink if the preferred_event_link does not exist it the event.
     warn_threshold: The number of minutes until an event starts before a
         warning is displayed to notify the user; e.g. if warn_threshold is 30
         and an event is starting in 30 minutes or less, a notification will be
@@ -163,6 +169,7 @@ class Py3status:
     # available configuration parameters
     auth_token = "~/.config/py3status/google_calendar.auth_token"
     blacklist_events = []
+    browser_invocation = "xdg-open {}"
     button_open = 3
     button_refresh = 2
     button_toggle = 1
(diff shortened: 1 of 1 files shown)


py3status/modules/google_calendar.py
```diff
@@ -397,7 +406,10 @@ def _build_response(self):
             event_dict["summary"] = event.get("summary")
             event_dict["location"] = event.get("location")
             event_dict["description"] = event.get("description")
-            self.event_urls.append(event["htmlLink"])
+            if self.preferred_event_link == "hangoutLink":
+                self.event_urls.append(event.get("hangoutLink", event.get("htmlLink")))
```
ultrabug, line 410: Instead of the if / else, couldn't we just try to pick any preferred_event_link and fallback like this?

```suggestion
            self.event_urls.append(event.get(self.preferred_event_link, event.get("htmlLink")))
```

This way we could actually honor any kind of link provided the user knows the key he's interested in (a link to the google api to find out other options in the docstring could be cool)?

ultrabug (changes requested):
Thank you so much for your first contribution @ben75020 !

py3status/modules/google_calendar.py
ben75020: done, also added additional information about the response status.
Once the PR is merged, I'll offer another change on adding a "?" prefix on upcoming events where you didn't decided if you want to come, or not.

py3status/modules/google_calendar.py
```diff
@@ -51,8 +56,17 @@
         (default False)
     num_events: The maximum number of events to display.
         (default 3)
+    preferred_event_link: link to open in the browser.
+        accepted values :
```
ultrabug, line 60: I'm sorry if my previous comment was not clear enough but here is where I expected to be less strict and point to any accepted value found (or which could be added later) in the google API response (just like you did with the response config below

So instead of listing accepted values, list them as examples but state clearly that any value present in the xxxx.xxxx google api object is valid there and add a link to its docs?

ultrabug:
Also, you'll need to fix those CI reports please

Thanks mate!

py3status/modules/google_calendar.py
ben75020: Unfortunately, we cannot tell "any value present in this SPECIFIC-FIELD", because we can retrieve a value from various fields : 
- htmlLink is located at .htmlLink of the Google API Object
- hangoutLink is located at .hangoutsLink of the Google API Object.

There is no obvious rule that can be given for (any value of the following object) -  this is why I considered safer for new users to explicitly tell what are the possible values. 
Maybe tomorrow Google will introduce a .newServiceURL in their scheme, at another location.

I'm sorry if I'm not clear enough --  I don't see how to document better, I have included  a link to Google Calendar API's to the top of the documentation.

Please feel free to reach me in DM if required.

ultrabug: thanks!

PR #2021, opened 2021-03-07
Wind direction added for weather_osm module

leroyz84:
I've added an option for wind direction.
Degrees doesn't say much for me, but i care about where the wind is coming from. I've put the names on it based on range direction

Used this table for the mapping : https://www.researchgate.net/figure/The-main-compass-directions-of-the-wind-used-for-clustering-the-wind-direction_tbl1_252953902


ultrabug:
TIL about wind degrees and compass direction :+1: thanks a lot for your first contribution @leroyz84 :)

leroyz84:
Thank you for your compliment, merging it and for this great software :)

PR #2022, opened 2021-03-09
Fix loading modules from entry_point that take a parameter

mcgillij:
fix for: https://github.com/ultrabug/py3status/issues/2020

I'm now able to load modules that take a parameter from `~/.i3/pystatus/`, from `py3status/modules` and those installed with pip/poetry that are in the same env.


mcgillij:
Very nice I'll close my PR.

Issue #2018, opened 2021-03-06
Can't run the py3status from NixOS


squiter:
I'm following the instructions in the documentation, but I can't get the bar work.

When I try to run `py3status` from terminal, I got this message:

```
Traceback (most recent call last):
  File "/nix/store/s6har47fhkqny6azr5clsavn9nbidqf4-python3.8-py3status-3.28/bin/.py3status-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/s6har47fhkqny6azr5clsavn9nbidqf4-python3.8-py3status-3.28/lib/python3.8/site-packages/py3status/__init__.py", line 27, in main
    from py3status.core import Py3statusWrapper
  File "/nix/store/s6har47fhkqny6azr5clsavn9nbidqf4-python3.8-py3status-3.28/lib/python3.8/site-packages/py3status/core.py", line 2, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

This is probably a problem with my NixOS configuration, but if anyone can help me here, I'll appreciate it.

ultrabug:
you mean @JohnAZoidberg @lasers :)

ultrabug:
I'm sorry @squiter I'm absolutely not at par with NixOS :(

Friendly ping to @thiagokokada who's renown for his NixOS powers too!

thiagokokada:
Can't say I have any idea what is happening here without more details on how `py3status` is being setup/started, but there is still this documentation that should be a good start: https://github.com/ultrabug/py3status/blob/master/doc/intro.rst#note-on-nixos

P.S.: just a note that I don't use py3status anymore. However I am gladly to help about those NixOS issues if I can.

thiagokokada:
I helped @squiter on Slack, but basically:

```
(python3Packages.py3status.overrideAttrs (oldAttrs: {
  propagatedBuildInputs = with python3Packages; oldAttrs.propagatedBuildInputs ++ [ setuptools  ];
}))
```

thiagokokada:
Opened a PR fixing the instructions. The lack of the `oldAttrs.propagatedBuildInputs` removed the `setuptools` dependency accidentally.

So not, there is no need to inject `setuptools`  directly, just patching the receipt in the correct way.

ultrabug:
Thanks a lot @thiagokokada, impressive as always!

PR #2024, opened 2021-03-09
Fix NixOS instructions

thiagokokada:
See issue #2018.

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -99,24 +99,26 @@ To have it globally persistent add to your NixOS configuration file py3status as
 .. code-block:: nix
 
     (python3Packages.py3status.overrideAttrs (oldAttrs: {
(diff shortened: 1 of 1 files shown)


doc/intro.rst
```diff
@@ -99,24 +99,26 @@ To have it globally persistent add to your NixOS configuration file py3status as
 .. code-block:: nix
 
     (python3Packages.py3status.overrideAttrs (oldAttrs: {
-      propagatedBuildInputs = [ python3Packages.pytz python3Packages.tzlocal ];
+      propagatedBuildInputs = with python3Packages;[ pytz tzlocal ] ++ oldAttrs.propagatedBuildInputs;
```
thiagokokada, line 102: We could use other packages as example here, since `pytz` and `tzlocal` is already included in the NixOS derivation.

ultrabug:
Impressive, that's so obscure to me :+1: thanks a lot !

PR #2030, opened 2021-03-19
structure: start moving v3 code logic to its own versioned folder

ultrabug:
this is the first step into making py3status multi engine capable (v3 and v4)


ultrabug:
not ready

PR #2031, opened 2021-03-19
Revert time.perf_counter back to time.time in clock

### LINKED ISSUE #2028: py3.time_in is off by ~20 seconds if sync_to=60 (regression)
**Describe the bug**
I found that after commit e431f5ec300def69f3015d3b2928894d3ed6bfac, clock widget is off by e.g. 20 seconds, if I'm using only hours an minutes.

**Your py3status version**
py3status version 3.33 (python 3.8.7) on sway

**To Reproduce**
Steps to reproduce the behavior along with your py3status configuration (filter out any private values):
1. Add module 'clock'
2. Configure module like this 'clock {
   format = ["{Local}", "{America/Los_Angeles}"]
   format_time = "%Y-%m-%d %H:%M %a"
}'
3. Run py3status
4. See that minutes are changed e.g. 20 seconds later.

**Expected behavior**
Minutes should change synchronously to real clock.

**Screenshots**
N/A

**Additional context**
I've tried to reproduce this issue in python cli:
```
>>> datetime.now(); requested = time.perf_counter()
datetime.datetime(2021, 3, 16, 23, 30, 19, 244439)
>>> r1 = (requested + 60) - (requested % 60)
>>> r1 - requested
1.356851185002597
```
This means, that time_in will wait for 1 second, if its ..:19.
`next_run_in = r1 - requested` could be simplified to `next_run_in = 60 - time.perf_counter % 60`. As perf counter could start at any wall clock second, this will cause a delay.

Here is another example:
```
>>> datetime.now(); requested = time.perf_counter()
datetime.datetime(2021, 3, 16, 23, 32, 41, 105984)
>>> r1 = (requested + 60) - (requested % 60)
>>> r1 - requested
39.49525859300047
```
it will wait until 41+39 = ...:20 again.

In py3status before this commit (when time.time() was used), it worked perfectly:
```
>>> datetime.now(); requested = time.time()
datetime.datetime(2021, 3, 16, 23, 34, 41, 519036)
>>> r1 = (requested + 60) - (requested % 60)
>>> r1 - requested
18.480836868286133
```
My guess is that the shift is the same, when perf counter was started. So, in my case, perf counter should be started at ...:20

P.S. I saw similar issue https://github.com/ultrabug/py3status/issues/2025, but it was closed as cosmetic, and I think this is worthy to be raised again.

P.P.S. 1-second interval should be broken as well, but it's much harder to see the delay in this case.

P.P.P.S perf counter should never be used for wall-clock-like things, it's only purpose - to measure some intervals of time on the same machine, e.g `t = time.perf_counter(); some_work; time_diff = time.perf_counter() - t`

eumiro:
Using `time.perf_counter()` to decide when to update the clock was my wrong decision in #1996 because its zero does not necessarily start at xx:xx:00. Revert to `time.time()`.

Thanks to @ya-isakov for reporting the bug
Resolves #2028

### DIFF
--- a/py3status/modules/clock.py
+++ b/py3status/modules/clock.py
@@ -184,7 +184,7 @@ class Py3status:
             self.active = 0
 
         # reset the cycle time
-        self._cycle_time = time.perf_counter() + self.cycle
+        self._cycle_time = time.time() + self.cycle
 
     def _get_timezone(self, tz):
         """
@@ -217,7 +217,7 @@ class Py3status:
     def _change_active(self, diff):
         self.active = (self.active + diff) % len(self.format)
         # reset the cycle time
-        self._cycle_time = time.perf_counter() + self.cycle
+        self._cycle_time = time.time() + self.cycle
         # save the active format
         timezone = self.format[self.active]
         self.py3.storage_set("timezone", timezone)
@@ -241,9 +241,9 @@ class Py3status:
     def clock(self):
 
         # cycling
-        if self.cycle and time.perf_counter() >= self._cycle_time:
+        if self.cycle and time.time() >= self._cycle_time:
             self._change_active(1)
-            self._cycle_time = time.perf_counter() + self.cycle
+            self._cycle_time = time.time() + self.cycle
 
         # update our times
         times = {}


ultrabug:
@eumiro do you have an idea to make py3status synchronize its schedule mechanism to the round of a minute? like 12:20:00

This would allow all clock related modules to look synchronous when a minute change, just like i3status does

eumiro:
We have the real problem only if `cycle == 60`, because otherwise we update it every second and that does always look “okay”, right? And we have no other cycles than 0/1 and 60. So if it is 60, we'll have to calculate the number of seconds until the next full minute. It did work that way before I broke it with `perf_counter`, didn't it?

And did you actually consider having `py3status` work asynchronously using coroutines?

ultrabug:
> It did work that way before I broke it with perf_counter, didn't it?

We never implemented a way to "sync" **all time related modules** to the 00 seconds overall clock no

> And did you actually consider having py3status work asynchronously using coroutines?

Hell yeah, I also have a working code base and as a matter a fact I opened #2030 to start this in the main repo. I'm changing strategy and will try to allow users select the v4 engine if they want until we can switch default and later deprecate v3

I still need to move more synchronous / v3 related code base to the v3 folder and then we will need to introduce a new option to allow switching engine

ya-isakov:
It seems that this commit doesn't fix the main issue, as it is in the time_in function, which is called here: https://github.com/ultrabug/py3status/pull/2031/files#diff-f9614e7a4b792c9623120c813eedfc9af89228525b52cba89b4fb2ad00976333L297.

I think, time_in can be changed like that:
```
if sync_to:
            requested = (requested + sync_to) - (time.time() % sync_to)
```
This way, it will add exact amount of seconds till the next minute.

ultrabug:
Brilliant @ya-isakov thanks, it indeed works like a charm; your fix actually fixed my own desync of tztime modules...

> We never implemented a way to "sync" all time related modules to the 00 seconds overall clock no

That's absolutely false... I was mislead by my own experience and forgot about sync_to on the time_in function... but the main problem was that I was suffering from the same issue as @ya-isakov 

Everything looks fine to me now!

PR #2037, opened 2021-04-08
Adapted to kdeconnect version 20.12.3

Urban82:
Adapted kdeconnector module to new interface on version 20.12.3

Battery and notifications methods were moved to a subfolder under device_id folder.


ultrabug:
Thanks for your contribution @Urban82 and I apologize for the answer delay.

To my understanding, this change is not backward compatible for users stuck with older kdeconnector and we usually don't like to break things for existing users :/

Do you think this code could be made backward compatible?

Urban82:
Hi @ultrabug, I think in this way it should work with older versions, but I have no way to check it at the moment because I cannot downgrade my installed version.

ultrabug:
Thanks @Urban82 ! I'll allow myself some cosmetic modifications after merging :+1:

PR #2050, opened 2021-06-10
Added support for displaying wind speeds in knots

tknapstad:
The knot is the unit of windspeed used in avionics and nautical settings. This PR adds support for knot as a windspeed unit.


ultrabug:
Thanks a lot for your (first!) contribution @tknapstad !

PR #2053, opened 2021-06-22
Work around bug in spotifyd

benjaminvdb:
There's [a bug in spotifyd](https://github.com/Spotifyd/spotifyd/issues/890) that causes the `PlayPause` toggle command not to work. To work around this, I've added a check for the current state of the player before sending either a `Play` or a `Pause` command, which effectively toggles the player.

### DIFF
--- a/py3status/modules/spotify.py
+++ b/py3status/modules/spotify.py
@@ -129,6 +129,13 @@ class Py3status:
         expression = expression.replace("META_WORDS_HERE", meta_words)
         return re.compile(expression, re.IGNORECASE)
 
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks for your contribution @benjaminvdb :) I'll merge and fix black formatting + add a note on the code logic pointing to the issue you mentioned for later reference

benjaminvdb:
@ultrabug Thanks! I completely forgot to check the contribution guidelines so I didn't know about Black. I'll remember to check the guidelines next time :smile:

PR #2055, opened 2021-07-06
_format_sunrise and _format_sunset as aware dt

msk-box:
Because naive datetime objects are treated by many datetime methods as local times, it would be better to make those objects aware to show correct time.
Also if we would like to force UTC it is recommended way to create an object representing a specific timestamp in UTC is by calling datetime.fromtimestamp(timestamp, tz=timezone.utc)


ultrabug:
You're right, this is confusing thank you for your first contribution @msk-box

PR #2057, opened 2021-07-10
Update clementine.py

ChoiZ:
Add https for internet radio streams over ssl.


lasers (commented):
Not necessary. We have `"http" in title`.
```python
~ ❯ python
Python 3.8.8 (default, Feb 19 2021, 11:04:50) [GCC 9.3.0] on linux

>>> title = "https://github.com/ultrabug/py3status"
>>> "http" in title
True

>>> ("http" in "http", "http" in "https")
(True, True)
```

PR #2065, opened 2021-09-17
Add configurable status string to battery_level module

oghm2:
Add the option to have a configurable status string (rather than an icon) in the output. (Roughly) this functionality is available in the original i3status.


ultrabug:
Thanks @oghm2 sorry if I was not clear, but can we loose the `string` part on the variable names too?

Like `format_status_bad` instead of `format_status_string_bad`

Issue #2059, opened 2021-07-27
time and tztime not updating on suspend/resume cycle


aplund:
I use the tztime module (which uses the i3status command) and when I wake the machine from suspend the displayed time is wrong for a minute.  I don't display seconds ("%H:%M" format).  i3status seems to give good output when I run 'i3status -c /tmp/py3status_deadbeef'.  There must be some problem in translation in py3status under this time jump condition.

**Version**
py3status version 3.38 (python 3.9.6) on i3

(this lies, I'm on sway).

**To Reproduce**

1. Add module 'time'
2. format = '%H:%M"
3. Run py3status on sway (well, that's all I've tested it on).
4. Suspend machine
5. Resume after some long time
6. See old time displayed for a minute

**Expected behaviour**
The time should be refreshed upon resume

ultrabug:
Hi @aplund 

I think this is because your format does not use `%S` and as such py3status does not trigger an immediate refresh of i3status output, could you tell me if it works if you add seconds to the format output?

Thanks

ultrabug:
Hi again, FYI I tried your format and I sent a SIGTSTP (pause) then a SIGCONT (continue) as this is what happens when going into suspend mode... but I cannot reproduce your problem I'm afraid (i3wm here)

You should get those signals on suspend/resume and it should be confirmed in your logs if you run py3status with the `-d` flag, can you see that?

aplund:
I added a string fmt with the "%S".  Here's a screenshot after a suspend resume cycle.

Seems the tztime bits without "%S" don't update immediately.

![20210805_12h33m13s_grim](https://user-images.githubusercontent.com/1369436/128281909-61a9108d-7fbd-4516-ae4c-557c8f6da07b.png)

Note there is a 1 hour 41 minute discrepancy between the formats without "%S" compared the the last one with a "%S".

The config for the tztime elements is:

```
tztime brisbane {
        format = "%H:%M %Z"
        timezone = "Australia/Brisbane"
}

tztime berlin {
        format = "%H:%M %Z"
        timezone = "Europe/Berlin"
}

tztime local {
        format = "%Y-%m-%d %H:%M:%S %Z"
```

aplund:
I did an strace on what the py3status thread that reads from i3status sees.  I've also taken a screenshot shortly after resume.

The reading from the child process seems to be working fine.

[...]

PR #2068, opened 2021-10-05
inhibit conflicting SIGCONT SIGTSTP that can be received too fast

ultrabug:
some OS seem to be sending a SIGCONT and SIGTSTP when resuming
from sleep or from some udev events

their order is not always clear and this leads to blocking
py3status when SIGTSTP is received last

we inhibit any SIGTSTP coming withing 1sec after any SIGCONT


ultrabug:
@aplund FYI, this fixes the problem I reproduce in #2059 

I will merge this soon and release 3.40

PR #2069, opened 2021-10-10
whoami module: Add {hostname} placeholder to output

oghm2:
As per the title. Allows format strings like "{username}@{hostname}" which I find useful with multiple machines.


ultrabug:
Nice, thanks!

Issue #2076, opened 2021-11-08
[Mpris] BrokenDBus is broken in multible ways (click event, init metadata, subscription disconnect handling)


valdur55:
**Describe the bug**
 BrokenDBus is broken in multible ways:

-  Clicking on mpris module while chrome/chromium playing is stopping statusbar update
- Metadata and playbackstatus is none on BrokenDBusMpris init.
- player_on_change function gets called by old no existent player_id and it will cause module crash

**Your py3status version**
3.40 . Problem is also in older versions

**To Reproduce**
**Clicking on mpris module while chrome/chromium playing is stopping statusbar update**

1.  Open media in chrome/chromium
2. Click on mpris module
3. Check if mediaplayer is now paused or not

**Metadata and playbackstatus is none on BrokenDBusMpris init.**

1. Open media in chrome/chromium
2. Check mpris player status
3. Reload i3wm
Check if player status is same or not. Expect playerstatus "stopped"

**player_on_change function gets called by old no existent player_id and it will cause module crash**

1. Open media in chrome/chromium
2. Close all chrome/chromium windows
3. Open media in chrome/chromium
4. Check if py3status values are updated or not

ultrabug:
@valdur55 I guess we can close this one thanks to you

PR #2077, opened 2021-11-08
Mpris bugfix

valdur55:
Intial scope:
Draft to fix issues which described in bugreport #2076
BrokenDbus supports now Play/Pause/Stop/Next/Prev media commands.
Player metadata and playbackStatus gets updated during BrokenDBusMpris init.
BrokenDbus PlayerOnChange subscription is disconnected during player_remove.
For easier debugging I made some exceptions more exceptional and reduced player_add and player_remove calls count. 

Current scope from commit 94e5650
Calculate ptime only when time is in format.
Use one callback to handle all players PlayerOnChange event.
PlayerOnChange event handler really knows signal sender.
self._player_details.get("id") - > self._player_details.get("_id")


py3status/modules/mpris.py
```diff
@@ -256,15 +256,15 @@ def _get_text(self):
         ptime = None
 
         if hasattr(self._player, "Position") and self.py3.format_contains(
```
valdur55, line 258: This line black formatting is ugly.

valdur55: Force push broke things... This comment isn't at corret line.

valdur55:
For now I have used it for a while. Time for conclusion.

**Key changes for mpris module:**
1. Using mpris2 and pydbus as backend which allows handle chrome/chromium browser as same way as other players. ( Probably best fix for #1827 )
2. Added config button_next_player and button_prev_player which switches between same status players as top priority player
3. Added config player_hide_non_canplay for hiding chrome/chromium mpris instances which otherwise hides only when chrome/chromium is closed after playing some media.
4. Added button_switch_to_top_player for switching top priority player.Player priority and name index handles corretnly vlc and chrome.instanceXXX.
5. Requesting for mediaplayer metadata only when it is really used.
   It is in two stages. 
       1. If no metadata used mpris will deny all metadata player_on_change calls. 
       2. If some metadata is used then asked metadata ammont is checked against format. (format_contains).
6. Subscribing one function for handling mediaplayer properties change and function knows signal sender player.
7. Sending play/pause/next/prev/stop signal to player will trigger update only once after mediaplayer has reported new state.

**Additional out of mpris scope changes:**
* events.py After i3wm reload first mouse event triggered unexpected error.
* core.py Drop negative update_due to prevent endless loop which causes high cpu. More info: #2085

artizirk:
Tested this on my computer, seems to work well!

ultrabug:
Thanks a lot for your hard work @valdur55

PR #2083, opened 2021-11-27
net_iplist: support parsing iface name with "@" symbol

blahgeek:
https://unix.stackexchange.com/questions/358982/network-interface-name-has-with-at-sign-what-is-it


ultrabug:
LGTM, thank you!

PR #2086, opened 2022-01-13
window module: fix crash on sway / wlroots compositors

steinex:
Sway / wlroots compositors have special overlay windows that never get
focus and are floating above everything else. Examples for this are
bemenu or notification daemons.

In some cases it was possible to trigger an update in the window module,
e.g. by calling bemenu twice which crashed the module. I'm not entirely
sure why that happened but this PR fixes it.

All credits go out to @valdur55 who actually authored the patch and
helped me debugging it in the first place.

### DIFF
--- a/py3status/modules/window.py
+++ b/py3status/modules/window.py
@@ -81,10 +81,14 @@ class I3ipc(Ipc):
         i3.main()
 
     def clear_title(self, i3, event=None):
-        self.update(i3.get_tree().find_focused())
+        focused = i3.get_tree().find_focused()
+        if focused:
+            self.update(focused)
 
     def change_title(self, i3, event=None):
         focused = i3.get_tree().find_focused()
+        if not focused:
+            return
 
         # hide title on containers with window title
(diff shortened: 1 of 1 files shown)


valdur55:
As  I remember for get into buggy state you had to use bemenu binding two times in row.
Error was: "AttributeError: 'NoneType' object has no attribute 'name'

At first point you asked how to fix crash occuring in line 104 https://github.com/ultrabug/py3status/blob/80ec45a9db0712b0de55f83291c321f6fceb6a6d/py3status/modules/window.py#L102-L104

Then I looked into places where update function is called out and found possible crash in line 92: https://github.com/ultrabug/py3status/blob/80ec45a9db0712b0de55f83291c321f6fceb6a6d/py3status/modules/window.py#L86-L92

Root cause of problem is wrong assumption of i3.get_tree().find_focused() return types.
It can renturn None or focused item from tree.

steinex:
Superseded by #2087 which also fixes the problem mentioned here.

Issue #2085, opened 2022-01-04
High cpu load after clicking mrpis pause.


valdur55:
**Describe the bug**
High cpu load by py3status. 
update_due gets negative value at line https://github.com/ultrabug/py3status/blob/master/py3status/core.py#L1031
It causes endless and rapid loop which moves to negative direction.
I made branch where issue is always occuring: https://github.com/valdur55/py3status/tree/mpris-cpu-load
And added one if check which reports: "Negative update_due occured. Reseting to 1" when issue happens.

Using VLC as media player (maybe it matters).

**Your py3status version**
3.40 (python 3.8.10)

**To Reproduce**
Steps to reproduce the behavior along with your py3status configuration (filter out any private values):
1. Add module 'mpris'
2. Configure module like this '{time} {player}'   (with my branch no need for that)
3. Run py3status.
4. Make player to playing state
5. Change player to paused state
6. High CPU load occurs.  ( in my mpris-cpu-load there will be log message "Negative update_due occured. Reseting to 1")

**Expected behavior**
CPU load should come down because paused state player will not update it's time value.

valdur55:
Did some analysis and found data flow:
When mediaplayer is playing, cache_until is set to  time.perf_counter() + 0.5
When mediaplayer gets non-playing state,  cache_until is set to -1
Update request from playing state is still up and when update time comes it isn't checking if cache_until is -1 and gets into endless loop for waiting module update/change.


Data for update_due values during high cpu load:
Update due -8.550362663998385
Update due -8.561496337002609
Update due -8.614066078007454
Update due -8.62588996499835
Update due -8.678577004000545
Update due -8.731410777996643
Update due -8.742878290009685
Update due -8.7540116480086
Update due -8.806623505006428

valdur55:
Made bug occuring more noticable... And here is full log: https://termbin.com/m4h0v
Mpris module is using normal time_perf() + 0.5 update and nagbar will noify about due_time which are between -0.1 and 0.
Critical place: https://github.com/valdur55/py3status/blob/mpris-cpu-load/py3status/core.py#L406

valdur55:
Made update at my test branch and made module which consist only logic for changing cache_timeout and string which updates on every module call. ( https://github.com/valdur55/py3status/blob/mpris-cpu-load/py3status/modules/autoupdate_sandbox.py )
To repro issue. You need to click once left click to start updating and second click for stopping auto update.
[...]

PR #2092, opened 2022-01-18
[core] Drop and log negative new_update_due which causes high cpu load and endless loop (Address #2085)

valdur55:
(Address #2085)


ultrabug:
I'm merging this so you get the credit of your finding but will be modifying it slightly thank you!

PR #2097, opened 2022-01-31
weather_owm module: fix forecast by using new onecall api endpoint

### LINKED ISSUE #2096: weather_owm: forecast_days controls 3-hour intervals, not daily intervals
The documentation suggests that `forecast_days` controls days, but in reality, it returns 3-hour intervals. The current forecast uses [the 5 Day / 3 Hour API](https://openweathermap.org/forecast5), so the data returned in `list` is separated by 3 hours, and not by days. Either:

* Change [_get_forecast](https://github.com/ultrabug/py3status/blob/master/py3status/modules/weather_owm.py#L543) to somehow take `cnt: 8*self.forecast_days` instead, and then take every eighth interval, or
* Use the [One Call API](https://openweathermap.org/api/one-call-api), which has a `daily` field. This is probably better, but will involve more code change, as the output format is different.

ultrabug:
thx to @cjquines

closes #2096

### DIFF
--- a/py3status/modules/weather_owm.py
+++ b/py3status/modules/weather_owm.py
@@ -266,6 +266,7 @@ import datetime
 # API information
 OWM_CURR_ENDPOINT = "https://api.openweathermap.org/data/2.5/weather?"
 OWM_FUTURE_ENDPOINT = "https://api.openweathermap.org/data/2.5/forecast?"
+OWM_ONECALL_ENDPOINT = "https://api.openweathermap.org/data/2.5/onecall?"
 IP_ENDPOINT = "http://geo.ultrabug.fr"
 
 # Paths of information to extract from JSON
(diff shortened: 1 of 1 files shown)


lucilanga:
There are some problems with onecall implementation.

If I may comment on this change, what you basically done here is replace _get_weather() with onecall['daily'][0] for current weather.
As per documentation the list onecall['daily'] contains: Daily **forecast** for 7 days.
We cannot use forecast for day 0 as current weather, because day 0 values are just forecasted values (eg. min temperature of the day, max temperature of the day and average of both values).
We could use onecall['current'] but it's not a drop-in replacement, the structure differs slightly from onecall['daily'] (eg. we do not have gust on wind)
On the other hand we cannot go back to _get_weather() either, because of onecall changes in the code and it wouldn't make sense to use 2 different API calls anyway.

ultrabug:
would you like to propose a PR for a better fix @lucilanga ?

ultrabug:
@lucilanga please could you check https://github.com/ultrabug/py3status/pull/2112 ?

PR #2098, opened 2022-02-06
arch_updates module: paru exits with a non-zero status if there are no updates

itaton:
Fixes update checks with paru.
Unlike yay, paru exits with return value `1` after `paru -Qua`, if there are no AUR updates available (conforming to the behavior of `pacman` itself)
Thus we should return `0` if `-Qua` exits with `1`.


ultrabug:
Thank you mate!

PR #2108, opened 2022-03-10
[core] Proper prevention for negative update_due and log it as warning.

valdur55:
Thx Ultrabug for courage to look deeply into problem.


ultrabug:
Thanks

PR #2109, opened 2022-03-10
Try to fix root cause of negative update_due

valdur55:
Auto update makes assumption that there isn't interruption between updates.
Now when module sends CACHE_FOREVER then system checks if pending update_due is from module (from previous auto update cycle) and then sets next update_due.... 
Address #2085

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -297,10 +297,17 @@ class Py3statusWrapper:
     def clear_timeout_due(self, module):
         old = self.timeout_queue_lookup_previous.get(module, None)
         if old and old == self.timeout_due:
-            if len(self.timeout_keys):
-                self.timeout_due = self.timeout_keys[0]
-            else:
-                self.timeout_due = None
+            self.set_new_timeout_due()
+
+    def set_new_timeout_due(self):
+        # sort keys so earliest is first
+        self.timeout_keys.sort()
+
+        # when is next timeout due?
+        try:
+            self.timeout_due = self.timeout_keys[0]
+        except IndexError:
+            self.timeout_due = None
 
     def timeout_process_add_queue(self, module, cache_time):
         """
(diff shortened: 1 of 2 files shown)


py3status/core.py
```diff
@@ -326,18 +346,13 @@ def timeout_process_add_queue(self, module, cache_time):
...
-                try:
-                    self.timeout_due = self.timeout_keys[0]
-                except IndexError:
-                    self.timeout_due = None
+                self._set_new_timeout_due()
             else:
                 self.timeout_queue[cache_time].add(module)
             # note that the module is in the timeout_queue
             self.timeout_queue_lookup[module] = cache_time
+            self.timeout_queue_lookup_previous[module] = cache_time
```
ultrabug, line 355: Looks to me that `self.timeout_queue_lookup_previous[module]` should also be cleared when `self.timeout_queue_lookup[module]` and not ONLY when `cache_time == Py3.CACHE_FOREVER:` ; else this dict will only grow in memory

valdur55: Well. We can clear it after we use it's value otherwise there will be no pointer for detecting ongoing timeout_due for module. Or there is other possibilities to overcome that issue?
Did you test your ideas by yourself or you just shared feeling?

valdur55: Definetly we need hold old values because at the runtime there isn't available data at dict self.timeout_queue_lookup.
Well after opening  debugger i noticed variable timeout_running which needs some attention.

ultrabug: Yes, but if the module gets processed, it can safely be dropped from timeout_queue_lookup right?

So, this could be added on L367 I guess

ultrabug: Actually I'm stupid sorry, the key is the module so this won't pile up ...

PR #2112, opened 2022-03-21
weather_owm: dont use onecall for better current day forecast

ultrabug:
thx to @lucilanga


ultrabug:
@mlmatlock FYI as per https://github.com/ultrabug/py3status/pull/2097#issuecomment-1067771255

lucilanga:
Hey ultra,
thanks for working on this.
So you basically brought _get_weather back as _get_current_weather for current day.
There is nothing wrong with that, as a matter fact I am under the impression that _get_current_weather is somewhat more accurate than onecall['current'].
However as we already use onecall for foercast wouldn't be cheaper in terms of api calls to use a single call and parse 'current' data? .. just asking?

Come to think of it, perhaps one might want to  display just  current_weather or just the forecast? .. which indeed makes sense to have separate calls. 

Lastly I was going to work on a PR but got sidetracked and to be perfectly honest I'm not very accustomed with github's PRs.

ultrabug:
>  However as we already use onecall for foercast wouldn't be cheaper in terms of api calls to use a single call and parse 'current' data? .. just asking?

it would, but we actually need it to get a more precise city location back from OWM which is not included in the onecall API. So having both helps for accuracy.

> Lastly I was going to work on a PR but got sidetracked and to be perfectly honest I'm not very accustomed with github's PRs.

np

py3status/modules/weather_owm.py
```diff
@@ -627,9 +638,15 @@ def _format_snow(self, wthr):
...
     def _format_wind(self, wthr):
         wind = {
-            "deg": self._jpath(wthr, OWM_WIND_DEG, ""),
-            "gust": self._jpath(wthr, OWM_WIND_GUST, ""),
-            "speed": self._jpath(wthr, OWM_WIND_SPEED, ""),
+            "deg": self._jpath(
+                wthr, OWM_WIND_DEG, self._jpath(wthr, OWN_CURRENT_WIND_DEG, "")
+            ),
+            "gust": self._jpath(
+                wthr, OWM_WIND_GUST, self._jpath(wthr, OWN_CURRENT_WIND_GUST, "")
```
lucilanga, line 645: if gust/speed are not found  why replace with "" ?
Below checks will then fail:

# Speed and Gust
msec_speed = wind["speed"] if ("speed" in wind) else 0
msec_gust = wind["gust"] if ("gust" in wind) else 0

and later you'll end up with non-float value.

mlmatlock:
Thanks, @ultrabug and @lucilanga!

I downloaded the updated module file and loaded it up.  First restart, nothing; so I ran the module in test mode.

@lucilanga was right about the double quotes in the wind defs:

```
can't multiply sequence by non-int of type 'float'.
Traceback
TypeError: can't multiply sequence by non-int of type 'float'
  File "/usr/lib/python3.10/site-packages/py3status/module.py", line 948, in run
    response = method()
  File "/home/merell/.config/py3status/modules/weather2.py", line 855, in weather_owm
    text = self._format(
  File "/home/merell/.config/py3status/modules/weather2.py", line 809, in _format
    today = self._format_dict(current_wthr, city, country)
  File "/home/merell/.config/py3status/modules/weather2.py", line 791, in _format_dict
    "wind": self._format_wind(wthr),
  File "/home/merell/.config/py3status/modules/weather2.py", line 645, in _format_wind
    "gust": msec_gust * FT_FROM_METER,
```
On a lark, I changed all of the double quotes in wind defs to 0 (zero), module passed self test, and loaded properly in the running py3status bar.  It appears to be displaying proper current conditions, *and* the proper forecast icons.

I'll revisit this post in a couple of hours, after a few cache timeouts, and with the sun going down, hopefully some condition changes as well.

py3status/modules/weather_owm.py
ultrabug: good point, ty

PR #2120, opened 2022-04-07
[mpris] Add player_shortname for displaying player busname

valdur55:
It uses name from busname: org.mpris.MediaPlayer2.vlc will use "vlc" instead of player "VLC mediaplayer"


ultrabug:
Thanks!

PR #2121, opened 2022-04-11
[xrandr] fix missing " at the end of line (example section).

valdur55:
Found it when I looked config examples for issue #2056


ultrabug:
Thanks!

PR #2124, opened 2022-04-22
Migrate to new Twitch API, fix #2123

### LINKED ISSUE #2123: Twitch "kraken" API is deprecated and has been disabled
**Describe the bug**
The twitch module doesn't work anymore, because the "kraken" API has been disabled after being deprecated two years ago.

**Your py3status version**
py3status version 3.43 (python 3.9.7) on i3

**To Reproduce**
1. Add module 'twitch'
2. Configure module like this
     ```
        twitch {
                client_id = "..."
                stream_name = "iskall85"
        }
     ```
3. Run py3status

**Expected behavior**
Module behaves as documented and displays the status of the streamer.

**Additional context**
[deprecation notice](https://discuss.dev.twitch.tv/t/legacy-twitch-api-v5-i-e-kraken-shutdown-reminder-february-28-2022/36589)

I'll be looking into providing a PR implementing the new API and the new auth-flow tonight.

jpicht:
This PR fixes the twitch module, that relied on an API that is no longer available. (closes #2123)

While I was at it, I documented the fields that were there all along (though they now have a prefix, depending on which endpoint provides them), documented the new fields provided by the new API and added the option to fetch stream tags. Some of the fields probably don't seem very useful at first glance, but might be interesting for some, if combined with `?if=` expressions, so I left them in.

### DIFF
--- a/py3status/modules/twitch.py
+++ b/py3status/modules/twitch.py
@@ -6,15 +6,41 @@ Configuration parameters:
         (default 60)
     client_id: Your client id. Create your own key at https://dev.twitch.tv
         (default None)
+    client_secret: Your client secret.
+        (default None)
     format: Display format when online
         (default "{display_name} is live!")
     format_offline: Display format when offline
         (default "{display_name} is offline.")
     stream_name: name of streamer(twitch.tv/<stream_name>)
         (default None)
+    trace: enable trace level debugging
+        (default False)
 
 Format placeholders:
-    {display_name} streamer display name, eg Ultrabug
+    {display_name} User's display name., eg Ultrabug
+    {is_streaming} (bool) True if streaming, fields prefixed with stream_ are available.
+    {user_id} User's id
+    {user_login} User's login name, eg xisumavoid
+    {user_display_name} (same as {display_name})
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @jpicht thanks for your first contribution to py3status! :)

I've allowed the CI to run to help you work through getting all our QA requirements green first

Thanks

py3status/modules/twitch.py
```diff
@@ -6,15 +6,52 @@
...
         (default None)
+    client_secret: Your client secret.
+        (default None)
     format: Display format when online
         (default "{display_name} is live!")
     format_offline: Display format when offline
         (default "{display_name} is offline.")
+    format_tag: Tag formatting
+        (default "{name}")
+    locale: List of locales to try for tag translations, eg. ["cs-cz", "en-uk", "en-us"]. If none is specified, auto-detect from environment, with a fallback to "en-us".
```
ultrabug, line 17: can we make this plural since this is a list please?

jpicht: done

ultrabug:
Thanks a lot @jpicht !

PR #2125, opened 2022-05-29
mail: fix pathlib adoption

hblanks:
os.path.expandvars() returns a string, not a Path. So, for
commit 37603bf3 to have worked, we needed to expand vars
first into a string, then convert to a path, then call
Path.expanduser().

Observed traceback prior  to the fix for added context, from current master branch:

```
2022-05-29 14:48:49 WARNING Exception in `mail` post_config_hook().
2022-05-29 14:48:49 INFO Traceback
AttributeError: 'str' object has no attribute 'exists'
  File "...py3status.venv/lib/python3.9/site-packages/py3status/module.py", line 143, in prepare_module
  File "...py3status.venv/lib/python3.9/site-packages/py3status/modules/mail.py", line 221, in post_config_hook
    if not path.exists():
```

### DIFF
--- a/py3status/modules/mail.py
+++ b/py3status/modules/mail.py
@@ -216,7 +216,8 @@ class Py3status:
                             if "path" not in account:
                                 raise Exception(STRING_MISSING.format(mail, "path"))
                             path = Path(
-                                os.path.expandvars(account["path"])).expanduser()
+                                os.path.expandvars(account["path"])
+                            ).expanduser()
                             if not path.exists():
                                 path = f"path: {path}"
                                 raise Exception(STRING_MISSING.format(mail, path))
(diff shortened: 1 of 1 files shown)


hblanks:
hi! I ran black on the fie I changed and pushed the diff, but I couldn't find any docs on how to run your CI steps locally. . . . and github won't run the build for me as I'm a first time contributor.

I did install tox and tox-gh-actions in a virtualenv and then run tox. It was at least happy with the python 3.9 build.

So, let me know if it's OK! Thanks for your work.

hblanks:
> @hblanks just run make qa and watch the magic please +1

Yeah. .. that's essentially what I did (`virtualenv /tmp/foo; /tmp/foo/bin/pip3 install tox; /tmp/foo/bin/tox`) and as noted it passed. My only local interpreter is 3.9 as noted.

Anyway - looks like the build is happy but had a transient failure of some kind in unpacking for one of the trials:

actions

I don't think I'm able to retry builds for you but if you're able to retry [CI #452](https://github.com/ultrabug/py3status/runs/6703018872?check_suite_focus=true) then you this PR should go green for you.

ultrabug:
Thanks!

PR #2131, opened 2022-06-22
ignore ZFS ARC cache in memory usage

anarcat:
Workstations with a ZFS filesystem will report an abnormal memory
usage due to the ZFS ARC cache. For example, on my workstation, the
ARC cache takes up 6GB out of the 16GB core memory. Total memory
usage, as reported by py3status, is close to 87%, quite alarming, but
when we *know* about the ZFS cache, this is no big deal.

Note that tools like `top(1)` also get it wrong, but `htop(1)` and
`btop(1)` don't. I actually took this idea from btop.

I didn't make this customizable because there's one of two things
here, either:

 1. the user doesn't have a ZFS setup and this will silently
    not change anything, or;

 2. the use *has* a ZFS setup and they will almost *certainly* want to
    count that out of their memory usage

btop(1) actually makes this configurable, but I think that's
superfluous: who would want to count that in anyways?

Obviously, this is kind of a flaw in the (Linux?) implementation of ZFS:
ZFS should use the normal caching systems of the operating system. But
considering how different that cache is, that might be practical (the
Linux memory cache is a LRU cache, and ZFS is an ARC cache).

I suspect this may also work on other ZFS implementations, but I
haven't tested those.

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -107,6 +107,7 @@ from fnmatch import fnmatch
 from json import loads
 from os import getloadavg
 from pathlib import Path
+import re
 
 INVALID_CPU_TEMP_UNIT = "invalid cpu_temp_unit"
 STRING_NOT_INSTALLED = "not installed"
@@ -365,12 +366,14 @@ class Py3status:
         of used memory, and units of mem (KiB, MiB, GiB).
         """
         if memory:
+            zfs_cache = self._get_zfs_arc_size()
             total_mem_kib = meminfo["MemTotal:"]
             used_mem_kib = (
                 total_mem_kib
                 - meminfo["MemFree:"]
                 - (
                     meminfo["Buffers:"]
+                    + zfs_cache
                     + meminfo["Cached:"]
                     + (meminfo["SReclaimable:"] - meminfo["Shmem:"])
                 )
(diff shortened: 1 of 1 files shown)


ultrabug:
Hi @anarcat I sincerely apologize for the delay in getting back to you...

Thanks for this nice contribution, I don't know why some checks did not run on your PR. Did you make sure that this commit pass the py3status tests?

I agree with you in this added computation not needing to be configurable **but** I think we should catch all Exceptions when reading the ARC cache to accommodate with possible different setups. I'd rather report a bad number than break users' RAM reporting which they've been happy with so far.

Could you fix that please?

anarcat:
> Thanks for this nice contribution, I don't know why some checks did not run on your PR. Did you make sure that this commit pass the py3status tests?

i did not, sorry.

> I agree with you in this added computation not needing to be configurable **but** I think we should catch all Exceptions when reading the ARC cache to accommodate with possible different setups. I'd rather report a bad number than break users' RAM reporting which they've been happy with so far.

What are you refering to here? Maybe this block?

```
                    m = ZFS_SIZE_REGEX.match(line)
                  if m:
                        return int(m.group(1))/1024
```

i guess i could handle `m.group` better and handle exceptions on failures to cast as an `int()`, anything else you had in mind?

ultrabug:
L423 you could maybe just catch any Exception and ignore it

anarcat:
isn't that considered bad practice?

anarcat:
ack, wilco.

anarcat:
i split the difference: i broadened the `FileNotFoundError` into `OSError`, which should catch any open/read-related error, and added a ValueError for the `int()` cast. I cannot think of *any* other exception this might raise, so we won't raise any with this *except* if someone changes the code in the future and introduces some Deprecation or SyntaxError....

i am okay with switching to a plain `Exception` if you really insist, but I cannot in good conscience just do that on my own. :p

anarcat:
oh, and i just ran the test suite - looks like i had overlooked that in the first time, and now it passes. (i had to run `black`, otherwise things were okay.)

ultrabug:
Congratulations, and most importantly thank you for your first contribution to py3status!

PR #2137, opened 2022-08-17
adds module audiosink

flyingapfopenguin:
This PR introduces a new module to display the current default audiosink (from pulseaudio) and let the user toggle it via left click.

The user can ignore sinks they never use via config.
Additionally short names for the devices can be configured such that no long device names get printed.

### DIFF
--- a/py3status/modules/audiosink.py
+++ b/py3status/modules/audiosink.py
@@ -92,7 +92,7 @@ class Py3status:
 
     def audiosink(self):
         composites = [
-            {"full_text": self._to_string(self._get_state()),},
+            {"full_text": self._to_string(self._get_state()), },
         ]
         audiosink = self.py3.composite_create(composites)
         cached_until = self.py3.time_in(0, offset=1)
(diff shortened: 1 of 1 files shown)


ultrabug:
Thank you @flyingapfopenguin that's an interesting module indeed!

Could you start by making sure tests pass please?

flyingapfopenguin:
I tried to make the test pass, but failed.

In my original commit `7cd4e3c` black wants three lines merged, with no white space behind the comma after the the only dictionary entry. The next commit `7cec30d` applies exactly this diff, which leads to FLAKE8 (I don't know this tool) missing this white space. Adding it via `cc3d25e` makes black complain again, which is consistent to blacks behaviour considering my original commit.

I'm sorry, but I don't know how to win this game.

ultrabug:
Ok let me try this myself then.

ultrabug:
Thanks a lot again, that's definitely a module that I'll be using everyday :D

PR #2138, opened 2022-08-23
Report git commit information consistently

aplund:
The reporting of an underlying git source does not work in a range of
instances.  For example, if the repo is part of a submodule, then no git
information will be reported.

Using the python git module gives more consistent information about what
git repo that we are running from.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -542,6 +542,50 @@ class Py3statusWrapper:
                 msg = f'Loading module "{module}" failed ({err}).'
                 self.report_exception(msg, level="warning")
 
+    def _log_gitversion(self):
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks for your time in improving py3status mate; let's make QA happy first :+1:

ultrabug:
@aplund the `git` package is not a requirement of py3status so your proposal will break for those who don't have it installed.

While I understand you want to improve the current detection, I'd like to ask that at least this PR does not make a regression about the current state.

When the git package is not installed, can you fallback to the previous way of reporting git revision?

ultrabug:
Thanks a lot @aplund !

PR #2140, opened 2022-09-05
Module audiosink improvements

AnwariasEu:
Move from pacmd to pactl (libpulse).
This allows pipewire-pulse implementations to be used with this module as well.

Rename placeholder with a more expressive name.

### DIFF
--- a/py3status/modules/audiosink.py
+++ b/py3status/modules/audiosink.py
@@ -7,12 +7,12 @@ Configuration parameters:
     display_name_mapping: dictionary mapping devices names to display names
         (default {})
     format: display format for this module
-        (default '{audiosink}')
+        (default '{sinkname}')
     sinks_to_ignore: list of devices names to ignore
         (default [])
 
 Format placeholders:
-    {audiosink} comma seperated list of (display) names of default sink(s)
+    {sinkname} short name of current default sink
 
 Requires:
     libpulse: A featureful, general-purpose sound server (client library)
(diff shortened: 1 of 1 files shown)


ultrabug:
Hello

Sorry for the late reply

It seems that `pactl -f json` outputs invalid JSON like `"balance":0,00` which unfortunately breaks the module with `JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 430 (char 429)`

This is deceptive I guess...

AnwariasEu:
I have never had non-valid json output.
Which version of pcatl are your running?

Maybe it's an upstream bug?

Full JSON output should have been merged in https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/497

ultrabug:
> Maybe it's an upstream bug?

I guess it is

```
➜  ~ pactl --version
pactl 15.99.1
Compiled with libpulse 15.99.0
Linked with libpulse 15.99.0
```

ultrabug:
Looking at your linked MR on pulseaudio I figured that the C output encoder "respects" `LC_NUMERIC` which is being localized to `fr_FR.UTF-8` in my case.

And indeed, if I use `LC_NUMERIC="en_US.UTF-8" pactl -f json list sinks` I get valid JSON...

AnwariasEu:
I have 16.1.

It seems that full json support was merged in 16.0 ([Release Notes](https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/16.0/#pactlcanshowinformationinjsonformat)).
Can you try to upgrade and check again?

ultrabug:
FYI : https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/issues/1398

ultrabug:
>  Can you try to upgrade and check again?

Well I'm following Ubuntu jammy updates just like most of anyone else. I do this on purpose to be in line with what I believe most users run.

So I'm afraid that solving this that way won't be something I'm willing to support for all py3status users.

ultrabug:
@AnwariasEu how do you want to move on here please?

ultrabug:
Closing, pulseaudio timeout, if they wake up someday we'll reopen

PR #2149, opened 2022-10-04
[window] Window without title will not produce error on i3msg mode.

valdur55:
(For example Pluma closing confirmation window)


ultrabug:
Thanks

PR #2153, opened 2022-10-07
Change the reading timeout to infinity

aplund:
I'm not exactly sure for the reason for the 500ms timeout in the events thread, but it seems like an unnecessary source of wakeups.

I've been running this patch on my system for a while now and nothing seems different, except the events thread doesn't wake up twice a second.


ultrabug:
Thanks @aplund indeed this seems to come from a long time ago... 

Let me run this myself for a while, will merge & release by the end of the month

PR #2158, opened 2022-11-02
battery_level: Allow icon to not use charging_character

devkev:
I want to `{icon}` to only display the info from `blocks`, even when the battery is charging, but currently there is no way to do this (AFAICT).  This change allows the user to obtain that behaviour, if they explicitly configure `charging_character = None`.


ultrabug:
Thanks @devkev , would you please add that behavior as a comment to the given parameter in the docstring?

It's important to make sure that this feature is discoverable by everyone :+1:

PR #2165, opened 2022-12-08
volume_status: smarter initialization logic with retry, remove start_delay setting

joanbm:
This is an iteration over #2136, dealing with the case where the audio backend is not ready when i3status launches.

PR #2139 attempted to fix it by adding a configurable delay, which in my experience is not ideal, because (1) you need to know about it (by running into the issue) and (2) you need to tune it, with a too short delay being flaky and a too long delay being annoyingly slow on most cases.

Instead, add a 'smart' retry logic at startup, allowing for up to ~30 seconds of worst-case delay, but initializing instantly if possible.

This PR also fixes an unrelated bug (which I discovered while testing another approach at this): If a module has a property which throws an exception, the module will fail to load. Plus a test for it.


ultrabug:
Thank you for your nice contribution @joanbm ; I fixed the CI requirements, please rebase

joanbm:
> Thank you for your nice contribution @joanbm ; I fixed the CI requirements, please rebase

Done!

joanbm:
> Thank you for your nice contribution @joanbm ; I fixed the CI requirements, please rebase

Done!

PR #2172, opened 2023-01-14
do_not_disturb module: Use 'makoctl mode' to check current mode.

valdur55:
Requested change by irc user Steinex.


ultrabug:
Thanks

PR #2181, opened 2023-04-16
vpn_status module: add wireguard support

javiertury:
Add wireguard support. These connections are identified by `Type: wireguard` instead of the `Vpn` attribute.


ultrabug:
Thanks a lot @javiertury !

PR #2189, opened 2023-05-27
Minor patch to get the local timezone.

sixy6e:
This PR is a small patch to resolve the issue detailed in:

https://github.com/ultrabug/py3status/issues/2188

The `ZoneInfo` class doesn't seem to cater for local timezones (most likely no file detailing local information can be loaded), and the [datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) module indicates the default call returns the local timezone, eg `datetime.now(None)`.

Looking into `ZoneInfo` and I couldn't see much in the way for specifics regarding local timezones. So the safest thing seemed to return None, and explicitly handle it later.

### DIFF
--- a/py3status/modules/clock.py
+++ b/py3status/modules/clock.py
@@ -189,10 +189,7 @@ class Py3status:
         """
         # special Local timezone
         if tz == "Local":
-            try:
-                return zoneinfo.ZoneInfo("localtime")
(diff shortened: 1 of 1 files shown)


lasers (approved):
Works.

ultrabug:
Thank you for your first contribution @sixy6e !

PR #2197, opened 2023-06-14
added timezone information to times implicitly using local system tim…

oaken-source:
the changes introduced in #2188 and #2189 will produce native datetime objects, which will cause the %Z format specifier to unconditionally expand to the empty string. This may not be desirable. Alternatively, it is possible to convert the datetime object to a timezone datetime object relative to the local timezone using `astimezone()`, and then the timezone name is available from `tzname()`.

This would cause the %Z and %z parameters to be correctly populated.

Alternatively, it should be documented somewhere that to cause %Z and %z to be expanded correctly, the timezone needs to be specified explicitly in the py3status config file.

### DIFF
--- a/py3status/modules/clock.py
+++ b/py3status/modules/clock.py
@@ -250,10 +250,11 @@ class Py3status:
 
                 # special case for handling Local timezone
                 if zone is None:
-                    timezone, tzname = None, None
+                    t = t.astimezone()
+                    timezone = t.tzname()
                 else:
                     timezone = zone.key
-                    tzname = timezone.split("/")[-1].replace("_", " ")
+                tzname = timezone.split("/")[-1].replace("_", " ")
 
                 if self.multiple_tz:
                     name_unclear = tzname


lasers (approved):
```diff
diff --git a/py3status/modules/clock.py b/py3status/modules/clock.py
index 84fa36e..3a825ee 100644
--- a/py3status/modules/clock.py
+++ b/py3status/modules/clock.py
@@ -110,12 +110,7 @@ class Py3status:
     button_reset = 3
     cycle = 0
     format = "{Local}"
-    format_time = [
-        "[{name_unclear} ]%c",
-        "[{name_unclear} ]%x %X",
-        "[{name_unclear} ]%a %H:%M",
-        "[{name_unclear} ]{icon}",
-    ]
+    format_time = ["%z, %Z"]
     locale = None
     round_to_nearest_block = True
 ```
```shell
❯ python3 clock.py --term
, 
,
,
```
vs new
```shell
❯ python3 clock.py --term
-0500, CDT
-0500, CDT
-0500, CDT
```

ultrabug:
:+1:

PR #2207, opened 2023-06-23
ns_checker module: replace query() with resolve()

bbusse:
This fixes a dnspython deprecation warning:
"please use dns.resolver.Resolver.resolve() instead"


ultrabug:
Thank you!

PR #2209, opened 2023-07-11
refactor(core): drop deprecrated gevent support

ultrabug:
it's time to move on and drop gevent


lasers (approved):
Existing `-g`, `--gevent` will stop this for users. I think that's fine anyway.

ultrabug:
Yes I bet very few users are using it nowadays + they are power users anyway; thanks mate!

PR #2233, opened 2024-01-28
add brightnessctl support to backlight module

GabriF:
Hello, I added support to brightnessctl into backlight module. This is necessary because light is no longer maintained and brightnessctl is supported on wayland.


ultrabug:
Thanks for this nice first contribution @GabriF !

PR #2234, opened 2024-02-16
music modules: implement sanitize_words support for all music modules

dpeukert:
This PR adds the sanitize_words feature from the spotify module to other music modules. The common functionality has been added to the Py3 module.

I've tested all the modified modules and everything seems to be working as it should.

### DIFF
--- a/py3status/constants.py
+++ b/py3status/constants.py
@@ -270,3 +270,16 @@ RETIRED_MODULES = {
 MARKUP_LANGUAGES = ["pango", "none"]
 
 ON_ERROR_VALUES = ["hide", "show"]
+
+DEFAULT_SANITIZE_WORDS = [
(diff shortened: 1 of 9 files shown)


dpeukert:
Forgot to run tests, changing to a draft for now.

dpeukert:
I've fixed the ordering errors, but I wasn't able to fix the docstring errors, which seem to be caused by the default value for sanitize_words being a constant. Not sure how to fix that.

ultrabug:
@dpeukert would you consider switching to using what @lasers proposes directly?

dpeukert:
@ultrabug Apart from the ability to be able to configure the sanitized words (some people may want to exlude stuff like "(Live)", some may not), I don't really mind either way.

PR #2240, opened 2024-03-12
fix volume_status for default source with pamixer

oceyral:
Hi,

#2228 and #2230 introduced a regression in `volume_status` when using it with the default source.

Example configuration chunk :

```
order += "volume_status mic"

volume_status "mic" {
      is_input = True
      cache_timeout = 1
}
```

With this configuration, instead of selecting the default source, `volume_status` selects the default sink by calling `pamixer --allow-boost --get-mute --get-volume`

This PR adds `--default-source` to the command in case no device is specified and the option `is_input` is True.

### DIFF
--- a/py3status/modules/volume_status.py
+++ b/py3status/modules/volume_status.py
@@ -171,10 +171,13 @@ class Amixer(Audio):
 
 class Pamixer(Audio):
     def setup(self, parent):
-        is_input = "--source" if self.is_input else "--sink"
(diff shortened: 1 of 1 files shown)


syyyr:
Sorry for the regression. I have tested this patch, and my issue (the one I addressed in the PRs) hasn't reappeared.

ultrabug:
Thanks!

PR #2242, opened 2024-03-27
py3: add new helper: get_replacements_list

lasers:
An alternative to PRs listed below.

I copied `thresholds` code. Minimal test. Allow users to be more specific.

Get placeholders_list first so we can target existing placeholders rather than all placeholders.
I don't know if this PR is the optimal solution, just an alternative.

Code untested on this player.. I tested on `nvidia_smi`.  Sorry if you ran into issues.

This is just an idea / experiment.

Please test this.

```python
# Ought to do dict (perform on matched string placeholders)
    replacements = {
        "artist": [
            "([\(\[][^)\]]*?(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version)[^)\]]*?[\)\]])",
            "([\-,;/])([^\-,;/])*(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version).*"
        ],
    }

# Ought to do list. (perform on all string placeholders)
    replacements = [
            "([\(\[][^)\]]*?(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version)[^)\]]*?[\)\]])",
            "([\-,;/])([^\-,;/])*(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version).*"
        ]
```

Solves https://github.com/ultrabug/py3status/pull/2234#issuecomment-2017989862.

Closes https://github.com/ultrabug/py3status/pull/2234.
Closes https://github.com/ultrabug/py3status/pull/2237.
#

### DIFF
--- a/py3status/modules/cmus.py
+++ b/py3status/modules/cmus.py
@@ -16,6 +16,7 @@ Configuration parameters:
         *(default '[\?if=is_started [\?if=is_playing > ][\?if=is_paused \|\| ]'
         '[\?if=is_stopped .. ][[{artist}][\?soft  - ][{title}]'
         '|\?show cmus: waiting for user input]]')*
+    replacements: specify a list/dict of string placeholders to modify (default None)
     sleep_timeout: sleep interval for this module. when cmus is not running,
         this interval will be used. this allows some flexible timing where one
         might want to refresh constantly with some placeholders... or to refresh
@@ -100,6 +101,7 @@ class Py3status:
         r"[\?if=is_stopped .. ][[{artist}][\?soft  - ][{title}]"
         r"|\?show cmus: waiting for user input]]"
     )
+    replacements = None
     sleep_timeout = 20
 
     def post_config_hook(self):
@@ -109,6 +111,7 @@ class Py3status:
         self.color_stopped = self.py3.COLOR_STOPPED or self.py3.COLOR_BAD
         self.color_paused = self.py3.COLOR_PAUSED or self.py3.COLOR_DEGRADED
         self.color_playing = self.py3.COLOR_PLAYING or self.py3.COLOR_GOOD
+        self.replacements_init = self.py3.get_replacements_list(self.format)
 
     def _seconds_to_time(self, value):
         m, s = divmod(int(value), 60)
@@ -165,14 +168,14 @@ class Py3status:
         cached_until = self.sleep_timeout
         color = self.py3.COLOR_BAD
 
-        is_started, data = self._get_cmus_data()
+        is_started, cmus_data = self._get_cmus_data()
 
         if is_started:
             cached_until = self.cache_timeout
-            data = self._organize_data(data)
-            data = self._manipulate_data(data)
+            cmus_data = self._organize_data(cmus_data)
+            cmus_data = self._manipulate_data(cmus_data)
 
-            status = data.get("status")
+            status = cmus_data.get("status")
             if status == "playing":
                 is_playing = True
                 color = self.color_playing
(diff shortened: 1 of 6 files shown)


lasers:
Ideally, we should separate this into `artist` and `album`. Who volunteer? 
```
"([\(\[][^)\]]*?(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version)[^)\]]*?[\)\]])",
"([\-,;/])([^\-,;/])*(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version).*",
```

lasers:
Merged verison of the two regex patterns according to GPT. Untested.
```
replacements = [
    r"([\(\[\-,;/][^)\],;/]*?(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version)[^)\],;/]*?[\)\]])"
]
```

lasers:
@ultrabug Review plz.

lasers:
I didn't add this `mpris` and `mpd_status`.

I don't need this for `cmus`, `deadbeef`, `moc`... This PR might be better understood with more examples... is which why I added few more music modules. I can drop them... and users could contribute them later just easily... or more likely easier to leave them in... and work on `mpris`, `mpd_status` later. Idk.

ultrabug:
Hi @lasers sorry for the delay

I agree this approach is sane and works good enough.

I would have argued that we could have done the replace work within py3 core so that modules didn't have to all implement the same replace loop but hey, let's move forward

valdur55:
Used regex from spotify example config:
```
    replacements = {
        "album": [("\s?[\(\[\-,;/][^)\],;/]*?(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version)[^)\],;/]*[\)\]]?", "")],
        "title": [("\s?[\(\[\-,;/][^)\],;/]*?(bonus|demo|edit|explicit|extended|feat|mono|remaster|stereo|version)[^)\],;/]*[\)\]]?", "")]
    }
```
Some findings by testing it with some youtube videos:
"Queen – Bohemian Rhapsody (Official Video Remastered)" -> "Queen – Bohemian Rhapsody"
"Soul Reaver 1 & 2 Remastered - Before You Buy" -> it is broken. doesn't cut.
"Tomb Raider IV-VI Remastered - Announce Trailer | PS5 & PS4 Games" -> "Tomb Raider IV"

Generally looks promising, but most difficult part is to iron out good regex,
And it should be given to user by example.

After implementing code, got one qestion. Is there difference between using self.replacements_init vs self._replacements_init in module code?
For me public properies are meant for config parameters which comes outside of module. Code works in both ways.

lasers:
>"Soul Reaver 1 & 2 Remastered - Before You Buy" -> it is broken. doesn't cut.

That's why we have user-configurable regex now. I tried what I can to consolidate both regexes so the example config would look this clean. Does this issue occur the same with old regexes? If yes, then it's perfect, but the example can be changed later on.
 
This is also why I think it's sensible not to default `replacements` with existing regexes (for spotify and others).

>Is there difference between using self.replacements_init vs self._replacements_init in module code?

`self._replacements_init` (in py3) is used to cache compiled regex objects for performance reasons since we would be searching / replacing placeholders all time. Compile once on startup. The real reason is that I just copied it.

`self.replacements_init` (in modules) is a list of keys from user-config `replacements` (dict). If it is not a dict, it will give a list of placeholders in `format_player` (or whatever you need replacing). It's usually one or two keys (with dict).

>For me public properies are meant for config parameters which comes outside of module. Code works in both ways.

Can you explain more?

Issue #2255, opened 2024-08-27
Playerctl shows idle chromium/chrome media player after media closed.


valdur55:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I open video in chrome/chromium and close tab and media player keeps idleing on statusbar

**Your py3status version**
py3status version 3.59 (python 3.12.4) on sway

**Describe the solution you'd like**
I prefer to not see stopped chromium/chrome media player when no media is playing in chrome/chromium

PR #2256, opened 2024-08-27
playerctl module: hide non-can-play players and use chrome and chromium as default value.

valdur55:
Proposes improvement for issue #2255.

### DIFF
--- a/py3status/modules/playerctl.py
+++ b/py3status/modules/playerctl.py
@@ -24,6 +24,8 @@ Configuration parameters:
         *(default '[\?color=status [\?if=status=Playing > ][\?if=status=Paused \|\| ]'
         '[\?if=status=Stopped .. ][[{artist}][\?soft  - ][{title}|{player}]]]')*
     format_player_separator: show separator if more than one player (default ' ')
+    player_hide_non_canplay: For hiding players which doesn't close mpris interface after media closing
+        (default ['chromium', 'chrome'])
     players: list of players to track. An empty list tracks all players (default [])
     seek_delta: time (in seconds) to change the playback's position by (default 5)
     thresholds: specify color thresholds to use for different placeholders
@@ -98,6 +100,7 @@ class Py3status:
         r"[\?if=status=Stopped .. ][[{artist}][\?soft  - ][{title}|{player}]]]"
     )
     format_player_separator = " "
+    player_hide_non_canplay = ["chromium", "chrome"]
     players = []
     seek_delta = 5
     thresholds = {"status": [("Playing", "good"), ("Paused", "degraded"), ("Stopped", "bad")]}
@@ -273,6 +276,12 @@ class Py3status:
         players = []
         cached_until = self.py3.CACHE_FOREVER
         for player in tracked_players:
+            if (
+                not player.props.can_play
+                and player.props.player_name in self.player_hide_non_canplay
+            ):
+                continue
+
             player_data = self._get_player_data(player)
 
             # Check if the player should cause the module to continuously update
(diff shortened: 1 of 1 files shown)


py3status/modules/playerctl.py
```diff
@@ -273,6 +276,12 @@ def playerctl(self):
         players = []
         cached_until = self.py3.CACHE_FOREVER
         for player in tracked_players:
+            if (
+                not player.props.can_play
```
lasers, line 280: Without digging in,  Is `not player.props.can_play` alone enough without `player_hide_non_canplay` config.... This seems like an mpris issue on chrome/chromium side.

Do chrome/chromium have this flag on `true` when it's playing?

valdur55: Actually there are other chromium based browsers like brave which has same bug. Chromium issue tracker has already bug in their system: https://issues.chromium.org/issues/40703847
Even the issue is from older raport states same issue: https://github.com/JasonLG1979/gnome-shell-extension-mpris-indicator-button/issues/10

Added some pictures about different "can" properties:

State when chromium is playing media
image

State when chromium media is paused:
image

State when chromium mpris is idle state:
image

VLC is opened but no media opened (just opened vlc, no media selected):
image

VLC is opened, media loaded and stopped playing media.
image

I don't wan't to blindly hide all players with can_play == false state because I prefer to see VLC media player on statusbar just after I opened it as empty player.

It littlebit seems like hide_non_can_play is too technical so even we can simplify it to level: hide media players stopped state and user provides list on media players which he want to hide.

I searched some chromium source code and found out where this all "can" properties false comes. I suspect that line is this: https://source.chromium.org/chromium/chromium/src/+/main:components/system_media_controls/linux/system_media_controls_linux.cc;l=288

lasers: I have been out of scene with py3status. Just sharing thoughts.

I think it doesn't benefit users to display empty players. If anything, this is maybe something that should be added in first place to ensure better quality control/behavior.

I do think `hide_non_can_play` config is too specific that partially addresses chromium-based players and not that general...  I would like an approach that that users can allow/deny (all) `can_xxx_xxx` configs (either by a dict config or new placeholders). New placeholders that exposes boolean like this... is maybe hard for users to come up with a new `format_player` that excludes those players... Dict may be be easier.

>I prefer to see VLC media player on statusbar just after I opened it as empty player.

It should be easier to just ignore players that doesn't have `can_play` flag true because of things like... You saying you don't like this empty-player thing on chrome/chromium #2255 but are okay with this on vlc... I think it would make more sense to avoid all empty-players in first place... to keep it simple.

This approach also may omit the need for adding `|{player}` at end of `format_player` config too.

Try it for a while. See if this small but simple tweak works quite well.

lasers:
Is it fixable with `format_player`?

```python
# Hide on empty title, no player.
format_player = "[\?if=title [\?color=status [\?if=status=Playing > ][\?if=status=Paused \|\| ]"
format_player += "[\?if=status=Stopped .. ][[{artist}][\?soft  - ]{title}]]]"
```

ultrabug:
@valdur55 do you want to follow up here before I do a release?

valdur55:
Oh no, I totaly forgot this thread.

I tested different desktop envs capabilities to handle this issue.
KDE only showed playing media players
Xfce showing m3edia players as is
Mate -no mpris status in sound menu
Gnome (https://extensions.gnome.org/extension/4928/mpris-label/) showed playing/paused media players.

Looks like winner is solution to hide paused media players, but there is one catch.
If module supports to stop media player then it is unexpected to hide stopped media player.
So looks like most pratical solution is to hide all non can_play media players.


@lasers your format_player magic didn't work for me. Player with media didn't show up.

lasers (approved):

ultrabug:
Thanks, will make a release soon

valdur55:
Hey.
Actually looks like I should drop also same parameter on mpris side, so it will hide all non can_play media players by default.
I will make pr for that ASAP when I am at home.

PR #2257, opened 2024-08-27
mpris module: improve documentation and set chrome and chromium as default player_hide_non_canplay value.

valdur55:
To improve first user UX i decided to put chrome/chromium as default players to hide on idle status. (#2255)
Removed (experimental) feature flags because they are not so experimental anymore.
Corrected requirments list.


valdur55:
Dropped default params part and. Point is to update documentation and change default behaviour in different time.

ultrabug:
Thanks!

PR #2258, opened 2024-09-02
playerctl: Add python-gobject dependency

jdholtz:
This is a dependency of the module and doesn't come with py3status by default. I used the same description as the usbguard module.


lasers (approved):

ultrabug:
Thanks a lot!

PR #2264, opened 2024-10-14
kdeconnector: replace pydbus with dbus-python

valdur55:
Tested it with 
majnaro linux python 3.12.6, kdeconnect-cli 24.08
It worked with _bat, _not, _con path

ubuntu 20.04  python3.8.10, kdeconnector-cli 1.4.0
It worked with old path.

Commit initiative from #2192


ultrabug:
Thanks a lot!

PR #2265, opened 2024-10-16
wwan: replace pydbus with dbus-python

valdur55:
Replacing pydbus with dbus-python required specific oldschool hardware usage.
Used Huawei E303 wwan adapter whcih I rescued years ago from second hand shop :).

Commit initiative from #2192


ultrabug:
Thanks a lot!

Issue #2260, opened 2024-09-20
[Feature request] Using IPinfo.io for IP information (IP Geolocation + ASN)


abdullahdevrel:
**Is your feature request related to a problem? Please describe.**
The current IP geolocation API service does not provide SSL/HTTPS encryption, has rate limits for high-frequency requests, and is based on ISP/ASN-reported location information.

**Your py3status version**

```
py3status version 3.59 (python 3.8.10) on sway
```

**Describe the solution you'd like**
I would like to recommend ipinfo.io for IP data. 

- Provides highly accurate IP geolocation backed by our probe network infrastructure. ([Blog post](https://ipinfo.io/blog/probe-network-how-we-make-sure-our-data-is-accurate/) and [accuracy posts](https://community.ipinfo.io/tag/accuracy))
- Free tier access token supports 50,000 requests per month.
- IPinfo supports HTTPS/SSL encryption on all endpoints.
- Tokenless access grants 1,000 requests per day.
- No rate limit for repeated API requests. You can make all the allocated requests in one go. 
- Includes ASN and hostname information. Supports bulk enrichment through a single API endpoint. We also have a comprehensive integration library.

Usage example:

```
curl https://ipinfo.io/8.8.8.8
```

```
{
    "ip": "8.8.8.8",
    "hostname": "dns.google",
    "anycast": true,
    "city": "Mountain View",
    "region": "California",
    "country": "US",
    "loc": "37.4056,-122.0775",
    "org": "AS15169 Google LLC",
    "postal": "94043",
    "timezone": "America/Los_Angeles"
}
```

Documentation: https://ipinfo.io/developers

**Describe alternatives you've considered**

I have not considered any alternatives.

---

Consider including IPinfo in the project for accurate data and generous free access. Please let me know what you think. Thank you very much.

ultrabug:
Thanks for letting us know @abdullahdevrel . If any user wants this in, they're (or you are) welcome to contribute that ofc.

abdullahdevrel:
@ultrabug Thank you for considering the request. Do you want to replace the existing providers? I think our service is superior to both of them: it offers more accurate results, is reliable, has better limits, and is HTTPS/SSL secure.

- https://github.com/ultrabug/py3status/blob/2422a43bff20cb229513bd494be125a32ce226bf/py3status/modules/getjson.py#L32
[...]

PR #2277, opened 2024-12-24
whatismyip: add an example with more url_geo choices

lasers:
Adding an example with url_geo choices... This ought to satisfy all parties.

Closes https://github.com/ultrabug/py3status/issues/2259
Closes https://github.com/ultrabug/py3status/issues/2260


ultrabug:
Thanks

PR #2279, opened 2024-12-26
Fix TypeError on xsel module

esuarezsantana:
If `log_file` is `None`, then actual source will raise a `TypeError`:

``` 
TypeError: expected str, bytes or os.PathLike object, not NoneType
```


lasers (approved):
Yes. Thank you. Introduced by https://github.com/ultrabug/py3status/pull/1978.

ultrabug:
Thanks for your contribution @esuarezsantana !

PR #2282, opened 2025-01-17
external_script module: allow scripts to specify urgent

devkev:
I had a need for an external script to specify not just the output color, but also whether or not the output should be "urgent".

### DIFF
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -5,6 +5,8 @@ Display output of any executable script set by `script_path`. Only the first
 two lines of output will be used. The first line is used as the displayed
 text. If the output has two or more lines, the second line is set as the text
 color (and should hence be a valid hex color code such as #FF0000 for red).
+If the second line is `urgent`, or has `!` prefixing the hex color, then the
+"urgent" flag will be set.
 The script should not have any parameters, but it could work.
 
 Configuration parameters:
@@ -76,8 +78,14 @@ class Py3status:
             output_lines = self.output.splitlines()
             if len(output_lines) > 1:
                 output_color = output_lines[1]
-                if re.search(r"^#[0-9a-fA-F]{6}$", output_color):
-                    response["color"] = output_color
+                if output_color == "urgent":
+                    response["urgent"] = True
+                elif re.search(r"^!?#[0-9a-fA-F]{6}$", output_color):
+                    if output_color[0] == "!":
+                        response["urgent"] = True
+                        response["color"] = output_color[1:]
+                    else:
+                        response["color"] = output_color
         except self.py3.CommandError as e:
             # something went wrong show error to user
             output = e.output or e.error


py3status/modules/external_script.py
```diff
@@ -5,6 +5,8 @@
 two lines of output will be used. The first line is used as the displayed
 text. If the output has two or more lines, the second line is set as the text
 color (and should hence be a valid hex color code such as #FF0000 for red).
+If the second line is `urgent`, or has `!` prefixing the hex color, then the
+"urgent" flag will be set.
```
lasers, line 9: It has been a long while since I used i3bar. Bear with me. Why just check for `urgent` on second line like `#ff0000,urgent` or `deepskyblue,urgent` or just `urgent`? The last one may just give it default due to invalid color name (untested).

As of right now, users can't have urgent with color names.

devkev: I also haven't looked at any of this in a long time.

I don't know if it's meaningful to set the "urgent" property without also setting "color" - perhaps i3bar uses some default, or perhaps it ignores the "urgent" property in that case.  I wouldn't mind if the "second line being 'urgent'" part was removed (I'm not using it in any of my configs).

Regarding non-hex colors (ie. color names), my reading of the original code is that only hex colors are supported on the second line anyway - if the line doesn't match the regex, then it gets ignored.

lasers: > only hex colors are supported on the second line anyway...

You're right. I kinda forget that we can specify `color_` / read color names in config. This doesn't not apply to module codes / `output` placeholder.

Like you said, checking for `urgent` alone on second line could be more than enough. Can you test it?

devkev: When I was writing this, setting "urgent" alone didn't result in any change or styling being applied by i3bar.  But, like I said, maybe that's because I didn't set some default, I'm not sure.  Whereas using `!#ff0000` to set both "urgent" and "color" did work as expected in i3bar, and that's what I'm using in my configs now.

So if anything, I would suggest keeping the known-good `!#rrggbb` format, and getting rid of the questionable `urgent` format.  I think the `!#rrggbb` format is more flexible anyway - even if it turns out that there is some way of getting `urgent` to work, the same result must still be achievable via `!#rrggbb` anyway.

lasers: I have no dog in this game, but I ponder if the ideal way would be to check for `urgent` on seperate line instead. That's all.

Supposedly if a user have a need to run a different `cache_timeout` or something else next based on the script output, then that could go on fourth line, et cetera. Anyhow, if you think we should keep it as-is, then I guess we could do that as-is too.

The new code would mean checking second line for color (assigning if not empty)... and seems more simplified than checking for either `urgent` or `!`.

Although `!#ff0000` looks more simplified (from script output perspective), I think this would be more clean. I was thinking something simple and straight to keep urgent and color separate from each others... like...
```python
>>> lines = ["first_line", "#ff0000", "urgent"]
>>> any([x for x in lines[1:] if x == "urgent"])
True

>>> lines = ["first_line", "urgent"]
>>> any([x for x in lines[1:] if x == "urgent"])
True
```
This code snippet could be for color too, rather than trying to parse settings from a simplified `[!]#hex` without separators. Idk.

devkev: Ohhh, now I understand what you're driving at.  I have no problem with this.

Taking that idea to its logical extreme would be some kind of structured format like JSON, but that would complicate things a lot (for both the module and the external scripts), so I definitely want to stay away from that.

My only concern is that if more and more extra lines of context/metadata keep get added, that's fine, until someone wants to set only the `cache_timeout` but not the color or urgent.  Blank lines could work, but things could get tricky if that actually has semantic meaning for some future option (as opposed to meaning "missing/absent").

Instead, I suggest keeping all this info on the second line still, but as whitespace separated tokens.  That is, something like:
```
if len(output_lines) > 1:
    words = output_lines[1].split()
    for word in words:
        if re.search(r"^#[0-9a-fA-F]{6}$", word):
            response["color"] = word
        elif word == "urgent":
            response["urgent"] = True
```
How do you feel about this?  It would allow any of
* `#rrggbb`
* `#rrggbb urgent`
* `urgent #rrggbb`
* `urgent`

which is still easy to read/understand, easy to output from the script, simple and easy to parse in the module, and nicely extensible in case we later want to add an `elif word.startswith("cache_timeout=")` (or whatever).

lasers: Yeah, getting this module to load/merge JSON output would be the absolutely best thing. Your solution looks fine. We can roll with that.

Just wanting you to see what I was trying to potentially avoid. `#rrggbb urgent` is more clear than `!#rrggbb` too. Let's do this. Fine by me.

ultrabug:
Thanks to you both for your contribution and iteration on ideas!

devkev:
Uhhh, I was just about to upload the revised code... I'll open a new PR for it.

ultrabug:
@devkev then I read lasers conclusion too fast I'm sorry, thx

PR #2283, opened 2025-01-21
external_script module: adjust how urgent is specified

devkev:


devkev:
The code is as discussed and agreed on #2282.  The updated documentation hasn't been reviewed.  I tested the change and it seems to work fine.

ultrabug:
sounds clear enough, thanks

PR #2284, opened 2025-01-21
mpris: use new helper: get_replacements_list introduced with (#2242)

valdur55:


py3status/modules/mpris.py
```diff
@@ -280,6 +281,10 @@ def metadata(self, metadata=None):
 
             self._metadata["nowplaying"] = metadata.get("vlc:nowplaying", None)
 
+            for x in self.parent.replacements_init:
+                if x in self._metadata and self._metadata[x]:
```
lasers, line 285: For what it's worth, `replace()` should skip on empty values too.

https://github.com/ultrabug/py3status/blob/c2d623ae43c05c7528410700e67eb6b79a6dcaaf/py3status/py3.py#L1271-L1272

ultrabug:
Thx!

PR #2287, opened 2025-05-28
playerctl: Ensure duration and position are always integers

jdholtz:
Occasionally, a player could return its position/duration in the song as a float (e.g. SMPlayer), but the module expects it to be an integer (causing an error while formatting the time). This commit ensures that the position/duration will always be an integer before formatting the time, as both arguments to `divmod` will always be integers (thus, returning an integer).


ultrabug:
Thank you!

PR #2288, opened 2025-05-31
moc: fix type mismatch in moc_data

joanbm:
moc.py was recently refactored by f648cb607f1c ("py3: add new helper: get_replacements_list"). This causes it to fail when MOC is started, because the code has two variables:
* `moc_data`, which is a `str`.
* `data`, which is a `dict`. But the refactor accidentally handles `moc_data` as a `dict`.

This commit fixes the issue, and unifies `data` and `moc_data`, using the same pattern that is already being used in cmus.py.

### DIFF
--- a/py3status/modules/moc.py
+++ b/py3status/modules/moc.py
@@ -110,22 +110,25 @@ class Py3status:
             is_started = False
         return is_started, data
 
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks a lot!

PR #2289, opened 2025-06-26
lm_sensors: update broken json parser

lasers:
This changes how we parse broken json blocks from upstream. Instead of splitting chunks based on `}\n{` , this just run every line until we get a valid json chunk before we add it to a dictionary.

Ought to address https://github.com/ultrabug/py3status/pull/2226#issuecomment-3002191080.  /cc @MacGyverNL for testing.


MacGyverNL:
Can confirm this works, although with this splitting, my case never hits the codepath with JSONDecodeErrors. So I can't testify to the effectiveness / robustness of building the chunk across multiple lines.

lasers:
Noted! It's a better solution. Please try it again.

MacGyverNL:
Yeah this should work, once you get the `^` and `$` right. I think what's confusing you is where to match start- and end-of-line, especially in multiline mode. These aren't indicators of where the pattern starts, they're match points you'd normally use to match only at the start and end of a line. The way I read the documentation, in multiline mode they still match right before and after each newline. So in principle, because we're explicitly matching the newline, you don't actually need them at all. `self.pattern = re.compile(r'(?<=})\n(?={)', re.MULTILINE)` suffices (and works, already tested it).

lasers:
Thank you for your help / looking into this. We should be good here now.

lasers:
@ultrabug \o/

Issue #2290, opened 2025-07-08
fedora_update: output of dnf command changed (fedora 41 / dnf 5.x) - security updates no longer noticed


girst:
**Describe the bug**
the `fedora_updates` module with `check_security = True` no longer turns red when security updates are available.
py3status parses `dnf updateinfo` to get at that information, but this command both changed syntax and ouput. it now is called `dnf updateinfo summary`, and its output looks like this:

```
Available advisory information summary:
Security    : 16
  Critical  : 2
  Important : 5
  Moderate  : 7
  Low       : 1
  Other     : 1
Bugfix      : 20
Enhancement : 14
Other       : 8
```

**Your py3status version**
py3status version 3.59 (python 3.13.3) on i3
(i verified the fedora module is unchanged in master)

**To Reproduce**
0. use fedora 41 or higher
1. Add module fedora_updates using the default module config
2. wait for security updates to become available

**Expected behavior**
the output of the module should turn red, but it stays yellow.

PR #2291, opened 2025-07-18
dnf_updates: new module from scratch

lasers:
Just updating this module from scratch. I dunno. Untested.
* Omit `color` in the response in favor of more colors via `self.py3.safe_format`
* Thresholds added
* Additional placeholders for extra customizations such as custom names/icons
  * Default `format` should mimic `check_security = True` behavior
  * Disable `check_security = True` behavior by using `format = "DNF [\?color=available {available}]`
    * I hope this is right. I don't know.
* For legacy theme, use...
  *  `format = "[\?if=security&color=bad DNF: {available}|\?color=available DNF: {available}]"`

Closes https://github.com/ultrabug/py3status/issues/2290. /cc @girst for testing.


girst:
additional note: --json was only introduced with dnf5 (at least it isn't 
present in my fc38 vm). keeping the old module might be a good idea for 
rhel/rocky/etc, who also use (old) dnf in versions 8 and 9. that said, 
who knows how few people use these OSes on their desktops, and i3/sway 
on top of that.

lasers:
Made a tweak to give default `0`  values... ie Users see `0` instead of `{security}` when no available security.

>not a fan of the non-colored 'DNF' text and preferred the
previous everything-the-same-color. the latter better matches the other
(both py3status and upstream/i3status) modules.

I heard you. I hadn't used py3status for while now. When I did, I was exposing modules with new color defaults because py3status used to not be able to display them.... and to keep modules in unity with each others. Also, some modules works better with sensible theme/colors such as `lm_sensors`, `nvidia_smi`, etc.

I just made the consistency tweaks. Modules like `sysdata` can benefit from colors too.

lasers:
(Testers) Do we need to change `self.py3.command_output` with something else?

girst:
On Tue, Aug 26, 2025 at 03:47:08AM -0700, Ultrabug wrote:
>ultrabug left a comment (ultrabug/py3status#2291)
>@girst need you to validate this module, I dont have a fedora box, 
>cheers

the module works, but i'll reiterate my gripe about the colorization.

https://imgur.com/Qii8S9A.png

modules: 3x imap, old fedora_updates, new dnf_updates, wifi, ethernet.

note how all of them color the whole module, but dnf_updates only colors 
the number. this is IMO worse, as especialy with single-digit updates it 
is easy to miss that something is available.

girst:
argh! it seems, github swallows some of my emails sent to this discussion.

one from earlier that did not make it:

```
Date: Sat, 19 Jul 2025 16:06:18 +0200
Subject: Re: [ultrabug/py3status] dnf_updates: new module from scratch (PR #2291)

you might have misunderstood me. my comment was about how the text 'DNF' does
not get colored in the same color that the number of outstanding updates is,
not about the global color pallette.

the few colored modules i use (ethernet, wifi, battery, imap) all use the same
color for the whole text of the module, not just color part of it. this has
the (positive, imo) effect that more pixels get colored, making it easier to
spot at a glance.

with <white>DNF</white> <yellow>1</yellow>, only 1 of 5 chars is in the
degraded color. what i would've preferred is <yellow>DNF: 1</yellow>.
```

just now:

```
Date: Wed, 3 Sep 2025 12:02:22 +0200
Subject: Re: [ultrabug/py3status] dnf_updates: new module from scratch (PR #2291)

module breaks every time after resuming from standby:

2025-09-03 11:59:49 WARNING Instance `dnf_updates`, user method `dnf_updates` failed (CommandError) py3.py line 1021.
2025-09-03 11:59:49 INFO Traceback
CommandError: Command `dnf updateinfo list --json` returned non-zero exit status 1
  File "/home/t/projects/py3status/py3status/module.py", line 952, in run
    response = method()
  File "/home/t/projects/py3status/py3status/modules/dnf_updates.py", line 70, in dnf_updates
    dnf_data = self._get_dnf_data()
  File "/home/t/projects/py3status/py3status/modules/dnf_updates.py", line 61, in _get_dnf_data
    updates = loads(self.py3.command_output("dnf updateinfo list --json"))
                    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/t/projects/py3status/py3status/py3.py", line 1021, in command_output
    raise exceptions.CommandError(
    )
```

lasers:
This is what you want.
```python
# i3status theme
format = "[\?if=security&color=bad DNF: {available}"
format += "|\?color=available DNF: {available}]"

# i3status theme - hide module when no available updates
format = "[\?if=available [\?if=security&color=bad DNF: {available}"
format += "|\?color=available DNF: {available}]]"
```
This is what I was trying to support in py3status formats.
```python
general {
    # These will be used if not supplied by a module
    color = '#FFFFFF'
    color_good = '#00FF00'
    color_bad = '#FF0000'
    color_degraded = '#FFFF00'
}
```
Imho,  py3status formats should include headers.
Imho,  py3status formats should include headers that also can be colorized by `color = '#FFFFFF'`.

Also, i3status can't colorize. This is customization issue. Py3status could benefit from having theme support... such as users specifying `theme = "i3status"` in py3status config section that would update default `format*` configs. This mean we would have to take on few theme files, blah. I would suggest to keep at taking theme files minimal too.

- `3x imap`: it looks like you configured those formats to mimic i3status.
- old `fedora_updates`: default format with forced colors.
- new `dnf_updates`: it looks like you would configure this format to mimic i3status.
- `wifi`, `ethernet`: it looks like you configured those formats (with trimmed headers) -- Maybe i3status modules.
-- `py3status/wifi.py` have legacy code to force colors just in case if `\?color=xxx` is not in the format.

I can add i3status theme (what you want above) in `Examples` to make this easy copy/paste. I feel like this would be the easiest way to address your i3status concern.

PR #2296, opened 2025-11-07
Implemented onecall 401 failure fallback

Jacotsu:
If a user has an OWM api token but no associated subscription with it, the onecall API will fail with a 401 error and no weather will be shown.

This commit will ignore the onecall api failure and will show just the current weather with the city name


ultrabug:
Thanks @Jacotsu please make sure your PR checks the CI

Issue #2297, opened 2025-11-16
py3status not showing output of conky modules


johanbluecreek:
**Describe the bug**

When starting py3status with an old config that used to work, it now just displays module name instead of output, e.g.
```
$ py3status -d -l debug.log
py3status: trying to auto-detect output_format setting
py3status: auto-detected "term"
eps py3status EntryPoints(())
2025-11-16 20:29:15
 load   | 2025-11-16 20:29:15
conky loadavg | 2025-11-16 20:29:15
conky loadavg | 2025-11-16 20:29:16
conky loadavg | 2025-11-16 20:29:17
conky loadavg | 2025-11-16 20:29:18
conky loadavg | 2025-11-16 20:29:19
conky loadavg | 2025-11-16 20:29:20
conky loadavg | 2025-11-16 20:29:21
```

**Your py3status version**

```
$ py3status -v
py3status version 3.62 (python 3.13.5) on sway
```

I run Gentoo, which has version "[3.61-r1](https://gitweb.gentoo.org/repo/gentoo.git/tree/x11-misc/py3status/py3status-3.61-r1.ebuild)", that is pyt3status version 3.61 Gentoo revision 1. Since this version did not work, I also applied a patch from tag `3.61` to `3.62`, which did not help, so I also applied a patch with all commits from `3.62` to latest commit at the time (601302514527e4ec88a07d59381686522d815f6b), as well. This did not work either.

**To Reproduce**
Steps to reproduce the behavior along with your py3status configuration (filter out any private values):
1. Add module `conky loadavg` (although any conky module I tried gives the same issue (tried `info`, `cpu`, `memory`, `filesystem`))
2. Configure module like this
```
general {
  colors = true
  interval = 5
}

# display load averages
order += "conky loadavg"
conky loadavg {
  format = ' load '
  format += '[\?color=lightgreen {loadavg 1} ]'
  format += '[\?color=lightgreen {loadavg 2} ]'
  format += '[\?color=lightgreen {loadavg 3}] '
}

order += "tztime local"
tztime local {
  color = '#E0CA3C'
  format = "%Y-%m-%d %H:%M:%S"
}
```
3. Run py3status
```
$ py3status -d -l debug.log
```
4. See ~~error~~ faulty output
```
$ py3status -d -l debug.log
py3status: trying to auto-detect output_format setting
py3status: auto-detected "term"
eps py3status EntryPoints(())
2025-11-16 20:29:15
 load   | 2025-11-16 20:29:15
[...]

PR #2298, opened 2025-11-16
conky: update ignored_conky_outputs

lasers:
Ignore `x11 session running`... and `wayland session running`.

Closes https://github.com/ultrabug/py3status/issues/2297.


ultrabug:
thanks

PR #2305, opened 2025-12-09
module battery_level: apply color_good on good discharge

aguilbau:
No color is currently set when battery is discharging and good threshold is met. This patch fixes it, applying `color_good` in this case.


aguilbau:
I think this PR can be closed.
I have non default `color`, and I use it for charging / decharging with a good threshold / fully charged. I could not get it to work without this patch, because I didn't know I could configure default color in a `general` block ; after doing so, I have the colors I wanted. Thank you !

PR #2312, opened 2026-01-02
add new module: i3status

lasers:
Just toying and OK, now we have a working module.

This addresses one of the [RFC] version 4.0 goals in https://github.com/ultrabug/py3status/issues/1584.

> **make i3status a module like any other**
>    * would be launched seamlessly if config use a i3status module
>    * but would be removed from core

Config slightly different, but probably acceptable. Core still intact. This does not have to be merged.

### DIFF
--- a/docs/dev-guide/the-py3-helper.md
+++ b/docs/dev-guide/the-py3-helper.md
@@ -268,6 +268,10 @@ This is useful because we just get simple placeholder without any
 formatting that may be applied to them
 eg ``'{placeholder:.2f}'`` will give ``['{placeholder}']``
 
+### get_config(name, default=None)
+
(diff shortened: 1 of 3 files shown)


lasers:
@ultrabug Ready for review.

lasers:
1) Enhanced -- allowing users to change per-module format `prefix {output} suffix`. 
2) Py3status would lose its `...an extensible i3status wrapper...` feature/catchphrase.
3) Converting user config from `core.i3status` to `modules.i3status` on the fly... needs to be implemented?

lasers:
TODO: See if we can run on_clicks + universal module options via py3 helpers.

PR #2317, opened 2026-02-08
add new module: librelinkup

lasers:
Hi, I made a new module `libre`.

libre

### DIFF
--- a/py3status/modules/libre.py
+++ b/py3status/modules/libre.py
@@ -1,0 +1,208 @@
+r"""
+Display glucose readings from LibreLinkUp-connected FreeStyle Libre sensor users.
+
+LibreLinkUp provide glucose readings up to every five minutes. Designed
+to help diabetes patients keep track of their blood glucose levels with ease.
+
+Configuration parameters:
+    cache_timeout: refresh interval for this module (default 300)
+    format: display format for this module (default '{format_patient}')
+    format_datetime: specify strftime characters to format (default {"timestamp": "%-I:%M %p"})
+    format_patient: display format for Libre patients
+        *(default '{first_name} [\?color=mg_dl {mg_dl} mg/dL {trend_arrow}]'
+        ' [\?color=darkgrey {timestamp}]')*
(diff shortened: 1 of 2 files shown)


lasers:
Changed threshold values (+orange). Tested OK.

lasers:
Ready to merge.

ultrabug:
Why not name it more explicitly `librelinkup` please?

lasers:
You're absolutely right! I will fix this.
[WARNING] Initially made and tested, but currently not testable, but code probably LGTM anyway.

ultrabug:
@lasers will you use it? because I try to refrain to get modules in tree with no users

lasers:
I'm not using it, but I can tell you this module is much better than `dexcom`.

Because from what I gather, Dexcom doesn't have sharing in place like LibreLinkUp (for family, friends, caretakers, et cetera). Dexcom requires you to sign in as yourself (your user, pass) anywhere, not just this module... which is kinda bad when it come to wanting to share glucose number with others.

Also, ChatGPT.
>According to the latest estimates from the International Diabetes Federation:
>
>About 11.1% of adults worldwide (ages 20–79) have diabetes.
>That is approximately 589 million people, or about 1 in every 9 adults.
>
>Additional facts:
>
>More than 90% of diabetes cases are type 2 diabetes.
>By 2050, the number of adults with diabetes is projected to increase to 853 million (about 1 in 8 adults).
>
>So, the current worldwide prevalence of diabetes is approximately 11.1% of the adult population.

I think this module will get users. I just make many modules like `conky`, `lm_sensors`, `nvidia_smi`, `mail` and the modules will reach users eventually, then the users will get to try them eventually too, then users will decide if they're useful or not.

I don't think/know if users are using `thunderbird_todos`, `transmission`, `hueshift`, `hddtemp`, `timewarrior`, and `wanda_the_fish` (i still like this silly module anyway).

I understand not wanting to merge junk modules. I just don't think this is a junk module.

ultrabug:
Thanks, I'll trust your judgment and hope indeed that this module will help someone like many others do!

PR #2318, opened 2026-03-02
parse_config: add timeout to file --mime-encoding subprocess call

nadzyah:
`parse_config.py` calls `file --mime-encoding` with no timeout. On some architectures (like s390x), this can hang when the magic database is slow to load, blocking py3status startup indefinitely.

The fix adds `timeout=3` to `check_output()` and handles `TimeoutExpired` the same way as `CalledProcessError` (fall back to utf-8).

I've initially discovered this issue while trying to fix failing tests for the `py3status` status package in Ubuntu and Debian

### DIFF
--- a/py3status/parse_config.py
+++ b/py3status/parse_config.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 from importlib import util
 from pathlib import Path
 from string import Template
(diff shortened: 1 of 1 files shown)


ultrabug:
Sorry for my latency @nadzyah 

That's a great finding, thank you very much!

PR #2320, opened 2026-03-26
external_script: always pass a string to safe_format

### LINKED ISSUE #2319: external_script can pass non-string to safe_format
**Describe the bug**
In external_script, the script's output is [passed as the format to safe_format](https://github.com/ultrabug/py3status/blob/c2d623ae43c05c7528410700e67eb6b79a6dcaaf/py3status/modules/external_script.py#L116) to get the "composite" field.  However, the [output might have been converted to an int or float](https://github.com/ultrabug/py3status/blob/c2d623ae43c05c7528410700e67eb6b79a6dcaaf/py3status/modules/external_script.py#L102-L109).  This leads to errors
```
Invalid format `123` (expected string or bytes-like object) (TypeError) external_script.py line 116.
```
polluting stderr.

**Your py3status version**
Problem appears to still be present in master.

**To Reproduce**
I expect it should show up for a module such as
```
external_script "foo" {
    script_path = "/bin/echo 123"
}
```

**Expected behavior**
The error message shouldn't occur just because the script happened to output something that looks numeric.

**Additional context**
I don't understand what "composite" is or what it's for, but I've worked around the problem by passing `f"{output}"` to `safe_format`, instead of `output`, since this will always be a string regardless of whether `output` is a string or a number.  But there might be a more appropriate way of handling it.

devkev:
Fixes #2319

### DIFF
--- a/py3status/modules/external_script.py
+++ b/py3status/modules/external_script.py
@@ -113,7 +113,7 @@ class Py3status:
         script_data = {
             "output": output,
             "lines": len(output_lines),
-            "composite": self.py3.safe_format(output),
+            "composite": self.py3.safe_format(str(output)),
         }
         response["full_text"] = self.py3.safe_format(self.format, script_data)
         return response


devkev:
CI failure seems unrelated?

ultrabug:
thanks @devkev please rebase for CI fix

PR #2322, opened 2026-04-11
vpn_status: Better VPN detection - prevent libvirt false-positive

### LINKED ISSUE #2321: vpn_status crashes when a VM network interface exists
## Bug description

`vpn_status` crashes when a `libvirt` virtual network interface exists.
Instead of the VPN status, the literal `vpn_status` string is printed on the bar.
It seems the cause is that the network interface created by `libvirt` is of type `"tun"`.

### Version

`py3status` installed from Debian 13 (trixie) repository.

```sh
$ py3status --version
py3status version 3.61 (python 3.13.5) on sway
```

## Log report

```sh
2026-04-08 16:47:20 WARNING Instance `vpn_status`, user method `vpn_status` failed.
2026-04-08 16:47:20 INFO Traceback
KeyError: 'ipv4'
  File "/usr/lib/python3/dist-packages/py3status/module.py", line 943, in run
    response = method()
  File "/usr/lib/python3/dist-packages/py3status/modules/vpn_status.py", line 175, in vpn_status
    vpn_info = self._get_vpn_status()
  File "/usr/lib/python3/dist-packages/py3status/modules/vpn_status.py", line 129, in _get_vpn_status
    ipv4, ipv6 = self._get_ips(bus, properties["Connection"])
                 ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/py3status/modules/vpn_status.py", line 142, in _get_ips
    ipv4 = self._get_ip(settings["ipv4"])
                        ~~~~~~~~^^^^^^^^
```

## Steps to reproduce

1. Have libvirt/QEMU installed with a virtual network (`default` NAT network creates the bridge `virbr0`)
2. Start a VM, which creates the `vnet tun` interface
3. Add `vpn_status` module with default config
4. Module fails, showing literal "vpn_status" text in the bar

## Expected behavior

Module should ignore non-VPN `tun` interfaces and show "VPN: no" when no VPN is active.

## Additional context

The active interface:

```sh
$ nmcli --fields NAME,TYPE,DEVICE connection show --active | grep "tun"
vnet1               tun       vnet1
```

### Preliminary fix

Removing the `tun` check from https://github.com/ultrabug/py3status/blob/fe39761/py3status/modules/vpn_status.py#L128 seems to work, although I don't know if there are valid VPNs which would have `properties.get("Vpn")` as `False` and be of type `tun`.

```diff
-                 if properties.get("Vpn") or properties.get("Type") in ("wireguard", "tun"):
+                 if properties.get("Vpn") or properties.get("Type") == "wireguard":
```

Also, from the log traceback, it seems the `settings["ipv4"]` field is assumed to always exist, but that may not be the case.

jdholtz:
Fix #2321. A `tun` interface on Linux is more broad than just a VPN interface. It can also refer to interfaces used in a virtualized network (as shown in the issue). This PR uses a stronger VPN detection mechanism by making sure the interface is not marked as a port/slave in NetworkManager (indicating it is part of a broader virtualized network).

We can't rely on the `Vpn` property unfortunately because during testing, I found that an OpenVPN client does not have the VPN property marked as true in NetworkManager.

Also mentioned in the issue, I added better parsing of the IPs for an interface so it doesn't fail if either IP key doesn't exist (instead it would just return None for the respective IP).

### DIFF
--- a/py3status/modules/vpn_status.py
+++ b/py3status/modules/vpn_status.py
@@ -31,6 +31,7 @@ Color options:
 Requires:
     dbus-python: to interact with dbus
     pygobject: which in turn requires libcairo2-dev, libgirepository1.0-dev
+    NetworkManager: provides the org.freedesktop.NetworkManager DBus service
 
 @author Nathan Smith <nathan AT praisetopia.org>
 
@@ -125,8 +126,9 @@ class Py3status:
             interface = dbus.Interface(manager, "org.freedesktop.DBus.Properties")
             try:
                 properties = interface.GetAll("org.freedesktop.NetworkManager.Connection.Active")
-                if properties.get("Vpn") or properties.get("Type") in ("wireguard", "tun"):
-                    ipv4, ipv6 = self._get_ips(bus, properties["Connection"])
+                conn_settings = self._get_connection_settings(bus, properties["Connection"])
+                if self._is_vpn(properties, conn_settings):
+                    ipv4, ipv6 = self._get_ips(conn_settings)
                     vpns.append({"name": properties.get("Id"), "ipv4": ipv4, "ipv6": ipv6})
             except dbus.DBusException:
                 # the connection id has disappeared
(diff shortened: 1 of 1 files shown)


jdholtz:
CI failure might be because of an outdated Hatch version? I'm not exactly sure, I cannot reproduce it locally.

ultrabug:
Thanks a lot!

PR #2325, opened 2026-04-20
playerctl: apply button options selectively

### LINKED ISSUE #2324: playerctl module incorrectly handles defaults
**Describe the bug**
In the playerctl module, the code internally sets button 1 (@L89) as the default for play/pause and button 3 as the default for stop (@L94). However, when setting other actions to these buttons in the config, the defaults are not cleared, leading to the incorrect action being taken.

**Your py3status version**
`py3status version 3.63 (python 3.14.4) on sway`

**To Reproduce**
1. Add module 'playerctl'
2. Set option `button_next = 3`
3. Start playing a compatible source and right-click (i.e. button 3) on the source in your bar
4. The player stops instead of skipping to the next track, as the `stop` check comes earlier in the if-else-chain than `next` (@L331-334)

Can be worked around by setting `button_stop = x` in the config.

**Expected behavior**
`button_next = 3` should cause button 3 to skip the track, without needing to set `button_stop` manually.

lasers:
This clears out the default buttons.

I initially don't know whenever to leave `button_play_pause` alone... or to let users configure that themselves. If we leave `button_play_pause` at default, then an user can run into same problem with a different button. It is more clean not to assign any default buttons.

And this checks for duplicate buttons too.

/cc notifying @jdholtz author on buttons

Closes https://github.com/ultrabug/py3status/issues/2324

### DIFF
--- a/py3status/modules/playerctl.py
+++ b/py3status/modules/playerctl.py
@@ -56,6 +56,15 @@ Requires:
         bmp, xmms2, and others.
     python-gobject: Python Bindings for GLib/GObject/GIO/GTK+
 
+Examples:
+```
+# add buttons
+playerctl {
+    button_play_pause = 1
+    button_stop = 3
+}
+```
+
 @author jdholtz
 
 SAMPLE OUTPUT
@@ -76,6 +85,9 @@ import gi
 
 gi.require_version("Playerctl", "2.0")
 from gi.repository import GLib, Playerctl  # noqa e402
+
+STRING_DUPLICATE_BUTTONS = "duplicate buttons"
+DEFAULT_BUTTONS = {"button_play_pause": 1, "button_stop": 3}
 
 
 class Py3status:
@@ -116,6 +128,8 @@ class Py3status:
         }
 
     def post_config_hook(self):
+        # Initialize module settings
+        self._setup_buttons()
         self.thresholds_init = self.py3.get_color_names_list(self.format_player)
         self.replacements_init = self.py3.get_replacements_list(self.format_player)
         self.position = self.py3.format_contains(self.format_player, "position")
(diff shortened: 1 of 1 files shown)


ultrabug:
Thanks @lasers ; that PR will break users using the module with default settings for ages.

Could we make it to actually respect the user's setting instead of resetting everything to None like this? I'm confused.

lasers:
User requires `button_stop = None` in his/her config first in order to get a working `button_next = 3`.

User gave it a random number as a workaround/solution.

This can happen with any modules with default buttons. This one just has the largest button choices.

Do we suggest users to assign default buttons with something else first? 

My approach for this module with that many button choices was that there are no default buttons anymore and users just assign what they want to use.

Happy to hear your suggestions/solutions.

ultrabug:
I understand but I'm also wary of breaking something being used seamlessly by possibly a large number of users vs one having a specific case even if it's indeed not smooth for her/him.

In this case my favored approach would be :

- on the validation function step, check if user tried to use one of the default (1 or 3) button number by setting them on _another_ button
- if so, set the previously set button to None
- else, keep the default

Does it make sense mate?

lasers:
OK. I added a commit. I like the first one. It's more neutral in the long run. I think it also could lead to users having `button_play = 1` or `button_play_pause = 1` in their config and not needing `button_stop = 3`.

The second one contains your favored approach. Just more `post_config_hook` code. Everything seems to work fine on my side. Please review and if you like this, I'll squash em and update PR title+commit. :+1: 

P.S. This kind of button code validation could be applied to all modules with buttons in the eventual future, just not in modules themselves. It'll do for now.

ultrabug:
I think the second and current one is my favored yes

PR #2327, opened 2026-04-28
core: improve performance by calculating f-string only when it is used

valdur55:
Additional suggestion is to fail fast when calling disallowed logging level

I encountered high memory usage for long period of time (some days of usage).
Image

Normal memory usage:
image

I let claude code analyse standardized logging commit and found out some problems.
Main issue was related to f-strings which were calculated on every update and most likely never used (like debug level messages).

Claude code conclusion:
> The f-string is now evaluated on every update cycle — and inside log(), pformat() 
is unconditionally called on dicts/lists before any level check. In hot paths 
(every module update, every click event, every command-server connection), 
this creates continuous allocation pressure on the GC. Over long-running sessions 
this shows up as RSS growth that doesn't come back down (Python doesn't release arena
memory back to the OS easily — this is the "leak" you likely see on ps/htop).

### DIFF
--- a/py3status/command.py
+++ b/py3status/command.py
@@ -156,7 +156,8 @@ class CommandRunner:
                     if requested_name == name.split(" ")[0]:
                         found_modules.add(module_name)
 
-        self.py3_wrapper.log(f"found {found_modules}", level="debug")
+        if self.py3_wrapper.log_enabled("debug"):
+            self.py3_wrapper.log(f"found {found_modules}", level="debug")
         return found_modules
 
     def refresh(self, data):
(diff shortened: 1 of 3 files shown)


lasers:
It's certainly terrible to look at those added `if debug`.

Original logging PR called for using the logging module directly everywhere plus using lazy `%s` strings. One can look into adding ruff rule to prevent f-strings logging to reduce wasted f-string calculation.

I think the real issue is that we're logging everything through a busy `self.log` instead of breaking up into either using logging functions directly or create additional py3 logging (ie `self.py3.debug` pass if disabled, `self.py3.info`).  I looked at how we logged debug/warn code... we just print plain strings most times, so no need to check for arrays/lists/sets/etc and other functions. Those should be just for user-facing logging.

And to really improve performance, we should be able to disable logs completely too, no syslog, no info log, for most runs, which your proposal PR currently doesn't cover. This issue is acknowledged too.

Issue #2328, opened 2026-04-28
High memory usage


valdur55:
**Describe the bug**
I have encountered high memory usage for several weeks.
High memory usage:
<img width="1915" height="286" alt="Image" src="https://github.com/user-attachments/assets/e00ea43c-acee-4645-b423-923b9476167f" />

**Your py3status version**
Latest master

**To Reproduce**
Config: https://github.com/valdur55/dotfiles/blob/2d21700f1c43617b4276e55e2036054c93502b20/py3status/bar-top.conf
Use it in everyday life.

**Expected behavior**
Memory usage is kept like ~0.5%
<img width="1672" height="332" alt="image" src="https://github.com/user-attachments/assets/06d60b4e-26fe-4e63-a75d-847ed781470b" />

**Additional context**
First idea is to reduce f-string calculation in debug messages (#2327)

I also tested 3.63 from pip package but after some  time I noticed that I din't use i3ipc mode for window title.
Now I will just checkout 3.63 and look if memory goes up or not.

@ultrabug please do not release new version until it is solved.

valdur55:
Unfortunately pure 3.63 has memory issues.
Now I made monitoring script for it:
```
external_script memory {
    format = "MEM: {output}"
    script_path = "ps aux | grep py3status | grep top | awk -F ' ' '{print$4}'"
}
```

"grep top" is needed to differentiate between two bars.

valdur55:
sysdata finding doesn't fix memory leak

valdur55:
Found out root cause of it. I had pydevd and pydevd-pycharm installed at the same time. It is dependency of my custom debugging module. https://github.com/valdur55/dotfiles/blame/main/py3status/bar-debug-user-modules/pydev_debug.py
Seems like in 5 months ago I switched back to pydevd-pycharm module.

 It also kept module in connected state. (in reality pycharm debug server wasn't running at all)

It started to get weird when another computer with similar config didn't had memory issue.
Also when I clicked on it to forcefully reconnect debugger then memory eating stopped.


I took another mile to detect memory usage more deeply:
```
external_script memory {
    format = "MEM: {output}"
    script_path = "ps aux | grep py3status | grep top | awk -F ' ' '{print$4 \" \" $5 \" \" $6}'"
    cache_timeout = 1
}
```

PR #2329, opened 2026-04-30
sysdata: compile ZFS_SIZE_REGEX regex only once per module load

valdur55:
During analysis of memory issue (#2328) I asked claude code to analyze module.
It pointed out that this regex is built in every cycle (or even twice when swap and memory monitor is enabled).

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -111,6 +111,7 @@ from pathlib import Path
 
 INVALID_CPU_TEMP_UNIT = "invalid cpu_temp_unit"
 STRING_NOT_INSTALLED = "not installed"
+ZFS_SIZE_REGEX = re.compile(r"^size\s+\d+\s+(\d+)")
 
 
 class Py3status:
@@ -410,7 +411,6 @@ class Py3status:
 
     def _get_zfs_arc_size(self):
         """will raise OSError on failures"""
-        ZFS_SIZE_REGEX = re.compile(r"^size\s+\d+\s+(\d+)")
         try:
             with Path("/proc/spl/kstat/zfs/arcstats").open() as f:
                 for line in f.readlines():


lasers (approved):
1) Reality: Regex part is skipped on `swap`.
2) Sad reality: We try/except a guaranteed failure on every cycle on most setups with no zfs.

lasers:
> Sad reality: We try/except a guaranteed failure on every cycle on most setups with no zfs.

This just bypass try/except zfs size for setups with no zfs.

```diff
diff --git a/py3status/modules/sysdata.py b/py3status/modules/sysdata.py
index 573d134..2d72ee6 100644
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -111,6 +111,7 @@ from pathlib import Path
 
 INVALID_CPU_TEMP_UNIT = "invalid cpu_temp_unit"
 STRING_NOT_INSTALLED = "not installed"
+ZFS_ARCSTATS = "/proc/spl/kstat/zfs/arcstats"
 
 
 class Py3status:
@@ -271,6 +272,12 @@ class Py3status:
             },
         }
 
+        if self.init["mem"]:
+            if Path(ZFS_ARCSTATS).is_file():
+                self.zfs_size_regex = re.compile(r"^size\s+\d+\s+(\d+)")
+            else:
+                self._get_zfs_arc_size = self._return_zero
+
         if self.init["cpu_freq"]:
             name = sorted(self.init["cpu_freq"])[0]
             self.thresholds_init["legacy"]["cpu_freq"] = name
@@ -309,6 +316,10 @@ class Py3status:
             else:
                 self.init["cpu_temp"] = []
 
+    @staticmethod
+    def _return_zero():
+        return 0
+
     def _get_cpuinfo(self):
         with Path("/proc/cpuinfo").open() as f:
             return [float(line.split()[-1]) for line in f if "cpu MHz" in line]
@@ -410,11 +421,10 @@ class Py3status:
 
     def _get_zfs_arc_size(self):
         """will raise OSError on failures"""
-        ZFS_SIZE_REGEX = re.compile(r"^size\s+\d+\s+(\d+)")
         try:
-            with Path("/proc/spl/kstat/zfs/arcstats").open() as f:
+            with Path(ZFS_ARCSTATS).open() as f:
                 for line in f.readlines():
-                    m = ZFS_SIZE_REGEX.match(line)
+                    m = self.zfs_size_regex.match(line)
                     if m:
                         return int(m.group(1)) / 1024
         except (OSError, ValueError):

```

valdur55:
Yea, I noticed it but I didn't wan't to touch it at first. I was worried about to break it.
It would be better that I had shared finding... 
At first it seemed that this file can get readable in any time that's why it can be treated that way, but now it loos like proc path comes available during boot time and in machine run time readable status doesn't change.

PR #2331, opened 2026-06-05
feat: add headsetcontrol module

weberval:


weberval:
Please squash before merging!

weberval:
Labelled it as draft again, because I found a bug when a headset not connected anymore.

py3status/modules/headsetcontrol.py
```diff
@@ -0,0 +1,90 @@
+r"""
+Display headset connection state.
+
+This module displays the connection state and battery for supported headsets.
+
+Configuration parameters:
+    cache_timeout: How often we refresh this module in seconds (default 10)
+    format: Display format (default 'Headsetcontrol: {devices}')
+    format_battery: Display format for the battery level value (default '({battery_level}%)')
+    format_device: Display format for devices (default '{product} {battery}')
```
lasers, line 10: 1) format - Replace `{devices}` with `{format_device}`.
2) format - Drop `Headsetcontrol: ` prefix because we can use more useful `product` name in `format_device` instead.
3) Kill off format_battery config in favor of adding this in `format_device` instead... ie `format_device` with `"{product} {battery_level}%"`.
4) Add `thresholds` config too for `battery_level` colorizing. :-)
5) I want to default this colorizing like `lm_sensors` module... (after thresholds gets added)... `format_device = "[\?color=darkgray {product}][\?color=battery_level  {battery_level}%]"`.

py3status/modules/headsetcontrol.py
```diff
@@ -0,0 +1,90 @@
...
+    format_device_separator: Seperator if more than one device (default ' ')
+
+Format placeholders:
+    {devices}: The list of devices attached.
+    {vendor}: The vendor of a device as reported by headsetcontrol.
+    {product}: The product name of a device as reported by headsetcontrol.
+    {battery}: The battery level of a device, formated.
+    {battery_level}: The unformated battary level in percent as reported by headsetcontrol.
```
lasers, line 18: (nit) Clean up... Drop `, formatted` and `as reported by headsetcontrol`. and closing periods.

py3status/modules/headsetcontrol.py
```diff
@@ -0,0 +1,90 @@
...
+    # available configuration parameters
+    cache_timeout = 10
+    format = "Headsetcontrol: {devices}"
+    format_battery = "({battery_level}%)"
+    format_device = "{product} {battery}"
+    format_device_separator = " "
+
+    def _get_headset_data(self):
+        try:
+            headsetcontrol_raw_json = self.py3.command_output("headsetcontrol -o json")
```
lasers, line 54: What does `headsetcontrol -o json` output?

py3status/modules/headsetcontrol.py
```diff
@@ -0,0 +1,90 @@
...
+        for device in headset_data["devices"]:
+            headset_data_flat = {
+                "vendor": self.py3.safe_format(device["vendor"]),
+                "product": self.py3.safe_format(device["product"]),
+                "battery": "",
+            }
+            if "CAP_BATTERY_STATUS" in device["capabilities"]:
+                headset_data_flat["battery"] = self.py3.safe_format(
+                    self.format_battery, {"battery_level": device["battery"]["level"]}
+                )
```
lasers, line 73: Safeformatting single strings `vendor`, `product`, `battery_level` is not needed. Do it once at bottom for each `format_device` with `device` dict.

Also, OK, I see we have `capabilities` placeholder. I think we need to see how json looks like, but you probably can add `device = self.py3.flatten_dict(device, delimiter="_")` here and we just use new `battery_level` placeholder.

After flattening dict, run thresholds loop code (see `loadavg` module for example) then safeformat + append list. This way, we expose all placeholders... and if headsetcontrol add more, we already have them available.

EDIT: This module code should look similar to (simplified) `bluetooth` module code after everything is finished.

py3status/modules/headsetcontrol.py
```diff
@@ -0,0 +1,90 @@
...
+    format_device = "{product} {battery}"
+    format_device_separator = " "
+
+    def _get_headset_data(self):
+        try:
+            headsetcontrol_raw_json = self.py3.command_output("headsetcontrol -o json")
+            return json.loads(headsetcontrol_raw_json)
+        except (ValueError, UnicodeDecodeError):
+            self.py3.error("Headsetcontrol returned no valid JSON.")
+        except self.py3.CommandError:
```
lasers, line 58: We don't want to burden users with this `Headsetcontrol returned on valid JSON.` vague error. If an error occurred, the exception can be displayed on the status too (after right-clicking to expand message). Did `ValueError` and/or `UnicodeDecodeError` actually happen?

Did `self.py3.CommandError` happen because of no connected headsets?

Also, explain more about this.
> ... because I found a bug when a headset not connected anymore.

lasers (changes requested):
First iteration.

py3status/modules/headsetcontrol.py
weberval: Thanks for your input. 

> 1. format - Replace {devices} with {format_device}.
> 2. format - Drop Headsetcontrol: prefix because we can use more useful product name in format_device instead.

Sure, can do.

> 3. Kill off format_battery config in favor of adding this in format_device instead... ie format_device with "{product} {battery_level}%".

There might be devices that do not report a battery level. Therefore, I would like to keep the battery format configurable in a dedicated config option.

> 4. Add thresholds config too for battery_level colorizing. :-)
> 5. I want to default this colorizing like lm_sensors module... (after thresholds gets added)... format_device = "[\?color=darkgray {product}][\?color=battery_level  {battery_level}%]".

I would do this in a later stage. First I would like to have an MVP. If you would like to contribute it, I'm also happy to have it in the initial implementation.

py3status/modules/headsetcontrol.py
weberval: Something like this: 

```json
{
  "name": "HeadsetControl",
  "version": "3.1.0-84-gc8bbeb3",
  "api_version": "1.4",
  "hidapi_version": "0.15.0",
  "device_count": 1,
  "devices": [
    {
      "status": "success",
      "device": "Logitech G PRO X 2 LIGHTSPEED",
      "vendor": "Logitech",
      "product": "PRO X 2 LIGHTSPEED",
      "id_vendor": "0x046d",
      "id_product": "0x0af7",
      "capabilities": [
        "CAP_SIDETONE",
        "CAP_BATTERY_STATUS",
        "CAP_INACTIVE_TIME",
        "CAP_EQUALIZER_PRESET",
        "CAP_EQUALIZER",
        "CAP_PARAMETRIC_EQUALIZER"
      ],
      "capabilities_str": [
        "sidetone",
        "battery",
        "inactive time",
        "equalizer preset",
        "equalizer",
        "parametric equalizer"
      ],
      "battery": {
        "status": "BATTERY_AVAILABLE",
        "level": 50
      },
      "equalizer_presets_count": 5,
      "equalizer_presets": {
        "flat": [
          0.0,
          0.0,
          0.0,
          0.0,
          0.0
        ],
        "bass boost": [
          4.0,
          2.0,
          0.0,
          0.0,
          0.0
        ],
        "team chat": [
          -1.0,
          2.0,
          1.0,
          3.0,
          3.0
        ],
        "shooter": [
          -1.0,
          -1.0,
          4.0,
          3.0,
          2.0
        ],
        "moba": [
          0.0,
          1.0,
          1.0,
          2.0,
          4.0
        ]
      }
    }
  ]
}
```

py3status/modules/headsetcontrol.py
weberval: > Did ValueError and/or UnicodeDecodeError actually happen?

No, not during my tests. 

> Did self.py3.CommandError happen because of no connected headsets?

Yes. Unfortunately, headsetcontrol exits with a non-zero exit code if no devices are detected.

> Also, explain more about this.

I discovered, that headsetcotrol reports battery level -1 if a previously connected headset is disconnected, but the dongle is still available.

py3status/modules/headsetcontrol.py
lasers: For devices without or with negative (battery_level), you could use `"[\?if=battery_level>0 {product} {battery_level}%]"`... basically making this device disappear.

`"{product}[\?if=battery_level>0  {battery_level}%]"` makes device without battery_level appear too.

Adding brackets, ie `"[{battery_level}]"` won't appear if placeholder doesn't exist, is `None`, or is `""`.

py3status/modules/headsetcontrol.py
lasers:
```
  "name": "HeadsetControl",
  "version": "3.1.0-84-gc8bbeb3",
  "api_version": "1.4",
  "hidapi_version": "0.15.0",
  "device_count": 1,
  ```
Make it so all those can appear in `format`. Pop off `devices` into `format_device` loop.

py3status/modules/headsetcontrol.py
lasers: If you still get json with non-zero exit code, you might want `self.py3.CommandError as ....` exception output. Lot of examples in `arch_updates` module.

py3status/modules/headsetcontrol.py
lasers: I can contribute. Sure. I let you do this first. Also, I haz no headsets.

lasers: Contributed: https://github.com/lasers/py3status/tree/newmodule-headsetcontrol 

I'm working off your json... I cannot test disconnected error output (`ce.output`).

PR #2332, opened 2026-06-26
volume_status add wpctl support

valdur55:


valdur55:
Well. Looks ok now!

valdur55:
Need to drop manual device setting because device ID-s are dynamic.

lasers:
> Need to drop manual device setting because device ID-s are dynamic.

`device` doesn't have to be number. It depends on backend.

----------

According to wiki, you may be able to supply `device.name` or `node.name` to `device`.

> Choose a property that is unique and stable for use in matches rules. The device.name or node.name properties are recommended.
>
> Avoid using device.id, as it is dynamic and may change between sessions. 

_src_ https://wiki.archlinux.org/title/WirePlumber#Obtain_interface_name_for_rules_matching

EDIT: I assume it wouldn't work, but technically, you could create a small function that runs a command to see if that particular device name/node is active before running any other function.

valdur55:
Added quite silly wpctl status -n parser.
Already looked into pactl devices list and found out that device name format matches,
Thankfully wpctl doesn't care about device input/output type. Control commands are same, and therefore it is possible to have faulty config:  
device : output-device
is_input : TRUE
Is it worth it to prevent that happening?

Is "if self.dev_target:" guarding overkill or not? I personally don't like to send invalid shell command which might have  unexpected side effects. ( sending None as device id)

In general view, ready for review.

py3status/modules/volume_status.py
```diff
@@ -315,6 +316,81 @@ def toggle_mute(self):
...
+        # so re-resolve it on each access
+        if self.device is None:
+            return
+
+        # map a configured device name to its wpctl node id
+        try:
+            lines = self.command_output(["wpctl", "status", "-n"])
+        except CommandError:
+            self.dev_target = None
+            return
```
lasers: Just wondering. Why do we need try/except with verified installed `wpctl`? Is this faulty `None` config you mentioned?

EDIT: It's for `self.dev_target = None` checks?

py3status/modules/volume_status.py
```diff
@@ -315,6 +316,81 @@ def toggle_mute(self):
         self.run_cmd(["pactl", f"set-{self.device_type}-mute", self.device, "toggle"])
 
 
+class Wpctl(Audio):
+    def setup(self, parent):
+        if self.device:
+            # a named device is resolved on demand; see _resolve_device_id()
+            self.dev_target = None
+        else:
+            self.dev_target = "@DEFAULT_AUDIO_SOURCE@" if self.is_input else "@DEFAULT_AUDIO_SINK@"
```
lasers: (Experiment/Optimization) Without testing anything... I think maybe for "performance reason".

Try noop: pass function (and real `resolve_device_id` function)... Assign a correct function at setup to basically make this optimized when `device` is not in use... and potentially avoid try/except + send None issue.

(Nit) Self.device, self.dev_target, self._resolve_device_id is blending in... Can you rename to `self.target` and `self._resolve_target`?

lasers (commented):
>therefore it is possible to have faulty config

I hope we can honor config as close as possible, especially on various backends. If user placed bad/invalid config, then I suppose we would **__not__** ignore it. However, it's up to you because...

Personally, I think/wonder if `is_input` config is a terrible user config that should be just removed because maybe the backends can figure `is_input` out themselves from device name / device information (in each setup)... We would still keep the placeholder. 

Just not in this PR.

> Added quite silly wpctl status -n parser.

Apparently there is `pw-dump` too (located in `pipewire-utils`). Ideal for JSON scripting. Just FYI. I don't know if it's worth it. Might be more clean/more reliable than wpctl if `wpctl status -n` changed one day. EDIT: Not a right tool.

lasers (commented):
To better understand this issue, I coded too.

Please look at this. https://github.com/lasers/py3status/tree/volume_status-wpctl-wip

This is still your contribution. Take it if you find this to be better.

I noticed that...

`wpctl status -n` gives us same thing we need from `get_volume` call.... By using this data instead, we can skip extra subprocess call. One call, not two... on every refresh/volume changed, device changed, etc.

We use this output to find `target_id` too.
We also can use this output to find default device anyway. Not in use, but...

Microphone percentage/muted now works too... because of self._target_id.

Unrelated -- I think it would be useful in adding `format` and `format_device` and defaulting devices to (default `*` devices)... meaning `volume_status` can spawn default source+sink on the bar by default (or be silly and spawn all sources+sinks).  And convert code to use thresholds too. All in a different PR though.

valdur55:
> > therefore it is possible to have faulty config
> 
> I hope we can honor config as close as possible, especially on various backends. If user placed bad/invalid config, then I suppose we would **not** ignore it. However, it's up to you because...

My code was just PoC. Looks like your code handles SINK/SOURCE correctly.

> Personally, I think/wonder if `is_input` config is a terrible user config that should be just removed because maybe the backends can figure `is_input` out themselves from device name / device information (in each setup)... We would still keep the placeholder.

We still need it to differentiate between default SINK/SOURCE




> To better understand this issue, I coded too.
> 
> Please look at this. https://github.com/lasers/py3status/tree/volume_status-wpctl-wip
> 
> This is still your contribution. Take it if you find this to be better.


I tested your code. Looks like working OK; but seems like output type should be int not a float.


> I noticed that...
> 
> `wpctl status -n` gives us same thing we need from `get_volume` call.... By using this data instead, we can skip extra subprocess call. One call, not two... on every refresh/volume changed, device changed, etc.
> 
> We use this output to find `target_id` too. We also can use this output to find default device anyway. Not in use, but...

You are  right. It is possible to get needed info from status command. Nice work with parsing it!
With default device we don't need to parse status output, we can directly do get-volume command.


> Microphone percentage/muted now works too... because of self._target_id.
> 

Interesting bug. Microphone percentage/muted was working for me.

> Unrelated -- I think it would be useful in adding `format` and `format_device` and defaulting devices to (default `*` devices)... meaning `volume_status` can spawn default source+sink on the bar by default (or be silly and spawn all sources+sinks). And convert code to use thresholds too. All in a different PR though.

As I see that You added some formatting to debug output. It helps to differentiate between used backends.
Idea is interesting, useful for debugging, but default behaviour is to control default SINK. Controlling source should be optional. 
What I feel missing is hiding unactive source control. It feels like it can be done with format

Also we need to update documentation to list wpctl status -n command in list:
`    device: Device to use. Defaults value is backend dependent.
        "aplay -L", "pactl list sinks short", "pamixer --list-sinks"
        (default None)`

Also I found out that you don't need set volume upper limit in your at all, because wpctl can handle limiting itself, because limit is one argument of setting volume.

BYW: Wrapping up some code changes and I will push it soon.

py3status/modules/volume_status.py
```diff
@@ -315,6 +316,114 @@ def toggle_mute(self):
...
+            return {
+                "volume": self._format_volume(parts[1]),
+                "muted": "[muted]" in parts,
+            }
+        except (ValueError, IndexError):
+            return None
+
+    @staticmethod
+    def _format_volume(volume):
+        return int(float(volume) * 100)
```
lasers: Use `class Meta` + `:.0f` instead... giving users int/float volume choices.

Grep `update_placeholder_format` on modules for an example.

valdur55:
In general it is good to go review again.

During testing, it took time to recover computer restart. For example pactl doesn't have issue.

lasers (commented):
Just addressing one thing immediately. My PR didn't had this fix too.

py3status/modules/volume_status.py
valdur55: I don't agree. Volume is always integer.
Proof:
````
wpctl set-volume -l 1 @DEFAULT_AUDIO_SINK@ 0.201
wpctl get-volume @DEFAULT_AUDIO_SINK@
Volume: 0.20
wpctl set-volume -l 1 @DEFAULT_AUDIO_SINK@ 0.209
wpctl get-volume @DEFAULT_AUDIO_SINK@
Volume: 0.21
````
Also we show audio volume in scale of 0% - 100%. Wpctl happens to use scale 0.00 - 1.00 so we need to multiply with 100. As example shown, 0.209 gets rounded to 0.21 so volume output is always 2 decimal places.

EDIT: Updated code block formating.

Comment (lasers) [reply py3status/modules/volume_status.py]: Just for future reference, you can wrap commands/code in triple backticks. [code-blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks)

```shell
wpctl set-volume -l 1 @DEFAULT_AUDIO_SINK@ 0.201
> wpctl get-volume @DEFAULT_AUDIO_SINK@
Volume: 0.20
> wpctl set-volume -l 1 @DEFAULT_AUDIO_SINK@ 0.209
> wpctl get-volume @DEFAULT_AUDIO_SINK@
Volume: 0.21
```

When it comes to percentages, I really don't care if it should be int or float. Integer is fine for volume.

It's just that this approach is more ideal when it comes to (any) percentages. This module have other audio classes too, not just `wpctl`. If we default `volume` with rounded `.0f`, it's still going to be an integer.

It's more about how we already have something in place that permits placeholders to be int or float and that we should try to use this more instead of redundant hardcoding code everywhere.

lasers: Fine with skipping too. Generally, we don't try to hardcode percentages in modules.

py3status/modules/volume_status.py
```diff
@@ -315,6 +316,114 @@ def toggle_mute(self):
...
+        parts = (
+            self.command_output(["wpctl", "get-volume", self.selected_device_id]).lower().split()
+        )
+        try:
+            return {
+                "volume": self._format_volume(parts[1]),
+                "muted": "[muted]" in parts,
+            }
+        except (ValueError, IndexError):
+            return None
```
lasers, line 428: Imho, no need for this exception handling if it didn't happen. I rather it happened first so we can get an open issue and fix it. EDIT: ~~Or parse it better.~~ I think your parsing is fine.

valdur55:
After computer reboot it takes some time to wpctl settle.
So it will give following output after boot:
```
2026-07-19 18:46:38 INFO [modules.volume_status]
PipeWire 'pipewire-0' [1.6.7, valdur@vlap, cookie:2147790589]
 └─ Clients:
        33. xdg-desktop-portal                  [1.6.7, valdur@vlap, pid:4522]
        34. WirePlumber                         [1.6.7, valdur@vlap, pid:4732]
        35. wpctl                               [1.6.7, valdur@vlap, pid:4760]

Audio
 ├─ Devices:
 │
 ├─ Sinks:
 │
 ├─ Sources:
 │
 ├─ Filters:
 │
 └─ Streams:

Video
 ├─ Devices:
 │
 ├─ Sinks:
 │
 ├─ Sources:
 │
 ├─ Filters:
 │
 └─ Streams:

Settings
 └─ Default Configured Devices:

2026-07-19 18:46:48 INFO [modules.volume_status]
PipeWire 'pipewire-0' [1.6.7, valdur@vlap, cookie:2147790589]
 └─ Clients:
        29. wpctl                               [1.6.7, valdur@vlap, pid:5206]
        33. xdg-desktop-portal                  [1.6.7, valdur@vlap, pid:4522]
        34. WirePlumber                         [1.6.7, valdur@vlap, pid:4732]
        40. xdg-desktop-portal-wlr              [1.6.7, valdur@vlap, pid:4777]
        44. pipewire                            [1.6.7, valdur@vlap, pid:4839]
        45. WirePlumber [export]                [1.6.7, valdur@vlap, pid:4732]
        64. Blueman                             [1.6.7, valdur@vlap, pid:4578]

Audio
 ├─ Devices:
 │      48. alsa_card.pci-0000_00_03.0          [alsa]
 │      49. alsa_card.pci-0000_00_1b.0          [alsa]
 │
 ├─ Sinks:
 │  *   47. alsa_output.pci-0000_00_1b.0.analog-stereo [vol: 0.55]
 │
 ├─ Sources:
 │  *   56. alsa_input.pci-0000_00_1b.0.analog-stereo [vol: 0.36]
 │
 ├─ Filters:
 │
 └─ Streams:

Video
 ├─ Devices:
 │      46. v4l2_device.pci-0000_00_14.0-usb-0_6_1.0 [v4l2]
 │      57. v4l2_device.pci-0000_00_14.0-usb-0_6_1.0.2 [v4l2]
 │
 ├─ Sinks:
 │
 ├─ Sources:
 │  *   65. v4l2_input.pci-0000_00_14.0-usb-0_6_1.0
 │
 ├─ Filters:
 │
 └─ Streams:

Settings
 └─ Default Configured Devices:
```

WIth default device it gives IndexError and command output is empty string.

When having no devices connected/enabled then there is following output.

```
~ $ wpctl status -n                                                                                                                                                                                                                    22:07:34
PipeWire 'pipewire-0' [1.6.7, valdur@tark, cookie:1573491665]
 └─ Clients:
        32. xdg-desktop-portal                  [1.6.7, valdur@tark, pid:6483]
        33. WirePlumber                         [1.6.7, valdur@tark, pid:6594]
        37. xdg-desktop-portal-wlr              [1.6.7, valdur@tark, pid:6602]
        43. Blueman                             [1.6.7, valdur@tark, pid:6324]
        44. WirePlumber [export]                [1.6.7, valdur@tark, pid:6594]
        45. pipewire                            [1.6.7, valdur@tark, pid:6681]
        63. Google Chrome input                 [1.6.7, valdur@tark, pid:7668]
        68. wpctl                               [1.6.7, valdur@tark, pid:232064]
        91. PulseAudio Volume Control           [1.6.7, valdur@tark, pid:231141]

Audio
 ├─ Devices:
 │      49. alsa_card.pci-0000_01_00.1          [alsa]
 │      50. alsa_card.pci-0000_00_1b.0          [alsa]
 │
 ├─ Sinks:
 │  *   60. auto_null                           [vol: 1.00]
 │
 ├─ Sources:
 │
 ├─ Filters:
 │
 └─ Streams:
        88. PulseAudio Volume Control
             73. monitor_FL
             81. input_FR
             82. input_FL
             86. monitor_FR
        89. PulseAudio Volume Control
             62. input_FL        < Dummy Output:monitor_FL	[paused]
             76. monitor_FL
             78. input_FR        < Dummy Output:monitor_FR	[paused]
             92. monitor_FR

Video
 ├─ Devices:
 │
 ├─ Sinks:
 │
 ├─ Sources:
 │
 ├─ Filters:
 │
 └─ Streams:

Settings
 └─ Default Configured Devices:
         0. Audio/Sink    alsa_output.pci-0000_00_1b.0.analog-stereo
```

When no audio source connected, I get an error:
```
~ $ wpctl get-volume @DEFAULT_AUDIO_SOURCE@ 
Translate ID error: '-1' is not a valid ID (returned by default-nodes-api)
```
Which ends up as empty response and therefore IndexError.

According to first result after  reboot I decided to get default aduio sink get-volume output until it is not empty string, otherwise I raise error which is captured backend retry logic.

For get-volume output, IndexError is needed but I don't fully get when it will get ValueError.  Also what purposa has it for pactl volume status?

lasers:
```python
# wpctl
>>> float("[MUTED]")
  File "<python-input-4>", line 1, in <module>
    float("[MUTED]")
    ~~~~~^^^^^^^^^^^
ValueError: could not convert string to float: '[MUTED]'
```
```python
# pamixer
>>> line = "50 true new"
>>> muted, perc = line.split()
  File "<python-input-6>", line 1, in <module>
    muted, perc = line.split()
    ^^^^^^^^^^^
ValueError: too many values to unpack (expected 2, got 3) 

# paxmier
>>> line = ""
>>> muted, perc = line.split()
  File "<python-input-2>", line 1, in <module>
    muted, perc = line.split()
    ^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 0)
```
EDIT: This hides exceptions from users, which in turns, wouldn't file issues. The only way we can know about them is if we can avoid adding try/exceptions everywhere (often, modules only). If outputs are consistent throughout versions, then we likely do not need them.

py3status/modules/volume_status.py
```diff
@@ -315,6 +316,122 @@ def toggle_mute(self):
...
+
+    @staticmethod
+    def _format_volume(volume):
+        return float(volume) * 100
```
lasers: Inline this on `volume`.

valdur55: Done

lasers (approved):
> re: Inline `volume`

I forget you had two calls in place to format same volume. That's the only part I didn't like (I mean two different calls, not formatting). Still clean to inline them anyway. I assume the verbose output would have skipped missing source on startup until it's ready after couple cycles. If not, it could be coded that way.

Unrelated. My reading is that Pamixer class can do `int(None)` on scrolling up/down... which would throw `TypeError` when the command failed... and maybe it hadn't... Possibly no need to try/except that command with `ValueError` after all.

Nice new volume class added. 👍

valdur55:
Well, using float(volume) * 100 instead of _volume_format.
We had before even  round(float(volume) * 100) which was too much to repeat, wihtout round it is OK to inline them.


I made empty response discovery by logging out get-volume command output.
Stripped down some other modules loading:
```
026-07-19 19:10:57 INFO [core] ========
2026-07-19 19:10:57 INFO [core] starting py3status version 3.63 python 3.14.6
2026-07-19 19:10:57 INFO [core] running within git repo
2026-07-19 19:10:57 INFO [core] git branch: volume_status_wpctl
2026-07-19 19:10:57 INFO [core] git commit: 942d8cb refs/heads/volume_status_wpctl onto 518d11a71053ebbac3cf2df585db04aeedb5287c
2026-07-19 19:10:57 INFO [core] config file: /home/valdur/.config/py3status/bar-top.conf
2026-07-19 19:10:57 INFO [core] window manager: sway
2026-07-19 19:10:57 INFO [core] path-included module paths: ['/home/valdur/dotfiles/py3status/bar-top-user-modules']
---------------------------------------------
2026-07-19 19:10:57 INFO [modules.lm_sensors] starting module
2026-07-19 19:10:57 INFO [modules.kdeconnector vphone] starting module
2026-07-19 19:10:57 INFO [modules.kdeconnector nordphone] starting module
2026-07-19 19:10:57 INFO [modules.volume_status] 
2026-07-19 19:10:58 INFO [modules.scratchpad] starting module
2026-07-19 19:10:58 INFO [modules.window] starting module
2026-07-19 19:10:58 INFO [modules.volume_status] 
volume: 0.55

2026-07-19 19:10:58 INFO [modules.volume_status] starting module
2026-07-19 19:10:58 INFO [modules.volume_status] 
volume: 0.55

2026-07-19 19:11:09 INFO [modules.volume_status] 
volume: 0.55

2026-07-19 19:11:19 INFO [modules.volume_status] 
volume: 0.55

2026-07-19 19:11:29 INFO [modules.volume_status] 
volume: 0.55

```

Uh oh. percentage formatting for format_muted isn't actually working :/.
When using following muted config line:
    format_muted = "({percentage}%)"
 It started to show floating values again. Somehow 55 ends up always like 55.00000000001 which showed me that something went  wrong.
 Added it to debug output to simplify research.
 
Damn. how I missed it before. It come out when i updated other computer codebase.

valdur55:
Now it looks better!.

valdur55:
@lasers hey! Can You confirm that the muted mode percentage also is int?

lasers:
@valdur55 I don't see any float `{percentage}` in both formats. `Int` confirmed everywhere.

ultrabug:
is this PR ready and okay for you both?

valdur55:
OK for me.

lasers:
Yes, OK.

ultrabug:
Thank you!

PR #2333, opened 2026-07-01
log: decouple py3status logging

lasers:
Working with codex off this https://github.com/ultrabug/py3status/pull/2327#issuecomment-4351974940 

> It's certainly terrible to look at those added if debug.

Direct logging avoids this.

> Original logging PR called for using the logging module directly everywhere plus using lazy %s strings. One can look into adding ruff rule to prevent f-strings logging to reduce wasted f-string calculation.

Addressed.
Lint ruled added -- https://docs.astral.sh/ruff/rules/logging-f-string/ 

> I think the real issue is that we're logging everything through a busy self.log instead of breaking up into either using logging functions directly or create additional py3 logging (ie self.py3.debug pass if disabled, self.py3.info). I looked at how we logged debug/warn code... we just print plain strings most times, so no need to check for arrays/lists/sets/etc and other functions. Those should be just for user-facing logging.

See above. `self.py3.log()` is now for user-facing module logging. We also could remove `self.py3_wrapper.log()` in favor of direct logging... for performance reasons since we don't really need wrapper / additional checks.

> And to really improve performance, we should be able to disable logs completely too, no syslog, no info log, for most runs, which your proposal PR currently doesn't cover. This issue is acknowledged too.

I haven't looked into this.

> @valdur55 Additional suggestion is to fail fast when calling disallowed logging level

Added in `self.py3.log()`.

1) I still have to review more. I just want to toss this PR out and see how you think/feel about this...  
2) I think core code should stick with direct logging and user-facing modules can stick with `self.py3.log()`
3) This changed lot of things. For instance, we now can see logging names (via `__name__`) rather than piping everything through `self.py3.log()` / `self.py3_wrapper.log()` functions.

Here's the benchmarks.

```
Summary (median µs/call)

benchmark                      | master | logging-memory-leak | logging-split
-------------------------------+--------+---------------------+--------------
internal_debug_disabled_simple | 0.96   | 0.70                | 0.18
internal_debug_disabled_dict   | 10.68  | 0.70                | 0.18
internal_info_enabled_simple   | 5.81   | 6.37                | 4.92
py3_debug_disabled_simple      | 1.20   | 0.78                | 0.35
py3_debug_disabled_dict        | 10.88  | 0.78                | 0.35
py3_info_enabled_simple        | 6.21   | 7.13                | 5.79
```

Replaces / Closes https://github.com/ultrabug/py3status/pull/2327

### DIFF
--- a/py3status/command.py
+++ b/py3status/command.py
@@ -1,9 +1,12 @@
 import argparse
 import json
+import logging
 import os
 import socket
 import threading
 from pathlib import Path
+
+logger = logging.getLogger(__name__)
 
 SERVER_ADDRESS = "/tmp/py3status_uds"
 MAX_SIZE = 1024
@@ -156,7 +159,7 @@ class CommandRunner:
                     if requested_name == name.split(" ")[0]:
                         found_modules.add(module_name)
 
-        self.py3_wrapper.log(f"found {found_modules}", level="debug")
+        logger.debug("matched modules: %s", sorted(found_modules))
         return found_modules
 
     def refresh(self, data):
@@ -168,7 +171,7 @@ class CommandRunner:
         update_i3status = False
         for module_name in self.find_modules(modules):
             module = self.py3_wrapper.output_modules[module_name]
-            self.py3_wrapper.log(f"refresh {module}", level="debug")
+            logger.debug("refreshing module '%s'", module_name)
             if module["type"] == "py3status":
                 module["module"].force_update()
             else:
@@ -194,7 +197,7 @@ class CommandRunner:
             for name, message in CLICK_OPTIONS:
                 event[name] = data.get(name)
 
-            self.py3_wrapper.log(event, level="debug")
+            logger.debug("dispatching click event %s", event)
             # trigger the event
             self.py3_wrapper.events_thread.dispatch_event(event)
 
@@ -203,7 +206,7 @@ class CommandRunner:
         check the given command and send to the correct dispatcher
         """
         command = data.get("command")
-        self.py3_wrapper.log(f"Running remote command {command}", level="debug")
+        logger.debug("running command '%s'", command)
         if command == "refresh":
             self.refresh(data)
         elif command == "refresh_all":
@@ -237,8 +240,7 @@ class CommandServer(threading.Thread):
         # Create a UDS socket
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         sock.bind(server_address.as_posix())
-
-        self.py3_wrapper.log(f"Unix domain socket at {server_address}", level="debug")
+        logger.debug("socket listening on %s", server_address)
 
         # Listen for incoming connections
         sock.listen(1)
@@ -262,25 +264,20 @@ class CommandServer(threading.Thread):
         while True:
             try:
                 data = None
-                # Wait for a connection
-                self.py3_wrapper.log("waiting for a connection", level="debug")
-
-                connection, client_address = self.sock.accept()
+                connection, _client_address = self.sock.accept()
+                logger.debug("connection accepted")
                 try:
-                    self.py3_wrapper.log("connection from", level="debug")
-
                     data = connection.recv(MAX_SIZE)
                     if data:
                         data = json.loads(data.decode("utf-8"))
-                        self.py3_wrapper.log(f"received {data}", level="debug")
+                        logger.debug("received payload %s", data)
                         self.command_runner.run_command(data)
                 finally:
                     # Clean up the connection
                     connection.close()
             except Exception:
                 if data:
-                    self.py3_wrapper.log("Command error")
-                    self.py3_wrapper.log(data)
+                    logger.error("payload: %s", data)
                 self.py3_wrapper.report_exception("command failed")
 
 
--- a/py3status/constants.py
+++ b/py3status/constants.py
@@ -13,24 +13,34 @@ GENERAL_DEFAULTS = {
 LOGGING_CONFIG = {
     "version": 1,
     "disable_existing_loggers": False,
+    "filters": {"shortname": {"()": "py3status.log.ShortnameFilter"}},
     "handlers": {
         "syslog": {
-            "formatter": "default",
+            "formatter": "syslog",
+            "filters": ["shortname"],
             "class": "logging.handlers.SysLogHandler",
             "address": "/dev/log",
         },
     },
     "formatters": {
-        "default": {
-            "format": "%(asctime)s %(levelname)s %(message)s",
+        "syslog": {
+            "format": "[%(shortname)s] %(message)s",
+        },
+        "logfile": {
+            "format": "%(asctime)s %(levelname)s [%(shortname)s] %(message)s",
             "datefmt": "%Y-%m-%d %H:%M:%S",
-        }
+        },
     },
     "root": {"handlers": ["syslog"], "level": "INFO"},
 }
 
 LOGGING_LOG_FILE_CONFIG = {
-    "__log_file": {"formatter": "default", "class": "logging.FileHandler", "filename": None}
+    "__log_file": {
+        "formatter": "logfile",
+        "filters": ["shortname"],
+        "class": "logging.FileHandler",
+        "filename": None,
+    }
 }
 
 LOGGING_LOG_LEVELS = {
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -6,18 +6,18 @@ import sys
 import time
 from collections import deque
 from pathlib import Path
-from pprint import pformat
 from signal import SIGCONT, SIGTERM, SIGTSTP, SIGUSR1, Signals, signal
 from subprocess import Popen
 from threading import Event, Thread
 from traceback import extract_tb, format_stack, format_tb
 
 from py3status.command import CommandServer
-from py3status.constants import LOGGING_CONFIG, LOGGING_LOG_FILE_CONFIG, LOGGING_LOG_LEVELS
+from py3status.constants import LOGGING_CONFIG, LOGGING_LOG_FILE_CONFIG
 from py3status.events import Events
 from py3status.formatter import expand_color
 from py3status.helpers import print_stderr
 from py3status.i3status import I3status
+from py3status.log import module_logger_name, resolve_log_level
 from py3status.module import Module
 from py3status.output import OutputFormat
 from py3status.parse_config import process_config
@@ -39,6 +39,7 @@ CONFIG_SPECIAL_SECTIONS = [
 
 ENTRY_POINT_NAME = "py3status"
 ENTRY_POINT_KEY = "entry_point"
+logger = logging.getLogger(__name__)
 
 
 class Runner(Thread):
@@ -105,7 +106,7 @@ class CheckI3StatusThread(Task):
         if not self.i3status_thread.is_alive():
             err = self.i3status_thread.error
             if not err:
-                err = "I3status died horribly."
+                err = "i3status died horribly"
             self.notify_user(err)
         else:
             # check again in 5 seconds
@@ -165,7 +166,7 @@ class Common:
             param = expand_color(param.lower(), self.none_setting)
         return param
 
-    def report_exception(self, msg, notify_user=True, level="error", error_frame=None):
+    def report_exception(self, msg, notify_user=True, level="error", error_frame=None, name=None):
         """
         Report details of an exception to the user.
         This should only be called within an except: block Details of the
@@ -217,15 +218,16 @@ class Common:
                     if found:
                         break
             # all done!  create our message.
-            msg = "{} ({}) {} line {}.".format(msg, exc_type.__name__, filename, line_no)
+            msg = "{} ({}) {} line {}".format(msg, exc_type.__name__, filename, line_no)
         except:  # noqa e722
-            # something went wrong report what we can.
-            msg = f"{msg}."
+            # something went wrong, report msg as-is.
+            pass
         finally:
             # delete tb!
             del tb
         # log the exception and notify user
-        self.py3_wrapper.log(msg, "warning")
+        tmp_logger = logging.getLogger(name) if name else logger
+        tmp_logger.log(resolve_log_level(level), msg)
         if traceback:
             # if debug is not in the config  then we are at an early stage of
             # running py3status and logging is not yet available so output the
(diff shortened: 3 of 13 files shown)


lasers:
Ready for review, but do not merge.
```
Summary (median us/call)
benchmark                      | master | logging-memory-leak | logging-split
-------------------------------+--------+---------------------+--------------
internal_debug_disabled_simple | 0.81   | 0.60                | 0.14         
internal_debug_disabled_dict   | 9.66   | 0.59                | 0.14         
internal_info_enabled_simple   | 5.27   | 5.38                | 4.39         
py3_debug_disabled_simple      | 1.11   | 0.65                | 0.33         
py3_debug_disabled_dict        | 10.09  | 0.66                | 0.33         
py3_info_enabled_simple        | 5.63   | 6.28                | 5.11
```

ultrabug:
Thanks @lasers on the principle I like it very much. Will try to review soon (vacations)

py3status/command.py
```diff
@@ -203,7 +206,7 @@ def run_command(self, data):
         check the given command and send to the correct dispatcher
         """
         command = data.get("command")
-        self.py3_wrapper.log(f"Running remote command {command}", level="debug")
+        logger.debug("Running remote command %s", command)
```
ultrabug:
```suggestion
        logger.debug("running remote command %s", command)
```

py3status/command.py
```diff
@@ -238,7 +241,7 @@ def __init__(self, py3_wrapper):
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         sock.bind(server_address.as_posix())
 
-        self.py3_wrapper.log(f"Unix domain socket at {server_address}", level="debug")
+        logger.debug("Unix domain socket at %s", server_address)
```
ultrabug:
```suggestion
        logger.debug("unix domain socket at %s", server_address)
```

py3status/command.py
```diff
@@ -263,24 +266,23 @@ def run(self):
...
+                        logger.debug("received %s", data)
                         self.command_runner.run_command(data)
                 finally:
                     # Clean up the connection
                     connection.close()
             except Exception:
                 if data:
-                    self.py3_wrapper.log("Command error")
-                    self.py3_wrapper.log(data)
+                    logger.info("command error: %s", data)
```
ultrabug: maybe warning?

py3status/constants.py
```diff
@@ -13,24 +13,31 @@
...
             "formatter": "default",
+            "filters": ["shortname"],
             "class": "logging.handlers.SysLogHandler",
             "address": "/dev/log",
         },
     },
     "formatters": {
         "default": {
-            "format": "%(asctime)s %(levelname)s %(message)s",
+            "format": "%(asctime)s %(levelname)s [%(shortname)s] %(message)s",
```
ultrabug, line 30: I think we should get rid of asctime since in modern distros journald adds its own timestamp

py3status/core.py
```diff
@@ -225,15 +226,16 @@ def report_exception(self, msg, notify_user=True, level="error", error_frame=Non
...
+        tmp_logger.log(resolve_log_level(level), msg)
         if traceback:
             # if debug is not in the config  then we are at an early stage of
             # running py3status and logging is not yet available so output the
             # error to STDERR so it can be seen
             if "debug" not in self.config:
                 print_stderr("\n".join(traceback))
             elif self.config.get("log_file"):
-                self.py3_wrapper.log("".join(["Traceback\n"] + traceback))
+                tmp_logger.info("Traceback\n%s", "".join(traceback))
```
ultrabug:
```suggestion
                tmp_logger.info("traceback\n%s", "".join(traceback))
```

py3status/core.py
```diff
@@ -464,13 +466,15 @@ def _get_entry_point_based_modules(self):
             try:
                 module = entry_point.load()
             except Exception as err:
-                self.log(f"entry_point '{entry_point}' error: {err}")
+                logger.info("entry_point '%r' error: %s", entry_point, err)
```
ultrabug: maybe warning or error?

py3status/core.py
```diff
@@ -564,7 +568,7 @@ def _log_gitversion(self):
         if not git_path.exists():
             return
 
-        self.log("Running within git repo")
+        logger.info("Running within git repo")
```
ultrabug:
```suggestion
        logger.info("running within git repo")
```

py3status/core.py
```diff
@@ -581,24 +585,24 @@ def _log_gitversion(self):
                 with (git_path / "HEAD").open() as f:
                     out = f.readline()
             except OSError:
-                self.log(
-                    "Unable to read git HEAD. " "Use python git package for more repo information"
+                logger.warning(
+                    "Unable to read git HEAD. Use python git package for more repo information"
```
ultrabug:
```suggestion
                    "unable to read git HEAD. Use python git package for more repo information"
```

py3status/log.py
```diff
@@ -0,0 +1,33 @@
+import logging
+from pprint import pformat
+
+from py3status.constants import LOGGING_LOG_LEVELS
+
+
+class ShortnameFilter(logging.Filter):
+    def filter(self, record):
```
ultrabug, line 22: please add an example input record and output

py3status/log.py
```diff
@@ -0,0 +1,33 @@
...
+        # use local mapping instead to support python 3.9+
+        level = LOGGING_LOG_LEVELS.get(level.upper(), logging.DEBUG)
+    return level
+
+
+def log_message(logger, message, level):
+    level = resolve_log_level(level)
+    if not logger.isEnabledFor(level):
+        return
+    # nicely format logs if we can using pretty print
```
ultrabug:
```suggestion
    # nicely format logs if we can use pretty print
```

py3status/module.py
```diff
@@ -152,9 +154,10 @@ def prepare_module(self):
                     f"{e}",
                 ]
                 self.runtime_error(self.error_messages[1], "post_config_hook")
-                msg = f"Exception in `{self.module_full_name}` post_config_hook() : {self.error_messages}"
-                self._py3_wrapper.report_exception(msg, notify_user=False)
-                self._py3_wrapper.log(f"terminating module {self.module_full_name}")
+                self._py3_wrapper.report_exception(
+                    "post_config_hook failed", notify_user=False, name=self.module_full_name
+                )
+                self._logger.info("terminating module")
```
ultrabug: maybe warning?

py3status/module.py
```diff
@@ -1035,10 +1049,12 @@ def run(self):
                         )
 
                 except Exception as e:
-                    msg = "Instance `{}`, user method `{}` failed"
-                    msg = msg.format(self.module_full_name, meth)
                     if not self.testing:
-                        self._py3_wrapper.report_exception(msg, notify_user=False)
+                        self._py3_wrapper.report_exception(
+                            f"user method '{meth}' failed",
```
ultrabug, line 1079: we want this f-string?

py3status/storage.py
```diff
@@ -46,9 +49,7 @@ def init(self, py3_wrapper):
         except OSError:
             pass
 
-        self.py3_wrapper.log(f"storage_path: {self.storage_path}")
-        if self.data:
-            self.py3_wrapper.log(f"storage_data: {self.data}")
+        logger.info("path: %s", self.storage_path)
```
ultrabug, line 50: maybe we could have kept a log for self.data but only the keys AND only in debug level?

ultrabug (changes requested):
not tested yet, just pure code review, looking promising so far!

py3status/constants.py
lasers: I think we need to leave it alone for `log_file`.

py3status/module.py
lasers: Not logger-related. Otherwise, lint `logging-f-string (G004)` would have caught this.

py3status/storage.py
lasers: I went with `logger.debug()` at first, but it's still messy / leaking information. Fine with keys.

lasers:
Wanting to make it clear in the logs about detecting available modules: `default modules`, `user-included modules`, and `entrypoint modules`.... I saw questionable code relating to entrypoint so I need to install and test entrypoint modules eventually.

Your suggestions had been largely included.

py3status/module.py
lasers: post_config_hook failed... Error.

py3status/constants.py
ultrabug: Then add a formatter for logfile itself? tbh I don't think much ppl use it and most rely on distro's logger.

Default should be distro's default behavior AKA syslog/journald

ultrabug:
Now we've been that far it looks like:

1. Py3statusWrapper.log() is dead code to be removed
2. module_test.py will not print anything anymore so we should call logging.basicConfig(level=logging.DEBUG) to get back useful debug infos

py3status/constants.py
lasers: Yes. Customization is one of the reasons why we added logger. Formatter `logfile` added. ;-)

py3status/command.py
lasers: I thought I replied. `post_config_hook failed` made this an error.

lasers:
> 1) Py3statusWrapper.log() is dead code to be removed

Dead, but I wasn't sure if we should remove it. Now dead and gone.

> 2) module_test.py will not print anything anymore so we should call logging.basicConfig(level=logging.DEBUG) to get back useful debug infos

OK. Added. New arg also added `--log-level [LEVEL]` such as `python loadavg.py --log-level debug # (default info)`. Filter also added to silence method noises occurring in `module.py` on debug level during module_testing.

--- 
Cleaning up i3status, other files, and especially module importing code to make it easy/compact to log. Additional code in `entry_points`-related added to permit loading only once, not again and again for logging (and maybe one more other thing).

Renaming variables `user_modules` to something a bit more explicit/clear... `discoverable_modules` to illustrate combined `path included modules` and `entry point modules`. Also, using more  variables such as `path_included_modules` and `entry_point` too. 

Ready for review again.
---
Same benchmarks as last two. Nothing significant changed.
```
Summary (median us/call)
benchmark                      | master | logging-memory-leak | logging-split
-------------------------------+--------+---------------------+--------------
internal_debug_disabled_simple | 0.82   | 0.59                | 0.14         
internal_debug_disabled_dict   | 9.83   | 0.59                | 0.14         
internal_info_enabled_simple   | 5.32   | 5.40                | 4.38         
py3_debug_disabled_simple      | 1.11   | 0.64                | 0.33         
py3_debug_disabled_dict        | 9.98   | 0.64                | 0.33         
py3_info_enabled_simple        | 5.69   | 6.30                | 5.05         
```

ultrabug:
So far so good, tried with journald and it's really an improvement!

I'm wondering about those logs tho, looks like we catch modules' loaded modules own logs but without referencing them:

```
DEBUG [requests] Making request: POST https://oauth2.googleapis.com/token
```

This comes from my google_calendar module but we only see it as `requests`.

I will give the logfile option a try

ultrabug:
Tested logfile, do user really expect to have BOTH syslog and the defined logfile being fed by py3status logs? I did not but maybe it's just I'm old and confused :)

lasers:
>This comes from my google_calendar module but we only see it as requests.

Nothing turned up in grepping `Making request`. That one came from [google.auth.transport.requests](https://google-auth.readthedocs.io/en/latest/_modules/google/auth/transport/requests.html).

The real culrpit is `shortname` filter with no guards. 

> record.shortname = record.name.rsplit(".", 1)[-1]

Prefix added to make sure it only happens on py3status files. I also modified py3staus modules to have `py3status.modules.` namespace... So  you tell me if you like shortname `modules.sysdata` prefix.. or you prefer `sysdata`. The latter introduces additional `leafname` field.

> do user really expect to have BOTH syslog and the defined logfile being fed by py3status logs?

AFAIK, same behavior. Logfile isn't in use by default.... Styling / formatting for both syslog and logfile are being assigned to stay backward compatibility with `--log-file`. We backward compatible `--debug` too.
EDIT: If you're asking me that we should only allow one or another, this breaks ~~old~~ same behavior. Idc.

py3status/constants.py
```diff
@@ -13,24 +13,34 @@
...
+            "filters": ["shortname"],
             "class": "logging.handlers.SysLogHandler",
             "address": "/dev/log",
         },
     },
     "formatters": {
-        "default": {
-            "format": "%(asctime)s %(levelname)s %(message)s",
+        "syslog": {
+            "format": "%(levelname)s [%(shortname)s] %(message)s",
```
ultrabug:
```suggestion
            "format": "[%(shortname)s] %(message)s",
```

I'd like to suggest we let journald handle levels so we don't need to log them explicitly

ultrabug:
>  So you tell me if you like shortname modules.sysdata prefix.. or you prefer sysdata.

I like modules.sysdata

ultrabug:
> EDIT: If you're asking me that we should only allow one or another

I think we should default to syslog but if log-file is used we should only use that log file; so make it exclusive

py3status/constants.py
lasers: Gotcha. Not logging level names explicitly.

lasers:
> I think we should default to syslog but if log-file is used we should only use that log file; so make it exclusive

Done.

ultrabug:
Tested OK, almost there mate will do another code review ASAP

ultrabug:
My turn with AI

*One thing I'd still raise, since this commit touches exactly these lines: _setup_logging() starts from init_logging_config = dict(LOGGING_CONFIG) — a shallow copy. _deep_merge(), the handler insertion, and the root-handlers rewrite all mutate the nested dicts, which are shared with the module-level LOGGING_CONFIG constant. In a single normal run that's invisible, but if _setup_logging() ever runs twice in one process (tests, restarts-in-process), state from the first call pollutes the constant. It's a pre-existing wart inherited from master, but since this commit is rewriting the handler-wiring logic anyway, switching to copy.deepcopy(LOGGING_CONFIG) would be a one-line hardening with zero downside.*

wdyt? sounds reasonable tho not super necessary

lasers:
I think it is not necessary because since we only run it once _(to set up logging config)_ per process. Also, this is what happened with your `https://github.com/ultrabug/py3status/issues/859` and my `https://github.com/ultrabug/py3status/issues/1062`. We also don't do anything with logging config after that.

lasers:
Right now, we only have to keep an eye out on trivial typos, inconsistency style, less-than-ideal or redundant messages.

Code for loading-once/logging entry-point modules is ugly because it's sharing code with path-included modules and maybe default modules too, but it gets the logging part done reasonably nice without loading repeatedly. One day, I may want to see if we can make it more lean/clean in future. I don't have more to do with this.

Don't forget to review feedback/mergeability on all those other PRs too.

ultrabug:
I agree, sounds reasonable like this indeed thanks for the hard work.

Will give it a last try and start merging

ultrabug:
Thank you so much, let's move on!

PR #2344, opened 2026-07-19
convert i3status wrapper to modules

lasers:
Merge `i3status` module first https://github.com/ultrabug/py3status/pull/2312.

-----

This converts user config from `core.i3status` wrapper to `modules.i3status` modules on the fly.

To keep backward compatibility, this would generate as many `modules.i3status` instances as it can -- for all i3status modules found in `core.i3status` wrapper... allowing users to retain their existing module positions and such.

Users would be expected to update their config eventually. This can be done by moving `core.i3status` module configs into `modules.i3status`'s config named `modules` where it holds a list of modules and their settings.

See `i3status` module for an example.


lasers:
@ultrabug Ready for review.

ultrabug:
I'll stop with current master and do a release first with logging etc before moving on, hope that's okay

ultrabug:
@lasers before looking at the code I have trouble understanding the real need for this? What's the point? Why is it needed? Please share your plan / thoughts.

lasers:
`modules.i3status` is just a module.

What this does... it finds and updates existing i3status wrapper configs (ie)
```python
wireless wlan0 {
        format_up = "W: (%quality at %essid, %bitrate) %ip"
        format_down = "W: down"
        color = '#3fff99'
}

ethernet eth0 {
        format_up = "E: %ip (%speed)"
        format_down = "E: down"
        color = '#3fff99'
}

order += "wireless _first_"
order += "load"
order += "ethernet _first"
```     
... and convert them to generated `modules.i3status` configs, load `modules.i3status` instances each. 
```python
# /tmp/py3status-i3status_67kvtcff.conf
general {
    output_format = "i3bar"
}

order += "wireless _first_"
wireless _first_ {
    format_up = "W: (%quality at %essid, %bitrate) %ip"
    format_down = "W: down"
}
```
```python
# /tmp/py3status-i3status__b8w95k8.conf
general {
    output_format = "i3bar"
}

order += "load"
load {
    format = "%5min"
}
```
```python
# /tmp/py3status-i3status_gkiba_ny.conf
general {
    output_format = "i3bar"
}

order += "ethernet _first_"
ethernet _first_ {
    format_up = "E: %ip (%speed)"
    format_down = "E: down"
}
```

It's more of a backward compatibility thing. Leave it in for few years or something. Idk... or we can do a hard stop to tell users to convert to a new `modules.i3status` module because we no longer do i3status wrapper.

lasers:
@ultrabug My plan is to provide you with options. That's it.

1) Merge `newmodule-i3status` pr first (it's a module).
a) Backward compatibility pr (it's a config updater from raw i3status configs to `modules.i3status` configs).
b) Retire i3status wrapper immediately pr https://github.com/lasers/py3status/tree/retire-i3status-wrapper 

Do 1) then a) or b).

The nice plan is to carry on with the converted, having users notice that there is a new module _for i3status???_ and they should gracefully update their config... and move everything over. 

The mean plan is to not merge this, and skip raw i3status modules because we don't run i3status wrapper anymore, only `modules.i3status` configs.

Third plan is... I can make `RETIRED_MODULES` for all raw i3status modules too.
> i3status wrapper has been deprecated. Please use i3status module instead.

My opinion was that we do nice plan first then maybe we replace nice plan with third plan (or just stop) later.
EDIT: If you want to retire them immediately, we can do that (third plan). https://github.com/lasers/py3status/tree/retire-i3status-wrapper 
EDIT: Maybe we should just retire them.

ultrabug:
>  load modules.i3status instances each.

RFC 4.0 I think I mentioned this indeed BUT I'd rather them all grouped and we spawn only ONE i3status process instead of many.

lasers:
This PR is stale, but I think I am finished with this on my side. I put lot of thoughts and efforts into this.. With AI help too. The problem is possibly spending next 6 months hashing this out. I don't even want to do that. Lol.

Gimme powers and I'll get many things done in few days. 💀

Also, that's not how modules/instances works though. Currently, If I run multiple conky modules, they all run a process each. You want a wrapper and you do have it. Tradeoffs. Also, it's somehow not practical to add conky wrapper into py3status. Why is it practical to keep i3status in py3status? And why do you want i3status to go from wrapper to module? 

My stance is that there would be no deprecation warnings ever. Too many people uses i3status with py3status. Too many people know about py3status being a i3status wrapper. There will be muscle memory or curiosity to do  i3status setup. People will just run py3status with default i3status.conf too. I don't want to annoy or fight people with deprecation. Death by a thousand paper cuts. That's why this exists. To just get their modules up and running.

If you want ONE i3status process, then you update your config... Use py3status's new i3status module... Put your i3status modules in there. In `modules` list of dicts. This approach, I think, would avoid issues in the long run  _(over i3status-as-a-container-module-with-nested-config-sections)_ but is not user friendly or same as config sections.

Other thing is that we have to try and honor i3status options, py3status options such as `border`, `color`, etc. Also, `on_click` configs.  If they gets grouped together, then that will break something... Module position /  `order +=` is probably one of them.  i3status modules are excluded from several py3status features simply because they're not py3status modules.

If you don't use many of those, then great, it can be grouped together, but definitely not on the wrapper->module config conversion on the fly with whatever configs they have. Too many things that can go wrong / restricted in little ways.

Making i3status a py3status module means no more reviewing or keeping i3status stuff in core/event/module/etc. I have tried my best to place everything in i3status module (and compat/i3status.py). I also want to speedrun cleaning lot of things so I can go back to living under my rock. 🪨

lasers:
I will break this up too.

PR #2346, opened 2026-07-21
remove module: transmission

lasers:
(qol) This PR removes a module `transmission`.

Somewhat broken, but it still works with minor fixes applied. Also, I don't think anybody uses this due to filename lengths and such taking up bar space. Probably a junk module to be removed soon. Not to mention `transmission-remote` does have `-j` (JSON) option that's far better than slicing strings... See below for why.  

> python3 transmission.py --term
 0.00  Stop 1 ed      ubuntu-26.10-snapshot1-live-server-s390x.iso 25.%
 0.00  Idle 1 ubuntu-26.10-snapshot1-live-server-s390x.iso 25.%
 0.00  Down 1 oading  ubuntu-26.10-snapshot1-live-server-s390x.iso 27.%


lasers:
@ultrabug This then `pr#2332`. I'll fix volume conflicts in my prs (if any).

PR #2348, opened 2026-07-21
sysdata: drop legacy config compatibility

lasers:
Merge `modules: drop legacy` first https://github.com/ultrabug/py3status/pull/2350.

----

(qol cleanup) This removes legacy stuff. We kept them long enough.

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -134,81 +134,12 @@ class Py3status:
     thresholds = [(0, "good"), (40, "degraded"), (75, "bad")]
 
     class Meta:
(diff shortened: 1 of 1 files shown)


lasers:
@ultrabug This can come after `modules: drop legacy config compatibility` pr.

Sysdata contains additional legacy code not found in `class Meta` for old threshold placeholders is which why this is separated from other modules, but really, this was made first, then I thought, yeah, we should deprecate legacy config compatibility in all modules next.

lasers:
Squashed into https://github.com/ultrabug/py3status/pull/2350.

PR #2349, opened 2026-07-21
sysdata: streamline metric collection

lasers:
Review ~~`sysdata: drop legacy` first https://github.com/ultrabug/py3status/pull/2348.~~ https://github.com/ultrabug/py3status/pull/2350 first.

-----

(qol cleanup) OK. I wanted to do this for some time... to move them into functions and keep updating with returned dicts in the loop, but codex says it's better to mutate everything instead... that way we don't have any really short-lived dict initialization and stuff like that. OK, we'll mutate.

I made minor change to `arc_get_size`... assuming that zfs, loaded on startup in /proc or not, will either stay there indefinitely or not. Users without zfs gets `0`... and no zfs regex is needed because the line looks very easy to parse.

Moved lot of things to `post_config_hook` too so we can stick with minimal functions including optional last one `max_used_percent`.

In short, `sysdata` can be 1-20% faster than before compared to master. It is a wash with _all_ placeholders. The expensive one is `lm_sensors` call for `{cpu_temp}` which I did not touch... mainly because we had various issues/prs wanting to fix `{cpu_temp}`.
- `(nolink to various cpu_temp issues/prs) https://github.com/ultrabug/py3status/pull/2029`.

Benchmark.
```
branch             case         mean_seconds      vs_master
-----------------------------------------------------------
master             default       0.000094523       baseline
master             wide          0.003934389       baseline
sysdata-update     default       0.000081974  13.28% faster
sysdata-update     wide          0.003869354   1.65% faster
```

### DIFF
--- a/py3status/modules/sysdata.py
+++ b/py3status/modules/sysdata.py
@@ -134,81 +134,12 @@ class Py3status:
     thresholds = [(0, "good"), (40, "degraded"), (75, "bad")]
 
     class Meta:
-        def update_deprecated_placeholder_format(config):
-            padding = config.get("padding", 0)
-            precision = config.get("precision", 2)
-            format_vals = ":{padding}.{precision}f".format(padding=padding, precision=precision)
-            return {
-                "cpu_freq_avg": format_vals,
-                "cpu_freq_max": format_vals,
-                "cpu_usage": format_vals,
-                "cpu_used_percent": format_vals,
-                "cpu_temp": format_vals,
-                "load1": format_vals,
-                "load5": format_vals,
-                "load15": format_vals,
-                "mem_total": format_vals,
-                "mem_used": format_vals,
-                "mem_used_percent": format_vals,
-                "mem_free": format_vals,
-                "mem_free_percent": format_vals,
-                "swap_total": format_vals,
-                "swap_used": format_vals,
-                "swap_used_percent": format_vals,
-                "swap_free": format_vals,
-                "swap_free_percent": format_vals,
-            }
-
-        deprecated = {
-            "rename": [
-                {
-                    "param": "temp_unit",
-                    "new": "cpu_temp_unit",
-                    "msg": "obsolete parameter use `cpu_temp_unit`",
-                },
-            ],
-            "rename_placeholder": [
-                {
-                    "placeholder": "temp_unit",
-                    "new": "cpu_temp_unit",
-                    "format_strings": ["format"],
-                },
-                {
-                    "placeholder": "cpu_usage",
-                    "new": "cpu_used_percent",
-                    "format_strings": ["format"],
-                },
-                {
-                    "placeholder": "mem_unit",
-                    "new": "mem_total_unit",
-                    "format_strings": ["format"],
-                },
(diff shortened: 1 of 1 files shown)


lasers:
@anarcat for potential zfs testing

anarcat:
hey, i'm sorry but i can't test this as i don't run zfs on my desktop anymore and i don't even run py3status either. :)

py3status/modules/sysdata.py
```diff
@@ -235,14 +233,14 @@ def update_deprecated_placeholder_format(config):
...
     def post_config_hook(self):
-        self.first_run = True
         self.init = {"meminfo": [], "stat": []}
         names_and_matches = [
             ("cpu_freq", ["cpu_freq_avg", "cpu_freq_max"]),
             ("cpu_temp", "cpu_temp"),
             ("cpu_percent", "cpu_used_percent"),
             ("cpu_per_core", "format_cpu"),
             ("load", "load*"),
+            ("max_used_percent", "max_used_percent"),
```
ultrabug: please add this to the docstring "Format placeholders:"

ultrabug:
```suggestion
            ("max_used_percent", ["max_used_percent", "max_cpu_mem"]),
```

ultrabug: else max_cpu_mem is now silently dropped no?

py3status/modules/sysdata.py
```diff
@@ -408,22 +431,19 @@ def _calc_mem_info(self, unit, meminfo, memory):
         )
 
     def _get_meminfo(self, head=28):
-        with Path("/proc/meminfo").open() as f:
+        with self.paths["meminfo"].open() as f:
             info = [next(f).split() for _ in range(head)]
             return {fields[0]: float(fields[1]) for fields in info}
 
     def _get_zfs_arc_size(self):
```
ultrabug, line 436: once again here I'd have kept the guard try/except in case the file does not exist or inaccessible instead of crashing the module

ultrabug (changes requested):

py3status/modules/sysdata.py
lasers: not silently dropped.

color only. it was never a placeholder. still supported in legacy thresholds (which would had been removed in `modules: drop legacy config compatibility` pr first thing).

to simplify/stay truthy, i exposed this as `max_used_percent` placeholder.

py3status/modules/sysdata.py
lasers: okay, we try it until it breaks. most users don't use zfs desktop.

py3status/modules/sysdata.py
lasers: `max_cpu_mem` added. This pr was stacked on that `modules: drop legacy config compatibility` pr (which doesn't have `max_cpu_mem` anymore) and that this pr now determine funcs based on placeholders/color names. yeah. :)

p.s. max_cpu_mem is very old too.

ultrabug:
Thanks

PR #2350, opened 2026-07-21
modules: drop legacy config compatibility

lasers:
(cleanup) This removes legacy stuff. We kept them long enough.

### DIFF
--- a/py3status/modules/arch_updates.py
+++ b/py3status/modules/arch_updates.py
@@ -44,9 +44,6 @@ class Py3status:
     format = None
     hide_if_zero = False
 
-    class Meta:
-        deprecated = {"remove": [{"param": "include_aur", "msg": "obsolete"}]}
-
     def post_config_hook(self):
         helper = {
             "pacman": self.py3.check_commands(["checkupdates"]),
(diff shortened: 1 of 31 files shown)


lasers:
@ultrabug This can come after `parse_config: improve config loading` pr.

Some bad decisions were definitely made (ie replacing `format_*` with `icon_*`) in rather simple modules. The best decisions placed (almost) everything in a single `format`. Oh well.

lasers:
@ultrabug What next? This.

ultrabug:
After re-reading RFC 4.0 I'm not sure this is what I want tbh. I'd expect the users' config to be automagically upgraded to those new config params instead of breaking until they manually switch over.

There's no real harm in keeping this. I'm not settled on this yet :(

lasers:
Sure, we can move on to next one. 

Just rebased `sysdata: streamline metric collection - #2349` without this pr.

lasers:
I'm working on config source updater that would nag users to update their configs first... allowing faster config deprecation... and to keep deprecated codes out of modules immediately. Wait until that one gets merged in first.

I'm just stopping / capping at 30 prs right now. :grin:

PR #2354, opened 2026-07-21
clean up python2 leftovers

lasers:
We got rid of python2.7 eons ago, but we still got few more... So I asked my buddy J.A.R.V.I.S. to seek and destroy any legacy python2 compatibility leftovers it can find in the repository except for... explicitly skipped files.
- py3status/i3status.py
- py3status/modules/i3pystatus.py
- py3status/modules/mpd_status.py

1) Got rid of python2-related unicde tests.
2) Dropped pickle 2/3 compatibility version (2) in favor of default version (4). The highest version is currently (5) and will not work with `python 3.10` anyway.
3) Skimmed through the diffs (comments, modifications) and they looked okay.
4) Python2 code still lives in `i3status.py` and I hope to replace this with `modules.i3status` module.

### DIFF
--- a/docs/dev-guide/the-py3-helper.md
+++ b/docs/dev-guide/the-py3-helper.md
@@ -71,7 +71,7 @@ If a string is passed then that command will be checked for.
 
 ### command_output(command, shell=False, capture_stderr=False, localized=False)
 
-Run a command and return its output as unicode.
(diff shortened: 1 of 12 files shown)


ultrabug:
Thank you

PR #2357, opened 2026-07-23
add new module: sysinfo

lasers:
Hi, I made a new module `sysinfo`. This just displays static system information. 

image
Basically, I thought why don't we clean this place up a bit by just consolidating static modules?

- `uname`
- `whoami`
- ~~`static_string`~~

I have added a lot of new static placeholders, but removed lot of them afterward too... except for handful ones. ~~Compare commits to find original ones.~~ See https://github.com/lasers/py3status/commit/379ee76d74e07a5f5baac376514c75d8ca38ded6  instead.

I rather users add new ones themselves if they want/need it. I know `static_string` technically could be part of this module too, but... it is not... and we probably should leave that one alone. 

I'm just tossing this module out. I think `sysinfo` is a nice wordplay on `sysdata`. Kept new placeholders `{python_version}`, `{py3status_version}` and `{wm_name}` in to make it a more useful module. Doesn't have to be merged.

EDIT: One more thing, it is easier to remember static system information are all kept in one place. EDIT: One more thing again, it's nice to establish a ~~place~~ module for static information, not necessarily system information (ie sysinfo). I thought about moving all three to `static` module instead, but it feels odd, but not wrong. I want to reduce redundant/minimal codes/modules and keep/make the fat/good ones. Idk. Just sharing thoughts AND FEELINGS!


lasers:
No need to wait on this PR. Review when convenient. Random pick by default.

py3status/modules/sysinfo.py
```diff
@@ -0,0 +1,132 @@
...
+
+        if not self.format:
+            keys, values = (("version",), (None, ""))
+            key, value = choice(
+                [(k, v) for k, v in self.sys.items() if k not in keys and v not in values]
+            )
+            self.format = r"{key} [\?color=darkgray {value}]"
+            self.sys.update(key=key, value=value)
+
+        if self.py3._py3_wrapper.config.get("testing"):
```
ultrabug, line 114: what is this testing? something you forgot to remove maybe?

py3status/modules/sysinfo.py
```diff
@@ -0,0 +1,132 @@
+"""
+Display system information.
+
+Configuration parameters:
+    format: display format for this module, otherwise random (default None)
```
ultrabug, line 5: I'd rather provide a sensible default than random stuff tbh

ultrabug:
I like the gist of having system infos in one place yeah. I wouldn't go too far on the static thoughts tho. Keep it simple as it is proposed here with sensible default including the uname and whoami placeholders that are being replaced at least.

py3status/modules/sysinfo.py
lasers: (debug) Run `python3 sysinfo.py --term --log-level debug` in the term to see actual placeholders.

lasers:
> Keep it simple as it is proposed here with sensible default including the uname and whoami placeholders that are being replaced at least.

@ultrabug Re: RFC 4.0, where you want seamless configuration updates.... configuration reader should backup and update user config on the fly.

I have a wip branch that basically rewrite config on the fly. Promise. It is stacked after i3status stuff and intermixed with retiring modules (and) rewriting i3status wrapper to modules.

This module is basically part of this (a concept). My thinking is that `sysinfo` should stay random (fun, i know) because the `format` is not set... and users can contribute more static stuff down the road. Just set `format` to get what you want to stop this.

Plus we're also going to make `uname` and `whoami` work flawlessly with and without their formats too, generated from `sysinfo` module because both has same `format` and placeholders.

Maybe it's best to hold this off until the end.

lasers:
I looked at this again.

It's not intended to be a total config drop-in replacement. It's for mostly modules being renamed, has same format, and optional placeholder renames. Very simple.

The goal is to basically translate similar modules like uname, whoami, scratchpad, window, etc on the fly.

For deeper ones, we have module deprecation (class Meta) that we probably should use... possibly in combination with this.

Can confirm `order+=whoami` and `order+= uname` worked fine out of the box with and without module config sections... both generated from sysinfo module.

PR #2359, opened 2026-07-30
conky: log module errors and process exits

lasers:
This fixes a minor issue where errors ~~and process exits~~ are displayed on the bar just fine, but not logged.

I noticed this issue when I was working on i3status stuff.
`i3status` module is similar to `conky` module so I'm just fixing this too.

### DIFF
--- a/py3status/modules/conky.py
+++ b/py3status/modules/conky.py
@@ -322,9 +322,10 @@ bar
 ]
 """
 
+from contextlib import suppress
 from json import dumps
 from pathlib import Path
-from subprocess import PIPE, STDOUT, Popen
+from subprocess import PIPE, STDOUT, Popen, TimeoutExpired
 from tempfile import NamedTemporaryFile
 from threading import Thread
 
@@ -384,8 +385,8 @@ class Py3status:
 
         # skip invalid conky errors
         self.ignored_conky_outputs = [
-            "conky: invalid setting of type 'table'",
-            "conky: FOUND:",
+            "invalid setting of type 'table'",
+            "FOUND:",
             "session running",
         ]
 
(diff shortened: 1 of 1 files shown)


py3status/modules/conky.py
```diff
@@ -393,29 +394,51 @@ def post_config_hook(self):
...
         self.py3.update()
 
     def _start_loop(self):
         try:
-            self.process = Popen(self.conky_command, stdout=PIPE, stderr=STDOUT)
-            while True:
-                line = self.process.stdout.readline().decode()
-                if self.process.poll() is not None or "conky:" in line:
-                    if any(x in line for x in self.ignored_conky_outputs):
+            self.process = Popen(self.conky_command, stdout=PIPE, stderr=STDOUT, text=True)
```
ultrabug:
```suggestion
            self.process = Popen(self.conky_command, stdout=PIPE, stderr=STDOUT, text=True, encoding="utf-8", errors="replace")
```

Quoting AI 

```
The old .decode() was hardcoded UTF-8. text=True uses locale.getpreferredencoding(False), so under LANG=C/POSIX this becomes ASCII and any non-ASCII conky output (process names from {top name}, {execi} output, non-ASCII bar chars) now raises UnicodeDecodeError where it previously decoded fine. i3bar-spawned processes inherit whatever locale the session has, so this isn't hypothetical.
```

py3status/modules/conky.py
```diff
@@ -393,29 +394,51 @@ def post_config_hook(self):
...
+            while self.running:
+                line = self.process.stdout.readline()
+                process_stopped = self.process.poll() is not None
```
ultrabug:
```suggestion
                if not line:  # EOF
                    self.process.wait()
                    if self.running:
                        self._set_error("")
                    break
                process_stopped = self.process.poll() is not None
```

Quoting AI

```
At EOF readline() returns "" immediately, but poll() can still return None for a short window before the child is reaped. In that window process_stopped and conky_error are both false, so "" is treated as data: self.line = "" plus a py3.update() (blanking the bar), then the loop spins at full speed on empty reads until poll() catches up. This race exists on master too, so it's not a regression — but you're already in this code:
```

py3status/modules/conky.py
```diff
@@ -393,29 +394,51 @@ def post_config_hook(self):
         self.line = ""
         self.error = None
         self.process = None
+        self.running = True
         self.t = Thread(target=self._start_loop)
         self.t.daemon = True
         self.t.start()
 
+    def _set_error(self, error):
+        error = str(error).strip().removeprefix("conky").strip(" .:")
```
ultrabug, line 403:
```suggestion
        error = str(error).strip().removeprefix("conky:")
```

py3status/modules/conky.py
```diff
@@ -393,29 +394,51 @@ def post_config_hook(self):
...
+        if not error:
+            returncode = self.process.poll() if self.process else None
+            if returncode is None:
+                error = "stopped unexpectedly"
+            elif returncode < 0:
+                error = f"exiting due to signal {-returncode}"
+            else:
+                error = f"exiting due to code {returncode}"
+        self.error = error
+        self.py3.log(self.error, level="error")
```
ultrabug:
```suggestion
        self.py3.log(self.error, level=self.py3.LOG_ERROR)
```

ultrabug (changes requested):

py3status/modules/conky.py
lasers: Done.

EDIT: 
> At EOF readline() returns "" immediately, but poll() can still return None for a short window before the child is reaped. In that window process_stopped and conky_error are both false, so "" is treated as data: self.line = "" plus a py3.update() (blanking the bar), then the loop spins at full speed on empty reads until poll() catches up. This race exists on master too, so it's not a regression — but you're already in this code:

py3status/modules/conky.py
lasers: Error is based on leading `conky:` string. I remove `conky`, then I do `strip(asdfghjkl)` one more time to clear whitespace because I haven't looked if `conky` is consistent with its error output.

Your suggestion still need something like `removeprefix("conky:").strip()` too because of probably whitespace. This was me indrectly saying, ok, we need to strip `" "`, `":"` and eh, let's do `"."` too for nice bar/logging.

lasers:
@ultrabug Fixed behavior when `startswith("conky:")` errors occurs, errors gets displayed, logged, but the process still run... assuming users fix config and restart/reload soon enough.... Now we just kill the process too.

- logged (new)
- kill process (very new)

py3status/modules/conky.py
lasers: Done. FYi, you can use `>` instead to quote. Yes, I scrolled all the way to end!

EDIT:
> The old .decode() was hardcoded UTF-8. text=True uses locale.getpreferredencoding(False), so under LANG=C/POSIX this becomes ASCII and any non-ASCII conky output (process names from {top name}, {execi} output, non-ASCII bar chars) now raises UnicodeDecodeError where it previously decoded fine. i3bar-spawned processes inherit whatever locale the session has, so this isn't hypothetical.

ultrabug:
Thank you

PR #2365, opened 2026-08-07
py3.safe_format: remove max_width parameter

lasers:
Not easy to say this. I respect you too. Hear this out.

 This reverts https://github.com/ultrabug/py3status/pull/2113 in favor of something that exists for long time.

I particularly don't remember my comments there. I found this odd when I was working on `threshold_update` and went looking at the history. I feel like this was a rushed PR. My original suggestion wasn't even attempted.

1)  `mpris`'s default format means the the buttons are the first one getting cut off but wait, okay, you skip ones with `index`, then that mean `max_width` can be inaccurate, or a little unclear, but okay, I understand this.

2) This was made specifically for `mpris` only and may have been made fast to address user's feature request with `mpris` outside of this space.

3) It is not easy to add this config in several modules, but it is much easier to safeformat modules and just having users inline `\?max_length=5`  in their formats (anywhere) if they want it.

4) Most likely, no other modules will use this implementation.

EDIT: I just want this to be looked at... again. Users should use (simplified) suggestion below instead.
```diff
diff --git a/py3status/modules/static_string.py b/py3status/modules/static_string.py
index 1c26059..db53fbe 100644
--- a/py3status/modules/static_string.py
+++ b/py3status/modules/static_string.py
@@ -15,7 +15,7 @@ class Py3status:
     """ """
 
     # available configuration parameters
-    format = "Hello, world!"
+    format = "\?max_length=5 Hello, world!"
 
     def static_string(self):
         return {
```
```shell
$ python static_string.py --term
Hello
Hello
^C⏎
```

### DIFF
--- a/docs/dev-guide/the-py3-helper.md
+++ b/docs/dev-guide/the-py3-helper.md
@@ -382,7 +382,7 @@ be added automatically following the convention:
 
 returns: HttpResponse
 
-### safe_format(format_string, param_dict=None, force_composite=False, attr_getter=None, max_width=None)
+### safe_format(format_string, param_dict=None, force_composite=False, attr_getter=None)
 
 Parser for advanced formatting.
 
@@ -434,9 +434,6 @@ returned.
 
 attr_getter is a function that will when called with an attribute name
 as a parameter will return a value.
-
-max_width lets you to control the total max width of 'full_text' the
-module is allowed to output on the bar.
 
 ### stop_sound()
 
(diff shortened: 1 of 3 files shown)


ultrabug:
The idea behind max_width was to start working on supporting native `short_text` from the [i3bar protocol](https://i3wm.org/docs/i3bar-protocol.html) which I sadly never added support on py3status.

I always wanted a simple and generic way to have py3status be clever and add default support for short text output (using max_width) which users could also override by adding their own `short_format` too

lasers:
How does that work with composites? Each block would have its own `full_text` and optionally `short_text` too. With  `short_text`, one only can have a single color since `short_text` is only good for one dull block. 

Does i3bar just turn on `short_text` for all blocks everywhere or only those that went through the invisible wall (i.e having a processed composite shredded into a leading short_text + trailing full_text pieces?)

We'll figure it out eventually. This just ain't it (imh0).

lasers:
@ultrabug py3status will have `short_text` support ;)

PR #2367, opened 2026-08-08
py3, modules: support native separators

lasers:
~~(Hack)~~ Every once in a while, I wonder why we can't have nice things... like real separators.

We safeformat our separators everywhere so technically `safeformat` is for strings, not `None`, but if we do this way, we can return `None` back straight to `composite_join()` next where it will add real separators there.

I do it this way just because if we don't, we will need to do something like this for all modules w/ separators.
```python
separator = self.format_separator
if separator is not None:
    separator = self.py3.safe_format(separator)

output = self.py3.composite_join(separator, items)    
```
or one liner
```python
output = self.py3.composite_join(None if format_separator is None else self.py3.safe_format(format_separator), items)
```
---- 

`lm_sensors` : Add this for chips for for sensors to see diff.
```diff
diff --git a/py3status/modules/lm_sensors.py b/py3status/modules/lm_sensors.py
index d48e355..8dccb1e 100644
--- a/py3status/modules/lm_sensors.py
+++ b/py3status/modules/lm_sensors.py
@@ -176,9 +176,10 @@ class Py3status:
     chips = []
     format = "{format_chip}"
     format_chip = "{name} {format_sensor}"
-    format_chip_separator = " "
+    format_chip_separator = None
     format_sensor = r"[\?color=darkgray {name}] [\?color=auto.input&show {input}]"
-    format_sensor_separator = " "
+    format_sensor_separator = None
     sensors = []
     thresholds = {"auto.input": True}
 
```
Idk if we need this. Idk if we should merge this. Cool feature tho.
Changes `safe_format`'s `format_string` too. Hack.

---

Pls you no merge. Need 2 think more.

### DIFF
--- a/docs/dev-guide/the-py3-helper.md
+++ b/docs/dev-guide/the-py3-helper.md
@@ -438,6 +438,15 @@ as a parameter will return a value.
 max_width lets you to control the total max width of 'full_text' the
 module is allowed to output on the bar.
 
+### safe_join(separator, items)
+
+Join items using a format string, Composite, or native separators.
+
+String separators are processed by the formatter. Boolean separators
+control whether native separators are used between visible items.
+
+A Composite object will be returned.
+
 ### stop_sound()
 
 Stops any currently playing sounds for this module.
--- a/py3status/composite.py
+++ b/py3status/composite.py
@@ -95,7 +95,21 @@ class Composite:
             diff = item.copy()
             del diff["full_text"]
 
-            if diff == diff_last or (item["full_text"].strip() == "" and item_last):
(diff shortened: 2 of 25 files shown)


lasers:
No longer a hack, I think.

Use `self.py3.safe_join()` when it's for separators... permitting users to initialize `None`.

lasers:
Fixed `google_calendar` to get rid of the `sep` index and zero-based event numbers (which may have opened incorrect URLs, maybe... I don't know) in favor of using event_id instead.

Need testers. This is the last module AFAIK.

PR #2370, opened 2026-08-10
module: support i3bar min_width string values

lasers:
This fixes `min_width = "WWWWWWWW"`.

```python
order += "static_string wall"
order += "static_string width_pixels"
order += "static_string width_text"
order += "static_string wall"

static_string wall {
    format = "\|"
}

static_string width_pixels {
    format = "200"
    min_width = 200
    align = "center"
}

# fix this
static_string width_text {
    align = "center"
    format = "W"
    min_width = "WWWWWWWW"
}
```

---

From  https://i3wm.org/docs/i3bar-protocol.html
> **min_width**
>
> The minimum width (in pixels) of the block. If the content of the full_text key take less space than the specified min_width, the block will be padded to the left and/or the right side, according to the align key. This is useful when you want to prevent the whole status line to shift when value take more or less space between each iteration.
>
>The value can also be a string. In this case, the width of the text given by min_width determines the minimum width of the block. This is useful when you want to set a sensible minimum width regardless of which font you are using, and at what particular size.

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -167,7 +167,8 @@ The following options will work on `i3`.
   is not reached. Either `left` (default), `center`, or `right`.
 - `background`: Specify a background color for py3status modules.
 - `markup`: Specify how modules should be parsed.
-- `min_width`: Specify a minimum width of pixels for modules.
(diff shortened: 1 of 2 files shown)


lasers:
@ultrabug Ready for review.

PR #2371, opened 2026-08-10
module: document i3bar and non-i3bar separators

lasers:
~~This addresses `HACK: separator` issue I ran into... when I added  universal/i3bar_gaps module options years ago.~~

* Applied the fix at first, but ultimately, we need to keep the code unchanged.
* Keep `general` on `True` for `separator` config and just document.

### DIFF
--- a/py3status/core.py
+++ b/py3status/core.py
@@ -137,11 +137,11 @@ class Common:
         self.none_setting = NoneSetting()
         self.config = py3_wrapper.config
 
-    def get_config_attribute(self, name, attribute):
+    def get_config_attribute(self, name, attribute, general=True):
         """
         Look for the attribute in the config.  Start with the named module and
-        then walk up through any containing group and then try the general
-        section of the config.
+        then walk up through any containing group and the py3status section.
+        Finally, try the general section unless `general` is false.
         """
 
         # A user can set a param to None in the config to prevent a param
@@ -158,7 +158,7 @@ class Common:
         if hasattr(param, "none_setting"):
             # check py3status config section
             param = config["py3status"].get(attribute, self.none_setting)
-        if hasattr(param, "none_setting"):
+        if hasattr(param, "none_setting") and general:
             # check py3status general section
             param = config["general"].get(attribute, self.none_setting)
         if param and (attribute == "color" or attribute.startswith("color_")):
@@ -776,7 +776,7 @@ class Py3statusWrapper:
             self.config["py3_config"]["general"]["output_format"]
         )
 
-        # determine the output separator, if needed
+        # configure the separator for non-i3bar output formats
         color_separator = None
         if self.config["py3_config"]["general"]["colors"]:
             color_separator = self.config["py3_config"]["general"]["color_separator"]
--- a/py3status/module.py
+++ b/py3status/module.py
@@ -346,17 +346,18 @@ class Module:
                     raise ValueError(err)
                 self.i3bar_module_options["align"] = align
 
-        separator = fn(self.module_full_name, "separator")
+        separator = fn(self.module_full_name, "separator", True)
         if not hasattr(separator, "none_setting"):
-            # HACK: separator is a valid setting in the general section
-            # of the configuration. but it's a string, not a boolean.
-            # revisit how i3status and py3status differ in this regard.
+            # separator is a valid setting in the general section
+            # (a string for non-i3bar output formats); we usually
+            # want boolean from module or py3status section.
+            # ----
             # if not isinstance(separator, bool):
-
             #     err = "Invalid `separator` attribute, should be a boolean. "
             #     err += f"Got `{separator}`."
             #     raise TypeError(err)
             # self.i3bar_module_options["separator"] = separator
+            # ----
             if isinstance(separator, bool):
                 self.i3bar_module_options["separator"] = separator
 


lasers:
@ultrabug Made the fix, but not applied. Changed nothing.

Code just get an arg to skip general section now.

lasers:
@ultrabug Ready for review.

ultrabug:
While your edits of comments look appropriate I don't see the point of adding a positional argument that is used only once in the whole code base (and silently defaulted elsewhere). Do we really need that signature change? Maybe just comments and we're done?

lasers:
It's the right fix signature for skipping wrong separator  if users had both separator set (one in i3status's general config, one in py3status config aka universal options).

It's not likely for users to set this, but we handled it fine first time by inverted validating (ie accepting right value type instead).

We don't need this because I think it mostly got sorted out early in parse_config... and we assumed there are no attribute issue between module/py3status and general... which made me think we may not really need lot of general-related code stuff (like this) everywhere.

If both module/py3status + general accepts same type but different values, then we are screwed without this.

Burning question.
Does py3status do module/py3status/general for py3status modules... or just module/py3status? As in... Is general config a core part of py3status, or do we just do this to support i3status wrapper?

I asked because of i3status-related changes I had been messing with... and if i3status-wrapper is gone (ie moved) then we technically can destroy general early after giving i3status (python and/or module) what it need from user config.

lasers:
I don't disagree with you. Just a trivial correctness fix for something not right now.

Just close this PR instead if you don't want this in. Doesn't change anything either way.

ultrabug:
> Is general config a core part of py3status, or do we just do this to support i3status wrapper?

We aim at being 100% in line with i3status config yes so we have to keep general

ultrabug:
After adding my comment I thought that maybe you had a pending PR somewhere else using it, is that the case?

lasers:
No pending PR that explicitly uses this.

Just like `tests: fix formatter color coverage` pr, I was just looking at HACKS, TODO, FIXME, etc. I remembered having issue with this and went to review it again. 

I added `i3status = "python"` flag to py3status config section for `add native i3status-compatible modules` pr. I think it grabbed the value from config (via parse_config) instead of this anyway.

Do you have Discord?

----

> We aim at being 100% in line with i3status config yes so we have to keep general

Meaning maybe we should get rid of late general-related code because we would have applied whatever i3status (wrapper/python/module) needed at startup... and apply known settings (like colors if set) to py3status config at startup... and this function would only check module/container/py3status... no more `general`. Early initialization/optimization stuff.

Does that make sense? Otherwise, users who switched back to i3status can have invalid py3status-only configs in general config because I'm not sure if we should support having py3status-only configs in general config too.

lasers:
```python
general {
  separator = True
  i3status = "python"
}
```
i3status doesn't complain about `separator = True`. Fine. We got lucky.
i3status complained with `no such option 'i3status'`.

If i3status complained that `separator` is not a string, then this fix is needed.
EDIT: It looks like i3status just parsed `True` as string `True` anyway.

lasers:
> We aim at being 100% in line with i3status config yes so we have to keep general

Does this mean we should try to block non-i3status features in `general` config too? See below. py3status-only feature.

```python
general {
    min_length = 10
    position = 'center'
}
```
```bash
$ i3status -c ~/.config/py3status/config
no such option 'min_length'

$ i3status -c ~/.config/py3status/config
no such option 'position'
```

This fix can be applied to those two (or three). Otherwise, they stay inside `general` config... and it would be fine as long as they stayed with py3status.

This is why I am not sure if we should get rid of `general` early so we don't accidentally support py3status-only configs in `general`... or just use flags? We tends not to use general config for anything other than startup parsing.

lasers:
Hmm. i3status reports invalid configs like...

`/path/to/i3status/config5: no such option 'py3status'`
`/path/to/i3status/config5: no such option 'static_string`
... which makes it impossible... so users need to comment them out.

Now I wonder about colors.
```python
general {
    color_good = "lightskyblue"
}
order += "load"
```
```bash
$ i3status
Bad color format
```

py3status read `lightskyblue` just fine. Now, do we change code so that `color_good = "lightskyblue"`  in general doesn't work too... enforcing users to keep py3status-only features in `py3status` config? 

Sorry about all those trivial but correctness questions on the relationship between general + py3status config sections.

lasers:
I hold off on this because of `i3status: add proxy/container modules` pr and other prs that should be reviewed first. This pr is just a nothingburger right now.

My big question/goal/discovery with this is that there are two "global" config... `general` and `py3status`... and whenever `general` is really a core part of py3status... or just for those few i3status configs (colors_*, interval, markup)... Very few, really.... then maybe we should put everything in py3status config section (rather than having both).

If i3status modules can be contained to a container (it currently do both containers + bare i3status modules) then we could enable `general=False` for most of our module options (see a list below in a photo).... suggesting users to move them to py3status config section from now on... if any. Phrasing out global `general` config early (if wanted) is possible in parse_config.... but still supported for i3status only (the general configs would be dealt with early)

Screenshot_20260911_220635

I'm not trying to cause headaches. This is headache for me too... because of several levels and/or possible duplicates...  having to parse correctly with newly i3status container's own general {} -> py3status{} -> top-level general {}... and possibly not forgetting group {}... but some part is going to be supported too, gets scoped to configs inside i3status general, and not just any configs.

Keeping py3status as i3status wrapper is a forever pain (due to keeping compat), but at least it can be out of core/command/etc and can be ran as real py3status modules.

I do not mind cleaner/faster deprecation by containing i3status settings+bare i3status modules down to i3status containers only... but this mean basically stop doing i3status wrapper. It's not that I'm against i3status wrapper. I'm just trying to make py3status better than before (in my own eyes).

It doesn't have to go. It just mean we can do first one too. 
```python
# bare i3status modules
group system1 {
	load {}
	cpu_temperature 0 {}
	sysinfo {}
	conky {}
}
```
```python
# i3status container
group system2 {
    i3status {
        load {}
        cpu_temperature 0 {}
    }
    sysinfo {}
    conky {}
}
```
I think I will get this resolved soon either way anyway.

PR #2372, opened 2026-08-10
tests: fix formatter color coverage

lasers:
This fixes few wrong color tests (ie some color tests wasn't consistent with real bar colors).


ultrabug:
This time I tried it full AI, what do you think?

2026-08-13-220611_1125x1466_scrot

lasers:
Done. Probably should had stayed with `\?if=value<0 hide_this_block_please` instead of (color?) `hidden` thresholds.

PR #2374, opened 2026-08-14
core: support terminal notifications

lasers:
This prints nagbar notifications in terminal. This also drop reload advice.

1) People don't always use `i3wm`. You forget `sway`. 
2) People don't always bind `Mod+Shift+R`. Their setups can be different.
3) I'm often testing py3status in terminal so I add support for terminal notifications too.
4) Reload advice doesn't make sense in terminal anyway.
5) Clean as a whistle.


lasers:
@ultrabug Ready for review.

PR #2377, opened 2026-08-26
add native i3status-compatible modules

lasers:
i3status-banner

### Native, subprocess-free reimplementation of i3status's bundled modules.

py3status/i3status/modules/ now ships Python reimplementations of all 16 i3status 2.15 bundled modules (battery, cpu_temperature, cpu_usage, ddate, disk, ethernet, ipv6, load, memory, path_exists, read_file, run_watch, time, tztime, volume, wireless), talking to the kernel/hardware directly (ALSA and PulseAudio via ctypes, nl80211 via netlink, ethtool via ioctl, /proc and /sys directly) instead of shelling out to the real i3status binary.

Dispatch between the real i3status wrapper and these native modules is per-section: a module's config section can set `python = True`/`False` to force one or the other, and `py3status { i3status = "python" }` sets the default for all 16 module names at once when neither is set.
```python
py3status {
    i3status = "python"  # global
}

ethernet eth0 {
    format_up = "E: %ip (%speed)"  # python
}

```
```python
py3status {
    # Intentionally left blank
}

volume master {
    python = True
    format = "♪: %volume"  # python
    mixer = "Master"
}

cpu_temperature 0 {
    format = "T: %degrees C"  # wrapper
    max_threshold = 80
}

tztime local {
    format = "%H:%M %Z"  # wrapper
}

tztime berlin {
        python = True
        format = "%Y-%m-%d %H:%M:%S %Z"  # python
        timezone = "Europe/Berlin"
}

```

Since several of these modules derive part of their config from their i3status order-entry's instance/title rather than an ordinary config key (eg `disk "/home" { }` implies path="/home", `battery 0 { }` implies number=0), translate_instance()/instance_config_key() in py3status/i3status/helpers.py fill that in automatically for the native modules, and parse_config.py strips those same instance-derived keys (and the `python` directive itself) before delegating to the real i3status wrapper, since real i3status hard-errors on any config key it doesn't recognize.

Each module's docstring documents its deliberate deviations from the real i3status binary (bugs it doesn't reproduce, options it doesn't support, defaults it adds for a better native-config experience), most verified against the actual i3status 2.15 source and binary.


lasers:
(Note) This is usable right now. I don't know your plan. See if you like my plan.

* I don't want to support dual i3status (wrapper + python).
* This was aimed to be a drop-in replacement (not finalized).
     * 99.5% feature parity with real i3status -- all configs.
          * 0.5% is BSD / untested / unverified stuff.
     * 75% features parity with py3status -- such as post_config_hook, exception, color, cache_timeout.
       * i3status-python modules can run individually.

* Merge newmodule-i3status
  * An alternative option -- the one with real i3status + 100% py3status customization + features.
  * Good way to verify differences (if any) between real i3status and i3status-python.
  * You can run everything off one process with this too.
  * Issues arise when wanting separate on_clicks or universal_module options. For that, run i3status instances instead... or not caring that if sole on_click works with all i3status outputs.

* Remove i3status-wrapper and its residues. Lot of residues.
  * Remove some more i3status-wrapper residues.

Ideally, we should remove i3status-wrapper + residues first as much as we can -- then cleaning up this pr again to drop dual i3status support (such as guards/translations/extras/etc).

PR #2382, opened 2026-09-04
docs: switch to zensical

lasers:
This switches to zensical.

I initially went with full page, but to stay modern ,we should use its search instead of full page ctrl+f.

~~1) Let me know if you want to prefer to preserve or omit doc dirs. I don't see this turning into much more (and wouldn't mind omitting).~~
~~2) What to do with main page `README.md`? It's mostly static / stale. I can make it minimal and just point it to 
zensical docs.~~
_See different comment below._

`Py3 References` are now generated from docstrings. No more manually updating docs in two different places and no more drifting docs. Some codes also has been refactored to make documenting values (such as a list of config candidates) possible. I have applied macros everywhere I can find. I have been cleaning up too.

I paint this with py3status baby blue too. Thx.


lasers:
No feedback yet. I went with `keep-things-simple` approach by omitting internal docs dirs and making main page README minimal _just like modules/README_ so we don't have too much of stale/inconsistency documentations. It is hard to keep them ~~synchronized~~ updated _between site and GitHub README/markdowns._

All documentations changes would be done for the site, not GitHub README/markdowns.

Let me know what you think of it... and if we should make any other change.

P.S. Zensical is a rust drop-in replacement that's like 5x faster and this is just getting py3status docs sorted away ahead of mkdocs 2.0 milestone (no plugins, several changes, et cetera).

lasers:
TODO/NOTE: Some things can be refactored to make more macros possible.

lasers:
Finished adding more macros. Ready for review.

PR #2386, opened 2026-09-13
getjson: support custom request headers

bensynapse:
I maintain Live Tennis API, and I should say plainly why I want this: our own API authenticates with an `X-API-Key` header, so `getjson` cannot currently talk to it. The feature is generic and applies to any header-authenticated API, but the motivation is mine, and you should weigh it on that basis.

`getjson` currently supports Basic authentication but cannot send headers required by APIs using `X-API-Key` or Bearer tokens. Add an optional `headers` dictionary and pass it to the existing HTTP helper. Copy it for each request so the helper's User-Agent and Basic-auth additions do not modify the configuration. The module documentation includes an example and the existing Basic-auth precedence.

Verified with `hatch run all`: 253 tests passed on each of Python 3.10, 3.11, 3.12, 3.13 and 3.14; isort, Black and Ruff passed. Seven new tests use a real localhost HTTP server through the project's module loader, request helper and formatter, covering API-key/Bearer headers, defaults, Basic auth and configuration preservation across requests and instances.


py3status/modules/getjson.py
```diff
@@ -51,8 +52,18 @@
...
+# authenticate using an API-key header
+getjson {
+    url = 'https://example.com/api/status'
+    headers = {'X-API-Key': 'your-api-key', 'Accept': 'application/json'}
+    format = '{status}'
+}
 ```
 
+If both `username` and `password` are set, Basic authentication takes precedence
+over an `Authorization` header supplied in `headers`.
```
lasers: This should belong in newly created `Notes:` section just before `Examples:` section.

lasers (changes requested):

py3status/modules/getjson.py
bensynapse: Moved the Basic-auth precedence guidance into a new `Notes:` section immediately before `Examples:` in 32656ea. `hatch run all` passes: 253 tests on each Python 3.10–3.14, plus isort, Black and Ruff.

